PR #18, opened 2026-04-20
fix: map missing template cleanup locator to not found

A55566:
## Summary
- map ErrTemplateCleanupLocatorMissing to ErrorCode_NotFound in deleteTemplate
- update delete template HTTP handler test expectation accordingly

## Why
A missing historical cleanup locator indicates stale or missing template cleanup state rather than an invalid client parameter, so returning NotFound better matches the actual condition and expected API behavior.


chenhengqi (commented):
Please add your signed-off-by tag and describe the purpose of this change in commit message.

up2wing:
> Please add your signed-off-by tag and describe the purpose of this change in commit message.

Yeap, maybe we should add the DCO check in workflows.
@fslongjin

fslongjin:
> > Please add your signed-off-by tag and describe the purpose of this change in commit message.
> 
> Yeap, maybe we should add the DCO check in workflows. @fslongjin



> > Please add your signed-off-by tag and describe the purpose of this change in commit message.
> 
> Yeap, maybe we should add the DCO check in workflows. @fslongjin

Sure, I'll add it later~

PR #34, opened 2026-04-21
feat(dev-env): add scripts for syncing binaries, log collection, and autostart management

fslongjin:
- Add sync_to_vm.sh: build and push binaries to VM with automatic rollback on failure
- Add copy_logs.sh: package and copy /data/log from guest to host
- Add cube-autostart.sh: enable/disable/status the systemd autostart unit from host
- Add internal/setup_autostart.sh: install cube-sandbox-oneclick.service inside VM
- Update README.md and README_zh.md with comprehensive documentation and troubleshooting
- Update .gitignore to ignore *.tar.gz


dev-env/.gitignore
```diff
@@ -1 +1,2 @@
 .workdir/
+*.tar.gz
```
chenhengqi: No newline at the end ?

dev-env/copy_logs.sh
```diff
@@ -0,0 +1,135 @@
+#!/usr/bin/env bash
+
```
chenhengqi, line 21: Could you please add comments for all these scripts ? Describe what the script does and how to use it.

dev-env/.gitignore
fslongjin: fixed

dev-env/copy_logs.sh
fslongjin: ok

PR #57, opened 2026-04-23
fix(kernel-sync): ensure latest shared vmlinux is used when recreating templates

### LINKED ISSUE #33: [Bug Report] Template redo may still use an old vmlinux after cube-kernel-scf/vmlinux is replaced
## Summary

After a template is created from an image, replacing `cube-kernel-scf/vmlinux` may not take effect in later `redo` runs. Redo can keep using the old per-artifact kernel file under `cubebox_os_image/<artifactID>/<artifactID>.vm` instead of the updated shared `vmlinux`.

## Environment

- CubeSandbox version / commit: c439bb513f5124d4d9389451b31b8aeb87ab539c
- Relevant component: `CubeMaster`, `Cubelet`

## Steps to Reproduce

1. Create a template from a fixed container image and let it complete successfully.
2. Verify that `cubebox_os_image/<artifactID>/<artifactID>.vm` already exists on the node.
3. Replace `cube-kernel-scf/vmlinux`.
4. Run `redo` for the same template without changing the source image digest or template parameters.
5. Observe which kernel file is actually used.

## Expected Behavior

Redo should pick up the new `cube-kernel-scf/vmlinux`, or at least should not silently keep reusing the old per-artifact kernel.

## Actual Behavior

Redo can still use:

```text
cubebox_os_image/<artifactID>/<artifactID>.vm
```

instead of:

```text
cube-kernel-scf/vmlinux
```

## Additional Context

This looks like a real code-level bug for three main reasons:

1. **Redo often reuses the previous artifact** instead of rebuilding it.
   - `CubeMaster/pkg/templatecenter/template_image.go:568`
   - `CubeMaster/pkg/templatecenter/template_image.go:1094`

2. **Artifact fingerprinting does not include kernel state**, so changing only `cube-kernel-scf/vmlinux` does not force a new artifact.
   - `CubeMaster/pkg/templatecenter/template_image.go:1781`
   - `CubeMaster/pkg/templatecenter/template_image.go:1222`

3. **Node-side kernel copy is effectively copy-once**: if `<artifactID>.vm` already exists, the updated shared kernel is not copied over it.
   - `Cubelet/internal/cube/server/images/ext4image/utils.go:119`
   - `Cubelet/internal/cube/server/images/ext4image/utils.go:124`
   - `Cubelet/internal/cube/server/images/ext4image/utils.go:127`
   - `Cubelet/internal/cube/server/images/ext4image/utils.go:135`

Runtime also clearly prefers the per-artifact kernel path under `cubebox_os_image` once `AppImageID` is set:

- `Cubelet/services/cubebox/cube_container_create.go:235`
- `Cubelet/plugins/cbri/cubeboxcbri/cubebox.go:153`
- `Cubelet/plugins/cbri/cubeboxcbri/cubebox.go:380`

There is already a unit test showing that the shared kernel update does **not** overwrite the existing per-artifact kernel:

- `Cubelet/internal/cube/server/images/ext4image/utils_test.go:17`
- `Cubelet/internal/cube/server/images/ext4image/utils_test.go:55`
- `Cubelet/internal/cube/server/images/ext4image/utils_test.go:56`

## Suggested Fix Directions

1. Include kernel identity (for example, `cube-kernel-scf/vmlinux` SHA256) in `TemplateSpecFingerprint`.
2. Refresh the existing `<artifactID>.vm` when it does not match the current shared kernel.
3. Ensure redo cleanup really removes the node-local per-artifact kernel before redistribution when needed.

fslongjin:
- Add`SyncKernelFile` to compare SHA256 and refresh target kernel when it differs from shared kernel
- Simplify`ensureKernelFile`to delegate to`pmem.SyncKernelFile`
- Call`syncLatestKernelForImage`in`CreateSandbox` when creating snapshots to ensure latest kernel is used
- Update tests to verify kernel refresh behavior

Close: https://github.com/TencentCloud/CubeSandbox/issues/33

### DIFF
--- a/Cubelet/internal/cube/server/images/ext4image/utils.go
+++ b/Cubelet/internal/cube/server/images/ext4image/utils.go
@@ -116,33 +116,7 @@ func tryDownloadPmemFile(ctx context.Context, imagePath string, spec *cubeimages
 }
 
 func ensureKernelFile(ctx context.Context, instanceType, imageRef string) error {
-	kernelPath := pmem.GetRawKernelFilePath(instanceType, imageRef)
-	exist, err := utils.FileExistAndValid(kernelPath)
-	if err != nil {
-		log.G(ctx).Warnf("kernel file %s validation failed, try copy: %v", kernelPath, err)
-	}
-	if exist {
-		return nil
-	}
-	sharedKernelPath := pmem.GetSharedKernelFilePath()
-	sharedExist, err := utils.FileExistAndValid(sharedKernelPath)
-	if err != nil {
-		return fmt.Errorf("local shared kernel validation failed: %w", err)
-	}
-	if !sharedExist {
-		return fmt.Errorf("local shared kernel not found: %s", sharedKernelPath)
-	}
-	if err := copyLocalKernelAtomically(ctx, sharedKernelPath, kernelPath); err != nil {
-		return err
-	}
-	exist, err = utils.FileExistAndValid(kernelPath)
-	if err != nil {
-		return fmt.Errorf("copied kernel file %s validation failed: %v", kernelPath, err)
-	}
-	if !exist {
-		return fmt.Errorf("copied kernel file %s not exist", kernelPath)
-	}
-	return nil
+	return pmem.SyncKernelFile(ctx, pmem.GetSharedKernelFilePath(), pmem.GetRawKernelFilePath(instanceType, imageRef))
 }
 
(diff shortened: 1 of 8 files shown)


dushulin:
信件已收到      【自动回复】

fslongjin:
> 信件已收到      【自动回复】

@dushulin Hi, since you are watching this repository, could you please turn off your email auto-reply to avoid generating unnecessary notifications/comments?

PR #58, opened 2026-04-23
fix(one-click): support non-eth0 NIC and stable CubeProxy CAROOT

### LINKED ISSUE #15: [Bug Report] in one-click installation, the network-agent process got killed
I follow the quick start guide and use the one click install command, but i awlays got this installation error.

Image
and the network-agent.log is like this 

```
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x28 pc=0xa5ac04]

goroutine 1 [running]:
main.main()
        /workspace/network-agent/cmd/network-agent/main.go:113 +0xcc4
```
Could you guys help me to solve this error, thanks

chux0519:
## Summary

This PR improves one-click deployment robustness on cloud hosts where the primary NIC is not `eth0`, and makes CubeProxy certificate generation deterministic across restarts.

## Motivation

In AWS and other environments, the primary interface is often `ens5`/`enp*` instead of `eth0`.
The current one-click flow may assume `eth0`, which can lead to incorrect node networking config.

Also, CubeProxy cert generation relied on implicit mkcert CA location. Setting a stable `CAROOT` makes behavior more predictable in automated deployments.

## Changes

1. Add primary NIC detection in `deploy/one-click/lib/common.sh`
   - New `detect_primary_interface()`:
      - Prefer `CUBE_SANDBOX_ETH_NAME` when explicitly provided
      - Otherwise use default route interface
      - Fallback to first non-loopback up interface
   - Update `detect_node_ip()` to detect IPv4 from the detected primary NIC first, then fallback to first global IPv4.

2. Wire NIC detection into install flow in `deploy/one-click/install.sh`
   - Detect/export `CUBE_SANDBOX_ETH_NAME`
   - Persist it into one-click runtime env file
   - Patch `Cubelet/config/config.toml` (`eth_name = ...`) during install when value is available

3. Stabilize mkcert CA root in `deploy/one-click/scripts/one-click/up-cube-proxy.sh`
   - Set/export `CAROOT` default to `${TOOLBOX_ROOT}/cubeproxy/caroot`
   - Ensure `CAROOT` directory exists before cert generation

## Validation

- `bash -n` passed for all modified scripts
- Function smoke test:
   - `detect_primary_interface` returns host primary NIC
   - `detect_node_ip` returns expected IPv4

## Compatibility

- Backward compatible for environments already using `eth0`
- Users can still override NIC explicitly via `CUBE_SANDBOX_ETH_NAME`

### DIFF
--- a/deploy/one-click/lib/common.sh
+++ b/deploy/one-click/lib/common.sh
@@ -243,17 +243,51 @@ detect_node_ip() {
 
   local detected_ip=""
   if command -v ip >/dev/null 2>&1; then
-    detected_ip="$(ip -4 addr show dev eth0 2>/dev/null \
+    local detected_iface
+    detected_iface="$(detect_primary_interface || true)"
+    if [[ -n "${detected_iface}" ]]; then
+      detected_ip="$(ip -4 addr show dev "${detected_iface}" 2>/dev/null \
+        | grep -oP 'inet \K[0-9.]+' | head -1 || true)"
+      if [[ -n "${detected_ip}" ]]; then
+        log "auto-detected node IP from ${detected_iface}: ${detected_ip}"
+        printf '%s\n' "${detected_ip}"
+        return 0
+      fi
+    fi
+
+    detected_ip="$(ip -4 addr show scope global 2>/dev/null \
       | grep -oP 'inet \K[0-9.]+' | head -1 || true)"
   fi
 
   if [[ -n "${detected_ip}" ]]; then
-    log "auto-detected node IP from eth0: ${detected_ip}"
+    log "auto-detected node IP from first global IPv4 address: ${detected_ip}"
     printf '%s\n' "${detected_ip}"
     return 0
   fi
 
-  die "cannot auto-detect node IP (eth0 not found or has no IPv4). Please set CUBE_SANDBOX_NODE_IP or pass --node-ip=<ip>"
+  die "cannot auto-detect node IP. Please set CUBE_SANDBOX_NODE_IP or pass --node-ip=<ip>"
+}
+
+detect_primary_interface() {
+  if [[ -n "${CUBE_SANDBOX_ETH_NAME:-}" ]]; then
+    printf '%s\n' "${CUBE_SANDBOX_ETH_NAME}"
+    return 0
+  fi
+
+  command -v ip >/dev/null 2>&1 || return 1
+
+  local iface
+  iface="$(ip -o -4 route show to default 2>/dev/null | awk '{print $5; exit}')"
+  if [[ -n "${iface}" ]]; then
+    printf '%s\n' "${iface}"
+    return 0
+  fi
+
+  iface="$(ip -o link show up 2>/dev/null \
+    | awk -F': ' '$2 != "lo" {print $2; exit}' \
+    | cut -d@ -f1)"
+  [[ -n "${iface}" ]] || return 1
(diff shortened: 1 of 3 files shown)


deploy/one-click/scripts/one-click/up-cube-proxy.sh
```diff
@@ -33,6 +33,8 @@ CUBE_PROXY_REDIS_IP="${CUBE_PROXY_REDIS_IP:-${CUBE_SANDBOX_NODE_IP}}"
 CUBE_PROXY_REDIS_PORT="${CUBE_PROXY_REDIS_PORT:-${CUBE_SANDBOX_REDIS_PORT:-6379}}"
 CUBE_PROXY_REDIS_PASSWORD="${CUBE_PROXY_REDIS_PASSWORD:-${CUBE_SANDBOX_REDIS_PASSWORD:-ceuhvu123}}"
 MKCERT_BUNDLED_BIN="${TOOLBOX_ROOT}/support/bin/mkcert"
+CAROOT="${CAROOT:-${TOOLBOX_ROOT}/cubeproxy/caroot}"
```
chenhengqi, line 36: Why do we need a `CAROOT` dir ? Where will it be used ?

deploy/one-click/lib/common.sh
```diff
@@ -243,17 +243,51 @@ detect_node_ip() {
...
+    log "auto-detected node IP from first global IPv4 address: ${detected_ip}"
     printf '%s\n' "${detected_ip}"
     return 0
   fi
 
-  die "cannot auto-detect node IP (eth0 not found or has no IPv4). Please set CUBE_SANDBOX_NODE_IP or pass --node-ip=<ip>"
+  die "cannot auto-detect node IP. Please set CUBE_SANDBOX_NODE_IP or pass --node-ip=<ip>"
+}
+
+detect_primary_interface() {
```
chenhengqi, line 271: Please add comments for this function.

fslongjin (changes requested):
Thanks for the contribution~ Just a tiny change based on the bot's note is all that's needed

deploy/one-click/scripts/one-click/up-cube-proxy.sh
chux0519: CAROOT was added because we hit real TLS failures when using CubeProxy’s mkcert certs from SDK clients.

Observed error on client side:
ssl.SSLCertVerificationError: [SSL: CERTIFICATE_VERIFY_FAILED] unable to get local issuer certificate
(when calling https://<sandbox-id>.cube.app).

Root cause:

- one-click script runs as root (require_root) and mkcert otherwise writes CA to a user-dependent default path,
- clients then often load SSL_CERT_FILE / NODE_EXTRA_CA_CERTS from a different user context,
- resulting in loading the wrong rootCA.pem.

So we pin:
CAROOT="${TOOLBOX_ROOT}/cubeproxy/caroot"
to make CA location deterministic and reproducible, and to make the correct rootCA.pem discoverable for client trust setup.

It is only used by mkcert during cert generation; Cube runtime does not directly consume CAROOT.

chenhengqi: I see, this will be consumed by the `mkcert` tool.

fslongjin (approved):

PR #59, opened 2026-04-23
fix(CubeMaster): align hostdir-mount annotation key with CubeAPI

dushulin:
CubeAPI (handlers/sandboxes.rs, const HOSTDIR_MOUNT_KEY) lifts the caller's metadata["host-mount"] onto the CreateSandbox annotation map under the key "host-mount". CubeMaster's injectHostDirMounts was reading "hostdir-mount" instead, so every host-mount request was silently dropped (only logged as "annotation absent or empty, skip").

Switch the constant to "host-mount" to match CubeAPI, and note the cross-component invariant in a comment so future renames keep the two sides in lockstep.

### DIFF
--- a/CubeMaster/pkg/service/sandbox/hostdir_mount.go
+++ b/CubeMaster/pkg/service/sandbox/hostdir_mount.go
@@ -16,7 +16,12 @@ import (
 )
 
 const (
-	AnnotationHostDirMount = "hostdir-mount"
+	// AnnotationHostDirMount must match the annotation key that CubeAPI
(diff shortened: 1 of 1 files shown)


dushulin:
cc @fslongjin help review, thanks
btw This pipeline seems to have failed. Could it be due to my authority?

chenhengqi (approved):
Good catch. Thanks.

fslongjin:
> cc @fslongjin help review, thanks btw This pipeline seems to have failed. Could it be due to my authority?

I'll fix the workflow later~

fslongjin (approved):

PR #60, opened 2026-04-23
fix(Cubelet): propagate host mount events into cubelet's mount namespace

dushulin:
Previously, volumes mounted on the host after cubelet had started were invisible inside cubelet's persisted mount namespace. The first sandbox requesting such a path failed with
  prepareHostDirVolume: bind mount /mnt/<vol> -> /data/cubelet/hostdir/…:
  no such file or directory
and the only known recovery was to drain sandboxes, unmount the persisted ns file, and restart cubelet so a fresh ns picked up the host's current mount tree.

Root cause: in newCubeMnt, the sleep placeholder was spawned with both Cloneflags: CLONE_NEWNS and Unshareflags: CLONE_NEWNS. Go's stdlib applies an automatic mount("none", "/", MS_REC|MS_PRIVATE) after the unshare step, which detached the new ns's root from the host's shared peer group. Once private, subsequent `mount --make-rslave /` is a no-op because there is no peer group to slave to.

Fix:
  * Drop Unshareflags, keep only Cloneflags. CLONE_NEWNS via clone() preserves the shared-propagation peer group that the kernel copies from the parent ns, so the new ns's "/" stays in shared:1.
  * Add `mount --make-rslave /` in the new ns (via nsenter) before persisting it. The root becomes a one-way slave of the host's "/" peer group, so subsequent host mounts propagate in, while mounts the shim performs inside this ns do not flow back out.

Verified on 4-node cluster:
  - cubelet ns /proc/self/mountinfo shows `/` with `master:1`
  - `mount -t tmpfs tmpfs /mnt/new` on host becomes instantly visible in cubelet ns with real content
  - sandbox with metadata[host-mount] pointing at a post-start mount succeeds and binds the real overlay, not the empty underlying dir
  - existing JFS mount / sandbox flow regressed clean

### DIFF
--- a/Cubelet/cmd/cubelet/main.go
+++ b/Cubelet/cmd/cubelet/main.go
@@ -161,9 +161,16 @@ func newCubeMnt() error {
 		cmd.Stdin = os.Stdin
 		cmd.Stdout = os.Stdout
 		cmd.Stderr = os.Stderr
+		// Use Cloneflags only (not Unshareflags) so that Go does not run
+		// its automatic post-unshare mount("none", "/", MS_REC|MS_PRIVATE)
+		// (see runtime.forkAndExecInChild1 in the Go stdlib). CLONE_NEWNS
+		// in Cloneflags still gives us a fresh mount namespace, and the
+		// kernel preserves the shared-propagation peer group relationship
+		// for mounts copied from the parent. That is a prerequisite for
+		// the subsequent `mount --make-rslave /` to slave this ns's root
+		// back to the host's shared group.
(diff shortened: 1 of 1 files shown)


ls-ggg:
@dushulin Thanks for the fix，could you please add a  ⁠Signed-off-by⁠  line to your commit message as required by our CONTRIBUTING.md

dushulin:
> @dushulin Thanks for the fix，could you please add a ⁠Signed-off-by⁠ line to your commit message as required by our CONTRIBUTING.md

I have added sign, plz review again, thanks

PR #70, opened 2026-04-23
cubeapi: map missing sandbox template to bad request on create

### LINKED ISSUE #10: [Bug Report] CubeAPI returns HTTP 500 instead of 400/404 when creating sandbox with non-existent template
# CubeAPI returns HTTP 500 instead of 400/404 when creating sandbox with non-existent template

## Summary

When creating a sandbox via the CubeAPI (E2B-compatible layer) with a template ID that does not exist, the API returns **HTTP 500 Internal Server Error**. It should return **HTTP 400 Bad Request** or **HTTP 404 Not Found** instead.

## Root Cause

The error flow involves two layers:

1. **CubeMaster** (internal API): When a sandbox is created with a non-existent template, `GetDefinition()` in `pkg/templatecenter/store.go` catches `gorm.ErrRecordNotFound` and returns `ErrTemplateNotFound`. The `createSandbox` handler in `pkg/service/httpservice/cube/sandbox_create.go` wraps this as error code `130400` (`ErrorCode_MasterParamsError`) with message `"failed to get template param from store: template not found"`.

2. **CubeAPI** (Rust/axum gateway): In `src/handlers/sandboxes.rs` (around line 341-347), the `create_sandbox` handler converts all CubeMaster errors uniformly into `AppError::Internal`:

   ```rust
   resp.ret
       .into_result()
       .map_err(|e| AppError::Internal(anyhow::anyhow!(e.to_string())))?;
   ```

   The `is_not_found()` check in `CubeMasterError` only matches `ret_code == 130404`, but template-not-found during sandbox creation comes back as `130400`. So it falls through to `AppError::Internal`, which maps to HTTP 500.

## Comparison with other endpoints

For direct template operations (GET/DELETE `/cube/template`), CubeMaster correctly returns `130404` (`ErrorCode_NotFound`) when the template doesn't exist. CubeAPI recognizes `130404` via `is_not_found()` and returns HTTP 404. Only the sandbox creation path uses `130400` for this case, causing the inconsistency.

| Scenario | CubeMaster `ret_code` | CubeAPI HTTP Status | Expected HTTP Status |
|---|---|---|---|
| GET/DELETE template (not found) | `130404` | 404 | 404 ✅ |
| POST sandbox with non-existent template | `130400` | **500** | **400 or 404** ❌ |

## Suggested Fix

**Option A (CubeAPI side):** In `create_sandbox` in `src/handlers/sandboxes.rs`, check for `130400` or inspect the error message for "template not found" and return `AppError::BadRequest` or `AppError::NotFound` accordingly.

**Option B (CubeMaster side):** In `createSandbox` → `dealCubeboxCreateReqWithTemplate` in `pkg/service/httpservice/cube/sandbox_create.go`, detect `ErrTemplateNotFound` specifically and return `130404` instead of `130400`, making it consistent with other template endpoints.

**Option C (both):** Apply Option B for consistency at the CubeMaster layer, and also add `is_not_found()` handling in CubeAPI's `create_sandbox` as a defense-in-depth measure.

## Affected Files

- `CubeMaster/pkg/service/httpservice/cube/sandbox_create.go`
- `CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go`
- `CubeAPI/src/handlers/sandboxes.rs`
- `CubeAPI/src/error/mod.rs`

YangYuS8:
## Summary
- return a proper not-found error code from CubeMaster when sandbox creation references a missing template
- remove the previous CubeAPI-side string matching approach
- map the missing-template create-sandbox error according to the E2B API behavior for `POST /sandboxes`
## Problem
When creating a sandbox with a non-existent template, CubeMaster previously returned a params error and CubeAPI ended up returning an incorrect HTTP response.
## Fix
- classify the missing-template case in CubeMaster using the proper error code
- keep other template resolution failures as params errors
- update CubeAPI `create_sandbox` to rely on CubeMaster error codes instead of string matching
- map the `POST /sandboxes` missing-template case according to the E2B spec for this endpoint
Closes #10
## Validation
- `cd CubeAPI && cargo +1.88.0 test`
- `cd CubeAPI && cargo +1.88.0 check`
- `cd CubeMaster && go test ./pkg/templatecenter -run TestGetTemplateRequestAssignsRuntimeRequestID`
## Notes
- This PR intentionally keeps the fix scoped to sandbox creation.
- Existing unrelated test failures in `CubeMaster/pkg/service/httpservice/cube` were not expanded beyond the minimum necessary scope.

### DIFF
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -340,11 +340,21 @@ pub async fn create_sandbox(
 
     let resp = state.cubemaster.create_sandbox(&req).await.map_err(|e| {
         tracing::error!(error = %e, "create_sandbox: cubemaster error");
-        AppError::Internal(anyhow::anyhow!(e.to_string()))
+        if e.is_not_found() {
+            AppError::BadRequest(format!("template {} not found", body.template_id))
+        } else {
+            AppError::Internal(anyhow::anyhow!(e.to_string()))
+        }
     })?;
     resp.ret
         .into_result()
-        .map_err(|e| AppError::Internal(anyhow::anyhow!(e.to_string())))?;
+        .map_err(|e| {
+            if e.is_not_found() {
+                AppError::BadRequest(format!("template {} not found", body.template_id))
+            } else {
+                AppError::Internal(anyhow::anyhow!(e.to_string()))
+            }
+        })?;
 
     let sandbox_id = resp.sandbox_id.clone();
 
--- a/CubeMaster/pkg/service/httpservice/cube/sandbox_create.go
+++ b/CubeMaster/pkg/service/httpservice/cube/sandbox_create.go
@@ -16,7 +16,13 @@ import (
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/common"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/templatecenter"
 	"github.com/tencentcloud/CubeSandbox/cubelog"
+)
+
+var (
+	createSandboxDealCubeboxCreateReqWithTemplateFn = dealCubeboxCreateReqWithTemplate
+	createSandboxRunFn                              = sandbox.CreateSandbox
 )
 
 func createSandbox(w http.ResponseWriter, r *http.Request, rt *CubeLog.RequestTrace) interface{} {
@@ -44,16 +50,20 @@ func createSandbox(w http.ResponseWriter, r *http.Request, rt *CubeLog.RequestTr
 		"InstanceType": req.InstanceType,
 	}))
 
-	if err := dealCubeboxCreateReqWithTemplate(ctx, req); err != nil {
-		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
+	if err := createSandboxDealCubeboxCreateReqWithTemplateFn(ctx, req); err != nil {
+		retCode := errorcode.ErrorCode_MasterParamsError
+		if errors.Is(err, templatecenter.ErrTemplateNotFound) {
+			retCode = errorcode.ErrorCode_NotFound
+		}
+		rsp.Ret.RetCode = int(retCode)
 		rsp.Ret.RetMsg = err.Error()
-		rt.RetCode = int64(errorcode.ErrorCode_MasterParamsError)
+		rt.RetCode = int64(retCode)
 		log.G(ctx).Error(err)
 		return rsp
 	}
 
 	ctx = runInsReq2Affinity(ctx, req)
-	ret := sandbox.CreateSandbox(ctx, req)
+	ret := createSandboxRunFn(ctx, req)
 	rt.RetCode = int64(ret.Ret.RetCode)
 	return ret
 }
--- a/CubeMaster/pkg/service/httpservice/cube/sandbox_preview_test.go
+++ b/CubeMaster/pkg/service/httpservice/cube/sandbox_preview_test.go
@@ -16,6 +16,7 @@ import (
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+	CubeLog "github.com/tencentcloud/CubeSandbox/cubelog"
 )
 
 func TestPreviewSandboxReturnsResolvedRequests(t *testing.T) {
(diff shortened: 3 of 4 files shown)


YangYuS8:
The failed GitHub Action appears to come from the repository’s automated Claude review workflow permission check for external contributors. The log shows my actor permission is `read`, and the action exits with `Actor does not have write permissions to the repository`. The code change itself was validated locally with `cargo +1.88.0 test` and `cargo +1.88.0 check`.

CubeAPI/src/cubemaster/mod.rs
```diff
@@ -226,6 +226,16 @@ impl CubeMasterError {
         }
     }
 
+    /// True when sandbox creation failed because the referenced template does not exist.
+    pub fn is_template_not_found(&self) -> bool {
+        match self {
+            Self::Api { ret_code, ret_msg } => {
+                *ret_code == 130400 && ret_msg.to_lowercase().contains("template not found")
+            }
```
fslongjin: This is poorly written. The corresponding error codes should be returned in Cubemaster, rather than using string matching to determine them.

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -340,11 +340,21 @@ pub async fn create_sandbox(
...
     resp.ret
         .into_result()
-        .map_err(|e| AppError::Internal(anyhow::anyhow!(e.to_string())))?;
+        .map_err(|e| {
+            if e.is_template_not_found() {
+                AppError::NotFound(format!("template {} not found", body.template_id))
+            } else {
+                AppError::Internal(anyhow::anyhow!(e.to_string()))
+            }
+        })?;
```
fslongjin: Perhaps the return value should adhere to the conventions specified in https://github.com/e2b-dev/E2B/blob/main/spec/openapi.yml#L1949 ?

fslongjin (changes requested):

YangYuS8:
Addressed the requested changes in the latest update:

* moved the missing-template error classification into CubeMaster so CubeAPI no longer relies on string matching
* kept CubeAPI mapped by CubeMaster error codes
* aligned `POST /sandboxes` to return `400 Bad Request` for the missing-template case based on the E2B spec linked in review
* updated the PR title/body to match the current implementation

Validation run locally:

* `cd CubeAPI && cargo +1.88.0 test`
* `cd CubeAPI && cargo +1.88.0 check`

Please take another look when convenient. Thanks!

chenhengqi (changes requested):
* Please organize you commits logically, one commit per component.
* Please follow our AI-Generated Code Policy in contributing guide.

YangYuS8:
I reorganized the PR into two logical commits, one for CubeMaster and one for CubeAPI. I reviewed the AI-assisted changes myself and added my own Signed-off-by trailers according to the project’s AI-Generated Code Policy and DCO requirements.

CubeMaster/pkg/service/httpservice/cube/sandbox_preview_test.go
```diff
@@ -16,6 +16,7 @@ import (
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+	CubeLog "github.com/tencentcloud/CubeSandbox/cubelog"
```
chenhengqi, line 19: Why do we need this change?

CubeMaster/pkg/service/httpservice/cube/sandbox_create.go
```diff
@@ -16,9 +16,13 @@ import (
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/common"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/templatecenter"
 	"github.com/tencentcloud/CubeSandbox/cubelog"
 )
 
+var createSandboxDealCubeboxCreateReqWithTemplateFn = dealCubeboxCreateReqWithTemplate
```
chenhengqi: Please group variables:
```
var (
...
)
```

CubeMaster/pkg/service/httpservice/cube/sandbox_preview_test.go
YangYuS8: This import is needed because `sandbox_preview_test.go` already uses `CubeLog.RequestTrace{}` in existing tests. Without it, the `cube` package tests fail to compile before reaching the new sandbox creation tests. I kept the change minimal to only add the missing import.

CubeMaster/pkg/service/httpservice/cube/sandbox_create.go
YangYuS8: Updated. I grouped the test hook variables into a single `var (...)` block.

YangYuS8:
Hi @chenhengqi, gentle ping on this PR.

I’ve addressed the latest review comments and kept the PR organized as two signed-off commits, one for CubeMaster and one for CubeAPI. Please let me know if there is anything else I should adjust. Thanks!

PR #84, opened 2026-04-24
CubeProxy: switch apk mirror to tencentyun and remove debug packages

staryxchen:
## Problem

`apk add` in the Dockerfile uses the default upstream mirror
(dl-cdn.alpinelinux.org), which is unreliable from within the Tencent
Cloud build environment and frequently causes image build failures.

## Changes

- **Switch apk mirror**: replace `dl-cdn.alpinelinux.org` with
  `mirrors.tencentyun.com` (Tencent Cloud internal mirror, same network
  as the image registry)
- **Remove debug-only packages**: `tcpdump`, `vim`, `python3`, `py3-pip`,
  `bpftrace`, `ethtool`, `file`, `gawk`, `inetutils-telnet`, `iperf`,
  `iperf3`, `iproute2`, `netcat-openbsd`, `net-tools`, `sysstat`, and
  pip packages `pythonping` / `requests`
  - Verified: none of these are referenced in `start.sh`, `rotate_nginx_log.sh`,
    any `.lua` file, or `nginx.conf`
- **Keep**: `tzdata` (required for timezone setup), `curl`, `wget`
- **Add `--no-cache`** to avoid stale index artifacts in the image layer

## Impact

- Eliminates the network-related build failure
- Reduces image size significantly (removes Python runtime + ~15 debug tools)
- Reduces attack surface in the production image

### DIFF
--- a/CubeProxy/Dockerfile
+++ b/CubeProxy/Dockerfile
@@ -2,16 +2,10 @@ FROM cube-sandbox-image.tencentcloudcr.com/opensource/openresty:1.21.4.1-6-alpin
 
 LABEL maintainer="staryxchen <staryxchen@tencent.com>"
 
-RUN mkdir -p /data/log/cube-proxy/ \
+RUN sed -i 's/dl-cdn.alpinelinux.org/mirrors.tencentyun.com/g' /etc/apk/repositories \
+    && mkdir -p /data/log/cube-proxy/ \
     && apk update \
-    && apk add tcpdump \
-    && apk add vim \
-    && apk add tzdata \
-    && apk add python3 \
-    && apk add py3-pip \
-    && apk add bpftrace busybox curl ethtool file gawk inetutils-telnet \
-        iperf iperf3 iproute2 netcat-openbsd net-tools sysstat wget \
-    && pip3 install pythonping requests \
+    && apk add --no-cache tzdata curl wget \
     && cp /usr/share/zoneinfo/Asia/Shanghai /etc/localtime \
     && echo "Asia/Shanghai" > /etc/timezone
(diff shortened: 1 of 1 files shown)


staryxchen:
## CI Failure Explanation

The `review` job failed because `claude-code-action` internally checks for write permissions and exits with code 1 when the actor only has `read` access. This causes all external contributors (with `author_association: CONTRIBUTOR` or `NONE`) to see a red CI check, which is misleading.

### Root Cause

The workflow splits jobs using `author_association`, but `CONTRIBUTOR` (has previously submitted a PR) ≠ write permission. The action's internal permission guard then fails loudly.

### Suggested Fix

Change the `review` job condition from:

```yaml
if: github.event.pull_request.author_association != 'FIRST_TIME_CONTRIBUTOR'
```

to only run for actors with actual write access:

```yaml
if: contains(fromJSON('["OWNER", "MEMBER", "COLLABORATOR"]'), github.event.pull_request.author_association)
```

And update `first-time-review` to cover everyone else:

```yaml
if: "!contains(fromJSON('[\"OWNER\", \"MEMBER\", \"COLLABORATOR\"]'), github.event.pull_request.author_association)"
```

This way one job always runs and the other always skips — no external contributor will ever see a spurious CI failure.

fslongjin (approved):

chenhengqi:
We can't assume mirror of Tencent Cloud is better for all users, no?

staryxchen:
> We can't assume mirror of Tencent Cloud is better for all users, no?

Yes, we should provide a way for users experiencing network issues to override the mirror URL.

Issue #95, opened 2026-04-27
cubemastercli: default --probe budget of 30s is too tight for slower hosts


ishswar:
## Summary

When passing `--probe <port>` to `cubemastercli tpl create-from-image`, the generated Probe spec gives the in-guest application **30 seconds** to bind the probe port and answer `/health`. On nested-KVM hosts (and arguably on slow disks / loaded hosts in general) the s6 service tree in `ags-image/sandbox-code:latest` consistently needs longer than 30 s to bring the code-interpreter HTTP server up on `:49999`.

The result on nested KVM:

- `tpl create-from-image --probe 49999` fails at `CREATING_TEMPLATE` with `Get "http://<sandbox-ip>:49999/health": dial tcp ... connect: connection refused`.
- The same command without `--probe` "succeeds", but `cube-shim` takes the template snapshot right after vsock-ready, before the code interpreter has finished starting. Sandboxes restored from that template return `502 Bad Gateway` from cube-proxy on every later `run_code` call, because `:49999` was never bound at snapshot time.

## Where it's set

`CubeMaster/cmd/cubemastercli/commands/cubebox/template.go:1314-19`:

```go
overrides.Probe = &types.Probe{
    ProbeHandler: &types.ProbeHandler{...},
    TimeoutMs:        30000,    // 30s per HTTP attempt
    PeriodMs:         500,      // poll every 500ms
    FailureThreshold: 60,       // 60 consecutive failures -> fail
    SuccessThreshold: 1,
}
```

The wall-clock budget is 60 polls × 500 ms = **30 seconds**. The probe runs immediately after vsock-ready, so the budget has to cover the full s6 service tree boot + the code-interpreter HTTP server binding.

The CLI help text at line 768 also documents `timeout_ms=30000, period_ms=500`, which would need updating if the defaults change.

## Repro

On a nested-KVM dev host (or any host where the s6 tree in `sandbox-code:latest` needs >30 s to come up):

```bash
cubemastercli tpl create-from-image \
  --image ccr.ccs.tencentyun.com/ags-image/sandbox-code:latest \
  --writable-layer-size 1G \
  --expose-port 49999 --expose-port 49983 \
  --probe 49999
```

Phase reaches `CREATING_TEMPLATE` then fails with `connection refused` to the probe port.

## Proposed fix

Bump the defaults to a 120 s budget. PR coming.

[...]

PR #96, opened 2026-04-27
shim: bump VsockServerReady wait from 10s to 60s for nested-KVM hosts

### LINKED ISSUE #94: shim: VsockServerReady wait of 10s is too tight for nested-KVM hosts
## Summary

When running Cube Sandbox on a nested-KVM host (cloud-hypervisor inside a QEMU VM, e.g. the dev-env on a GCP VM with nested virtualization, or any "VM-in-VM" development setup), every `cubemastercli tpl create-from-image` fails with:

```
failed to create shim task: Others("Other: Create sandbox failed: Receive event timeout after 10000ms")
```

even though the microVM kernel inside is booting normally.

## Root cause

`CubeShim/shim/src/sandbox/sb.rs:780` waits up to **10 seconds** for the in-guest agent to send `VsockServerReady` after `start_vm`:

```rust
let ev = ch
    .wait_notify(Duration::from_nanos(1000 * 1000 * 1000 * 10 as u64))
    .await?;
```

Under nested KVM, the agent consistently takes ~18 s to come up because every page fault VM-exits twice. Measurements from `cube-shim-req.log` on a GCP n2 host with nested virtualization (cube-sandbox-one-click v0.1.1):

```
"vm ready, vsock is listening, cost:18152"
"vm ready, vsock is listening, cost:18690"
```

The kernel itself is fine — it just hasn't reached `VsockServerReady` within the 10 s window.

On bare metal this never trips because cold boot is well under 10 s. It's a soft assumption of bare-metal speed that breaks any nested setup.

## Impact

- Anyone trying to evaluate Cube on a dev VM with nested KVM (which is the documented `dev-env/` path for non-bare-metal hosts) is blocked — `tpl create-from-image` is the first command in the quickstart and it always fails.
- The error message is misleading; it suggests the sandbox is broken when in fact it just needs more time.

## Repro

On any nested-KVM host (e.g. a GCP VM with nested virtualization enabled, running the `dev-env/` flow):

```bash
  --writable-layer-size 1G \
  --expose-port 49999 --expose-port 49983 \
  --probe 49999
```

Watch `/data/log/CubeShim/cube-shim-req.log` — kernel boots, virtio devices activate, but the create call fails at the 10 s mark.

## Proposed fix

Bump the timeout to 60 s. PR coming.

ishswar:
## Summary
- raise the cold-boot `wait_notify` timeout in cube-shim from 10s to 60s
- unblock `tpl create-from-image` on nested-KVM dev hosts where the
  in-guest agent legitimately takes ~18s to signal `VsockServerReady`
- no behavioural change for bare-metal callers (cold boot is well
  under 10s there, so the wait still returns as soon as the agent is
  ready)

## Problem
On nested-KVM hosts (cloud-hypervisor running inside a QEMU dev VM,
e.g. on a GCP n2 with nested virtualization), the cold-boot wait at
`CubeShim/shim/src/sandbox/sb.rs:780` is hardcoded to 10 seconds:

```rust
.wait_notify(Duration::from_nanos(1000 * 1000 * 1000 * 10 as u64))
```

Every page fault inside a nested guest VM-exits twice, so a microVM
boot that takes ~1s on bare metal takes ~18s in this environment.
Measurements from `cube-shim-req.log` on a GCP n2:

```
"vm ready, vsock is listening, cost:18152"
"vm ready, vsock is listening, cost:18690"
```

The result: every `cubemastercli tpl create-from-image` fails with

```
failed to create shim task: Others("Other: Create sandbox failed:
Receive event timeout after 10000ms")
```

even though the microVM kernel and agent are starting fine. This blocks
the entire upstream `dev-env/` quickstart flow on nested-KVM dev hosts.

## Fix
Bump the timeout to 60 seconds. The wait still returns as soon as the
agent is ready, so steady-state latency is unchanged on every host.

A nicer follow-up would be to make this configurable (env var like
`CUBE_SHIM_BOOT_TIMEOUT_SECS`, defaulting to 10) so callers can tune
without recompiling. This PR keeps the scope to the minimum-viable fix
that unblocks the dev-env path; happy to do the env-var version as a
follow-up if maintainers prefer.

## Validation
- `cargo fmt --check` clean for the touched lines (the rustfmt diff
  reported on master at line 313 is unrelated to this change).
- `cargo build --release --bin containerd-shim-cube-rs` succeeds.
- `cubemastercli tpl create-from-image` (with `--probe 49999`) reaches
  `READY` consistently after deploying the patched binary, in repeated
  tests on:
    GCP n2-standard-4, nested virtualization enabled,
    OpenCloudOS 9 dev VM (per `dev-env/run_vm.sh`),
    cube-sandbox-one-click v0.1.1.
- Confirmed in `cube-shim-req.log` that the agent reports ready within
  the new window (cost:18152, 18690 ms in our runs).
- End-to-end `e2b-code-interpreter` Python SDK round-trip works after
  the change (warm `run_code` ~80–200 ms).

## Compatibility
- Bare-metal callers see no functional change. Cold boot completes in
  well under the new 60s window, the `wait_notify` returns as soon as
  the agent is ready, and steady-state behavior is identical.
- A buggy/wedged microVM that previously failed-fast at 10s will now
  fail at 60s. In our experience the 10s ceiling tends to mask real
  boot issues (genuinely-slow but eventually-correct boots) as flaky
  template builds rather than producing faster useful failures, so the
  longer ceiling seems like an improvement on slower hosts.

Closes #94

### DIFF
--- a/CubeShim/shim/src/sandbox/sb.rs
+++ b/CubeShim/shim/src/sandbox/sb.rs
@@ -777,7 +777,7 @@ impl SandBox {
             let ch = self.ch.as_mut().unwrap().lock().await;
             let start = Instant::now();
             let ev = ch
-                .wait_notify(Duration::from_nanos(1000 * 1000 * 1000 * 10 as u64))
+                .wait_notify(Duration::from_nanos(1000 * 1000 * 1000 * 60 as u64))
                 .await?;
 
             if CH::NotifyEvent::VsockServerReady != ev {


fslongjin:
Since nested virtualization is already enabled on the GCP n2-standard-4 instance, you simply need to install a system like RHEL, OpenCloudOS, or CentOS directly inside that VM (with XFS as the rootfs), and then proceed with the standard installation process. There is no need to run an additional 'dev-env' virtual machine. The performance issues you're encountering are due to your current setup, where the micro-VM is already at the third level of nested virtualization. I believe the >10s startup time is clearly an installation or configuration issue, rather than a problem with the VM boot speed itself.

ishswar:
Thanks @fslongjin — you were right, and the data backs it up. We re-tested on a fresh setup matching what you described and the boot-time gap is huge.

**Test environment:** GCP `n2-standard-4` with nested virtualization enabled, **Rocky Linux 9.7** (XFS rootfs) installed directly on the VM. Standard `online-install.sh`, **stock v0.1.2 with no patches applied** at all.

**Result:** `cubemastercli tpl create-from-image --probe 49999` (the exact upstream quickstart command) reaches `READY` cleanly. Both upstream defaults (10s shim wait, 30s probe) have plenty of headroom.

**`cost:NNNN` from `cube-shim-req.log`:**

| Setup | Nesting levels | microVM cold-boot |
|---|---|---|
| Original report (dev-env QEMU on Ubuntu host) | 3 | 18,152 ms / 18,690 ms |
| **Rocky direct (your suggestion)** | 2 | **2,067 ms** |

~9× speedup just by removing the dev-env QEMU layer. The dev-env path's third nesting level is what was blowing past the 10s ceiling — not a real upstream bug. Same for issue #95 — at 2 levels the 30s probe budget is plenty too (template reaches READY with `--probe 49999` consistently).

End-to-end e2b SDK round-trip on the Rocky setup:

| Operation | Round-trip |
|---|---|
| `Sandbox.create` (snapshot restore) | ~520 ms |
| First `run_code` (Jupyter cold-start) | ~540 ms |
| Cold pandas import + DataFrame + tables | ~380 ms |
| Warm `run_code` | ~50–80 ms |

So this PR fixes a symptom of an over-nested test environment, not a real problem. Happy to **close this PR** unless you'd like to keep the same change as a small dev-env quality-of-life note (a one-liner in `dev-env/README.md` warning that boot is much slower in nested setups). Let me know which you prefer.

---

**Side note (file separately if useful):** the standard install on a stock Rocky 9 host has a couple of unrelated rough edges that I worked around but didn't surface upstream yet. If you'd like Issues / PRs for them I'm happy to file:

1. `dnf install -y docker docker-compose` in `lib/common.sh::install_docker()` fails on Rocky/RHEL 9 — the `docker-compose` RPM doesn't exist there (Docker deprecated standalone docker-compose mid-2023; the RHEL-9 path is `docker-compose-plugin` invoked as `docker compose`).
2. `up-cube-proxy.sh::install_mkcert` checks `command -v mkcert` after copying to `/usr/local/bin/mkcert`, but Rocky's sudo `secure_path` doesn't include `/usr/local/bin` by default, so the check fails even though the binary is there.

Both are independent of the timeout discussion.

fslongjin:
> **Side note (file separately if useful):** the standard install on a stock Rocky 9 host has a couple of unrelated rough edges that I worked around but didn't surface upstream yet. If you'd like Issues / PRs for them I'm happy to file:
> 
> 1. `dnf install -y docker docker-compose` in `lib/common.sh::install_docker()` fails on Rocky/RHEL 9 — the `docker-compose` RPM doesn't exist there (Docker deprecated standalone docker-compose mid-2023; the RHEL-9 path is `docker-compose-plugin` invoked as `docker compose`).
> 2. `up-cube-proxy.sh::install_mkcert` checks `command -v mkcert` after copying to `/usr/local/bin/mkcert`, but Rocky's sudo `secure_path` doesn't include `/usr/local/bin` by default, so the check fails even though the binary is there.
> 
> Both are independent of the timeout discussion.

Thanks — both issues you pointed out are indeed valid. If you’re interested, feel free to submit a PR! 😊

fslongjin:
@ls-ggg What do you think about increasing the timeout duration? From the perspective of user environment compatibility, this PR seems totally fine to me.

ls-ggg:
@ishswar Thanks for the pr, I think the current 10s timeout should be enough for most cases, except for some extreme edge cases. Just a heads-up: we’re about to release a PVM solution for running the sandbox on regular VMs. Once that's out, we expect the nested-KVM setup won't be the preferred choice for most users anymore.

PR #97, opened 2026-04-27
cubemastercli: extend default --probe budget for slower hosts

### LINKED ISSUE #95: cubemastercli: default --probe budget of 30s is too tight for slower hosts

ishswar:
## Summary
- raise the default `--probe` budget from 30s to 120s in
  `cubemastercli tpl create-from-image`
- update the `--probe` CLI help text to match the new defaults
- unblock the upstream quickstart on nested-KVM dev hosts (and any
  host where the in-guest application needs >30s to bind the probe
  port)
- no behavioural change for hosts where the application binds quickly
  — the probe still succeeds on the first poll and returns

## Problem
The default Probe spec generated by `--probe <port>` at
`CubeMaster/cmd/cubemastercli/commands/cubebox/template.go:1314-19`:

```go
TimeoutMs:        30000   // 30s per HTTP attempt
PeriodMs:           500   // poll every 500ms
FailureThreshold:    60   // 60 consecutive failures -> fail
```

gives the in-guest application 30 seconds wall-clock to bind the probe
port and answer `/health`. The probe runs immediately after the
microVM signals vsock-ready, which means the 30s window has to cover:

  - the rest of microVM boot
  - the entire s6 service tree (dbus, sudo, x11-common, ...)
  - the code-interpreter HTTP server binding `:49999`

On nested-KVM dev hosts (cloud-hypervisor inside a QEMU VM), 30s is
not enough. The upstream quickstart command:

```bash
  --writable-layer-size 1G \
  --expose-port 49999 --expose-port 49983 \
  --probe 49999
```

fails at the `CREATING_TEMPLATE` phase with

```
Get "http://<sandbox-ip>:49999/health": dial tcp ... connect:
connection refused
```

even though the microVM is fine — the code interpreter is still
starting up.

Workaround: omit `--probe`. But then `cube-shim` snapshots the
sandbox right after vsock-ready, before the code interpreter is up.
Sandboxes restored from that template return 502 Bad Gateway from
cube-proxy on every `run_code` call, because the upstream service was
never running when the snapshot was taken.

## Fix
Bump the defaults to a 120s budget:

```go
TimeoutMs:        120000
PeriodMs:         1000
FailureThreshold: 120
```

Update the `--probe` flag's `Usage` text accordingly so `--help`
matches what the code actually does.

A nicer follow-up would be to surface `--probe-timeout-ms`,
`--probe-period-ms`, `--probe-failure-threshold` CLI flags so
callers can tune per call without recompiling. This PR keeps the scope
to adjusting the defaults so the upstream quickstart works as-is on
slower hosts; happy to do the per-call flag version as a follow-up if
maintainers prefer.

## Validation
- `gofmt -l` clean.
- `go vet ./cmd/cubemastercli/...` clean.
- `go build -ldflags="-s -w" -o cubemastercli ./cmd/cubemastercli`
  succeeds.
- After deploying the patched binary on a nested-KVM dev host, the
  upstream `tpl create-from-image --probe 49999` quickstart command
  reaches `READY` cleanly (status: SUCCEEDED, template_status: READY,
  artifact_status: READY).
- Confirmed end-to-end with `e2b-code-interpreter` Python SDK that
  sandboxes spawned from the resulting template have the code
  interpreter up at first call (warm `run_code` ~80–200 ms; no 502s).
- Tested on:
    GCP n2-standard-4, nested virtualization enabled,
    OpenCloudOS 9 dev VM (per `dev-env/run_vm.sh`),
    cube-sandbox-one-click v0.1.1.

## Compatibility
- Hosts where the in-guest application binds the probe port quickly
  see no functional change. The probe succeeds on the first poll; the
  ceiling is irrelevant.
- A genuinely-broken in-guest application (e.g. wrong probe port,
  missing service) now fails after 120s instead of 30s. This is the
  one case where behaviour changes — the failure is delayed by 90s.
  In our experience the 30s ceiling tends to mask "slow but eventually
  correct" boots as flaky template builds rather than producing
  useful fast failures, so the longer ceiling seems like an
  improvement.

Closes #95

### DIFF
--- a/CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
+++ b/CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
@@ -766,8 +766,9 @@ var TemplateCreateFromImageCommand = cli.Command{
 		cli.StringSliceFlag{Name: "arg", Usage: "override container CMD (args); repeat for multiple elements"},
 		cli.StringSliceFlag{Name: "env", Usage: "set environment variable, KEY=VALUE format; repeat for multiple envs"},
 		cli.StringSliceFlag{Name: "dns", Usage: "set container DNS nameserver; repeat for multiple servers"},
-		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); sets timeout_ms=30000, period_ms=500"},
+		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); see --probe-timeout, --probe-path"},
 		cli.StringFlag{Name: "probe-path", Value: "/health", Usage: "HTTP path for the readiness probe (default: /health); only effective when --probe is set"},
+		cli.DurationFlag{Name: "probe-timeout", Value: 120 * time.Second, Usage: "total probe wall-clock budget (e.g. 30s, 2m); only effective when --probe is set; default 120s suits nested-KVM hosts where the in-guest application takes longer than 30s to bind the probe port"},
 		cli.IntFlag{Name: "cpu", Value: 2000, Usage: "CPU millicores for the template container (default: 2000, i.e. 2 cores)"},
 		cli.IntFlag{Name: "memory", Value: 2000, Usage: "Memory for the template container in MB (default: 2000 MB)"},
 		cli.BoolFlag{Name: "json", Usage: "print raw json response"},
@@ -1315,6 +1316,16 @@ func parseContainerOverrides(c *cli.Context) (*types.ContainerOverrides, error) 
 		if probePath == "" {
 			probePath = "/health"
 		}
+		probeTimeout := c.Duration("probe-timeout")
+		if probeTimeout < time.Second {
+			return nil, fmt.Errorf("--probe-timeout must be at least 1s, got %s", probeTimeout)
+		}
+		// PeriodMs is fixed at 1s; FailureThreshold derives from the requested
+		// total budget (rounded up to whole seconds). TimeoutMs == total budget
+		// so it never trips before FailureThreshold does.
+		periodMs := int32(1000)
+		timeoutMs := int32(probeTimeout.Milliseconds())
+		failureThreshold := int32((probeTimeout + time.Second - 1) / time.Second)
 		host := ""
 		overrides.Probe = &types.Probe{
 			ProbeHandler: &types.ProbeHandler{
@@ -1324,9 +1335,9 @@ func parseContainerOverrides(c *cli.Context) (*types.ContainerOverrides, error) 
 					Host: &host,
 				},
 			},
-			TimeoutMs:        30000,
-			PeriodMs:         500,
-			FailureThreshold: 60,
+			TimeoutMs:        timeoutMs,
+			PeriodMs:         periodMs,
+			FailureThreshold: failureThreshold,
 			SuccessThreshold: 1,
 		}
 	}


CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -765,7 +765,7 @@ var TemplateCreateFromImageCommand = cli.Command{
 		cli.StringSliceFlag{Name: "cmd", Usage: "override container ENTRYPOINT (command); repeat for multiple elements, e.g. --cmd /bin/sh --cmd -c"},
 		cli.StringSliceFlag{Name: "arg", Usage: "override container CMD (args); repeat for multiple elements"},
 		cli.StringSliceFlag{Name: "env", Usage: "set environment variable, KEY=VALUE format; repeat for multiple envs"},
-		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); sets timeout_ms=30000, period_ms=500"},
+		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); sets timeout_ms=120000, period_ms=1000, failure_threshold=120"},
```
chenhengqi: How about adding `--probe-timeout` option ?

ishswar:
Thanks @chenhengqi — done. Pushed an amended commit just now that adds `--probe-timeout` taking a Go-duration string (e.g. `30s`, `2m`, `5m`).

Behaviour:

- `--probe 49999` (no `--probe-timeout`) → 120s budget (same default as the previous version of this PR — backward compatible with what reviewers already saw).
- `--probe 49999 --probe-timeout 30s` → matches the original upstream 30s budget for callers on fast hosts who want it back.
- `--probe 49999 --probe-timeout 5m` → useful for very slow / heavily-loaded hosts.

Implementation: `PeriodMs` is fixed at 1000ms; `FailureThreshold = ceil(budget_seconds)`; `TimeoutMs` is set to the total budget so the per-attempt timeout never trips before the failure-count ceiling does. Single user-visible knob → fewer wrong combinations to reason about.

Updated the `--probe` help text to reference `--probe-timeout` instead of baking specific numbers into it (avoids drift between defaults and docs).

Validation steps + test environment in the new commit message. CI should re-run automatically; if anything else looks off let me know.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -765,9 +765,9 @@ var TemplateCreateFromImageCommand = cli.Command{
 		cli.StringSliceFlag{Name: "cmd", Usage: "override container ENTRYPOINT (command); repeat for multiple elements, e.g. --cmd /bin/sh --cmd -c"},
 		cli.StringSliceFlag{Name: "arg", Usage: "override container CMD (args); repeat for multiple elements"},
 		cli.StringSliceFlag{Name: "env", Usage: "set environment variable, KEY=VALUE format; repeat for multiple envs"},
-		cli.StringSliceFlag{Name: "dns", Usage: "set container DNS nameserver; repeat for multiple servers"},
-		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); sets timeout_ms=30000, period_ms=500"},
+		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); see --probe-timeout, --probe-path"},
 		cli.StringFlag{Name: "probe-path", Value: "/health", Usage: "HTTP path for the readiness probe (default: /health); only effective when --probe is set"},
+		cli.StringFlag{Name: "probe-timeout", Value: "120s", Usage: "total probe wall-clock budget as a Go duration (e.g. \"30s\", \"2m\"); only effective when --probe is set; default 120s suits nested-KVM hosts where the in-guest application takes longer than 30s to bind the probe port"},
```
chenhengqi: Use `DurationFlag`?

ishswar:
Done — switched to `cli.DurationFlag` and dropped the manual `time.ParseDuration`. Help text picks up `(default: 2m0s)` from the flag automatically.

ishswar:
Tested the latest revision (`--probe-timeout` as `cli.DurationFlag` per your suggestion) on a 2-level GCP nested-KVM setup — Rocky Linux 9 with XFS, **stock v0.1.2 with this PR applied on top**. The flag works as expected: accepts Go-duration strings (`30s`, `2m`, etc.), defaults to 120s, no behaviour change for callers who don't pass it.

That said — on the 2-level setup the original 30s budget is also plenty (template reaches `READY` even with `--probe-timeout 30s`). So the value here is the configurability rather than the default bump. Up to you whether the additive `--probe-timeout` flag is worth merging on that basis; happy to close if you'd prefer to defer.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -1315,6 +1306,16 @@ func parseContainerOverrides(c *cli.Context) (*types.ContainerOverrides, error)
...
+		probeTimeout := c.Duration("probe-timeout")
+		if probeTimeout < time.Second {
+			return nil, fmt.Errorf("--probe-timeout must be at least 1s, got %s", probeTimeout)
+		}
+		// PeriodMs is fixed at 1s; FailureThreshold derives from the requested
+		// total budget (rounded up to whole seconds). TimeoutMs == total budget
+		// so it never trips before FailureThreshold does.
+		periodMs := int32(1000)
+		timeoutMs := int32(probeTimeout / time.Millisecond)
+		failureThreshold := int32((probeTimeout + time.Second - 1) / time.Second)
```
chenhengqi, line 1328: Use `Milliseconds()` method.

chenhengqi (changes requested):
Please rebase against latest master branch.

ishswar:
Done — rebased onto latest master (resolved the conflict with #98's `--dns` flag; both flags coexist) and switched to `probeTimeout.Milliseconds()` per your suggestion. Build/vet clean, helps text shows the new `--dns` and `--probe-timeout` side by side.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -1315,6 +1316,16 @@ func parseContainerOverrides(c *cli.Context) (*types.ContainerOverrides, error)
...
+		probeTimeout := c.Duration("probe-timeout")
+		if probeTimeout < time.Second {
+			return nil, fmt.Errorf("--probe-timeout must be at least 1s, got %s", probeTimeout)
+		}
+		// PeriodMs is fixed at 1s; FailureThreshold derives from the requested
+		// total budget (rounded up to whole seconds). TimeoutMs == total budget
+		// so it never trips before FailureThreshold does.
+		periodMs := int32(1000)
+		timeoutMs := int32(probeTimeout.Milliseconds())
+		failureThreshold := int32((probeTimeout + time.Second - 1) / time.Second)
```
chenhengqi, line 1328: You've checked probeTimeout above, so you can just do:
```golang
failureThreshold := int32(probeTimeout / time.Second)
```

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -766,8 +766,9 @@ var TemplateCreateFromImageCommand = cli.Command{
 		cli.StringSliceFlag{Name: "arg", Usage: "override container CMD (args); repeat for multiple elements"},
 		cli.StringSliceFlag{Name: "env", Usage: "set environment variable, KEY=VALUE format; repeat for multiple envs"},
 		cli.StringSliceFlag{Name: "dns", Usage: "set container DNS nameserver; repeat for multiple servers"},
-		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); sets timeout_ms=30000, period_ms=500"},
+		cli.IntFlag{Name: "probe", Usage: "enable HTTP GET probe on the specified port (e.g. --probe 9000); see --probe-timeout, --probe-path"},
```
novahe: nit: The flag name `probe` is a bit ambiguous as it doesn't specify what value is expected (it could be interpreted as a boolean). Suggest renaming it to `probe-port` for better clarity, especially since it complements `--probe-timeout` and `--probe-path` quite well.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
ls-ggg: @ishswar Could you please address the suggestions and update the PR?

PR #98, opened 2026-04-27
cubemaster: support dns flag for template image creation

### LINKED ISSUE #12: [Feature Request] No DNS configuration support when creating templates via cubemastercli
## Problem / Motivation

`cubemastercli tpl create-from-image` provides no way to specify DNS configuration (nameservers, search domains, options). In isolated or private-network deployments, this makes it impossible to ensure correct DNS resolution for sandboxes created from such templates — for example, when a readiness probe needs to resolve an internal hostname.

## Proposed Solution

 and add the corresponding CLI flags:
 --dns value         nameserver; repeatable

YangYuS8:
## Summary

* add repeatable `--dns` support to `cubemastercli tpl create-from-image`
* propagate DNS server overrides into generated template container specs
* include DNS configuration in the template spec fingerprint

## Problem

`cubemastercli tpl create-from-image` previously had no way to specify DNS nameservers, even though the underlying container model already supports `DnsConfig`.

This makes it difficult to create templates for environments that require custom DNS resolution, such as private networks or internal service discovery.

## Fix

* add `DnsConfig` to `ContainerOverrides`
* parse repeated `--dns` values in the create-from-image CLI flow
* apply DNS server overrides when generating the template create request
* include DNS config in fingerprint generation so different DNS settings produce different template artifacts

Example:

```bash
  --dns 8.8.8.8 \
  --dns 1.1.1.1
```

Closes #12

## Validation

* `cd CubeMaster && go test ./cmd/cubemastercli/commands/cubebox -run DNS -count=1`
* `cd CubeMaster && go test ./pkg/templatecenter -run DNS -count=1`

## Notes

* This PR only adds nameserver support via `--dns`.
* It intentionally does not add `--dns-search` or `--dns-option` to keep the scope focused.

### DIFF
--- a/CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
+++ b/CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
@@ -765,6 +765,7 @@ var TemplateCreateFromImageCommand = cli.Command{
 		cli.StringSliceFlag{Name: "cmd", Usage: "override container ENTRYPOINT (command); repeat for multiple elements, e.g. --cmd /bin/sh --cmd -c"},
 		cli.StringSliceFlag{Name: "arg", Usage: "override container CMD (args); repeat for multiple elements"},
 		cli.StringSliceFlag{Name: "env", Usage: "set environment variable, KEY=VALUE format; repeat for multiple envs"},
+		cli.StringSliceFlag{Name: "dns", Usage: "set container DNS nameserver; repeat for multiple servers"},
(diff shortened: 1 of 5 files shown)


YangYuS8:
Thanks for the review. I addressed the actionable suggestions:

- added IP validation for repeated "--dns" values using "net.ParseIP"
- added tests for invalid DNS input and the no-DNS case
- expanded the fingerprint test to cover DNS vs no-DNS
- kept "dnsConfigOrNil" unchanged for consistency with the existing helper pattern

PR #107, opened 2026-04-27
migrate direct jsoniter usage to encoding/json

dkshitij29:
## Summary
- Replace direct `github.com/json-iterator/go` usage in `.go` source with `encoding/json`
- Update marshal/unmarshal callsites and byte-to-string boundaries where required
- Keep behavior equivalent while removing jsoniter-specific source usage
## Validation
- `git grep -nE 'github.com/json-iterator/go|jsoniter\.' -- '*.go'` returns no matches
- Linux container compile checks passed for key changed packages
- `pkg/store/cubebox` delete-container test failures were reproduced on clean `origin/master` baseline (pre-existing)


chenhengqi (changes requested):
Please read and follow our contribution guide.

dkshitij29:
Addressed based on feedback and aligned this PR with CONTRIBUTING guidance.

- Source migration is complete: direct `github.com/json-iterator/go` / `jsoniter.` usage in Go sources has been replaced with `encoding/json` calls (no alias indirection).
- All touched Go files were formatted with `gofmt`.
- Commit history was cleaned to remove intermediate/noise work so the branch now contains a single focused migration commit.

Targeted validation performed:

- `git grep -nE 'github.com/json-iterator/go|jsoniter\.' -- '*.go'` -> no matches.
- Linux container compile checks completed successfully for key changed packages.

Note on failing tests:

- `Cubelet/pkg/store/cubebox` failures in `TestStoreDeleteContainer*` were reproduced from clean `origin/master` in a separate worktree and Linux container, indicating this behavior is pre-existing and not introduced by this migration.
- Some storage tests can fail in unprivileged containers due to mount permission limits (`Operation not permitted`) unless a privileged runtime is used.

Please let me know if you want this split further into component-scoped commits.

dkshitij29:
Reopened after branch ref restore/rename issue. Latest migration commits are unchanged; requesting re-review.

Cubelet/cmd/cubecli/commands/cubebox/create.go
```diff
@@ -60,7 +59,8 @@ var Create = &cli.Command{
 				myPrint("create failure:%v", err)
 				os.Exit(1)
 			}
-			respStr, err := jsoniter.MarshalToString(rsp)
+
+			respStr, err := json.Marshal(rsp)
```
chenhengqi: Note that `json.Marshal()` returns a byte slice instead of string.

chenhengqi (changes requested):
Please make sure you review the code that AI generated before open a PR. Thanks.

Cubelet/cmd/cubecli/commands/cubebox/create.go
dkshitij29: Nice catch, Will need to work on full pr again to check if there are any other mistakes.

I have declared a new valriable called  "respBytes". 
respBytes, err := json.Marshal(rsp)
//error handling 
respStr := string(respBytes) // converting it back to str.

Hope that works. New to go. Typical beginner mistakes.

Cubelet/cmd/cubecli/commands/unsafe/volumedb.go
```diff
@@ -87,7 +87,7 @@ var volumedb = &cli.Command{
 		allSandboxVolumes := map[string]*createInfo{}
 		for k, v := range all {
 			bf := &createInfo{}
-			err = jsoniter.ConfigFastest.Unmarshal(v, bf)
+			err = json.Unmarshal(v, bf)
```
dkshitij29, line 91: Hey @chenhengqi. 
Had a query on this the old alias call had jsoniter.ConfigFastest.Unmarshal(v, bf) ->  json.Unmarshal(v, bf).

Will we run into any performance or memory allocation issues with increased object sizes in this specific path? Based on docs, ConfigFastest skips strict validations to prioritize speed, so I want to ensure the standard library will still meet our throughput requirements for these payloads.

Cubelet/pkg/apis/shimapi/devices.go
```diff
@@ -47,7 +48,7 @@ func (csc *cubeShimControl) AddDevices(ctx context.Context, devices []*shimtypes
 			API_UPDATE_ACTION_KEY: API_ACTION_HOT_PLUGIN_ADD,
 		})
 
-		b, err := jsoniter.MarshalToString(toAppendDevices)
+		b, err := json.Marshal(toAppendDevices)
```
dkshitij29, line 50: String conversion not required here:
there is a inline string conversion method called in the log entry. below the err handing.
logEntry = logEntry.WithField(API_UPDATE_ACTION_DATA_KEY, string(b))

CubeMaster/pkg/service/sandbox/types/types.go
```diff
@@ -553,12 +554,19 @@ const (
 	StartTime contextKey = "startTime"
 )
 
-var FastestJsoniter = jsoniter.Config{
```
dkshitij29: Using a custom struct block as a replacement here(AI suggested). @chenhengqi do you have any better suggestions or this is fine, wanted your feedback? 
func (stdJSONTool) Marshal(v interface{}) ([]byte, error) {
	var b bytes.Buffer
	enc := json.NewEncoder(&b)
	enc.SetEscapeHTML(false)
	if err := enc.Encode(v); err != nil {
		return nil, err
	}
	out := b.Bytes()
	if n := len(out); n > 0 && out[n-1] == '\n' {
		out = out[:n-1]
	}
	return out, nil
}

func (stdJSONTool) Unmarshal(data []byte, v interface{}) error {
	dec := json.NewDecoder(bytes.NewReader(data))
	dec.UseNumber()
	return dec.Decode(v)
}

var FastestJsoniter = stdJSONTool{}

PR #125, opened 2026-04-29
cubemastercli: Refactor CubeVS params parsing logic

chenhengqi:
Let's rely on urfave/cli's functionalities instead of reinvent custom parsing logic. With this change, we allow users to specify the following options:

  --allow-internet-access
  --allow-internet-access=false
  --allow-internet-access=true
  --allow-out-cidr "10.0.0.0/8"
  --deny-out-cidr "10.0.0.0/8"

But not:

  --allow-internet-access false
  --allow-internet-access true


fslongjin:
I found one blocking issue that should be addressed before merge.

### Blocking

`tpl create-from-image` regresses the previously supported `--allow-internet-access false` usage, and this also causes the following CIDR flags to stop being parsed.

- Relevant code: `CubeMaster/cmd/cubemastercli/commands/cubebox/net.go:16`, `CubeMaster/cmd/cubemastercli/commands/cubebox/template.go:644`
- This PR replaces the dedicated `mergeCreateFromImageCubeVSContextFlags(...)` path with the shared `parseCubeVSParams(...)` path for `create-from-image`
- However, this command still uses `urfave/cli` v1 `BoolFlag`. In that library, `--allow-internet-access false` does not parse `false` as the flag value; `false` becomes a positional argument instead
- As a result:
  - `c.Bool("allow-internet-access")` becomes `true`, which is the opposite of what the user explicitly requested
  - Once `flag.FlagSet` hits that positional argument, it stops parsing the remaining flags, so `--allow-out-cidr` / `--deny-out-cidr` are also dropped from the final request

I verified this locally with a minimal reproduction:

```text
--allow-internet-access false
=> IsSet=true, Bool=true, remaining arg0="false"

--allow-internet-access false --allow-out-cidr 172.16.0.0/16
=> Bool=true, allowOut=[]
```

This is a behavior regression. The previous implementation had `parseCreateFromImageExtraCubeVSFlags(...)` specifically to handle this case, and this PR removes that compatibility path. The related tests were removed at the same time, so the current test pass does not cover this regression.

Suggested fix:

1. Restore compatibility for `create-from-image` with `--allow-internet-access false`, or at minimum fail explicitly in the CLI instead of silently interpreting `false` as `true`
2. Add regression tests for this case, especially when combined with `--allow-out-cidr` / `--deny-out-cidr`

I do not recommend merging this as-is.

PR #127, opened 2026-04-30
fix(cube-master): panic if http server fails to start

yyyang91:
- Fix the conditional logic for port availability detection.
- Add explicit success logs to print the listening address upon startup.
- Add detailed error logs when the HTTP server fails to start.
- Enhance service observability for better troubleshooting.

### DIFF
--- a/CubeMaster/pkg/server/server.go
+++ b/CubeMaster/pkg/server/server.go
@@ -8,6 +8,8 @@ package server
 import (
 	"context"
 	"fmt"
+	stdlog "log"
+	"net"
 	"net/http"
 	"os"
 	"sync"
@@ -23,18 +25,19 @@ import (
 	metahttp "github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/meta"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/middleware"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/notify"
-	"github.com/tencentcloud/CubeSandbox/cubelog"
+	CubeLog "github.com/tencentcloud/CubeSandbox/cubelog"
 )
 
 type Server struct {
 	InternalHttpServer *internalHttp
+	ErrChan            chan error
 }
 
 func New(ctx context.Context, cfg *config.Config) (*Server, error) {
 	if cfg == nil || cfg.Common == nil {
 		return nil, errors.New("config is nil")
 	}
-	s := &Server{}
+	s := &Server{ErrChan: make(chan error, 1)}
 	var err error
 	s.InternalHttpServer, err = NewInternalHttp(ctx, cfg)
 	if err != nil {
@@ -111,13 +114,26 @@ func (s *internalHttp) registerHandlers() {
 }
 
 func (s *internalHttp) Start() error {
-	if err := s.ListenAndServe(); err != nil {
-		if err == http.ErrServerClosed {
-			return nil
-		}
-		return errors.WithStack(err)
+	addr := s.Addr
+	if addr == "" {
+		addr = ":http"
 	}
-	return nil
+
+	ln, err := net.Listen("tcp", addr)
+	if err != nil {
+		return err
+	}
+
+	actualAddr := ln.Addr().String()
+	CubeLog.Infof("cube-master listening on %s", actualAddr)
+	stdlog.Printf("cube-master listening on %s", actualAddr)
+
+	err = s.Serve(ln)
+	if err == http.ErrServerClosed {
+		return nil
+	}
+
+	return errors.WithStack(err)
 }
 
 func (s *Server) Run() {
@@ -125,6 +141,7 @@ func (s *Server) Run() {
 		go func() {
 			if err := s.InternalHttpServer.Start(); err != nil {
 				CubeLog.Errorf("ListenAndServe:%v", err)
+				s.ErrChan <- err
 			}
 		}()
 	}
--- a/CubeMaster/pkg/server/server.go
+++ b/CubeMaster/pkg/server/server.go
@@ -8,8 +8,6 @@ package server
 import (
 	"context"
 	"fmt"
-	stdlog "log"
-	"net"
 	"net/http"
 	"os"
 	"sync"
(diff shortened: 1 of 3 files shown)


chenhengqi:
What's the purpose of this change?

yyyang91:
> What's the purpose of this change?

This modification mainly addresses the following issues.

When the port used by the cubemaster service is occupied, cubemaster does not correctly print error logs. It is difficult to quickly identify the problem based on the logs. Even when it starts successfully, there is no success log similar to that of the cube API.
scree

CubeMaster/pkg/server/server.go
```diff
@@ -111,20 +114,34 @@ func (s *internalHttp) registerHandlers() {
...
-		if err == http.ErrServerClosed {
-			return nil
-		}
-		return errors.WithStack(err)
+	addr := s.Addr
+	if addr == "" {
+		addr = ":http"
+	}
+
+	ln, err := net.Listen("tcp", addr)
```
chenhengqi: Does this change work with cubemastercli?

CubeMaster/pkg/server/server.go
```diff
@@ -111,20 +114,34 @@ func (s *internalHttp) registerHandlers() {
...
+	if err != nil {
+		return err
 	}
-	return nil
+
+	actualAddr := ln.Addr().String()
+	CubeLog.Infof("cube-master listening on on %s", actualAddr)
+	stdlog.Printf("cube-master listening on on %s", actualAddr)
```
chenhengqi: Why do we need a stdlog?

CubeMaster/pkg/server/server.go
yyyang91: This logic is ported from the Go standard library ([net/http/server.go, v1.24.8](https://github.com/golang/go/blob/go1.24.8/src/net/http/server.go#L3337-L3351)).

I migrated the original implementation here to gain more granular control over the startup sequence. Specifically, it allows us to inject the startup/failure logs and port check logic directly into the serving loop, which addresses the observability requirements for cube-master.

CubeMaster/pkg/server/server.go
yyyang91: This is because:

1.  CubeLog.Infof only writes logs to a specific business log file /data/log/CubeMaster/cubemaster-req.log
2.  [start_with_pidfile fucniton](https://github.com/TencentCloud/CubeSandbox/blob/1ef7499e12bc99a632cf28e173f6a086f52a79f6/deploy/one-click/scripts/one-click/common.sh#L98-L126)  just redirect stdout/stderr to /var/log/cube-sandbox-one-click/cubemaster.log, we cannot find errors in /var/log/cube-sandbox-one-click/cubemaster.log

CubeMaster/pkg/server/server.go
chenhengqi: You did not answer my question.

CubeMaster/pkg/server/server.go
fslongjin: Only the logs under /data/log/CubeMaster/meet the requirements.

CubeMaster/pkg/server/server.go
yyyang91: Yes, it works

CubeMaster/pkg/server/server.go
yyyang91: But, the installation logs directed me to the /var/log/cube-sandbox-one-click directory to check for errors, You can view the screenshot. https://github.com/TencentCloud/CubeSandbox/pull/127#issuecomment-4352198471

yyyang91: Do you mean just keep stdlog when the server start error? If the server start successfully, just  logs to /data/log/CubeMaster/

chenhengqi: How about panic when the requested port is in use?

yyyang91: I also think a panic is more clearer. I didn't use it because I'm not sure about the project's policy on panicking. Additionally, I found out that CubeLog.Fatal doesn't exit the program.

yyyang91: > How about panic when the requested port is in use?

I think the root cause of the current issue is likely that CubeLog.Errorf does not trigger a panic or exit the program.😅

CubeMaster/pkg/server/server.go
chenhengqi: If the real issue is that the requested port is in use, we can simply panic here.

chenhengqi (changes requested):

CubeMaster/pkg/server/server.go
yyyang91: It seems that a panic only terminates the program, while a non-panic means the process is still running. To speed up troubleshooting, it's necessary to have logs indicating whether the service started successfully. Should we write success and error messages to `/var/log/cube-sandbox-one-click/cubemaster.log?  This is because when the installation script encounters an error, it prompts us to check the files in the directory.

chenhengqi: Panic will log to stderr and should be captured, no?

yyyang91: Panic will log to stderr and can be captured. but it will print the stack.  Is it ok to print the stack?

CubeMaster/pkg/server/server.go
chenhengqi: @fslongjin PTAL.

yyyang91:
I revert the preious commit, and add a commit just panic.

fslongjin:
I believe a better solution here would be to introduce a synchronization point​ before the startup completes. If an issue is detected, trigger a graceful shutdown​ instead of panicking outright. We should already have a graceful exit mechanism in place. @yyyang91

yyyang91:
> I believe a better solution here would be to introduce a synchronization point​ before the startup completes. If an issue is detected, trigger a graceful shutdown​ instead of panicking outright. We should already have a graceful exit mechanism in place. @yyyang91

 
On the master branch, when serverTmp.Run() occur errors  would not triger a graceful exit mechanism. 
https://github.com/TencentCloud/CubeSandbox/blob/09101b2339ddebb8267cf215d0a0f0b94bb50f12/CubeMaster/cmd/cubemaster/app/main.go#L112-L114


I revert the commit that can  tigger graceful exit yesterday.

yyyang91:
When the port is already in use, cube-master hangs silently instead of exiting. The graceful shutdown mechanism is not triggered, so the process just sits there. At the same time, no error is written to /var/log/cube-sandbox-one-click/cubemaster.log — the directory the installation script points you to when something goes wrong.
So when the install fails and the ops team follows the script's prompt to check logs, there is no trace of the startup failure anywhere.

What is the project's expected behavior here? CubeLog alone cannot emit errors to the path the installation script expects, so should the process exit immediately, trigger a graceful shutdown, or is there another mechanism to make startup failures visible in the installation script's diagnostic directory?

@chenhengqi @fslongjin 
Please confirm the direction and I'll implement accordingly.

chenhengqi (approved):
I am OK with this change.

yyyang91:
Hi maintainers, just checking if this PR is still worth keeping open.

It has already received an approval, so please let me know if anything else is needed from my side, or whether it would be better to close this PR.

chenhengqi:
cc @fslongjin

PR #133, opened 2026-05-03
bugfix(CubeProxy): solve stale routing by removing refresh-on-hit and implementing proactive cache invalidation

novahe:
#### 1. The Problem
`CubeProxy` suffered from stale routing issues when sandboxes were destroyed or migrated. The root cause was a "refresh-on-hit" pattern in the local routing cache:
- Every successful cache hit extended the entry's TTL.
- For active sandboxes, routing entries never naturally expired.
- When `CubeMaster` updated Redis with a new backend IP/Port, the Proxy would continue using the old cached data until a manual restart or an extremely long timeout.

#### 2. The Solution
This PR implements two major changes:
*   **Natural Expiry**: Removed the `cache:set` call on hits in `rewrite_phase.lua`. Routing entries now follow their original random TTL (60-300s).
*   **Proactive Invalidation**: Implemented a precise identification mechanism in `header_filter_phase.lua` to detect connection failures and clear the local cache immediately.

#### 3. Key Logic: Bytes vs. Status
To avoid "false positives" (app-level 5xx), we use `$upstream_bytes_received` as the primary signal:
- **Network Failure**: `upstream_bytes_received == 0`. Action: **Clear Cache.**
- **App-level Error**: `upstream_bytes_received > 0`. Action: **Keep Cache.**

Refactored with guard clauses for performance and detailed logging for observability.

#### 4. Verification Results (Manual E2E):
- [x] **Network Failures**: Confirmed connection failures trigger immediate cache invalidation.
- [x] **App-level 5xx**: Confirmed business 5xx responses preserve the cache.
- [x] **TTL Behavior**: Verified TTL no longer refreshes on hit.
- [x] **Convergence Speed**: Verified backend switching completes in a single request after Redis update.

### DIFF
--- a/CubeProxy/lua/header_filter_phase.lua
+++ b/CubeProxy/lua/header_filter_phase.lua
@@ -1,14 +1,47 @@
-if ngx.status ~= 200 then
-    if ngx.var.upstream_addr ~= nil and ngx.var.cube_retcode == "310200" then
-        local utils = require "utils"
-        local ok, err = utils:is_faulty_backend(ngx.var.backend_ip, false)
-        if err then
-            ngx.log(ngx.ERR, "LEVEL_WARN||", "check backend fault err: ", err)
+if ngx.status ~= 200 and ngx.var.upstream_addr and ngx.var.cube_retcode == "310200" then
+    local utils = require "utils"
+    local ok, err = utils:is_faulty_backend(ngx.var.backend_ip, false)
+    if err then
+        ngx.log(ngx.ERR, "LEVEL_WARN||", "check backend fault err: ", err)
+    end
+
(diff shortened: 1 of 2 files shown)


staryxchen:
Thanks for the contribution — the root cause analysis is spot-on. Cube doesn't support live migration yet (planned). Without it, the backend IP for a given sandbox ID won't change during its lifetime, meaning:

- Removing refresh-on-hit adds Redis round-trips without benefit.
- Proactive cache invalidation clears entries that Redis will repopulate with the same (dead) value.

I’d suggest we defer this until live migration support is underway, so the two can ship and be validated together. CC @chenhengqi

Issue #111, opened 2026-04-28
[Feature Request] Can cube support GPU sandboxes


dushulin:
## Problem / Motivation

Some evaluation tasks and rollover tasks require execution of cuda-related instructions

YangYuS8:
I’m interested in helping with GPU sandbox support, but this looks like a larger design-level feature involving device passthrough, scheduling, and runtime compatibility.

Would it be useful to start with a small design/proposal document that summarizes possible approaches, such as PCI passthrough / vGPU / NVIDIA runtime integration, and identifies which CubeSandbox components would need changes?

dushulin:
@YangYuS8 Thanks, We have GPU sandboxes scene internally, and I think there are few cloud platforms currently supporting GPU sandboxes, but there will be many needs in the future. I can simply sort out the poc goals and build a plan with the community. cc @fslongjin hi～What do you think of GPU sandboxes?

kinwin-ustc:
Contributions are welcome.
GPU support would make the sandbox rather heavyweight, sacrificing its capabilities such as fast startup, high concurrency and high density.
However, it is still a viable option for scenarios that do not require extremely high elasticity.

timeuser4:
Do you have any plans? I'd like to contribute to this discussion.

fslongjin:
I'll share some reference materials with you later today. We've already conducted some preliminary experiments in this area, which might be helpful for your work. We welcome collaboration on this.

@YangYuS8 @dushulin @timeuser4

fslongjin:
To implement GPU support, the following aspects are involved:

1. **CubeMaster**: Manage GPU information on the host machines, schedule based on resource requests, and set node metadata.
2. **Cubelet**:
   - Report node GPU information (model, count, usage, and association with sandboxes).
   - Configure the container's prestart hook.
   - Calculate the memory overhead required for GPUs and add it to the sandbox configuration.
   - Manage GPU cards: bind them to the VFIO driver, allocate and release resources, perform reconciliation, health checks, etc.
   - **Key Pitfall to Note**: GPU cards must only be released after the shim process has fully exited. This is because VFIO unpin operations can be slow.
3. **Guest Image**:
   - Add NVIDIA drivers and install the `nvidia-container-toolkit` (it is recommended to separate this into two images distinct from non-GPU instances).
[...]

PR #160, opened 2026-05-08
docs: add GPU sandbox design proposal

YangYuS8:
Related to #111.

This PR adds an initial GPU sandbox design proposal based on the discussion in #111.

It does not introduce implementation changes yet. The goal is to document the initial scope, non-goals, component responsibilities, lifecycle limitations, deployment limitations, and open questions before starting implementation work.


chenhengqi:
Instead of open a PR, could we start with a discussion?

chenhengqi:
https://github.com/TencentCloud/CubeSandbox/discussions

kinwin-ustc:
It seems that there are still many open questions to be discussed, and I think it is more appropriate to put it in the discussion group first and then settle it into a document

YangYuS8:
Thanks for the suggestion. I moved the proposal to a GitHub Discussion so the design can be reviewed and refined before becoming part of the docs:

https://github.com/TencentCloud/CubeSandbox/discussions/174

Closing this PR for now. Once the design direction is agreed on, I can open a follow-up docs or implementation PR.

PR #163, opened 2026-05-08
fix(network-agent): allow dots in sandboxID validation

rogeroger-yu:
## Summary

  - Fix overly strict sandboxID validation in network-agent's state_store that
    rejected any ID containing `.` (dot). IDs like `cubesandbox-python-slim-3.11-nydus_0`
    (containing version numbers) were incorrectly flagged as path traversal.
    Changed to only reject `/`, `\`, and `..` — consistent with Cubelet's
    `pathutil.ValidateID()`.
  - Fix ansible control role crash when `/etc/NetworkManager/dnsmasq.d` does not
    exist (hosts using systemd-resolved instead of NM-dnsmasq).
  - Skip nydus-image tarball download in nydus-uffd role when the binary is
    already pre-staged locally.

  ## Root Cause

  `state_store.go:path()` used `strings.ContainsAny(sandboxID, "/\\.")` which
  treats a single `.` the same as path separators. A dot in a version string
  (e.g. `3.11`) is perfectly safe — only `..` is a traversal risk.

  ## Test Plan

  - [x] Template creation with dotted image tag succeeds end-to-end
  - [x] Full multi-node ansible deploy passes (0 failures across 5 nodes)
  - [x] All compute nodes report HEALTHY after network-agent restart
  
  ## Original Error
  
```
<ie-training/cubesandbox-python-slim:3.11 --writable-layer-size 1G --storage-media nydus --expose-port 49983 --probe 49983 --probe-path /health --template-id cubesandbox-python-slim-3.11-nydus
job_id: d94ebd92-785e-4cfa-ba7d-b19e9df80aa8
template_id: cubesandbox-python-slim-3.11-nydus
attempt_no: 1
operation: CREATE
artifact_id:
status: PENDING
phase: PULLING
progress: 0%
distribution: 0/0 ready, 0 failed
root@ak-ai-engine-sandbox05:~/Workspace/code/cube-sandbox# /usr/local/services/cubetoolbox/CubeMaster/bin/cubemastercli tpl watch --job-id d94ebd92-785e-4cfa-ba7d-b19e9df80aa8
job_id: d94ebd92-785e-4cfa-ba7d-b19e9df80aa8
template_id: cubesandbox-python-slim-3.11-nydus
attempt_no: 1
operation: CREATE
artifact_id: rfs-85671e04f4cdfcd700f581a1
status: FAILED
phase: CREATING_TEMPLATE
progress: 100%
distribution: 1/1 ready, 0 failed
template_spec_fingerprint: 85671e04f4cdfcd700f581a19887ccdf7de95430f820b195973b99d2687f8ab3
template_status: FAILED
error: template cubesandbox-python-slim-3.11-nydus creation failed: network-agent EnsureNetwork failed: grpc_code=Unknown grpc_msg=invalid sandboxID "cubesandbox-python-slim-3.11-nydus_0": contains path separators or traversal characters
cubemastercli run fail: template cubesandbox-python-slim-3.11-nydus creation failed: network-agent EnsureNetwork failed: grpc_code=Unknown grpc_msg=invalid sandboxID "cubesandbox-python-slim-3.11-nydus_0": contains path separators or traversal characters
```

### DIFF
--- a/Cubelet/plugins/cube/internals/createid/plugin.go
+++ b/Cubelet/plugins/cube/internals/createid/plugin.go
@@ -48,13 +48,7 @@ func (l *local) Create(ctx context.Context, opts *workflow.CreateContext) error 
 	opts.SandboxID = utils.GenerateID()
 
 	if opts.IsCreateSnapshot() {
-
-		templateID, ok := opts.GetSnapshotTemplateID()
-		if !ok {
-			return ret.Err(errorcode.ErrorCode_InvalidParamFormat, "cube.master.appsnapshot.template.id should provide")
-		}
-
-		opts.SandboxID = templateID + "_" + "0"
+		opts.SandboxID = utils.GenerateID() + "_snapshot"
 	}
 	return nil
 }
--- a/network-agent/internal/service/state_store_test.go
+++ b/network-agent/internal/service/state_store_test.go
@@ -4,7 +4,54 @@
 
 package service
 
-import "testing"
+import (
+	"path/filepath"
+	"testing"
+)
+
+func TestStateStorePath(t *testing.T) {
+	store, err := newStateStore(t.TempDir())
+	if err != nil {
+		t.Fatalf("newStateStore error=%v", err)
+	}
+
+	tests := []struct {
+		name      string
+		sandboxID string
+		wantErr   bool
+	}{
+		{name: "valid simple id", sandboxID: "sb-1", wantErr: false},
+		{name: "valid uuid", sandboxID: "a1b2c3d4e5f6", wantErr: false},
+		{name: "valid snapshot id", sandboxID: "a1b2c3d4e5f6_snapshot", wantErr: false},
+		{name: "empty string", sandboxID: "", wantErr: true},
+		{name: "contains dot", sandboxID: "sb.1", wantErr: true},
+		{name: "forward slash", sandboxID: "a/b", wantErr: true},
+		{name: "backslash", sandboxID: "a\\b", wantErr: true},
+		{name: "dot prefix", sandboxID: ".hidden", wantErr: true},
+		{name: "double dot traversal", sandboxID: "..", wantErr: true},
+		{name: "oci image name", sandboxID: "registry.example.com/path/image:tag", wantErr: true},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			p, err := store.path(tt.sandboxID)
+			if tt.wantErr {
+				if err == nil {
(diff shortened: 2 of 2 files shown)


fslongjin:
@rogeroger-yu Hi, could you please add some unit tests to stabilize this behavior?

rogeroger-yu:
@fslongjin okey.

fslongjin:
After reviewing this, I don't think this is the right way to fix it. The sandbox ID should be designed as a UUID. The root cause here is that when CubeMaster generates the template, it sends the template OCI image name​ to Cubelet as the sandbox ID.
A possible solution would be to modify CubeMaster to either generate a UUID or pass down the template ID​ to the Cubelet layer.
Note:​ For normal sandbox creation requests, the sandbox ID is generated by Cubelet. However, the current step is creating a sandbox template​ (for app snapshots), which is why CubeMaster specifies the sandbox ID.

rogeroger-yu:
@fslongjin I agree with your judgment and resubmit a version. I will review it after my hard work.

Cubelet/plugins/cube/internals/createid/plugin.go
```diff
@@ -48,13 +48,10 @@ func (l *local) Create(ctx context.Context, opts *workflow.CreateContext) error
 	opts.SandboxID = utils.GenerateID()
 
 	if opts.IsCreateSnapshot() {
-
-		templateID, ok := opts.GetSnapshotTemplateID()
-		if !ok {
+		if _, ok := opts.GetSnapshotTemplateID(); !ok {
```
chenhengqi: This `if` statement should be dropped.

network-agent/internal/service/state_store_test.go
chenhengqi, line 1: Please update your commit message accordingly.

chenhengqi:
cc @fslongjin

fslongjin:
Hi @rogeroger-yu,

Thank you for your contribution and for the detailed PR description! We really appreciate you taking the time to investigate and fix the sandboxID validation issue.

However, after careful consideration, we've decided not to merge this change. Our reasoning is as follows:

We are currently working on open-sourcing the **snapshot rollback** and **sandbox clone** features. These features rely on **ID prefixes** to distinguish whether an ID refers to a template or a sandbox snapshot. Allowing user-customized template IDs (including those with dots) would conflict with this design, as it would make it harder to reliably differentiate between template IDs and snapshot IDs based on their prefixes.

The fix itself is correct and well-written — it's just that our upstream architecture is heading in a direction where we need stricter control over ID formats.

Thank you again for your effort and for using CubeSandbox! We hope to see more contributions from you in the future. 🙏

fslongjin:
Closing as not planned — see above for details.

PR #181, opened 2026-05-11
fix(cubecli): remove dead listmd command

chenggui53:
The `cubecli cubebox listmd` command was introduced when each cubelet plugin persisted its own bbolt metadata file under a plugin-scoped directory. Cubebox metadata has since been migrated to a shared multimeta store (`<state>/io.cubelet.cubemetastore.v1.mutilmeta/db/{b..k}/meta.db`, exposed via the gRPC `MultiMetaDBServer` API). The `listmd` implementation — which still targets the old per-plugin layout — has rotted in two ways that make it unsafe to keep:

1. **Destructive side effects.** Despite the `list`-style name, the loop calls `cubes.RecoverPod` on every entry. For sandboxes in the `Stopped` state, `RecoverContainer` issues `t.Delete(ctx, containerd.WithProcessKill)` (`Cubelet/plugins/cube/internals/cubes/restart.go:232`), which SIGKILLs and reaps the containerd task. Pod recovery is already owned by cubelet startup; an operator CLI has no business doing it.

2. **Concurrency-unsafe DB access.** `copyDb` does `cp -r` on the live cubebox bbolt tree into `os.TempDir()` and opens it through a package-level `dbHandle *utils.CubeStore`. Copying a live bbolt file while cubelet holds it open can yield a torn snapshot, and the package-level handle is unguarded against concurrent invocations.

The functionality is fully covered by already-shipping commands that talk to the live gRPC services:

  - `cubecli meta dbs`                      list registered buckets
  - `cubecli meta view -db cubebox sandbox/v1`   dump cubebox sandbox entries
  - `cubecli cubebox inspect <id>`               pretty-print one sandbox

These are strictly better: they require no path guessing, are read-only by contract, and cannot be broken by further plugin-layout changes.

Remove `listMetaData`, its `copyDb` helper, and the unused `sandBoxInfo`/`containerInfo` types it dragged along. Deregister it from the subcommand list in `cubebox.go`. Drop imports that become unused. No net-new code.

### DIFF
--- a/Cubelet/cmd/cubecli/commands/cubebox/cubebox.go
+++ b/Cubelet/cmd/cubecli/commands/cubebox/cubebox.go
@@ -15,7 +15,6 @@ var Command = &cli.Command{
 	Subcommands: []*cli.Command{
 		ListCommand,
 		ListSandboxCommand,
-		&listMetaData,
 		update,
 		Create,
 		&inspecMetaData,
--- a/Cubelet/cmd/cubecli/commands/cubebox/metadata.go
+++ b/Cubelet/cmd/cubecli/commands/cubebox/metadata.go
@@ -6,88 +6,13 @@ package cubebox
 
 import (
 	"fmt"
-	"log"
 	"strings"
 
-	"context"
-	"os"
-	"path"
-	"path/filepath"
-	"time"
-
-	containerd "github.com/containerd/containerd/v2/client"
-	"github.com/containerd/containerd/v2/pkg/namespaces"
-	"github.com/containerd/platforms"
-	jsoniter "github.com/json-iterator/go"
 	"github.com/urfave/cli/v2"
 
 	"github.com/tencentcloud/CubeSandbox/Cubelet/api/services/cubebox/v1"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/cmd/cubecli/commands"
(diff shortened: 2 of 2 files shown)


fslongjin:
Thanks for the cleanup. One correction on the PR description: the path `/data/cubelet/state/io.cubelet.internal.v1.cubebox/db` does still exist — the cubebox plugin creates and uses `db/meta.db` there (see `local.go:51-56`). Multimeta registers the existing `CubeStore` via gRPC but doesn't relocate the underlying data.

Suggest updating the description to reflect the actual reasons for removal: the command has destructive side effects (calls `RecoverPod` which can `ProcessKill` containers), concurrency issues (package-level `dbHandle`, unsafe live DB copy), and its functionality is fully covered by the safer `inspect` / `meta view` gRPC commands and the plugin-internal `RecoverAllCubebox` path.

LGTM to merge after the description update.

chenggui53:
Thanks @fslongjin — description rewritten around the two real reasons (destructive `RecoverPod`/`ProcessKill` + unsafe live `cp -r` on the bbolt tree). Please take another look.

fslongjin:
@chenggui53  Hello, DCO is needed. Please add a line "Signed-off-by" in your commit message~Thanks

chenggui53:
Hi @fslongjin , "Signed-off-by"  added in my commit, Please take another look ~

PR #191, opened 2026-05-11
fix(shim): use read_stderr for stderr stream forwarding

novahe:
Found that commands executed via exec were missing stderr output, and sometimes stdout output was randomly split into the stderr stream. 

Investigated and discovered that the forward_stderr logic was incorrectly polling read_stdout instead of read_stderr

### DIFF
--- a/CubeShim/shim/src/container/exec.rs
+++ b/CubeShim/shim/src/container/exec.rs
@@ -253,7 +253,7 @@ impl Exec {
         };
         let ctx = Context::default();
         loop {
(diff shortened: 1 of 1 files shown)


fslongjin (approved):

fslongjin:
Hello, please add DCO to your commit.

fslongjin (changes requested):
DCO needed

novahe:
PTAL @fslongjin

fslongjin (approved):

PR #194, opened 2026-05-12
fix(CubeMaster/templatecenter): serialize concurrent rootfs builds for the same artifact_id

dushulin:
## Motivation
  Quick-succession `cubemastercli tpl create-from-image` calls against the same OCI image spec race inside `ensureRootfsArtifact`: both goroutines resolve to the same `artifact_id`, share the same workDir/ext4Path, and one's `defer
  cleanupIntermediateArtifacts` wipes the ext4 file out from under the other. The surviving caller reaches `distributeRootfsArtifact` with a partial record (`ext4_size_bytes=0`, empty `download_token`, empty `master_node_ip`); `buildDownloadURL` falls back
   to `os.Hostname()`, and cubelet rejects the pull with `invalid size:0` — the template is marked FAILED with no clear signal of the underlying race.

  ## Changes
  - `artifactBuildLocks sync.Map` (keyed by `artifact_id`) serializes the full find-or-build flow in `ensureRootfsArtifact`. Granularity is `artifact_id` only — concurrent builds for different image specs still run in parallel.
  - Defense-in-depth guard at the top of `distributeRootfsArtifact`: refuse to push a `CreateImage` when the artifact record is obviously incomplete, returning a diagnostic error instead of relying on cubelet's opaque `invalid size:0`.

  ## Verification
  - `go build ./CubeMaster/pkg/templatecenter/...` clean
  - `go vet ./CubeMaster/pkg/templatecenter/...` clean
  - Reproduced on an internal deployment before the fix; after the fix, concurrent submits serialize as expected and no partial artifact reaches cubelet.

### DIFF
--- a/CubeMaster/pkg/templatecenter/artifact_build.go
+++ b/CubeMaster/pkg/templatecenter/artifact_build.go
@@ -9,6 +9,10 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"strings"
+	"sync"
+	"time"
+
 	"github.com/google/uuid"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/db/models"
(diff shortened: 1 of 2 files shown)


CubeMaster/pkg/templatecenter/template_image.go
```diff
@@ -1221,6 +1236,15 @@ func ensureRootfsArtifact(ctx context.Context, req *types.CreateTemplateFromImag
...
 	fingerprint := buildTemplateSpecFingerprint(req, source.digest)
 	artifactID := buildArtifactID(fingerprint)
+	// Serialize concurrent builds of the same artifactID. Without this, two
+	// submits of the same image spec race on workDir/storeDir/ext4Path; the
+	// losing goroutine's defer cleanup can wipe the ext4 file while the
+	// winner is still relying on it. See artifactBuildLocks comment for the
+	// full failure mode.
+	muV, _ := artifactBuildLocks.LoadOrStore(artifactID, &sync.Mutex{})
+	mu := muV.(*sync.Mutex)
+	mu.Lock()
```
chenhengqi: The second request will be blocked here. Can we return early here?

dushulin: Requests that do not currently obtain a lock will be stuck in the first state of the state machine (BUILDING_EXT4). After the job with the lock is completed, other tasks will instantly enter the next state. I understand what you mean is that when you find this situation, do the next job directly throw an exception at the api level to the cli and end the job?

dushulin: I think the advantage of this is that it is not intrusive to the state machine or the cli, but can cause too many coroutines in extreme cases

fslongjin:
Hello! Thank you for your contribution! This issue does exist. However, since we are currently refactoring the template creation functionality (we felt that binding it to cubemaster wasn't ideal), I plan to hold off on merging this PR for now. The new design will introduce an independent module for template management. I will apply your changes there and add you as a Co-author. Thanks again for your contribution!

dushulin:
> Hello! Thank you for your contribution! This issue does exist. However, since we are currently refactoring the template creation functionality (we felt that binding it to cubemaster wasn't ideal), I plan to hold off on merging this PR for now. The new design will introduce an independent module for template management. I will apply your changes there and add you as a Co-author. Thanks again for your contribution!您好！感谢您的贡献！这个问题确实存在。不过，由于我们目前正在重新设计模板创建功能（我们认为将其与 Cubemaster 集成并不理想），我打算暂时先不合并这个 PR。新的设计中，模板管理将作为一个独立的模块来实现。我会把您的修改应用到新系统中，并将您列为该模块的合著者。再次感谢您的贡献！

Thanks for your reply!

fslongjin (approved):

PR #205, opened 2026-05-12
docs(network): Document compute-node port allocation ranges

chenhengqi:
Clarify the 10000-19999 / 20000-29999 / 30000-65535 split between network-agent ip_local_port_range, CubeProxy sandbox access, and CubeVS SNAT, in both zh and en network docs.

Note that these port ranges are not configurable for now but should work for most use cases. We can revise this later.


chenhengqi:
cc @fslongjin

fslongjin (approved):

PR #207, opened 2026-05-12
Change default exposed port to 49983 for E2B compatibility

### LINKED ISSUE #204: [good first issue] Change default exposed port to 49983
Currently, the default exposed ports are 8080 and 32000 which are not useful for E2B compatible sandboxes.
Let's change it to 49983 instead.

# Related Components

* CubeMaster
* Cubelet
* network-agent

turtlemoji:
## Summary

  The default exposed ports (8080, 32000) are not useful for E2B-compatible sandboxes. This PR replaces them with 49983 across all three affected components.

  ### Changes

  - **CubeMaster**: Update `defaultTemplateExposedPorts()` and add tests verifying the new default and reserved port behavior
  - **Cubelet**: Update `DefaultExposedPorts`, shipped `config.toml`, and the validation error message to reflect one reserved port instead of two
  - **network-agent**: Update `DefaultConfig()` and add a test for the new default

  ## Test plan

  - [x] CubeMaster: `go test ./pkg/templatecenter/ -run TestDefaultTemplateExposedPorts`
  - [x] CubeMaster: `go test ./pkg/templatecenter/ -run TestCountCustomTemplateExposedPortsTreats49983AsReserved`
  - [x] Cubelet: `go test ./network/ -run TestDefaultExposedPortsContainsOnly49983`
  - [x] network-agent: `go test ./internal/service/ -run TestDefaultConfigExposedPorts`
  - [x] All existing exposed port tests still pass
  - [x] `gofmt` clean on all changed files

  Closes #204

### DIFF
--- a/Cubelet/config/config.toml
+++ b/Cubelet/config/config.toml
@@ -66,7 +66,6 @@ dynamic_config_path = "/usr/local/services/cubetoolbox/Cubelet/dynamicconf/conf.
     mvm_gw_mac_addr = "20:90:6f:cf:cf:cf"
     mvm_gw_dest_ip = "169.254.68.5"
     mvm_mtu = 1300
-    default_exposed_ports = [8080, 32000]
     # metric
     check_interval_in_sec = "5s"
     report_stat_interval_in_sec = "60s"
--- a/Cubelet/services/cubebox/check.go
+++ b/Cubelet/services/cubebox/check.go
@@ -14,7 +14,6 @@ import (
 	"github.com/containerd/fifo"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/api/services/cubebox/v1"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/api/services/errorcode/v1"
-	"github.com/tencentcloud/CubeSandbox/Cubelet/network"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/pkg/constants"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/pkg/log"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/pkg/recov"
@@ -48,26 +47,14 @@ func checkParam(ctx context.Context, realReq *cubebox.RunCubeSandboxRequest) err
 		}
 	}
 
-	reqPorts := 0
 	for _, p := range realReq.GetExposedPorts() {
 		if p <= 0 || p > 65535 {
 			return ret.Errorf(errorcode.ErrorCode_InvalidParamFormat, "invalid exposed port %d", p)
 		}
-		isDefault := false
-		for _, d := range network.DefaultExposedPorts {
-			if uint16(p) == d {
-				isDefault = true
-				break
-			}
-		}
-		if !isDefault {
-			reqPorts++
-		}
 	}
-
-	if reqPorts > 3 {
+	if len(realReq.GetExposedPorts()) > 4 {
 		return ret.Errorf(errorcode.ErrorCode_InvalidParamFormat,
-			"exposed ports should less or equal than 5(system reserves two of them)")
+			"exposed ports should be at most 4")
 	}
 
 	if err != nil {
--- a/CubeMaster/pkg/templatecenter/template_image.go
+++ b/CubeMaster/pkg/templatecenter/template_image.go
@@ -2272,8 +2272,7 @@ func countCustomTemplateExposedPorts(ports []int32) int {
 
 func defaultTemplateExposedPorts() map[int32]struct{} {
 	return map[int32]struct{}{
-		8080:  {},
-		32000: {},
+		49983: {},
 	}
 }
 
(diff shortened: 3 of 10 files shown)


Cubelet/config/config.toml
```diff
@@ -66,7 +66,7 @@ dynamic_config_path = "/usr/local/services/cubetoolbox/Cubelet/dynamicconf/conf.
     mvm_gw_mac_addr = "20:90:6f:cf:cf:cf"
     mvm_gw_dest_ip = "169.254.68.5"
     mvm_mtu = 1300
-    default_exposed_ports = [8080, 32000]
+    default_exposed_ports = [49983]
```
chenhengqi: Thanks for working on this issue.

I have a question in mind: Do we need these configurations in three components?

turtlemoji: Yes @chenhengqi, all three need updating because each component uses its own default independently — CubeMaster uses it for template validation, Cubelet for building port mappings, and network-agent for the actual network enforcement. If we only changed one, the others would still use the old ports and wouldn't match. That said, consolidating them into a single file / source of truth would be cleaner.

turtlemoji: For consolidating into a shared constant package tho, I recommend doing it on another PR. (Let me know 😃)

chenhengqi: If CubeMaster already set default exposed port to 49983 on template creation, Cubelet should be able to obtain it through realReq.GetExposedPorts().

turtlemoji: Ah, thats a better workflow than shared constant @chenhengqi, I think for that to work we will need to remove the overrides in cubelet and network-agent tho

turtlemoji: I can try to implement the changes, should I commit in this same pr with the same format? `component: desc`

chenhengqi: > Ah, thats a better workflow than shared constant @chenhengqi, I think for that to work we will need to remove the overrides in cubelet and network-agent tho

Right, I think we should remove those ports from cubelet and network-agent.

chenhengqi: > I can try to implement the changes, should I commit in this same pr with the same format? `component: desc`

Just do it in this PR.

turtlemoji: Main fixes on https://github.com/TencentCloud/CubeSandbox/pull/207/commits/5ca7d3199cb5bdb1fc0b5ff316f28cde804dd3e3 and https://github.com/TencentCloud/CubeSandbox/pull/207/commits/f9618d6d6dbeea8bee080f45f9c3707f68182b6a

Little correction and `go fmt` https://github.com/TencentCloud/CubeSandbox/pull/207/commits/e606b0e62afb483ac264f92485e9079aa4443cef
https://github.com/TencentCloud/CubeSandbox/pull/207/commits/32f0a425c43e622783c54692b20fa5680e353663

chenhengqi: Please reorganize your commits. One commit per component.

turtlemoji: Done @chenhengqi.

Cubelet/services/cubebox/check.go
```diff
@@ -48,26 +47,14 @@ func checkParam(ctx context.Context, realReq *cubebox.RunCubeSandboxRequest) err
...
-				break
-			}
-		}
-		if !isDefault {
-			reqPorts++
-		}
 	}
-
-	if reqPorts > 3 {
+	if len(realReq.GetExposedPorts()) > 4 {
```
chenhengqi, line 55: `> 5`?

turtlemoji: @chenhengqi 
  - Old: 2 defaults + 3 customs = 5 total
  - New: 1 default + 3 customs = 4 total
  
The old limit was 2 defaults + 3 custom = 5. Since we're down to 1 default (49983) and CubeMaster still allows max 3 custom ports (countCustomTemplateExposedPorts > 3), I went with 1 + 3 = 4. Happy to bump to 5 if you'd prefer to keep the old ceiling as to be more permissive.

turtlemoji: for the other 2 mentions of cubebot, 

2) Will push a new test, this is correct

3) This is consistent with the previous behavior. The old code also didn't dedupe; it counted each occurrence of a non-default port, so [9000, 9000, 9000, 9000] would have hit the > 3 custom limit and been rejected. Cubelet and network-agent both dedupe via maps downstream (portReq in plugin_tap.go, byContainerPort in normalizePortMappings), so duplicates don't cause double mappings — only the count-based validation rejects them.

turtlemoji: pushed test fix for 2) and reorganized commits

chenhengqi: > @chenhengqi
> 
> * Old: 2 defaults + 3 customs = 5 total
> * New: 1 default + 3 customs = 4 total
> 
> The old limit was 2 defaults + 3 custom = 5. Since we're down to 1 default (49983) and CubeMaster still allows max 3 custom ports (countCustomTemplateExposedPorts > 3), I went with 1 + 3 = 4. Happy to bump to 5 if you'd prefer to keep the old ceiling as to be more permissive.

Sounds reasonable. Let's keep it as 4 for now.

chenhengqi (approved):
LGTM. Thanks.

chenhengqi:
@fslongjin Please take a look.

Issue #172, opened 2026-05-09
Port conflicts occurred during deployment, and configuration changes did not take effect.


PERFECT-IT-D:
在裸金属执行curl -sL https://cnb.cool/CubeSandbox/CubeSandbox/-/git/raw/master/deploy/one-click/online-install.sh | MIRROR=cn bash命令，因为宿主机已经有mysql的3306端口了，所以导致端口冲突，改配置文件之后再次执行，配置没生效呢，还是用的3306

kinwin-ustc:
Can you provide more information? The specific operation process, such as what configuration file has been changed, how to check that it does not take effect, etc

luzhixing12345:
Same issue here, and I have found the cause.

## Steps to Reproduce

1. self build deployment, build `cube-sandbox-one-click-<version>.tar.gz` and `cp env.example .env`
2. by default `CUBE_SANDBOX_MYSQL_PORT=3306`, but if the host is already running MySQL, it will be using port 3306, causing a port conflict.
3. set `CUBE_SANDBOX_MYSQL_PORT=3308` to avoid port conflict, run `sudo ./install.sh` to install cube, but get ERROR

   ```bash
   $ sudo ./install.sh
   [one-click] loading env file: /home/lzx/CubeSandbox/deploy/one-click/dist/cube-sandbox-one-click-6ce4ddb/.env
   [one-click] using node IP: 192.168.1.123
   [one-click] using primary network interface: enp56s0f0
   ...
   [one-click-runtime] dependencies stopped
   /usr/local/services/cubetoolbox/scripts/one-click/up.sh: line 82: 385353 Killed                  nohup env -i PATH="${clean_path}"    HOME="${clean_home}" LANG="${clean_lang}" SHELL="/bin/bash" bash -c "${cmd}" > "${log_file}" 2>&1
   [one-click-runtime] ERROR: core services did not become ready, check logs under /var/log/cube-sandbox-one-click
   ```

4. check  /var/log/cube-sandbox-one-click/cubemaster.log

   ```bash
   $ cat /var/log/cube-sandbox-one-click/cubemaster.log
   panic: dial tcp 127.0.0.1:3306: connect: connection refused
   
   goroutine 1 [running]:
   github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/db.Init(0xc0002f0c01?)
           /workspace/CubeMaster/pkg/base/db/db.go:32 +0x8a
[...]

PR #210, opened 2026-05-12
feat(deploy): sync cubemaster custom ports with .env config

luzhixing12345:
Related to https://github.com/TencentCloud/CubeSandbox/issues/172

## Problem

During self-build deployment, users may change `CUBE_SANDBOX_MYSQL_PORT` or `CUBE_SANDBOX_REDIS_PORT` to avoid host port conflicts. However, CubeMaster still uses the default hardcoded addresses (`127.0.0.1:3306` and `127.0.0.1:6379`) from the static config file, causing service startup failures.

As a result, CubeMaster cannot connect to MySQL during initialization and the one-click installation fails.

## Solution

This PR patches the generated CubeMaster config during installation and synchronizes the MySQL/Redis addresses with the values provided in the one-click environment configuration.

### DIFF
--- a/deploy/one-click/install.sh
+++ b/deploy/one-click/install.sh
@@ -109,6 +109,20 @@ check_proxy_cert_preflight() {
   :
 }
 
+generate_cubemaster_config_ports() {
+  [[ "${DEPLOY_ROLE}" != "compute" ]] || return 0
+
+  local cfg="${PKG_ROOT}/CubeMaster/conf.yaml"
+  local mysql_port="${CUBE_SANDBOX_MYSQL_PORT:-3306}"
+  local redis_port="${CUBE_SANDBOX_REDIS_PORT:-6379}"
+
+  ensure_file "${cfg}"
+  sed -i \
+    -e "s|__CUBE_SANDBOX_MYSQL_PORT__|${mysql_port}|g" \
+    -e "s|__CUBE_SANDBOX_REDIS_PORT__|${redis_port}|g" \
+    "${cfg}"
+}
+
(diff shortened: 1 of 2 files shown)


deploy/one-click/install.sh
```diff
@@ -109,6 +109,23 @@ check_proxy_cert_preflight() {
   :
 }
 
+patch_cubemaster_config_ports() {
```
chenhengqi: patch? We should generate `conf.yaml` on the fly instead.

luzhixing12345: for now the script just simply copy the file

https://github.com/TencentCloud/CubeSandbox/blob/855756dca06ebee50579d0830c2f1723b1c8c7a1/deploy/one-click/build-release-bundle.sh#L228

How can we generate a YAML file based on the configuration? Is it necessary to introduce a new YAML parsing tool (such as yq)?

chenhengqi: I think the `cubemaster.yaml` should contain port placeholder and perform a replace before copying.

luzhixing12345: Modified, use`__CUBE_SANDBOX_MYSQL_PORT__`and`__CUBE_SANDBOX_REDIS_PORT__` placeholders in cubemaster.yaml

chenhengqi (approved):

fslongjin (approved):

fslongjin:
Thanks @luzhixing12345 for the great contribution! This fix nicely addresses the custom port sync issue during self-build deployment. Appreciate the detailed problem description and clean implementation. 🎉

PR #224, opened 2026-05-13
network-agent: fix gateway MAC selection

xiongxz:
Fix CubeVS outbound traffic using a non-gateway destination MAC.

`getGatewayMacAddr()` previously returned the first reachable IPv4 neighbor on the configured node interface. When the interface has multiple reachable neighbors, network-agent can select a non-gateway MAC address and program CubeVS with the wrong L2 next hop. In that case sandbox SNAT packets leave the host NIC but do not reach the default gateway, so connections remain in `syn_sent` and eventually time out.

This resolves the gateway IP from the interface default route first, then looks up the neighbor entry for that exact gateway IP.

The PR also adds tests for a reachable non-gateway neighbor appearing before the actual gateway and for the missing default route case.

## Validation

```bash
go test -run 'TestGetGatewayMacAddr|TestEnsureRouteToCubeDev' -v ./internal/service
```

Assisted-by: Codex:GPT-5

### DIFF
--- a/network-agent/internal/service/netdevice.go
+++ b/network-agent/internal/service/netdevice.go
@@ -21,10 +21,12 @@ import (
 
 var netlinkRouteReplace = netlink.RouteReplace
 var netlinkRouteListFiltered = netlink.RouteListFiltered
+var netlinkRouteList = netlink.RouteList
 var netlinkLinkByIndex = netlink.LinkByIndex
 var netlinkLinkByName = netlink.LinkByName
 var netlinkLinkList = netlink.LinkList
 var netlinkLinkDel = netlink.LinkDel
+var netlinkNeighList = netlink.NeighList
 var unixOpen = unix.Open
 var unixClose = unix.Close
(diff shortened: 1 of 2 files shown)


chenhengqi:
@xiongxz Thanks for your contribution. Please follow our contribution guide. Will review this PR when it ready.

xiongxz:
Updated the PR according to the contribution guide: added the DCO sign-off, refreshed the PR description, and marked it ready for review. Thanks!

xiongxz:
Updated according to the review comments: renamed the default-route helper to make the IPv4 scope explicit, and added tests for missing gateway neighbors, multiple default-route metrics, and accepted/rejected neighbor states.\n\nValidation:\n\n```bash\ngo test -run 'TestGetGatewayMacAddr|TestEnsureRouteToCubeDev|TestIsUsableGatewayNeighbor' -v ./internal/service\n```

chenhengqi (approved):
LGTM. Thanks.

PR #227, opened 2026-05-13
fix(template_image): add request_id field to prevent duplicate job creation

fslongjin:
- Add request_id column to template_image_job table for idempotent request tracking
- Add unique index on (request_id, operation) to prevent duplicate submissions
- Implement migration logic for existing records with legacy request IDs
- Refactor job record creation into helper functions for clarity


chenhengqi:
cc @hlz260

taylorcmoon (commented):
The pick

PR #234, opened 2026-05-13
perf: avoid SHA256 computation on every startup

fslongjin:
- Split SyncKernelFile into EnsureKernelFilePresent (copy only if missing) and RefreshKernelFile (force refresh with verification)
- Use EnsureKernelFilePresent in EnsurePmemFile to skip SHA256 comparison on normal startup
- Add materializeDistributedTemplateRuntimeFiles to refresh kernel after image distribution

Fixed the performance regression in PR https://github.com/TencentCloud/CubeSandbox/pull/57

Before this PR (v0.2.0) in PVM (`INTEL(R) XEON(R) PLATINUM 8576C`): 
image

After: 
image

### DIFF
--- a/Cubelet/internal/cube/server/images/image_distribution_handler.go
+++ b/Cubelet/internal/cube/server/images/image_distribution_handler.go
@@ -38,14 +38,16 @@ type imageDistributionHandler struct {
 
 func defaultTemplateImageSpec(ns string, template *templatetypes.TemplateImage) *cubeimages.ImageSpec {
 	annotations := map[string]string{}
-	if template != nil && template.StorageMedia == cubeimages.ImageStorageMediaType_ext4.String() {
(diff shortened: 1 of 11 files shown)


Cubelet/internal/cube/server/images/image_distribution_handler.go
```diff
@@ -34,6 +36,43 @@ type imageDistributionHandler struct {
...
+	annotations := map[string]string{}
+	if template != nil && template.StorageMedia == cubeimages.ImageStorageMediaType_ext4.String() {
+		annotations[constants.MasterAnnotationInstanceType] = cubebox.InstanceType_cubebox.String()
+	}
+	var storageMedia string
+	var imageRef string
+	if template != nil {
+		storageMedia = template.StorageMedia
+		imageRef = template.Image
+	}
```
chenhengqi, line 51: Nit:
* Group variables
* Single `if template != nil` statement

Issue #235, opened 2026-05-13
Cubelet/storage: `newExt4RawByReflinkCopy` fails with "bad magic number in superblock" under N=100 concurrent spawn


WaylandYang:
## Environment

- CubeSandbox v0.2.0 via official one-click installer (`cube-sandbox-one-click-9c16021/install.sh`)
- Host: bare-metal Intel i7-12700, Ubuntu 24.04, Linux 6.14
- `/data/cubelet` on XFS with `reflink=1` (verified via `xfs_info`), 30 GiB free
- `PoolType: cp_reflink_type` (probe in `checkPoolType` succeeded, so cubelet stayed on the reflink path)
- Template: `python:3.12-slim`, `--writable-layer-size 2Gi`

## Repro

```bash
# 1. Template
cubemastercli template create-from-image \
    --image python:3.12-slim \
    --template-id repro-pynp \
    --writable-layer-size 2Gi \
    --allow-internet-access

# 2. Fire 100 concurrent POST /sandboxes against cube-api,
#    then DELETE each on success.
```

## Observed

- Run A: **77/100** sandboxes spawned cleanly, wall-clock 20.3 s.
- Run B: **36/100** spawned cleanly, wall-clock 19.8 s.

Wall-clock is stable across runs; success rate is variable. Failed spawns log:

```
newExt4RawByReflinkCopy failed: e2fsck 1.47.0 (5-Feb-2023):
bad magic number in superblock
```

Variable success rate + stable wall-clock suggests a timing-sensitive
concurrency issue rather than a deterministic config problem.

## Analysis

I read `Cubelet/storage/{shell.go, pool.go, plugin.go}`. A few observations
that may be relevant — happy to PR fixes for the mechanical ones if you
agree on direction.

### 1. `cmdTimeout = 3 * time.Second` may be too short for multi-GiB ext4 ops

[`shell.go`](https://github.com/TencentCloud/CubeSandbox/blob/master/Cubelet/storage/shell.go) defines:

```go
const cmdTimeout = time.Second * 3
```

`newExt4RawByReflinkCopy` runs four commands under this single timeout:

```
cp --reflink=always base target
truncate -s <newsize> target
e2fsck -fy target
resize2fs target
```

For `--writable-layer-size 2Gi`, `target` is truncated to ~3 GiB. Under
N=100 concurrent invocations sharing one XFS volume, `e2fsck -fy` and
[...]

PR #236, opened 2026-05-14
storage: make cmdTimeout configurable via plugin config

WaylandYang:
Implements proposal **(1)** from #235.

Adds a new optional `cmd_timeout` field to the storage plugin's TOML config. When unset (or zero) the existing 3 s default is preserved exactly — no behavior change for current deployments.

## Motivation

`shell.go`'s `cp / truncate / e2fsck / resize2fs / mkfs.ext4` chain shares a single 3 s timeout per command. On the pre-formatted pool fast path the small images comfortably finish inside 3 s. On the live-create slow path (`writable_layer_size` that doesn't match `pool_default_format_size_list`), 3 s on multi-GiB images is tight under concurrent load — that's the contention window where the reflink-copy race in #235 surfaces.

Letting operators raise the per-command timeout from config is the smallest knob that lets the live-create path survive without recompiling cubelet.

## Config example

```toml
[plugins."io.cubelet.internal.v1.storage"]
  cmd_timeout = "30s"
```

- **Default value stays 3 s.** Bumping the default is a behavior change you may want to do separately after measuring P95 on a host that actually hits the slow path.
- **Single timeout shared by all the commands.** Splitting per-command (e.g. faster `cp`, slower `e2fsck`) is doable but expands scope; happy to do as a follow-up PR if you'd like.
- `shell.go`'s `cmdTimeout` is now `var` instead of `const`, set during `(*local).init`. Other styles (passing timeout into each function, plumbing through a `context`) are also options; this matches the existing package-level state in this package (`localStorage`, `defaultPoolSize`, ...).

## Out-of-scope observations

- `shell.go`'s three create-image functions (`newExt4BaseRaw`, `newExt4RawByCopy`, `newExt4RawByReflinkCopy`) all swallow stderr into a short string error without command index / file stat / elapsed time. Proposal **(2)** from #235 covers that — **separate PR coming, keeping this one focused.**

## Tested

- `go vet ./storage/...` — clean
- `go build ./storage/...` — clean
- `gofmt -l` — clean

### DIFF
--- a/Cubelet/storage/local.go
+++ b/Cubelet/storage/local.go
@@ -75,6 +75,7 @@ var (
 	defaultPoolTriggerIntervalInMs = 1000
 	defaultWarningPercent          = 100
 	defaultFormatSize              = "1Gi"
+	defaultCmdTimeout              = 3 * time.Second
 
 	unifiedStorageSize        = resource.MustParse(defaultFormatSize)
 	otherFormatSize           = "othersv2"
--- a/Cubelet/storage/plugin.go
+++ b/Cubelet/storage/plugin.go
@@ -50,6 +50,14 @@ type Config struct {
 
 	FreeInodesThreshold int32            `toml:"free_inodes_threshold"`
 	ReconcileInterval   tomlext.Duration `toml:"reconcile_interval"`
+
+	// CmdTimeout overrides the per-command timeout for utils.ExecV
+	// invocations in shell.go (cp / truncate / e2fsck / resize2fs /
+	// mkfs.ext4). Defaults to defaultCmdTimeout when zero. The slow
+	// ext4-create path on multi-GiB images can need noticeably more
+	// than the 3s default; this knob lets operators bump it without
+	// recompiling.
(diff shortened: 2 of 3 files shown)


Cubelet/storage/local.go
```diff
@@ -178,6 +179,10 @@ func (l *local) init(ic *plugin.InitContext) error {
 	if l.config.ReconcileInterval == 0 {
 		l.config.ReconcileInterval = tomlext.FromStdTime(5 * time.Minute)
 	}
+	if l.config.CmdTimeout == 0 {
```
hlz260: I think it would be more appropriate to move this initialization code to the InitFn in plugin.go

WaylandYang: Thanks @hlz260 — good catch, same ordering issue cubebot flagged. Pushed an update that moves `cmdTimeout` init into `InitFn` before `checkPoolType`, plus a non-negative guard and the comment cleanup cubebot also suggested. PTAL.

hlz260 (approved):
LGTM

PR #246, opened 2026-05-14
fix: change default artifact temp dir to storage partition to prevent…

### LINKED ISSUE #240: [Bug Report] 临时目录空间不足导致模版构建失败
# Issue: 临时目录空间不足导致模版构建失败

## 问题描述

CubeSandbox 在从 OCI 镜像创建模版时，模版创建进度卡在 `UNPACKING` 20% 左右，最终超时失败。

## 环境信息

- 磁盘布局：
  - 根分区 `/dev/vda2`: 50G，系统占用后仅剩 ~5G 空闲
  - 数据分区 `/dev/vdb`: 200G，挂载至 `/data/cubelet`，有 ~150G 空闲

### 1. 临时工作目录默认在根分区

CubeSandbox (cubemaster) 在创建模版时需要大量临时空间用于：
- 镜像解压
- rootfs 构建
- ext4 文件系统创建

默认情况下，临时工作目录 `CUBEMASTER_ROOTFS_ARTIFACT_DIR` 为 `/tmp/cubemaster-rootfs-artifacts/`，位于根分区。

### 2. 空间需求

对于一个 1.6GB 的 OCI 镜像，模版创建过程大约需要 8-10GB 的临时空间：
- 镜像层解压
- 中间文件
- ext4 镜像构建

根分区仅剩 5GB 空间，无法满足需求。

### 修改启动脚本 `/usr/local/services/cubetoolbox/scripts/one-click/up.sh`

添加 `CUBEMASTER_ROOTFS_ARTIFACT_DIR` 环境变量，将临时工作目录指向大分区：

```bash
# 新增：临时工作目录配置
CUBEMASTER_ROOTFS_ARTIFACT_DIR_DEFAULT="/data/cubelet/tmp/cubemaster-rootfs-artifacts"
CUBEMASTER_ROOTFS_ARTIFACT_DIR_CONFIGURED="${CUBEMASTER_ROOTFS_ARTIFACT_DIR:-}"
CUBEMASTER_ROOTFS_ARTIFACT_DIR="${CUBEMASTER_ROOTFS_ARTIFACT_DIR_CONFIGURED:-${CUBEMASTER_ROOTFS_ARTIFACT_DIR_DEFAULT}}"

# 配置临时工作目录（大分区）
if [[ -n "${CUBEMASTER_ROOTFS_ARTIFACT_DIR_CONFIGURED}" ]]; then
  mkdir -p "${CUBEMASTER_ROOTFS_ARTIFACT_DIR}"
  CUBEMASTER_ARTIFACT_DIR_EXPORT="export CUBEMASTER_ROOTFS_ARTIFACT_DIR=\"${CUBEMASTER_ROOTFS_ARTIFACT_DIR}\";"
elif mkdir -p "${CUBEMASTER_ROOTFS_ARTIFACT_DIR}" >/dev/null 2>&1; then
  CUBEMASTER_ARTIFACT_DIR_EXPORT="export CUBEMASTER_ROOTFS_ARTIFACT_DIR=\"${CUBEMASTER_ROOTFS_ARTIFACT_DIR}\";"
else
  log "cubemaster artifact work dir ${CUBEMASTER_ROOTFS_ARTIFACT_DIR} unavailable, using /tmp fallback"
fi

# 启动时传递环境变量
start_with_pidfile \
  "cubemaster" \
  "export CUBE_MASTER_CONFIG_PATH=\"${CUBEMASTER_CFG}\"; ${CUBEMASTER_ARTIFACT_STORE_EXPORT} ${CUBEMASTER_ARTIFACT_DIR_EXPORT} \"${CUBEMASTER_BIN}\""
```

## 环境变量说明

| 变量 | 用途 | 默认值 | 建议值 |
|------|------|--------|--------|
| `CUBEMASTER_ROOTFS_ARTIFACT_DIR` | 临时工作目录 | `/tmp/cubemaster-rootfs-artifacts` | `/data/cubelet/tmp/cubemaster-rootfs-artifacts` |
| `CUBEMASTER_ROOTFS_ARTIFACT_STORE_DIR` | 最终存储目录 | `/data/CubeMaster/storage` | `/data/CubeMaster/storage` |

## 建议改进

1. **文档更新**: 在部署文档中明确说明磁盘空间要求，建议根分区至少保留 20GB 空闲空间

2. **默认配置优化**: 建议官方安装脚本/配置默认将临时目录设置到大分区

3. **空间检查**: 在模版创建前检查临时目录可用空间，不足时提前报错而非卡住

4. **清理机制**: 添加失败模版的临时文件自动清理机制

## 验证方法

```bash
# 1. 检查环境变量是否生效
cat /proc/<cubemaster-pid>/environ | tr '\0' '\n' | grep CUBEMASTER_ROOTFS_ARTIFACT_DIR

# 2. 监控临时目录使用情况
watch -n 5 'du -sh /data/cubelet/tmp/cubemaster-rootfs-artifacts/*'

# 3. 检查模版创建进度
cubemastercli tpl ls
```

Jxt-Eli:
… disk full errors during extraction
Fixes #240

**What this PR does:**
This PR addresses the issue where template creation fails due to `/tmp` running out of space. 

Instead of defaulting to `os.TempDir()` (which would cause issues when disk space isn't enough), `artifactWorkRootDir()` in `template_image.go` will now dynamically default to creating a `tmp` folder inside the configured `artifactStoreRootDir` unless an environment variable is explicitly set by the user. 

**Why this is better:**
1. It prevents root partition disk exhaustion.
2. It improves performance by keeping intermediate files on the same disk as the final storage, avoiding cross-device copy penalties.

(Note: This is my first open-source contribution! Let me know if any changes are needed to the code style or structure.)

### DIFF
--- a/CubeMaster/pkg/templatecenter/template_image.go
+++ b/CubeMaster/pkg/templatecenter/template_image.go
@@ -2005,7 +2005,8 @@ func artifactWorkRootDir() string {
 	if value := strings.TrimSpace(os.Getenv("CUBEMASTER_ROOTFS_ARTIFACT_DIR")); value != "" {
 		return value
 	}
-	return filepath.Join(os.TempDir(), "cubemaster-rootfs-artifacts")
+	storeRoot := artifactStoreRootDir()
+	return filepath.Join(storeRoot, "tmp", "cubemaster-rootfs-artifacts")
 }
 
 func artifactStoreRootDir() string {
--- a/CubeMaster/pkg/templatecenter/template_image.go
+++ b/CubeMaster/pkg/templatecenter/template_image.go
@@ -72,6 +72,7 @@ const (
 	fallbackArtifactStoreDir   = "cubemaster-rootfs-artifacts-store"
 	rootfsWritableVolumeName   = "cube_rootfs_rw"
 	defaultDistributionWorkers = 4
+	requiredExtractionSpace    = 20 << 30
 )
 
 var getTemplateImageConfig = config.GetConfig
@@ -2000,13 +2001,29 @@ func artifactModelToInfo(record *models.RootfsArtifact) *types.RootfsArtifactInf
 		LastError:               record.LastError,
 	}
 }
+func checkPartition(tempPath string) error {
+	var stat syscall.Statfs_t
+	err := syscall.Statfs(tempPath, &stat)
+	if err != nil {
+		return fmt.Errorf("error! Failed to check disk space at %s: %w", tempPath, err)
+	}
+	available := stat.Bavail * uint64(stat.Bsize)
+	if available < requiredExtractionSpace {
+		return fmt.Errorf("error: not enough space on %s\nRequired 20GiB\n Available: %dGiB", tempPath, available>>30)
+	}
+	return nil
+}
 
 func artifactWorkRootDir() string {
+	tempDir := os.TempDir()
 	if value := strings.TrimSpace(os.Getenv("CUBEMASTER_ROOTFS_ARTIFACT_DIR")); value != "" {
 		return value
 	}
-	storeRoot := artifactStoreRootDir()
-	return filepath.Join(storeRoot, "tmp", "cubemaster-rootfs-artifacts")
+	if err := checkPartition(tempDir); err == nil{
+		return filepath.Join(tempDir, "cubemaster-rootfs-artifacts")
+	}
+	fallbackDir := artifactStoreRootDir()
+	return filepath.Join(fallbackDir, "tmp", "cubemaster-rootfs-artifacts")
 }
 
 func artifactStoreRootDir() string {
--- a/README.md
+++ b/README.md
@@ -147,7 +147,7 @@ Cube Sandbox requires an x86_64 Linux environment with KVM support — **WSL 2**
 > - **Bare-metal / physical machine users**: grab an x86_64 Linux physical machine, or rent a bare-metal server from a cloud provider.
 > - **Ordinary cloud VM users**: no bare-metal required — install the PVM host kernel to enable KVM on any standard cloud VM. See [PVM Deployment](./docs/guide/pvm-deploy.md).
 >
-> **NOTE on storage**: Ensure the partition hosting the CubeMaster storage directory (usually `/data`) has at least 20GB of free space for template creation.
+> **NOTE on storage**: Ensure the partition hosting the CubeMaster storage directory (usually `/data`) has at least 20GiB of free space for template creation.
 
 Once your environment is ready, launch your first sandbox in four steps:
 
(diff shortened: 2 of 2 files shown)


fslongjin:
Hi @Jxt-Eli,

Thanks for the PR! The DCO (Developer Certificate of Origin) check is failing. You need to add a `Signed-off-by` line to your commit message.

You can fix this by amending your commit:

```bash
git commit --amend --signoff
git push --force
```

This will append a `Signed-off-by: Your Name <your-email@example.com>` line to the commit message, which is required by the project's DCO policy. See [DCO](https://developercertificate.org/) for details.

fslongjin:
Hi @Jxt-Eli,

Thanks for the PR. I think the current approach works around the issue rather than addressing the root cause. Silently changing the default temp dir to the storage partition changes behavior in a way that may not be expected.

I think a more robust solution would be to add **disk space detection** — check available space before template creation and fail early with a clear error message if there isn't enough room, rather than letting it fail halfway with a confusing error. This way users know exactly what to fix (e.g., free up space or configure `CUBEMASTER_ROOTFS_ARTIFACT_DIR`).

Would you be interested in exploring this approach?

Jxt-Eli:
I've pushed the requested change and added a signed-off-by commit.

fslongjin:
Hi @Jxt-Eli, thanks for the update! The fallback logic (try `/tmp` first, then fall back to storage partition) is structurally much better now.

However, I still have concerns about `requiredExtractionSpace = 20 << 30` being a hardcoded arbitrary threshold. It's the same problem in a different form — different rootfs images vary wildly in size (some are a few hundred MB, others could be much larger). Hardcoding 20GB means:

- **For small images**: it's an artificial barrier — 5GB free might be plenty but the code falls back unnecessarily.
- **For very large images**: it could be optimistic — 20GB might still not be enough.

I think a more robust approach would be:

1. **Estimate required space from the actual artifact** — derive the needed space from the compressed image size (which is known from the image layer metadata) and apply an expansion factor (e.g., `compressedSize * 3` for temp space + final extracted size headroom). This makes the check dynamic and accurate per-template.

2. **Graceful error handling when disk runs out** — if extraction fails mid-way with `ENOSPC`, catch that error and return a clear actionable message like: `"extraction failed: not enough disk space on /tmp (needed X GB, available Y GB). Set CUBEMASTER_ROOTFS_ARTIFACT_DIR to a partition with more space."`

This would eliminate the hardcoded constant while still giving users early, actionable feedback. What do you think?

Jxt-Eli:
@fslongjin where do I find the size of the tar file from?
I initially thought I could use the `tarPath` value from `exportImageRootfs` to compute the value of the tar size for use in `checkPartition()` . but I cant use a global variable since I'd be risking data race problems if there are multiple concurrent instances of the file running. 

Furthermore, here is why I think estimating the size of the extracted file is impossible: 
-  The `checkPartition()` function I defined to help `artifactRootWorkDir` select a directory needs a tar path to estimate the size of the size of the extracted package
- `exportImageRootfs` is the only file that has the tar path, and before its able to get the tar path, it takes `workDir` as argument which is derived like this; 
	`workDir := filepath.Join(artifactWorkRootDir(), record.ArtifactID)`
- This implies that `exportImageRootfs` runs before `artifactWorkRootDir`, and this makes it impossible to get `tarPath` before the function that returns `tarPath` even runs.

Jxt-Eli:
@fslongjin 
If I could think of anything close to a solution, I think the user themselves should put in the size of their tar file so that the program gives the estimate and begins the extraction. Should I do it that way?

PR #254, opened 2026-05-14
Add Go SDK

nemo1105:
Add a standalone standard-library Go SDK for CubeSandbox with lifecycle, execution, command, file read, proxy transport, README, unit tests, and opt-in integration tests.

Autonomously-by: Codex:GPT-5

### DIFF
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -1,0 +1,159 @@
+# cubesandbox Go SDK
+
+Go SDK for [CubeSandbox](https://github.com/TencentCloud/CubeSandbox). It matches the current Python SDK surface: sandbox lifecycle, code execution, commands, and file reads only.
+
+## Install
+
+```bash
+go get github.com/tencentcloud/CubeSandbox/sdk/go
+```
+
+## Configuration
+
+```bash
+export CUBE_API_URL=http://127.0.0.1:3000
+export CUBE_TEMPLATE_ID=<your-template-id>
+
+# Optional remote data-plane access.
+export CUBE_PROXY_NODE_IP=<cubeproxy-node-ip>
+export CUBE_PROXY_PORT_HTTP=80
+export CUBE_PROXY_SCHEME=http
+export CUBE_SANDBOX_DOMAIN=cube.app
+```
+
+`NewConfigFromEnv` also accepts `E2B_API_URL` and `E2B_API_KEY`; `CUBE_API_URL` and `CUBE_API_KEY` take precedence.
+`CUBE_PROXY_SCHEME` supports `http` and `https`; when omitted, port `443` defaults to `https` and other ports default to `http`.
+
+## Create And Run Code
+
+```go
+package main
+
+import (
+	"context"
+	"fmt"
+
+	cubesandbox "github.com/tencentcloud/CubeSandbox/sdk/go"
+)
+
+func main() {
+	ctx := context.Background()
+	client := cubesandbox.NewClient(cubesandbox.NewConfigFromEnv())
+
+	sb, err := client.Create(ctx, cubesandbox.CreateOptions{})
+	if err != nil {
+		panic(err)
+	}
+	defer sb.Kill(ctx)
+
+	exec, err := sb.RunCode(ctx, "x = 41\nx + 1", cubesandbox.RunCodeOptions{})
+	if err != nil {
+		panic(err)
+	}
+	fmt.Println(exec.Text)
+}
+```
+
+## Commands
+
+```go
+result, err := sb.Commands().Run(ctx, "echo hello", cubesandbox.CommandOptions{})
+if err != nil {
+	panic(err)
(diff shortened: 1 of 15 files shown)


tinklone:
## `Commands.Run` should use envd Process API instead of Python code execution

The current `Commands.Run` implementation wraps shell commands in Python code (`subprocess.run(...)`) and executes them via the `/execute` endpoint. This is architecturally wrong — it should call envd's `process.Process/Start` (Connect Protocol) interface directly.

### How E2B does it (the correct approach)

E2B's official SDK calls:
```
POST http://49999-{sandboxID}.{domain}/process.Process/Start
```
With payload:
```json
{
  "process": {
    "cmd": "/bin/bash",
    "args": ["-l", "-c", "<user-command>"],
    "envs": {},
    "cwd": "<optional>"
  }
}
```

The response is a server-stream of `ProcessEvent` messages:
- `StartEvent` → gives PID
- `DataEvent` → stdout/stderr bytes
- `EndEvent` → exit_code, exited status

### Why the current approach is problematic

1. **Requires Python in the sandbox** — breaks for Go/Rust/C-only templates
2. **Performance overhead** — spawns a Python interpreter just to call `subprocess.run`
3. **Fragile exit code handling** — parsing return code from expression result text vs getting `EndEvent.exit_code` directly
4. **Inconsistent with E2B SDK** — the Go SDK should behave the same as the Python/JS SDKs

### Suggested fix

Implement a Connect Protocol (gRPC over HTTP/JSON) client for `process.Process/Start` in the Go SDK. The proto definition is at `spec/envd/process/process.proto`. The wire format is simple HTTP POST with JSON body + server-sent streaming response.

This also enables implementing `Commands.List`, `Commands.Kill`, `Commands.SendStdin`, and `Commands.Connect` — all available via the same envd process service.

Similarly, `Files.Read` should use the envd HTTP file API (`GET /files?path=...`) instead of generating Python `open()` code through `/execute`.

tinklone (approved):
## CR Review: New Commits (6b03855d + 17aad989)

**Summary**: Refactored Commands/Files from Python RunCode hack to native envd Process API + HTTP File API. Significant improvement in correctness and maintainability.

### ✅ Highlights

1. **Architecture**: `Commands` and `Files` now use independent `processStarter` / `fileReader` interfaces instead of shared `codeRunner` — cleaner SRP
2. **Correctness**: Exit code from `EndEvent` field instead of fragile stdout last-line parsing
3. **Connect protocol**: Proper envelope parsing, end-stream error handling, compressed flag rejection
4. **ProxyScheme**: Smart default (port 443 → https, else http) with explicit override support
5. **Test coverage**: New integration + unit tests with clean mock boundaries

### 🟡 Minor Suggestions (non-blocking)

| # | Location | Issue | Suggestion |
|---|----------|-------|------------|
| 1 | `envd.go:186` | `io.ErrUnexpectedEOF` treated as normal EOF in `readConnectEnvelope` | Should return error for mid-stream truncation (only `io.EOF` is clean end) |
| 2 | `envd.go:167` | No break after `sawEnd = true` in stream parse loop | After EndEvent, remaining frames are unexpected — consider breaking early |
| 3 | `envd.go:81` | Dual field `ExitCode` + `ExitCodeSnake` without comment | Add a brief comment explaining why both exist (backend version compat?) |
| 4 | `commands.go:21` | Empty map init for nil Envs | If envd accepts `"envs":null`, this is unnecessary; if not, keep it but add a comment |

### ❓ Questions for Author

1. **64MB envelope limit**: Could a large stdout `cat` OOM? Consider streaming for large outputs.
2. **`/bin/bash -l -c`**: Login shell loads `.bash_profile` etc. Is this intentional? May add latency in some images.
3. **`readFile` unbounded**: `io.ReadAll` without `LimitReader` — risk for very large files?

### Verdict

**LGTM** — core logic correct, tests thorough. Minor items can be follow-up.

tinklone:
Hi @wangchenglong-hj 👋

CI has two failing checks, one of which requires action on your side:

### ❌ DCO (Developer Certificate of Origin)

All 3 commits are missing the `Signed-off-by` line:

```
37b25030 Add Go SDK
6b03855d Fix Go SDK command exit code handling
17aad989 fix(go-sdk): use envd process and file APIs
```

To fix:

```bash
git rebase --signoff HEAD~3 && git push --force-with-lease
```

This will automatically append to each commit message:
```
Signed-off-by: wangchenglong-hj <wangchenglong-hj@huafang.com>
```

Reference: https://developercertificate.org/

---

### ⚠️ review check (no action needed from you)

This is a repo CI configuration issue (GitHub App token missing `app-id`). Unrelated to your PR code — known issue.

---

Code itself is LGTM ✅ — once the DCO sign-off is added, this is good to merge.

ls-ggg:
Could you please squash these commits into a single one? Thanks! @wangchenglong-hj

PR #256, opened 2026-05-14
Add aarch64 support across CubeSandbox

lkml-likexu:
# Add aarch64 support across CubeSandbox

## Motivation

CubeSandbox until now has been an x86_64-only stack: the builder image,
the MicroVM kernel pipeline, the hypervisor / shim / guest agent and
several Go components all carried hard-coded `x86_64` / `amd64`
assumptions. As we start onboarding ARM64 hosts (Ampere, Graviton,
Yitian, Apple Silicon dev boxes), running the same control plane and
sandbox runtime on aarch64 is a recurring blocker.

This PR brings CubeSandbox to a state where, on a stock aarch64 Linux
host, the builder image, the MicroVM kernel, the hypervisor, the shim,
the in-guest agent and the relevant Go services can all be built and
brought up. It also drags in two unrelated correctness fixes that
surfaced during the bring-up.

## Scope

The series is intentionally split per component, atomic per commit,
roughly along the layering of the stack:

1. **`deploy/aarch64`: reproducible MVM vmlinux build pipeline** —
   adds a self-contained Dockerfile + `build_mvm_vmlinux.sh` + pinned
   `mvm.config` / `mvm.cmdline` so anyone can reproduce the aarch64
   guest kernel with a single command, decoupled from any host distro.
2. **`docker/Dockerfile.builder`: multi-arch builder image** —
   resolves Go / protoc / Rust musl target / libseccomp cross knobs
   from a single `HOST_ARCH` (auto-detected) instead of the old
   amd64-only constants, so the build container itself works on both
   architectures.
3. **`hypervisor`: build and runtime regressions on aarch64** —
   realigns the KVM, snapshot and migration call sites, exposes
   `SysCtrl` over MMIO so the guest can still notify the shim of
   shutdown / reboot, and tightens cross-arch cfg gates.
4. **`shim`: enable aarch64 support for CubeShim** — generalizes the
   guest kernel cmdline (console, x86-only mitigation knobs) and the
   seccomp allow-lists used by the hypervisor / snapshot workers so
   that the shim can both boot a guest and pass syscall filtering on
   ARM64.
5. **`agent`: enable aarch64 for the guest agent** — drops the
   hard-coded `x86_64-unknown-linux-musl` triple in the build, and
   replaces the PIO-based "vsock server is ready" notification with
   an MMIO write into the SysCtrl region exposed by the hypervisor,
   matching what the shim listens for.
6. **`cubenet`: enable arm64 build for the cubevs eBPF loader** —
   adds an arm64-only Go file mirroring the existing loader API and
   embeds the already checked-in little-endian eBPF objects, so
   CubeNet builds and the cubevs data plane can be brought up on
   ARM64 without introducing a separate set of compiled artifacts.
7. **`cubelet`: image pull + overlay snapshot edge cases** — fixes a
   handful of latent bugs uncovered while exercising additional
   sandbox flavors during ARM64 bring-up: the parsed cube image spec
   was not threaded through the request context, the ext4 runtime
   artifacts were not refreshed after the rootfs was prepared, and
   the overlay snapshotter would panic on a `nil` `Labels` map. A
   regression test covers the snapshotter case.
8. **`cubemaster`: drop the legacy gomonkey v1 dependency** —
   consolidates onto `gomonkey/v2` (already used everywhere else),
   removing the unmaintained v1 module from `go.sum` and shrinking
   the dependency graph.

## Compatibility

- **x86_64**: no behavioural change is intended on any commit. Builds
  and runtime paths previously exercised on x86_64 are kept on the
  exact same code paths via `cfg(target_arch = "x86_64")` /
  `//go:build amd64` guards or via `target.'cfg(...)'.dependencies`
  in Cargo.
- **aarch64**: now a first-class build & runtime target. Unsupported
  architectures (anything other than amd64 / arm64) fail fast with a
  clear error in the builder image and the kernel pipeline.

## How to test

On an aarch64 Linux host with Docker:

```bash
# 1) Builder image (now multi-arch).
make builder-image

# 2) Reproducible aarch64 MicroVM kernel.
./deploy/aarch64/build_mvm_vmlinux.sh

# 3) Native build of the Rust + Go components inside the builder.
make hypervisor shim agent cubelet cubemaster
```

On an x86_64 host, the same commands should keep producing the same
artifacts as before this PR.


CubeNet/cubevs/bpf_arm64.go
```diff
@@ -0,0 +1,57 @@
+// SPDX-License-Identifier: Apache-2.0
```
chenhengqi, line 1: Have you tested the change to CubeVS? It seems wrong to me.

We will need at least:
* A new vmlinux.h for arm64
* Add go:generate command for arm64 in `cubevs.go`

lkml-likexu: I logged into the MVM container, and the network is working fine (e.g., `yum update`). If there are any other CubeVS specific tests, please let me know.

lkml-likexu:
We have tested this PR on the following platfroms:
- Kunpeng 920
- Apple M3 Series

Host Kernel:
- 5.4
- 6.6

If you encounter any issues on other arm64 platforms or kernels, please let me know—or better yet, we encourage you to contribute.

lisongqian:
Good job! Have you run run_integration_tests_aarch64.sh test in the hypervisor? Some device implementations in hypervisor only support x86, such as ivshmem. We need to enable the CI for aarch64.

kinwin-ustc:
I think we can divide it into several series, compiling-related cases and running-related cases, where the running-related ones can be submitted separately for better review. Some commits in this series, such as commit 0b3daae, have already been better fixed by PR #341

lkml-likexu:
> Good job! Have you run run_integration_tests_aarch64.sh test in the hypervisor? Some device implementations in hypervisor only support x86, such as ivshmem. We need to enable the CI for aarch64.干得好！你在虚拟机监控程序里跑过 run_integration_tests_aarch64.sh 测试吗？虚拟机监控程序中有些设备实现只支持 x86，比如 ivshmem。我们需要启用 aarch64 的 CI。

Regarding the hypervisor component, I believe we should migrate to the upstream cloud-hypervisor. This would allow us to unlock and leverage more of CH's features (in this case, arm64 support) and CI infrastructure, while consolidating open-source efforts.

lisongqian:
> Regarding the hypervisor component, I believe we should migrate to the upstream cloud-hypervisor. This would allow us to unlock and leverage more of CH's features (in this case, arm64 support) and CI infrastructure, while consolidating open-source efforts.

I guess you want to rebase to the latest upstream? I'm not sure if this is feasible. I prefer to cherry-pick the patch that we need :)

lkml-likexu:
For developers who want to try CubeSandbox on ARM, feel free to use and modify the patches in this PR. We encourage you to contribute new patches based on these (including component fixes and performance optimizations) back to the CubeSandbox community.

PR #269, opened 2026-05-14
fix(deploy): bind MySQL/Redis to localhost and switch CubeProxy to host networking

fslongjin:
## Summary
- Bind MySQL (3306) and Redis (6379) ports to `127.0.0.1` only, preventing external network access to these services with default credentials
- Switch CubeProxy from Docker bridge networking to `network_mode: host` so it can reach Redis via localhost (following CoreDNS's existing pattern)
- Add parameterized `nginx.conf.template` so CubeProxy nginx listens on configurable ports (default 443/80), maintaining the same external-facing behavior

## Security Impact
Before: MySQL and Redis were accessible on all network interfaces (`0.0.0.0`) with weak default credentials. Any client that could reach the host IP could connect.

After: Both services are only accessible from `127.0.0.1`. External connections are refused.

## Files Changed
| File | Change |
|------|--------|
| `deploy/one-click/support/docker-compose.yaml.template` | MySQL/Redis port binding → `127.0.0.1` |
| `deploy/one-click/cubeproxy/docker-compose.yaml.template` | `network_mode: host`, remove `ports`, add nginx.conf volume |
| `deploy/one-click/cubeproxy/nginx.conf.template` | **New**: parameterized nginx config with `__CUBE_PROXY_HTTPS_PORT__`/`__CUBE_PROXY_HTTP_PORT__` |
| `deploy/one-click/scripts/one-click/up-cube-proxy.sh` | Redis IP default → `127.0.0.1`, nginx template processing, port variables |
| `deploy/one-click/env.example` | Updated variables |

## Test plan
- [ ] Run full one-click install flow, verify all services start normally
- [ ] `ss -tlnp | grep -E '3306|6379'` confirms MySQL/Redis only on `127.0.0.1`
- [ ] Verify external IP connections to 3306/6379 are refused
- [ ] Verify CubeMaster connects to MySQL/Redis normally
- [ ] Verify CubeProxy connects to Redis and HTTPS proxy works
- [ ] `ss -tlnp | grep -E ':443|:80'` confirms CubeProxy listens on 443/80


deploy/one-click/cubeproxy/nginx.conf.template
```diff
@@ -0,0 +1,181 @@
+user root;
+worker_processes 12;
```
staryxchen: As a template, it's best not to hard-code it; use `auto` instead.

deploy/one-click/cubeproxy/nginx.conf.template
staryxchen, line 1: Please note that this is for reference only and should not be used directly in a production environment.

deploy/one-click/cubeproxy/nginx.conf.template
fslongjin: Addressed in 68fddf80. The `nginx.conf.template` is no longer hand-maintained: `deploy/one-click/build-release-bundle.sh` now derives it at build time from the upstream `CubeProxy/nginx.conf`, rewriting `worker_processes 12` to `worker_processes auto` (and parameterizing the listen ports / TLS cert paths) during generation. The single source of truth is upstream `CubeProxy/nginx.conf`.

deploy/one-click/cubeproxy/nginx.conf.template
fslongjin: Good point. The generator in `build-release-bundle.sh` now prepends an explicit header to the produced template, calling out that the file is auto-generated from `CubeProxy/nginx.conf`, that the configuration is for reference only and not production-ready as-is, and explaining the deliberate lack of an HTTP→HTTPS redirect on port 80. Generated header excerpt:

```
# NOTE: This file is auto-generated by deploy/one-click/build-release-bundle.sh
# from CubeProxy/nginx.conf. DO NOT edit by hand; modify the upstream
# CubeProxy/nginx.conf and re-run the bundle build.
#
# The configuration below is provided for reference only and is NOT meant for
# production use as-is. Tune worker_processes, buffer/limit sizes, timeouts and
# certificate paths according to your environment.
```

See 68fddf80.

fslongjin:
Thanks @staryxchen and the bot review. Pushed 68fddf80 to address the feedback.

### Core change: eliminate the duplicated nginx.conf

Instead of hand-maintaining `deploy/one-click/cubeproxy/nginx.conf.template` alongside the upstream `CubeProxy/nginx.conf`, the template is now **auto-generated at build time** by `deploy/one-click/build-release-bundle.sh` from the upstream file. The generator:

- rewrites `worker_processes 12` to `worker_processes auto`
- replaces the listen ports `8081` / `8080` with `__CUBE_PROXY_HTTP_PORT__` / `__CUBE_PROXY_HTTPS_PORT__`
- replaces the hard-coded SSL cert paths with `__CUBE_PROXY_SSL_CERT__` / `__CUBE_PROXY_SSL_KEY__`
- prepends a header that documents the auto-generated nature, the for-reference-only caveat, and the deliberate lack of an HTTP→HTTPS redirect
- asserts all four placeholders are present in the output (fail-fast if upstream changes break the sed pattern)

Single source of truth is now `CubeProxy/nginx.conf`.

### Per-item response

| # | Reviewer | Item | Resolution |
|---|----------|------|------------|
| 1 | staryxchen | `worker_processes 12` should be `auto` | Done — rewritten during template generation |
| 2 | staryxchen | "For reference only, not production-ready" | Done — generator prepends a header that says exactly that |
| 3 | bot | `user root;` runs nginx workers as root | Not changed in this PR. The directive is inherited from the upstream `CubeProxy/nginx.conf` and the cube-proxy image expects worker write access to `/data/log/cube-proxy/` and `/cache`. Switching to a non-privileged user requires coordinated changes in `CubeProxy/Dockerfile` (chown of those dirs) and is better handled as a follow-up so this PR stays scoped to the host-networking + secret-binding fix |
| 4 | bot | SSL cert paths hard-coded (`cube.app+3.pem`) | Done — parameterized via `CUBE_PROXY_SSL_CERT` / `CUBE_PROXY_SSL_KEY` env vars (defaults preserve mkcert's naming). When users override the names, `up-cube-proxy.sh` skips mkcert auto-provisioning and requires the cert files to be placed manually |
| 5 | bot | Readiness check only verifies HTTPS port | Done — `up-cube-proxy.sh` now waits for both HTTP and HTTPS to be listening, with separate per-port error messages on timeout |
| 6 | bot | No HTTP→HTTPS redirect | Intentional. Port 80 is used to proxy plain-HTTP backend traffic; redirecting would break those clients. The generated template header now documents this design |
| 7 | bot | `network_mode: host` port-conflict risk | Done — `up-cube-proxy.sh` now performs a preflight check on both HTTP/HTTPS ports before `compose up` and exits with a clear "port X already in use; cube-proxy uses host networking and requires it free" message instead of letting nginx fail later with a cryptic "address in use" |
| 8 | bot | `CUBE_PROXY_DNS_ANSWER_IP` still uses node IP | Already by design (sandbox-internal lookups need a reachable IP). Added a comment to `env.example` to make this explicit |

### Validation

- Bundle build via `bash deploy/one-click/build-release-bundle-builder.sh` completes cleanly.
- Generated `sandbox-package/cubeproxy/nginx.conf.template` was diffed against the previous hand-maintained template — only the three intended differences remain (header, `worker_processes auto`, SSL placeholders).
- Simulated runtime sed (default 443/80 + default cert names) produces the same final `nginx.conf` as before, modulo `worker_processes auto`.

### Files changed

- `deploy/one-click/build-release-bundle.sh` — new `generate_cube_proxy_nginx_template` function and call site
- `deploy/one-click/cubeproxy/nginx.conf.template` — **deleted** (now generated)
- `deploy/one-click/scripts/one-click/up-cube-proxy.sh` — SSL cert vars, port preflight, dual-port readiness
- `deploy/one-click/env.example` — `CUBE_PROXY_SSL_CERT` / `CUBE_PROXY_SSL_KEY` and `CUBE_PROXY_DNS_ANSWER_IP` notes

PR #277, opened 2026-05-15
Docs/add contributing zh

QingqLin:


ls-ggg:
@QingqLin Please fix the suggestions from cubebot and squash the three commits into a single one.

Issue #274, opened 2026-05-15
[Feature Request] Support EROFS for image Rootfs and Guest OS images


novahe:
####  Motivation
CubeSandbox currently relies on `ext4` for Rootfs and Guest OS images. However, `ext4` is not the most efficient choice for read-only, high-density AI sandbox environments. It leads to larger image sizes and suboptimal IO performance during massive concurrent cold starts.

#### Proposed Solution
Support **EROFS (Enhanced Read-Only File System)** as a modern alternative to `ext4` to achieve:
*   **Reduced Image Size**: Leverage transparent compression (LZ4) to significantly cut down distribution bandwidth and storage overhead.
*   **Performance Optimization**: Faster metadata lookups and lower IO latency during high-concurrency cold starts.
*   **Native Integrity**: A natively immutable filesystem that provides better security for read-only Rootfs.

#### Proposed Implementation
*   **CubeMaster**: Integrate `mkfs.erofs` into the image building pipeline and update metadata to support EROFS media type.
*   **Cubelet**: Remove the hardcoded block for EROFS and support dynamic `FsType` for pmem devices.
*   **CubeShim**: Allow `rootfstype=erofs` in the VM boot configuration.

#### Additional Context
This alignment follows industry best practices for container runtimes, similar to the work being done in Kata Containers: [kata-containers/pull/13019](https://github.com/kata-containers/kata-containers/pull/13019).
> The NVIDIA images go down from 1.2GB to less than 500MB.

kinwin-ustc:
I agree with your proposal, can you help submit a PR?

PR #283, opened 2026-05-16
docs: add design proposal for EROFS rootfs support

novahe:
## Description
This PR introduces a comprehensive design proposal for integrating **EROFS (Enhanced Read-Only File System)** into the CubeSandbox runtime for **container Rootfs artifacts**. The goal is to optimize storage efficiency, distribution speed, and cold-start performance for read-only layers.

### Key Highlights:
- **End-to-End Lifecycle**: Defines the integration from CubeMaster build pipeline to Cubelet distribution and CubeShim runtime mount execution.
- **Engineering Hardening**: Specifies mandatory `erofs-utils` versioning (>= 1.5.0) and xattr support to preserve OverlayFS metadata (Whiteouts).
- **Performance & Observability**: Outlines a detailed benchmarking plan and critical metrics (compression ratios, cold-start latency) for production readiness.

**Note**: In this proposal, the Guest OS rootfs remains using the existing **ext4** scheme to maintain stability, while workload Rootfs artifacts are extended to support EROFS.

## Related Issues
Relates to #274

### DIFF
--- a/docs/proposals/support-erofs-rootfs/README.md
+++ b/docs/proposals/support-erofs-rootfs/README.md
@@ -1,0 +1,620 @@
+# KEP: Container Rootfs Support for EROFS
+
+## Proposal Information
+
+| Field | Content |
+|------|------|
+| Status | Draft |
+| Related Issue | https://github.com/TencentCloud/CubeSandbox/issues/274 |
+| Target Components | CubeMaster, Cubelet, CubeShim |
+| Major API Changes | Container Rootfs artifact supports `erofs` |
+| Default Behavior | Keep `ext4` |
+
+## Release Signoff Checklist
+
+- [ ] Design details cover CubeMaster, Cubelet, CubeShim, and Guest kernel prerequisites.
+- [ ] Test plan covers unit tests, integration tests, end-to-end validation, and ext4 regression.
+- [ ] Upgrade, downgrade, and rollback strategies are described.
+- [ ] Production readiness issues cover enablement/disablement, observability, dependencies, scalability, and troubleshooting.
+- [ ] User documentation synced for `create-from-image --rootfs-fs-type erofs` usage.
+
+## Summary
+
+Cube Sandbox currently primarily adopts `ext4` as the image format for container Rootfs artifacts. While this approach is mature and compatible, `ext4` is not the **optimal solution** in terms of image size, distribution bandwidth, and cold-start I/O overhead for high-density AI Agent sandbox scenarios.
+
+This proposal introduces EROFS (Enhanced Read-Only File System) for container Rootfs artifacts:
+
+- Container Rootfs artifacts support being built as `erofs` format from OCI images.
+- CubeMaster, Cubelet, and CubeShim propagate the actual filesystem type via explicit `fs_type` metadata.
+- Guest OS rootfs remains using the existing `ext4` scheme and will not be switched to EROFS in this proposal.
+- Default behavior remains `ext4`, maintaining full compatibility with old templates, installers, and README paths.
+
+Note: the following data is based on LZ4-compressed EROFS results (`mkfs.erofs -zlz4`).
+
+| Object | ext4 | erofs | Gain |
+|------|------|-------|------|
+| Container Rootfs `sandbox-code:latest` | 4.7 GB | 2.6 GB | ~45% size reduction |
+
+## Motivation
+
+The base Rootfs in Cube Sandbox serves as an immutable lower layer at runtime, with actual writes handled by a writable layer. For this pattern, a read-only filesystem provides a **natural architectural fit**. EROFS significantly reduces distribution size through LZ4 transparent compression while providing immutable read-only semantics, effectively eliminating the risk of accidental writes or data drift in base images.
+
+In scenarios where a single node hosts massive sandbox instances, image size and cold-start I/O directly impact node pull time, disk footprint, concurrent creation latency, and failure recovery speed. Supporting EROFS for container Rootfs enhances resource efficiency in high-density deployments without changing the upper-layer E2B SDK usage.
+
+> [!IMPORTANT]
+> **Trade-offs**: Compared to `ext4 + DAX`, compressed EROFS has the following downsides:
+> 1. **No DAX Support**: No zero-copy feature; data must go through decompression and Page Cache.
+> 2. **Resource Overhead**: Decompression consumes CPU, and Page Cache usage may increase memory pressure.
+>
+> `ext4 + DAX` remains the **performance baseline**. Decisions should be based on a balance between storage gains and performance overhead.
+
+### Goals
+
+- CubeMaster supports building `erofs` container Rootfs artifacts from OCI images.
+- Explicitly pass the Rootfs artifact `fs_type` between CubeMaster, Cubelet, and CubeShim.
+- Cubelet supports downloading, caching, verifying, and injecting EROFS pmem images.
+- CubeShim generates correct Guest mount behavior based on the workload pmem's `fs_type`.
+- Maintain compatibility with existing ext4 default behavior, old templates, and old installers.
+
+### Non-Goals
+
+- Do not change the runtime writable layer to EROFS. EROFS is only for the read-only base Rootfs.
+- Do not support EROFS as a standard data volume.
+- Do not modify the Guest OS rootfs format, Guest OS remains as ext4.
+- Do not implement layered EROFS, cross-template deduplication, higher compression parameter tuning, or multi-version image reuse.
+
+## Proposal
+
+### User Stories
+
+#### Story 1: Template Creator uses EROFS Container Rootfs
+
+A template creator wants to create a smaller read-only base Rootfs based on the Code Interpreter image in the README:
+
+```bash
+cubemastercli tpl create-from-image \
+  --image cube-sandbox-int.tencentcloudcr.com/cube-sandbox/sandbox-code:latest \
+  --rootfs-fs-type erofs \
+  --writable-layer-size 1G \
+  --expose-port 49999 \
+  --expose-port 49983 \
+  --probe 49999
+```
+
+Once the template is READY, users still create sandboxes using the original E2B SDK method:
+
+```bash
+export E2B_API_URL="http://127.0.0.1:3000"
+export E2B_API_KEY="dummy"
+export CUBE_TEMPLATE_ID="<template-id>"
+export SSL_CERT_FILE="/root/.local/share/mkcert/rootCA.pem"
+```
+
+```python
+import os
+from e2b_code_interpreter import Sandbox
+
+with Sandbox.create(template=os.environ["CUBE_TEMPLATE_ID"]) as sandbox:
+    result = sandbox.run_code("print('Hello from Cube Sandbox on EROFS!')")
+    print(result)
+```
+
+#### Story 2: Node Ops troubleshoots container Rootfs format
+
+A node ops person wants to quickly confirm whether a template uses EROFS. The system should provide clear metadata and observability info:
+
+- `storage_media=erofs` is visible in the template request.
+- Artifact annotations include `cube.master.rootfs.artifact.fs_type=erofs`.
+- `fs_type=erofs` is included in the node's `cube.pmem` annotation.
+- The workload pmem's `fstype=erofs` is visible from the node or guest, while `/` inside the container remains an Overlay writable view.
+
+### Notes / Constraints / Caveats
+
+- **Guest OS rootfs**: Out of scope for this proposal and continues using the existing ext4.
+- **DAX Limitation**: EROFS compressed images do not use DAX by default. Compressed EROFS requires a decompression path and cannot simply reuse ext4's `ro,dax` mount options.
+- **Guest kernel Requirements**: Must support EROFS and LZ4 to mount the workload pmem inside the Guest.
+- **Fingerprint Conflict**: `ext4` and `erofs` artifacts must have different fingerprints to avoid accidental reuse of different formats for the same OCI image.
+- **erofs-utils version baseline**: **The build environment must use a fixed `erofs-utils` version, currently recommended as `>= 1.5.0`**. In practice, it should match the version in the CI image, and `mkfs.erofs -V` output should be logged to avoid inconsistent artifacts across multiple Master nodes.
+- **Preflight Check**: `mkfs.erofs` comes from `erofs-utils`; the build node needs an explicit preflight check, at least verifying the command exists, the version meets the baseline, LZ4 is supported, and xattrs are not disabled via `-x < 0`.
+
+### Risks and Mitigations
+
+| Risk | Impact | Mitigation |
+|------|------|------|
+| Guest kernel lacks EROFS/LZ4 built-in | workload pmem cannot be mounted | Check kernel config before build; e2e validation of Guest mount behavior. |
+| `mkfs.erofs` missing | Master build fails | Preflight check with clear instructions to install `erofs-utils`. |
+| `erofs-utils` version inconsistency | Same OCI image generates different checksums | Fix versions in CI and build nodes; preflight version comparison; record `mkfs.erofs` version in artifact metadata. |
+| xattr/whiteout loss | Deleted files reappear in container, Overlay semantics broken | `mkfs.erofs` forbids `-x < 0`; add whiteout and opaque directory e2e. |
+| ext4 and erofs artifact mixed up | Startup failure or checksum mismatch | Include fs type in fingerprint. |
+| EROFS mount parameters incompatible | workload pmem mount fails | **`erofs` workload mount options only use `ro` in the first phase, strictly excluding `dax`**. |
+| Old DB tables lack new fields | Read failure or empty fields | Explicitly treat empty `fs_type` as `ext4`; implement robust default values. |
+| Compression causes CPU spike | CPU jitter during high-concurrency cold starts | Record P50/P95 latency, CPU, and IO in e2e; implement creation rate limiting if necessary. |
+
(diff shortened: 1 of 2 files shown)


kinwin-ustc:
Thank you for your very detailed proposal. I think we can only support EROFS for container images, as the usage of guest images is extremely limited in our design. In typical scenarios, there should only be one guest image globally, which should be pre-deployed on each computing node and shared with all sandboxes through pmem. EROFS almost does not bring any benefits to the guest image, so we can focus on the container image scenario. In fact, we do use EROFS to support container images internally.

 @charliecgxu @ls-ggg @hlz260 @zhuangel Do you have any suggestions?

novahe:
I agree with you. Since multiple sandboxes share a single Guest OS rootfs, the gains are pretty minimal, and we'd lose the DAX capability. Let's keep our proposal focused on the container rootfs instead.

I've updated the version, PTAL @kinwin-ustc

charliecgxu:
This is a valuable and practical design proposal, and I really appreciate your efforts in exploring EROFS integration to further improve storage efficiency and cold startup performance for CubeSandbox.
Meanwhile I have a few points for further discussion:
Firstly, it would be nice to clarify whether we keep the traditional layered OCI image structure or adopt a flattened EROFS layout for workload rootfs. This decision will have impacts on our existing image building and distribution pipelines.
Secondly, if EROFS is utilized together with virtio-blk inside guest, we may face some constraints in resource sharing. In this way, it is hard to realize unified disk space sharing and host-side page cache sharing across different sandbox instances. The page cache will be isolated inside each guest system, which may bring extra storage and memory cost in large-scale scenarios, and cannot well take advantage of the shared cache capability of the current virtio-fs architecture.
Looking forward to your further ideas on these aspects.

novahe:
Thanks for your reply! These are indeed two critical issues (here are my current thoughts based on my understanding, and please feel free to correct me if I missed anything).

In this initial proposal, I kept it relatively simple: positioning EROFS as a **complementary option to ext4+DAX**, specifically tailored for scenarios where reducing disk footprint and network transfer size is the top priority. We have no intention of replacing ext4+DAX entirely. For memory-sensitive workloads that heavily rely on shared cache, ext4+DAX remains the go-to choice for performance.

**Layered OCI vs. Flattened EROFS Layout**
The containerd community's [EROFS Snapshotter](https://github.com/containerd/containerd/blob/main/docs/snapshotters/erofs.md) typically builds a `layer.erofs` for each OCI layer, adopting such a multi-device mounting scheme poses clear limitations under CubeSandbox's current MicroVM rootfs injection path:
* **Hypervisor Device Slot Limits:** Stacking multiple device layers requires dynamically hot-plugging a virtual block device (like vPMEM or virtio-blk) for each layer. Standard OCI images easily have 10-20+ layers, which would quickly exhaust the limited PCI/device slots in Cloud-Hypervisor (clh) and even lead to MicroVM boot failures.
* **Cold Start Overhead:** Forcing the Guest kernel to mount dozens of devices and stack them via OverlayFS introduces significant boot latency, which defeats the purpose of MicroVMs' ultra-fast cold starts.
* **Upstream MicroVM Image Format Limits:** While runc containers can easily leverage flattened filesystems, passing layered structures into MicroVMs usually requires an extra flattening mechanism or formats like [VMDK](https://github.com/containerd/nerdbox/issues/30). However, the proposal to support VMDK in Cloud-Hypervisor ([#7167](https://github.com/cloud-hypervisor/cloud-hypervisor/issues/7167)) is currently stalled.

**Page Cache Sharing & Resource Constraints (Compression vs. DAX)**
The Page Cache isolation issue is indeed present. I understand this as a fundamental technical trade-off: **ext4+DAX bypasses the Guest Page Cache to minimize duplicate cache overhead**, whereas **compressed EROFS focuses on extreme storage and bandwidth savings**.
Because compressed data must be decompressed within the Guest's memory, we cannot directly leverage the DAX path like ext4+DAX does. As a result, Page Cache is isolated within each Guest instance, which indeed increases memory cost under high-density deployments compared to virtio-fs or ext4+DAX.

**Future Evolution**
Internally, we are also actively exploring rootfs solutions for MicroVM scenarios. We've been particularly inspired by the work Nydus has done around native lazy-pulling (e.g., [EROFS over Fscache On-demand Loading Technology](https://www.alibabacloud.com/blog/faster-container-image-loading-speed-with-nydus-rafs-and-erofs_599012)) and are currently validating several related approaches.

One promising direction we see is an architecture combining EROFS, Nydus, and ublk/io_uring-based userspace block backends, enabling both efficient lazy pulling and high-density MicroVM deployments while minimizing userspace overhead and context-switch costs.

I heard that your team has also been utilizing EROFS for container rootfs in your environments. Could you share some of your experiences?

hsiangkao:
EROFS supports erofs+fsdax too, but as mentioned, compressed data cannot be used as DAX because compressed data cannot be directly used for mmap.
I think one solution is that we could convert EROFS compressed data on-the-fly on the host (the metadata is also generated/converted on demand) and provide a converted uncompressed filesystem so that it can be shared with guests instead.

EROFS over fscache is now deprecated, the future direction is to leverage fanotify pre-content hooks + eBPF to work out a similar approach.

kinwin-ustc:
If it is only a matter of replacing ext4 with erofs, using the compression mechanism of erofs to reduce disk storage and distribution bandwidth, and using the computing and memory resources inside the sandbox to exchange for storage resources on the host, this solution does not have universal value. First, the existing ext4 can perform layer-level deduplication at the host level, while as you described, if erofs uses the layer scheme, there will be issues with the number of devices, and using the flat scheme will lose the deduplication capability; second, if only the distribution bandwidth is considered, compressed distribution can be considered, and then decompressed locally into ext4, and now the network bandwidth of data centers is very large, basically already over 100Gbps, I believe that physical limitations will not be a big problem in terms of throughput. 

As for the greater value of EROFS, I believe it lies in utilizing the natural chunk mechanism of EROFS to perform chunk-level deduplication and lazy loading, which will bring significant benefits in terms of actual storage consumption, bandwidth consumption, and cold start speed.

hsiangkao:
> the existing ext4 can perform layer-level deduplication at the host level

how? you mean ext4 unpacking + overlayfs or some other alternatives? how does this approach pass into the guest, virtiofs? that performance will be bad. -- I don't get the overall point.

> if erofs uses the layer scheme, there will be issues with the number of devices, and using the flat scheme will lose the deduplication capability

I don't know if these words relate to CubeSandbox or just runC.
First, since Linux v6.12, erofs supports file-backed mounts, so if it's a runc usage, you could just mount the layer file directly without loopback block devices.

As for microVMs, currently many people are working on fsmerge/GPT + vmdk feature for virtio-blk/virtio-pmem, like:
https://github.com/containerd/nerdbox/issues/30
https://github.com/kata-containers/kata-containers/pull/12763
https://github.com/kata-containers/kata-containers/pull/12958
https://github.com/firecracker-microvm/firecracker/pull/5741
there won't be per-layer block/pmem devices in the guest for layer deduplication; only one block device with merged view for each instance is needed. (regarding to fsmerge, see: https://erofs.docs.kernel.org/en/latest/merging.html)

If FSDAX+erofs is enabled, the page cache will be shared in the per-layer basis (rather than per-image basis).

> and now the network bandwidth of data centers is very large, basically already over 100Gbps, I believe that physical limitations will not be a big problem in terms of throughput.

It only works within a private network, and object-storage-based OCI registries won't reach such limitation at all.
Also some users need to distribute/publish their images over the public network too, so network traffic saving + public network quality is important for some users.

kinwin-ustc:
>> the existing ext4 can perform layer-level deduplication at the host level
> how? you mean ext4 unpacking + overlayfs or some other alternatives? how does this approach pass into the guest, virtiofs? that performance will be bad. -- I don't get the overall point.

In the open-source version, we manage container rootfs on the host using the built-in image management mechanism of containerd, and present the images to the sandbox through virtiofs. This method has sufficient performance in actual deployment scenarios

> As for microVMs, currently many people are working on fsmerge/GPT + vmdk feature for virtio-blk/virtio-pmem, like:
> https://github.com/containerd/nerdbox/issues/30
> https://github.com/kata-containers/kata-containers/pull/12763
> https://github.com/kata-containers/kata-containers/pull/12958
> https://github.com/firecracker-microvm/firecracker/pull/5741
> there won't be per-layer block/pmem devices in the guest for layer deduplication; only one block device with merged view for each instance is needed. (regarding to fsmerge, see: https://erofs.docs.kernel.org/en/latest/merging.html)
> 
> If FSDAX+erofs is enabled, the page cache will be shared in the per-layer basis (rather than per-image basis).

I agree with many people's efforts to maintain layering in EROFS scenarios and show its merged view to the sandbox, but I believe that in actual production deployments, using EROFS's chunk mechanism, doing chunk-based deduplication, lazy loading, and distribution is a better solution.  Additionally, it also reduces the dependency on the network

hsiangkao:
> > > the existing ext4 can perform layer-level deduplication at the host level
> > > how? you mean ext4 unpacking + overlayfs or some other alternatives? how does this approach pass into the guest, virtiofs? that performance will be bad. -- I don't get the overall point.
> 
> In the open-source version, we manage container rootfs on the host using the built-in image management mechanism of containerd, and present the images to the sandbox through virtiofs. This method has sufficient performance in actual deployment scenarios

It sounds like what kata does as virtiofs passthrough fs, but I guess
you're open to enable containerd erofs snapshotter + erofs mount layers to CubeSandbox opensource implementation too?
since erofs snapshotter is already a builtin snapshotter in the containerd (since containerd 2.1), I guess supporting this snapshotter for CubeSandbox is not a bad thing?

> 
> > As for microVMs, currently many people are working on fsmerge/GPT + vmdk feature for virtio-blk/virtio-pmem, like:
> > [containerd/nerdbox#30](https://github.com/containerd/nerdbox/issues/30)
> > [kata-containers/kata-containers#12763](https://github.com/kata-containers/kata-containers/pull/12763)
> > [kata-containers/kata-containers#12958](https://github.com/kata-containers/kata-containers/pull/12958)
> > [firecracker-microvm/firecracker#5741](https://github.com/firecracker-microvm/firecracker/pull/5741)
> > there won't be per-layer block/pmem devices in the guest for layer deduplication; only one block device with merged view for each instance is needed. (regarding to fsmerge, see: https://erofs.docs.kernel.org/en/latest/merging.html)
> > If FSDAX+erofs is enabled, the page cache will be shared in the per-layer basis (rather than per-image basis).
> 
> I agree with many people's efforts to maintain layering in EROFS scenarios and show its merged view to the sandbox, but I believe that in actual production deployments, using EROFS's chunk mechanism, doing chunk-based deduplication, lazy loading, and distribution is a better solution. Additionally, it also reduces the dependency on the network

You could compose erofs images on demand with chunks of course (e.g. using vmdk mapping or some direct in-hourse approach) as a useful optimization, no question about that.

I don't know what @novahe thinks, personally I just hope we could have an opensource native erofs snapshotter support without virtiofs, etc.

novahe:
> As for the greater value of EROFS, I believe it lies in utilizing the natural chunk mechanism of EROFS to perform chunk-level deduplication and lazy loading, which will bring significant benefits in terms of actual storage consumption, bandwidth consumption, and cold start speed.

You’ve successfully convinced me. In fact, I agree with this point as well. If EROFS is introduced but doesn’t achieve block-level deduplication and deferred loading, its value isn’t very significant. It would be even better if the sandbox could share the cache between them.

> since erofs snapshotter is already a builtin snapshotter in the containerd (since containerd 2.1), I guess supporting this snapshotter for CubeSandbox is not a bad thing?

Currently, only block-level deduplication is implemented for containerd erofs snapshotter + virtiofs.

> I don't know what @novahe thinks, personally I just hope we could have an opensource native erofs snapshotter support without virtiofs, etc.

We are also currently trying to find such a solution.

Interestingly, I came across the following implementations, which seem quite interesting. At present, they appear to align with our aforementioned objectives. When I have time, I will give them a try：
https://github.com/dragonflyoss/nydus/pull/1921
https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8239

kinwin-ustc:
We may open-source our storage backend based on EROFS with deduplication capabilities and lazy loading, which is  better than this proposal

novahe:
> We may open-source our storage backend based on EROFS with deduplication capabilities and lazy loading, which is  better than this proposal

Is there an open source plan? I'll close the pr first.

rogeroger-yu:
> > As for the greater value of EROFS, I believe it lies in utilizing the natural chunk mechanism of EROFS to perform chunk-level deduplication and lazy loading, which will bring significant benefits in terms of actual storage consumption, bandwidth consumption, and cold start speed.
> 
> You’ve successfully convinced me. In fact, I agree with this point as well. If EROFS is introduced but doesn’t achieve block-level deduplication and deferred loading, its value isn’t very significant. It would be even better if the sandbox could share the cache between them.
> 
> > since erofs snapshotter is already a builtin snapshotter in the containerd (since containerd 2.1), I guess supporting this snapshotter for CubeSandbox is not a bad thing?
> 
> Currently, only block-level deduplication is implemented for containerd erofs snapshotter + virtiofs.
> 
> > I don't know what @novahe thinks, personally I just hope we could have an opensource native erofs snapshotter support without virtiofs, etc.
> 
> We are also currently trying to find such a solution.
> 
> Interestingly, I came across the following implementations, which seem quite interesting. At present, they appear to align with our aforementioned objectives. When I have time, I will give them a try： [dragonflyoss/nydus#1921](https://github.com/dragonflyoss/nydus/pull/1921) [cloud-hypervisor/cloud-hypervisor#8239](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8239)

This is doable, I have done something similar, another point to consider is how to share the snapshot

hsiangkao:
> > > As for the greater value of EROFS, I believe it lies in utilizing the natural chunk mechanism of EROFS to perform chunk-level deduplication and lazy loading, which will bring significant benefits in terms of actual storage consumption, bandwidth consumption, and cold start speed.
> > 
> > 
> > You’ve successfully convinced me. In fact, I agree with this point as well. If EROFS is introduced but doesn’t achieve block-level deduplication and deferred loading, its value isn’t very significant. It would be even better if the sandbox could share the cache between them.
> > > since erofs snapshotter is already a builtin snapshotter in the containerd (since containerd 2.1), I guess supporting this snapshotter for CubeSandbox is not a bad thing?
> > 
> > 
> > Currently, only block-level deduplication is implemented for containerd erofs snapshotter + virtiofs.
> > > I don't know what @novahe thinks, personally I just hope we could have an opensource native erofs snapshotter support without virtiofs, etc.
> > 
> > 
> > We are also currently trying to find such a solution.
> > Interestingly, I came across the following implementations, which seem quite interesting. At present, they appear to align with our aforementioned objectives. When I have time, I will give them a try： [dragonflyoss/nydus#1921](https://github.com/dragonflyoss/nydus/pull/1921) [cloud-hypervisor/cloud-hypervisor#8239](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8239)
> 
> This is doable, I have done something similar, another point to consider is how to share the snapshot

Could you share some details about "sharing the snapshot"?
-- currently we share the host page cache and use virtio-pmem to elimintate guest page cache.

rogeroger-yu:
> > > > As for the greater value of EROFS, I believe it lies in utilizing the natural chunk mechanism of EROFS to perform chunk-level deduplication and lazy loading, which will bring significant benefits in terms of actual storage consumption, bandwidth consumption, and cold start speed.
> > > 
> > > 
> > > You’ve successfully convinced me. In fact, I agree with this point as well. If EROFS is introduced but doesn’t achieve block-level deduplication and deferred loading, its value isn’t very significant. It would be even better if the sandbox could share the cache between them.
> > > > since erofs snapshotter is already a builtin snapshotter in the containerd (since containerd 2.1), I guess supporting this snapshotter for CubeSandbox is not a bad thing?
> > > 
> > > 
> > > Currently, only block-level deduplication is implemented for containerd erofs snapshotter + virtiofs.
> > > > I don't know what @novahe thinks, personally I just hope we could have an opensource native erofs snapshotter support without virtiofs, etc.
> > > 
> > > 
> > > We are also currently trying to find such a solution.
> > > Interestingly, I came across the following implementations, which seem quite interesting. At present, they appear to align with our aforementioned objectives. When I have time, I will give them a try： [dragonflyoss/nydus#1921](https://github.com/dragonflyoss/nydus/pull/1921) [cloud-hypervisor/cloud-hypervisor#8239](https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8239)
> > 
> > 
> > This is doable, I have done something similar, another point to consider is how to share the snapshot
> 
> could you share some details about "sharing the snapshot"? e.g. in which way you'd like to share? thanks.

- It is currently relatively simple and crude, directly storing distributed storage, but you need to consider SEEK_HOLE/SEEK_DATA sparse copy when uploading.
- virtio-pmem is only used on the rootfs，basic reuse of guest startup process

hsiangkao:
...
> > could you share some details about "sharing the snapshot"? e.g. in which way you'd like to share? thanks.
> 
> It is currently relatively simple and crude, directly storing distributed storage, but you need to consider SEEK_HOLE/SEEK_DATA sparse copy when uploading.

I thought you're refering to memory sharing, anyway.
yes, you could use distributed object storage for example to store duplicated chunks.

but since the memory is expensive, I think another charmful stuff is to save memory as much as possible.

rogeroger-yu:
> ...
> I thought you're refering to memory sharing, anyway. yes, you could use distributed object storage for example to store duplicated chunks.
> 
> but since the memory is expensive, I think another charmful stuff is to save memory as much as possible.

- roofs -> nydus + pmem -> distributed storage
- snapshot(memory) -> non-sparse upload -> distributed storage

PR #295, opened 2026-05-18
fix(proxy): disable buffering for sandbox streams

josh-cow:
Patch for `nginx` config:

  - Adds proxy_buffering off;
  - Adds proxy_request_buffering off;
  - Applies to both HTTP and HTTPS CubeProxy sandbox locations in CubeProxy/nginx.conf

Interactive `envd` Connect streams (PTY) need chunks flushed immediately. With CubeProxy’s global proxy_buffering on, long-lived PTY streams can receive input but not deliver output
  promptly.

  Validated:

  - `nginx -t` inside the running cube-proxy container passed.
  - After reload, direct `envd` Process/Start returned 200 and streamed the start frame immediately.

### DIFF
--- a/CubeProxy/nginx.conf
+++ b/CubeProxy/nginx.conf
@@ -107,6 +107,8 @@ http {
             proxy_send_timeout 7206s;
             proxy_read_timeout 7206s;
             proxy_connect_timeout 3s;
+            proxy_buffering off;
+            proxy_request_buffering off;
             proxy_set_header Host $http_host;
             proxy_set_header X-Real-IP $remote_addr;
             proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
@@ -163,6 +165,8 @@ http {
             proxy_send_timeout 7206s;
             proxy_read_timeout 7206s;
             proxy_connect_timeout 3s;
+            proxy_buffering off;
+            proxy_request_buffering off;
             # proxy_socket_keepalive may cause frequent vm exit
             # proxy_socket_keepalive on;
             proxy_set_header Host $http_host;


kinwin-ustc:
CubeSandbox is not just for envd, there are also scenarios where buffers are needed, and perhaps we need to clearly distinguish which requests require buffers and which are not

staryxchen:
Thanks for the contribution @josh-cow. `proxy_buffering` and `proxy_request_buffering` are two global directives that proxy all sandbox traffic, not just envd streams. So I suggesting fixing this at the response level via setting `X-Accel-Buffering: no` from envd, or by adding a dedicated streaming-only location.

josh-cow:
> Thanks for the contribution @josh-cow. `proxy_buffering` and `proxy_request_buffering` are two global directives that proxy all sandbox traffic, not just envd streams. So I suggesting fixing this at the response level via setting `X-Accel-Buffering: no` from envd, or by adding a dedicated streaming-only location.

@kinwin-ustc @staryxchen 
Make sense. are you on the plan ? Or may I draft the changes? Thanks.

kinwin-ustc:
I think you can continue to modify in this direction

staryxchen:
> > Thanks for the contribution @josh-cow. `proxy_buffering` and `proxy_request_buffering` are two global directives that proxy all sandbox traffic, not just envd streams. So I suggesting fixing this at the response level via setting `X-Accel-Buffering: no` from envd, or by adding a dedicated streaming-only location.
> 
> @kinwin-ustc @staryxchen Make sense. are you on the plan ? Or may I draft the changes? Thanks.

Hi @josh-cow, I have raised an issue (https://github.com/e2b-dev/infra/issues/2707) to envd, but haven’t received a response yet. So I think you can working on the Cube side (adding a dedicated streaming-only location) first to workaround.

PR #298, opened 2026-05-18
feat(template): support --user flag and respect image User field

woshihoujinxin:
- Add --user flag to cubemastercli tpl create-from-image command Supports both UID (e.g. --user 1000) and username (e.g. --user node)
- Apply Docker image User field as default RunAsUser/RunAsUsername in generateTemplateCreateRequest, previously imageCfg.User was parsed but ignored
- ContainerOverrides.SecurityContext from --user takes precedence over image User, consistent with env/cmd/workingdir override pattern

Fixes: container always runs as root because imageCfg.User was discarded


ls-ggg:
@woshihoujinxin 
Thanks your pr, please handle the feedback above and add the Signed-off-by line as per the PR conventions. 
Additionally, have you tested this locally to ensure the functionality is working as intended?

woshihoujinxin:
这个方案存在问题 暂不提交

PR #303, opened 2026-05-19
fix(templatecenter): strip canonical prefix from image digest

chenhengqi:
The IMAGE_INFO column in 'cubemastercli tpl ls' could render as
  name:tag@name@sha256:...
when docker inspect returned a populated RepoDigests array.

firstNonEmptyDigest() returned RepoDigests[0] verbatim (a canonical reference of the form 'name@sha256:...') and stored it in SourceImageDigest. composeImageInfo() then concatenated the source ref and that 'digest', producing the duplicated name and double '@'.

Fix the root cause by extracting only the digest portion of RepoDigests[0] before returning. Also harden composeImageInfo() to strip a 'name@' prefix from any digest argument so historical rows already persisted with the malformed value render correctly without a database migration.

Add a regression test covering the canonical-prefix case.

### DIFF
--- a/CubeMaster/pkg/templatecenter/store.go
+++ b/CubeMaster/pkg/templatecenter/store.go
@@ -680,6 +680,12 @@ func composeImageInfo(ref, digest string) string {
 	if imageDigest == "" {
 		return imageRef
 	}
+	// Tolerate historical rows where SourceImageDigest was stored as a
+	// full canonical reference ("name@sha256:..."). Strip the "name@"
+	// prefix so we never produce "name:tag@name@sha256:...".
+	if at := strings.Index(imageDigest, "@"); at >= 0 && at+1 < len(imageDigest) {
+		imageDigest = imageDigest[at+1:]
+	}
 	if strings.Contains(imageRef, "@") {
 		return imageRef
 	}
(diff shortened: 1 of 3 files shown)


fslongjin:
Fixed the broken test expectations after the digest normalization change.

What changed:
- Updated `TestPrepareSourceImageSkipsPullWhenImageExistsLocally` and `TestPrepareSourceImagePullsAfterLocalInspectMiss` to expect `sha256:abcd` instead of the old canonical `name@sha256:abcd` value.

Validation:
- `go test ./pkg/templatecenter -run 'TestPrepareSourceImageSkipsPullWhenImageExistsLocally|TestPrepareSourceImagePullsAfterLocalInspectMiss|TestComposeImageInfo' -count=1`
- Built manual package from this PR and deployed it to `root@cube-test-2 -p 36000`.
- Restarted with `/usr/local/services/cubetoolbox/scripts/one-click/down-with-deps.sh` and `up-with-deps.sh`.
- `quickcheck.sh` passed.
- Created a temporary template from image and confirmed the new job stores `source_image_digest` as plain `sha256:...` while `tpl list -o wide` renders `image:tag@sha256:...` without the duplicated image prefix.

Assisted-by: Cursor:GPT-5.5

PR #307, opened 2026-05-19
fix: make network-agent tap fd timeout configurable

yangjie1993:
## Problem

Cubelet hardcoded a 2s deadline when requesting tap fd from network-agent.
Under high concurrency (>40 sandboxes/node), requests queue behind the global
mutex (`s.mu`) in network-agent. Each lock hold takes ~50ms (dominated by
uncached `loadPinnedMap` bpffs I/O), causing requests to exceed the 2s deadline
with `i/o timeout` errors.

## Solution

Add `NetworkAgentTapFDTimeout` config field (`network_agent_tap_fd_timeout` in
TOML), defaulting to `2s` to preserve existing behavior. Operators can raise it
(e.g. `10s`) for high-concurrency deployments without recompiling.

### DIFF
--- a/Cubelet/config/config.toml
+++ b/Cubelet/config/config.toml
@@ -81,6 +81,7 @@ dynamic_config_path = "/usr/local/services/cubetoolbox/Cubelet/dynamicconf/conf.
     network_agent_tap_socket = "/tmp/cube/network-agent-tap.sock"
     network_agent_init_timeout = "120s"
     network_agent_retry_interval = "1s"
+    network_agent_tap_fd_timeout = "2s"
 
   [plugins."io.cubelet.internal.v1.storage"]
     #root_path=/data/cubelet/state/io.cubelet.internal.v1.storage
--- a/Cubelet/network/plugin_tap.go
+++ b/Cubelet/network/plugin_tap.go
@@ -275,11 +275,12 @@ type Config struct {
 
 	RootPath string `toml:"root_path"`
 
-	EnableNetworkAgent        bool             `toml:"enable_network_agent"`
-	NetworkAgentEndpoint      string           `toml:"network_agent_endpoint"`
-	NetworkAgentTapSocket     string           `toml:"network_agent_tap_socket"`
-	NetworkAgentInitTimeout   tomlext.Duration `toml:"network_agent_init_timeout"`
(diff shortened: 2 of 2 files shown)


Cubelet/config/config.toml
```diff
@@ -81,6 +81,7 @@ dynamic_config_path = "/usr/local/services/cubetoolbox/Cubelet/dynamicconf/conf.
     network_agent_tap_socket = "/tmp/cube/network-agent-tap.sock"
     network_agent_init_timeout = "120s"
     network_agent_retry_interval = "1s"
+    # network_agent_tap_fd_timeout = "10s"  # default: 2s; increase when node concurrency > 40
```
chenhengqi: Instead of comment it out here, let's add a good default value (like `2s`) here.

Cubelet/network/plugin_tap.go
```diff
@@ -956,7 +960,7 @@ func (l *local) registerNetworkAgentTapForPool(ctx context.Context, sandboxID st
 	if intf.IPAddr == nil {
 		return fmt.Errorf("shim network sandbox ip is empty")
 	}
-	file, err := requestNetworkAgentTapFile(l.Config.NetworkAgentTapSocket, sandboxID, intf.Name)
+	file, err := requestNetworkAgentTapFile(l.Config.NetworkAgentTapSocket, sandboxID, intf.Name, time.Duration(l.Config.NetworkAgentTapFDTimeout))
```
chenhengqi: Line too long. This would trigger a `lll` linter complaint.

chenhengqi (changes requested):
Left some comments below. Please also follow our contribution guide.

Cubelet/network/plugin_tap.go
yangjie1993: Changed the config to use an active default value `2s` instead of a comment
- Extracted the timeout to a local variable to fix the line length issue

chenhengqi (commented):
Looks good. Please squash your commits and add your signed-off-by tag.

yangjie1993:
> Looks good. Please squash your commits and add your signed-off-by tag.

done

chenhengqi (approved):

PR #310, opened 2026-05-19
docs(integrations): add OpenClaw integration guide (en + zh)

GongHening:
## Summary

  - Add OpenClaw integration guide — run OpenClaw Gateway inside a Cube MicroVM for kernel-level agent isolation
  - Both English (`docs/guide/integrations/openclaw.md`) and Chinese (`docs/zh/guide/integrations/openclaw.md`) versions included
  - Updated the published articles table in both language index pages

  ## What the guide covers

  1. **Setup** — Cube deployment, template creation (2G writable layer), environment variables
  2. **Integration diff** — native `openclaw gateway` vs. `Sandbox.create()` with side-by-side code comparison
  3. **Runnable demo** — full `run_openclaw.py` script that installs Node.js 24, configures OpenClaw, starts the gateway, and runs a
  health-check loop
  4. **Going further** — network isolation, timeouts, host directory mounts, pause/resume, pre-built custom image

  ## Key lessons baked in

  The guide is based on a real end-to-end deployment and captures non-obvious details:

  - Correct config sequence: minimal JSON → `openclaw doctor --fix` → `--bind loopback --allow-unconfigured`
  - Reliable process checking (`pgrep -x` / `grep -v grep` instead of bare `ps aux | grep`)
  - Startup script wrapper to avoid SDK timeouts on `nohup` commands
  - Template version must match CubeSandbox code version
  - `--writable-layer-size 2G` requirement for Node.js + OpenClaw


ls-ggg:
Thanks for your PR! Could you please build an OpenClaw image based on the official base image to showcase this demo?

PR #314, opened 2026-05-20
fix(network): resolve network resource leak on sandbox creation failure

novahe:
If sandbox creation fails after `EnsureNetwork` succeeds but before the allocation is persisted in allocationStore, subsequent cleanup workflows cannot detect the network allocation, leading to leaked TAP, IP, and port mappings on the `network-agent`.

This PR introduces a deferred rollback mechanism in tapPlugin.Create to automatically invoke ReleaseNetwork and unregister any partially set up TAP device if the creation returns an error.

### DIFF
--- a/Cubelet/network/plugin_tap.go
+++ b/Cubelet/network/plugin_tap.go
@@ -463,6 +463,26 @@ func (l *local) Create(ctx context.Context, opts *workflow.CreateContext) (err e
 		l.recordNetworkAgentFailure(naErr)
 		return ret.Errorf(errorcode.ErrorCode_CreateNetworkFailed, "network-agent EnsureNetwork failed: %s", classifyNetworkAgentError(naErr))
 	}
+
+	defer func() {
+		if err == nil {
+			return
+		}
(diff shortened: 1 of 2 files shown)


novahe:
PTAL @fslongjin @chenhengqi

fslongjin (approved):
Thanks for the fix! I traced the failure chain (EnsureNetwork succeeds → buildShimNetReqFromEnsureResponse / registerNetworkAgentTapForPool fails → allocationStore.Add is skipped → subsequent Destroy/CleanUp returns early via ErrorKeyNotFound → ReleaseNetwork is never called) and confirmed the leak is real. The deferred rollback is placed at the right layer, the detached context with a 15s timeout is the correct way to handle parent-context cancellation, NetworkHandle == SandboxID matches the network-agent convention, and ReleaseNetwork is idempotent on the agent side, so retries/no-op cases are safe. The new test reliably triggers the register-failure path by leaving NetworkAgentTapSocket empty.

LGTM.

novahe:
Can this PR be merged? @fslongjin

PR #317, opened 2026-05-20
fix(shim): allow delete on PAUSED sandboxes to prevent destroy timeout

chenggui53:
When a sandbox enters PAUSED state at CubeShim level (e.g., VM pause during Nydus operations), the destroy path fails because the delete handler rejects requests with "sandbox not in normal state". This causes the Cubelet destroy to fall back to stopTask(), where container.Task() blocks waiting for the sandbox mutex held by pause_vm(), hitting the 30-second ttrpc client timeout.

Remove the sb.paused() guard from the delete handler. During destroy, signal_container(SIGKILL) already returns Ok() without contacting the in-VM agent when the sandbox is not running (container/mod.rs:466-472), so deletion is safe even when the VM is paused. The agent will be cleaned up when the VM is eventually terminated.

### DIFF
--- a/CubeShim/shim/src/service/task_srv.rs
+++ b/CubeShim/shim/src/service/task_srv.rs
@@ -294,10 +294,9 @@ impl Task for TaskService {
             self.log.clone(),
         );
         let mut sb = self.sandbox.lock().await;
-        if sb.paused().await {
-            errf!(self.log, "sandbox not in normal state");
-            return Err(Others(format!("sandbox not in normal state")));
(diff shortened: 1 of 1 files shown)


ls-ggg:
@chenggui53 thanks for the pr，In the pause state, the guest is inaccessible, so APIs like kill, wait, delete, and destroy won't work. If we use a mock approach to return success, it might lead to data inconsistency. Therefore, the best approach is to resume it before destroying

PR #322, opened 2026-05-20
go sdk: move close cleanup to client

nemo1105:
Assisted-by: Codex:GPT-5

fix the first bug：https://github.com/TencentCloud/CubeSandbox/pull/254#issuecomment-4494120473

### DIFF
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -39,6 +39,7 @@ import (
 func main() {
 	ctx := context.Background()
 	client := cubesandbox.NewClient(cubesandbox.NewConfigFromEnv())
+	defer client.Close()
 
 	sb, err := client.Create(ctx, cubesandbox.CreateOptions{})
 	if err != nil {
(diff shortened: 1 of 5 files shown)


tinklone (approved):
LGTM! Clean fix for the Close() ownership issue I raised in #254.

The design is solid:
- `Client.Close()` is the right level for connection pool cleanup
- `Sandbox.Close()` properly deprecated with delegation
- Guard against double-close when `dataHTTP == controlHTTP` (shared via `WithHTTPClient`)
- `newControlHTTPClient` now clones `DefaultTransport` instead of sharing it — avoids global side effects
- Good test coverage with `closeCountingTransport`

Minor nit (non-blocking): multiple `Sandbox.Close()` calls on the same Client will invoke `CloseIdleConnections()` repeatedly (harmless but redundant). A `sync.Once` in `Client.Close` could prevent that, but since `Sandbox.Close` is deprecated this is fine as-is.

PR #326, opened 2026-05-21
feat(Cubelet): expose scheduler metrics as Prometheus gauges on /v1/metrics

chenggui53:
The scheduler metrics (CPU/mem quota usage, VM counts, NIC queues, realtime create/destroy requests) were only available as JSON from /v1/metrics/scheduler. Register them as Prometheus gauges via a custom prometheus.Collector so they are also scraped at /v1/metrics.

The JSON endpoint is preserved unchanged for backward compatibility.


kinwin-ustc:
can you squash it into one commit？

chenggui53:
@kinwin-ustc commits are squashed, please review ~

PR #333, opened 2026-05-21
fix(Cubelet): cleanup host-mount dirs after sandbox destroy

luzhixing12345:
Currently, sandbox destruction does not call `cleanupHostDirVolumes`, which causes host-mount related resources to remain on the host after the sandbox is removed.

As a result:

* directories under `/data/cubelet/hostdir/` are not cleaned up
* bind mount points created for `host-mount` remain mounted
* sandbox-specific directories continue to accumulate over time

The leaked mount points can be observed with:

```bash
sudo nsenter -t "$(pgrep -f '/Cubelet/bin/cubelet' | head -n1)" -m -- findmnt -n -o TARGET,SOURCE,FSTYPE,OPTIONS 
```

Example leaked mounts after sandbox destruction:

```bash
│ ├─/data/cubelet/hostdir/9847ba94c3b345c99222b30bf1ae0e9d/rw/hostdir-0
│ │     /dev/sda2[/tmp/rw]
│ │            ext4        rw,relatime,errors=remount-ro,stripe=16
│ └─/data/cubelet/hostdir/9847ba94c3b345c99222b30bf1ae0e9d/ro/hostdir-1
│       /dev/sda2[/tmp/ro]
│              ext4        ro,relatime,errors=remount-ro,stripe=16
```

This PR adds host-dir volume cleanup during sandbox destruction.


luzhixing12345:
Hi, may I ask when this PR might be merged?

PR #337, opened 2026-05-22
fix(Cubelet): use correct Cloud Hypervisor disk API endpoint

luzhixing12345:
Fix Cubelet disk hotplug requests to use the requested Cloud Hypervisor endpoint instead of always calling `vm.add-disk`. This makes `DelDisk` correctly call `vm.remove-device`.

Also switch to `http.NewRequestWithContext` so caller timeout/cancellation propagates to the HTTP request.


luzhixing12345:
Hi, may I ask when this PR might be merged?

PR #339, opened 2026-05-22
chore: remove deprecated rand.Seed calls

Piwriw:


fslongjin (approved):
LGTM. Removing the deprecated global rand.Seed calls is consistent with the current Go 1.24 toolchain, and the related node randomness test passed locally.

PR #341, opened 2026-05-23
Fix host-mount pause snapshot restore

nemo1105:
Host-mounted virtio-fs snapshots could be taken without preparing migration metadata. When the root inode lost migration state, resume failed with InvalidVirtioFsState for inode 1.

Prepare and finish backend serialization around snapshot/start migration, fail fast if the root inode cannot be serialized, keep non-root invalid markers, and commit pause/resume state only after VM operations succeed.

### DIFF
--- a/hypervisor/virtio-devices/src/fs.rs
+++ b/hypervisor/virtio-devices/src/fs.rs
@@ -321,7 +321,6 @@ impl FsEpollHandler {
         let mut limit_by_bytes = Wrapping(0);
 
         while let Some(desc_chain) = queue.pop_descriptor_chain(self.mem.memory()) {
-
             let head_index = desc_chain.head_index();
 
             let reader = Reader::new(desc_chain.memory(), desc_chain.clone())
@@ -559,6 +558,23 @@ unsafe impl ByteValued for VirtioFsConfig {}
 enum ServerType {
     PassthroughFs(Server<PassthroughFs>),
     PassthroughFsRo(Server<PassthroughFsRo>),
+}
+
+#[must_use]
+struct BackendSerializationGuard<'a> {
+    pre_serialization: &'a AtomicBool,
+}
+
+impl<'a> BackendSerializationGuard<'a> {
+    fn new(pre_serialization: &'a AtomicBool) -> Self {
+        Self { pre_serialization }
+    }
+}
+
+impl Drop for BackendSerializationGuard<'_> {
+    fn drop(&mut self) {
+        self.pre_serialization.store(false, Ordering::Release);
+    }
 }
 
 // Macro to forward method calls to the underlying server variant
@@ -835,6 +851,36 @@ impl Fs {
             )))
         }
     }
+
+    fn prepare_backend_serialization(
+        &self,
+        server: &ServerType,
+    ) -> std::result::Result<(), MigratableError> {
+        let start = Instant::now();
+        let cancel = Arc::new(AtomicBool::new(false));
+        debug!("serialize prepare");
+        server.prepare_serialization(cancel).map_err(|e| {
+            warn!("{} preserialization failed ({:?})", self.id, e);
+            MigratableError::Snapshot(anyhow!("serialize prepare failed: {:?}", e))
+        })?;
+        self.pre_serialization.store(true, Ordering::Release);
+        info!(
+            "{} preserialization success after {}(us)",
+            self.id,
+            start.elapsed().as_micros()
+        );
+        Ok(())
+    }
+
+    fn prepare_backend_serialization_if_needed(
+        &self,
+        server: &ServerType,
+    ) -> std::result::Result<(), MigratableError> {
+        if self.pre_serialization.load(Ordering::Acquire) {
+            return Ok(());
+        }
+        self.prepare_backend_serialization(server)
+    }
 }
 
 impl Drop for Fs {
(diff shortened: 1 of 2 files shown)


nemo1105:
检查OOM了，是不是执行检查的环境内存不足了？

```

thread 'common_parallel::test_focal_hypervisor_fw' panicked at tests/integration.rs:2497:14:
called `Result::unwrap()` on an `Err` value: Os { code: 12, kind: OutOfMemory, message: "Cannot allocate memory" }
stack backtrace:
   0: rust_begin_unwind
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/std/src/panicking.rs:647:5
   1: core::panicking::panic_fmt
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/panicking.rs:72:14
   2: core::result::unwrap_failed
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/result.rs:1649:5
   3: core::result::Result<T,E>::unwrap
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/result.rs:1073:23
   4: integration::common_parallel::test_simple_launch
             at ./tests/integration.rs:2487:25
   5: integration::common_parallel::test_focal_hypervisor_fw
             at ./tests/integration.rs:2466:9
   6: integration::common_parallel::test_focal_hypervisor_fw::{{closure}}
             at ./tests/integration.rs:2465:34
   7: core::ops::function::FnOnce::call_once
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/ops/function.rs:250:5
   8: core::ops::function::FnOnce::call_once
             at /rustc/aedd173a2c086e558c2b66d3743b344f977621a7/library/core/src/ops/function.rs:250:5
note: Some details are omitted, run with `RUST_BACKTRACE=full` for a verbose backtrace.
Successfully cleaned up guest temp dir: "/tmp/chHzS7y2"


failures:
    common_parallel::test_focal_hypervisor_fw

test result: FAILED. 6 passed; 1 failed; 0 ignored; 0 measured; 137 filtered out; finished in 231.63s

error: test failed, to rerun pass `--test integration`
```

zhuangel:
Adding the `prepare_serialization()` operation for `virtiofsd` within the hypervisor's `snapshot()` logic is crucial for ensuring the complete serialization of VirtioFS. Nice modification.

@nemo1105 May I ask whether the addition of `prepare_serialization` is sufficient to resolve the issues encountered previously? Since `prepare_serialization` already incorporates the pre-validation logic for inodes, it should no longer be necessary to include special handling logic for the root inode within `serialized::PassthroughFsV1`.

nemo1105:
Thanks for the review @zhuangel . I agree with your point.

After re-checking the snapshot and migration flow, the real fix is to make the virtio-fs snapshot path call prepare_serialization() before
serialize_data(). The pause-to-snapshot path does not go through start_migration(), so the backend migration metadata was not prepared
before snapshot serialization.

I have updated the patch accordingly: the root-inode-specific handling in PassthroughFsV1 serialization has been removed. Root inode state
is now handled by the normal prepare_serialization() contract, where it is prepared as RootNode before serialization.

zhuangel:
> Thanks for the review @zhuangel . I agree with your point.
> 
> After re-checking the snapshot and migration flow, the real fix is to make the virtio-fs snapshot path call prepare_serialization() before serialize_data(). The pause-to-snapshot path does not go through start_migration(), so the backend migration metadata was not prepared before snapshot serialization.
> 
> I have updated the patch accordingly: the root-inode-specific handling in PassthroughFsV1 serialization has been removed. Root inode state is now handled by the normal prepare_serialization() contract, where it is prepared as RootNode before serialization.

The virtiofs part modification LGTM. Thanks for your contribution.

zhuangel (approved):

CubeShim/shim/src/sandbox/sb.rs
```diff
@@ -1150,49 +1148,144 @@ impl SandBox {
...
 
-        ch.pause_vm_cube(format!("file://{}", snapshot_path).as_str())
-            .await?;
+        {
+            let ch = self.ch.as_mut().unwrap().lock().await;
+            if let Err(e) = ch
+                .pause_vm_cube(format!("file://{}", snapshot_path).as_str())
+                .await
+            {
+                drop(ch);
```
ls-ggg: drop(ch) is redundant

ls-ggg:
This modification doesn't seem related to host-mount. Also, if pause fails unexpectedly, resume might not work properly either. Could you please separate these two features into different PRs?

CubeShim/shim/src/sandbox/sb.rs
nemo1105: The code is clearer now.

nemo1105:
@ls-ggg Good point. The sandbox pause/resume state handling is a separate robustness change and not directly tied to the host-mount snapshot fix.
I’ll remove it from this PR and keep this PR focused on preparing virtio-fs backend serialization for pause-to-snapshot.

nemo1105:
Is there anything wrong with [Hypervisor Tests (x86-64) / Tests (x86-64) (gnu) (pull_request)](https://github.com/TencentCloud/CubeSandbox/actions/runs/26391398824/job/77681721826?pr=341) ?

PR #348, opened 2026-05-25
fix(nodemeta): preserve admin-configured MaxMvmNum across cubelet restarts

chenggui53:
When a cubelet registers with the scheduler, RegisterNode unconditionally overwrites MaxMvmNum from the request. This means any admin-configured limit (set directly in the database) is lost when the cubelet restarts, because the cubelet calculates its own value from available memory.

The fix preserves the existing MaxMvmNum when the incoming request doesn't specify one (<=0), allowing scheduler-side limits to survive cubelet restarts.

### DIFF
--- a/CubeMaster/pkg/nodemeta/service.go
+++ b/CubeMaster/pkg/nodemeta/service.go
@@ -129,6 +129,9 @@ func RegisterNode(ctx context.Context, req *RegisterNodeRequest) (*NodeSnapshot,
 	if req.HostIP == "" {
 		req.HostIP = req.NodeID
 	}
+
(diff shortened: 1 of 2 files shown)


fslongjin:
Thanks for the PR! I have a question about the usage scenario.

How do you currently set the cubelet's `MaxMvmNum` in CubeMaster — do you modify the database directly? We'd like to understand the workflow better to evaluate whether the fix covers the right case, especially given that `resolveHostMaxMvmNum` in the cubelet always returns a positive value (see the analysis in the review above).

If you're setting it via direct DB edits, we should also consider whether a proper API/UI for this configuration would be a better long-term approach.

PR #352, opened 2026-05-25
fix(network-agent): guard against nil service before Health check

chenggui53:
Add explicit nil and typed-nil checks after initService returns, before calling svc.Health(). Prevents SIGSEGV when initService returns a typed-nil service that passes the interface nil check but still dereferences nil on method call.


chenhengqi (changes requested):
This change does not make sense to me. It would be better to just panic.

PR #354, opened 2026-05-26
Improve pause/resume failure recovery

nemo1105:
## Summary

  Improve pause/resume failure recovery by delaying sandbox state transitions until the corresponding VM operation succeeds.

  ## Changes

  - Delay `Paused` / resume-related state updates until the VM operation has completed successfully.
  - Restore agent clients when pause setup fails before the sandbox is committed to the paused state.
  - Restart runtime monitors when pause-to-snapshot fails before the sandbox is committed to the paused state.
  - Keep sandbox runtime state consistent after partial pause/resume failures.

  ## Testing

  Not run locally.

  Assisted-by: Codex:GPT-5

### DIFF
--- a/CubeShim/shim/src/sandbox/sb.rs
+++ b/CubeShim/shim/src/sandbox/sb.rs
@@ -1134,11 +1134,37 @@ impl SandBox {
     }
 
     pub async fn create_snapshot(&self, snapshot_path: &str) -> CResult<()> {
-        let ch = self.ch.as_ref().unwrap().lock().await;
-        ch.pause_vm().await?;
-        ch.snapshot_vm(format!("file://{}", snapshot_path).as_str())
-            .await?;
-        ch.resume_vm().await
+        self.ensure_pause_vm_allowed().await?;
+
+        {
+            let ch = self.ch.as_ref().unwrap().lock().await;
+            ch.pause_vm().await?;
+        }
+        self.mark_pause_vm_committed().await;
+
+        let snapshot_result = {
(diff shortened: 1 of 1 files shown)


fslongjin:
@nemo1105 Thanks for the PR. Before we proceed with review, two things need to be addressed:

## 1. Please clarify the purpose of your changes in the PR description

The current description is fairly high-level ("Improve pause/resume failure recovery"), but the bot review surfaced several concerns that need explanation:

- The `wait_notify` error handling was downgraded from fatal (`?` propagation) to warn-only, which could **silently mask cases where the VM never actually stopped**.
- `create_snapshot()` completely bypasses the new state management logic you introduced.

Please elaborate: **What specific production issue did you encounter? What real-world scenario is this PR addressing?** This context is necessary for reviewers to evaluate whether the approach is correct and whether the notable behavioral changes are intentional.

## 2. Please run actual tests in your own cluster

The PR is marked "Not run locally" under Testing. This is not acceptable for a change touching the pause/resume state machine — a bug here would leave sandboxes in an inconsistent state.

Please test in your own environment:
- Run the existing unit tests for this module
- Manually exercise the key paths: pause → resume, pause → snapshot, pause failure → rollback, etc.
- Update the PR description with your test results

We cannot proceed with the review until both items are addressed. Thanks.

nemo1105:
@fslongjin I have test it in TencentCloud CVM using OpenCloudOS 9 OS. But I'm not sure this PR is perfact, so I'll close this PR.

PR #365, opened 2026-05-27
feat(python-sdk): add template creation API

tinklone:
## Summary
- add Python SDK models and client helpers for template creation APIs
- support stream-based template creation flows
- add unit coverage and opt-in E2E tests for template APIs

## Tests
- `python3 -m pytest tests -q` in `sdk/python` on cubesandbox002: 111 passed, 2 skipped
- Local validation skipped because this host does not have pytest installed

### DIFF
--- a/sdk/python/pyproject.toml
+++ b/sdk/python/pyproject.toml
@@ -10,6 +10,7 @@ license = { text = "Apache-2.0" }
 requires-python = ">=3.9"
 dependencies = [
     "httpx>=0.27",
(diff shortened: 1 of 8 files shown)


sdk/python/pyproject.toml
ls-ggg, line 7: Should we consider bumping the version number here?

fslongjin: > Should we consider bumping the version number here?

I'll bump it later

PR #378, opened 2026-05-27
fix(network-agent): prevent out-of-CIDR IP allocation on CIDR change

woshihoujinxin:
## Problem

When the CIDR in Cubelet config is changed and network-agent restarts, several code paths silently accepted old-CIDR IPs via `Assign()`, causing sandboxes to receive IPs from the wrong network segment.

## Changes

- **ipam**: add `Contains()` to check if IP falls within current CIDR range
- **recover()**: destroy kernel taps with out-of-range IPs instead of pooling them
- **reconcileState()**: reject sandbox with IP outside current CIDR
- **dequeueTapLocked()**: defensively destroy out-of-range taps from pool
- Wire CIDR and tap_init_num via environment variables
- Add comprehensive unit tests covering CIDR change scenarios

## Testing

- Unit tests in `cidr_change_test.go` cover recover, reconcileState, and dequeue paths
- IPAM `Contains()` tested in `ipam_test.go`

### DIFF
--- a/network-agent/cmd/network-agent/logging.go
+++ b/network-agent/cmd/network-agent/logging.go
@@ -67,9 +67,14 @@ func initLoggerWithDir(logDir, logLevel string, rollNum, rollSizeMB int) error {
 	// Keep stdlib log flags empty so the message body isn't double-prefixed with timestamps.
 	stdlog.SetOutput(&cubeLogStdWriter{})
 	stdlog.SetFlags(0)
+	// Get LocalIP safely
+	localIP := "unknown"
+	if CubeLog.LocalIP != "" {
+		localIP = CubeLog.LocalIP
+	}
 	CubeLog.WithContext(context.Background()).Infof(
 		"network-agent logger initialized: module=%s version=%s localip=%s level=%s logpath=%s",
-		defaultModuleName, resolveVersion(), CubeLog.LocalIP, strings.ToUpper(logLevel), logDir,
+		defaultModuleName, resolveVersion(), localIP, strings.ToUpper(logLevel), logDir,
 	)
 	return nil
 }
--- a/network-agent/cmd/network-agent/main.go
+++ b/network-agent/cmd/network-agent/main.go
@@ -12,6 +12,7 @@ import (
 	"os"
 	"os/signal"
 	"reflect"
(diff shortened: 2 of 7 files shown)


fslongjin:
The premise of this PR is wrong. Changing the local tap CIDR inherently breaks networking for all running sandboxes — their tap interfaces already have IPs from the old segment. Destroying those taps just kills their connectivity. There is no way to make this safe at runtime.

This is a deployment-time config that should never change on a live node. If the CIDR must change, drain the node and restart clean. Adding 600+ lines of validation and test code to handle a scenario that should not happen operationally only adds complexity to already-complex paths like `recover()`.

If anything is needed, it is a simple startup check: refuse to start if the CIDR in config does not match persisted state, and tell the operator to drain the node first. That is it.

What is the actual operational scenario that requires runtime CIDR changes?

chenhengqi:
I agree with @fslongjin . Let's focus on #379 instead.

chenhengqi (changes requested):
I think this change is not the right way.

woshihoujinxin:
The default CIDR conflicts with my K8s cluster network range, which is why I modify the CIDR after the initial installation. Changing the CIDR requires cleaning up the TAP network resources created during that first install. At this point, no sandboxes have been started yet. The scenario I'm describing is: 
  ▎ discovering a CIDR conflict after the initial installation that prevents the K8s cluster from replying to packets, forcing a CIDR change.

woshihoujinxin:
Additionally, if the TAP devices were created during installation and no sandbox is currently using them, they should be removed when the service is stopped. The current situation is: after installation, the system doesn't work, and when I try to uninstall and reinstall, the system has left behind a bunch of 
  ▎ orphaned TAP resources that prevent network-agent from starting after reinstallation.

chenhengqi:
> Additionally, if the TAP devices were created during installation and no sandbox is currently using them, they should be removed when the service is stopped. The current situation is: after installation, the system doesn't work, and when I try to uninstall and reinstall, the system has left behind a bunch of ▎ orphaned TAP resources that prevent network-agent from starting after reinstallation.

A reboot should work.

Let's stick to fixing the installation script. Thanks.

PR #379, opened 2026-05-27
feat(deploy): support CIDR and tap_init_num configuration via environment variables

woshihoujinxin:
## Problem

Operators had no way to override CIDR and tap pool size in one-click deployment without manually editing config files.

## Changes

- Add `CUBE_CIDR` and `CUBE_TAP_INIT_NUM` to `env.example`
- Propagate env vars through `install.sh`, `common.sh`, `up.sh`
- Cleanup TAP devices in `down-local.sh` to prevent restart race condition
- Add `test-cidr-replace.sh` for CIDR swap validation
- Update `Cubelet/config/config.toml` and `cubemaster.yaml` defaults

## Testing

- `test-cidr-replace.sh` validates CIDR swap scenario end-to-end


fslongjin:
I have a different perspective on this PR's approach and would like to discuss it.

## Core Issue

This PR splits the 4 CIDRs in `denyOut` into 4 separate environment variables, each with a semantically-named variable (`CUBE_SANDBOX_DENY_OUT_PRIVATE_A`, `CUBE_SANDBOX_DENY_OUT_CGN`, `CUBE_SANDBOX_DENY_OUT_PRIVATE_B`, `CUBE_SANDBOX_DENY_OUT_SANDBOX_CIDR`), and embeds 4 individual placeholders in the JSON template inside `cubemaster.yaml`.

This has several problems:

1. **Mismatch with the data model**: `DenyOut` is `[]string` in Go (see `cubebox.pb.go:2960`, `types/types.go:109`) — it's a **variable-length list**. But this PR turns it into a **fixed 4-element structure**, losing the list semantics entirely.

2. **Poor extensibility**: If an operator wants to add a 5th deny CIDR (e.g., `198.18.0.0/15` for network benchmarking), they need to modify the code — add a new placeholder, a new sed expression, and a new env var. If they want to reduce to 3 CIDRs, it's even more awkward because the JSON array structure can't have empty elements.

3. **Business semantics leaked into the deployment layer**: `denyOut` is a generic network policy list. Which specific CIDRs it contains is a configuration concern. But by naming each variable semantically (PRIVATE_A = RFC1918 10.0.0.0/8, CGN = RFC6598 100.64.0.0/10), this PR bakes those semantics into the deployment scripts — creating an implicit coupling that shouldn't exist.

## Suggested Approach

A better approach would be to **preserve the list as a whole**: use a **single placeholder** for the entire `denyOut` array in the template, and a **single env var** to carry the whole list.

### Recommended solution

**`cubemaster.yaml` template**:
```json
"denyOut": __CUBE_SANDBOX_DENY_OUT__
```

**`env.example`** (a single env var using JSON array format):
```
# Sandbox deny-out CIDRs, as a JSON array.
# These CIDRs will be blocked for outbound traffic from sandboxes.
# Default: deny RFC1918 private networks, CGN space, and the sandbox CIDR itself.
CUBE_SANDBOX_DENY_OUT=["10.0.0.0/8","100.64.0.0/10","172.16.0.0/12","192.168.0.0/18"]
```

**`common.sh` rendering logic** (simplified to a single sed):
```bash
sed -i "s|__CUBE_SANDBOX_DENY_OUT__|${CUBE_SANDBOX_DENY_OUT}|g" "${cfg}"
```

This way:
- `denyOut` remains a variable-length list — operators can freely add or remove CIDRs
- Adding a CIDR requires only changing the env var value, no code changes
- The data model and deployment config semantics are aligned

Optionally, you could add validation (e.g., via `jq`) to verify the env var is a valid JSON array of strings.

### Alternative

If JSON-in-an-env-var feels awkward, a comma-separated list works too, with sed assembling the JSON array:
```
CUBE_SANDBOX_DENY_OUT_CIDRS=10.0.0.0/8,100.64.0.0/10,172.16.0.0/12,192.168.0.0/18
```

Then construct the JSON array during replacement using `jq` or `printf` to format the CIDR list into a JSON array before inserting.

---

@woshihoujinxin Interested in your thoughts on this.

configs/single-node/cubemaster.yaml
```diff
@@ -40,7 +40,7 @@ req_template_conf:
     Poststop: true
     Prestop: true
   cube_box_req_template: >-
-    {"volumes":[{"name":"tmp","volume_source":{"empty_dir":{"medium":0}}}],"containers":[{"name":"cubebox-default","envs":[{"key":"TZ","value":"Asia/Shanghai"},{"key":"TERM","value":"xterm"}],"volume_mounts":[{"name":"tmp","container_path":"/"}],"security_context":{"privileged":true,"readonly_rootfs":false,"no_new_privs":false}}],"network_type":"tap","cubevs_context":{"allowInternetAccess":true,"denyOut":["10.0.0.0/8","100.64.0.0/10","172.16.0.0/12","192.168.0.0/18"]}}
+    {"volumes":[{"name":"tmp","volume_source":{"empty_dir":{"medium":0}}}],"containers":[{"name":"cubebox-default","envs":[{"key":"TZ","value":"Asia/Shanghai"},{"key":"TERM","value":"xterm"}],"volume_mounts":[{"name":"tmp","container_path":"/"}],"security_context":{"privileged":true,"readonly_rootfs":false,"no_new_privs":false}}],"network_type":"tap","cubevs_context":{"allowInternetAccess":true,"denyOut":["__CUBE_SANDBOX_DENY_OUT_PRIVATE_A__","__CUBE_SANDBOX_DENY_OUT_CGN__","__CUBE_SANDBOX_DENY_OUT_PRIVATE_B__","__CUBE_SANDBOX_DENY_OUT_SANDBOX_CIDR__"]}}
```
chenhengqi: I commented on your last PR. No need to change these CIDRs here.

PR #380, opened 2026-05-27
fix(cubelet): improve probe timeout error messages

woshihoujinxin:
## Problem

Probe timeout errors provided no context about how long the timeout was or what to check, making it hard for operators to diagnose issues.

## Changes

- Add elapsed time and actionable hints to timeout error messages
- Add `describeProbe()` to produce human-readable probe state summaries
- Add unit tests in `probe_describe_test.go`
- Update `go.sum` checksum for `containerd v2.2.2`

## Testing

- Unit tests in `probe_describe_test.go` cover all probe state descriptions


fslongjin (changes requested):
I don't think attributing probe timeout solely to network configuration is correct. Probe timeouts can have many root causes:

- The application inside the sandbox may not be ready in time (slow startup, resource contention, etc.)
- The probe target port may not be listening yet
- Resource constraints (CPU/memory/IO) slowing down the sandbox initialization

The updated message:

> "This may indicate a network connectivity issue between the host and sandbox. Check that the cube-dev interface and CIDR configuration are consistent."

Is too specific — it directs users to check cube-dev and CIDR when the real issue could be entirely unrelated to networking. This is misleading and could cause wasted debugging effort.

The addition of elapsed time and probe configuration details (via `describeProbe`) is a good improvement. I'd suggest removing the network-specific guidance and keeping the error message more neutral, for example:

> "Sandbox probe timed out after %v on %s. %s"

This still gives operators the key information (duration, target, probe config) without implying an incorrect root cause.

PR #383, opened 2026-05-27
fix(CubeMaster): retry mkfs.ext4 with larger image on failure

chenggui53:
The previous fixed 256 MiB overhead in createExt4Image was insufficient for images with many files or large content. For example, the gravitational.teleport image (3.71 GiB, 93244 files) requires ~477 MiB of ext4 metadata overhead (inode tables + journal + reserved blocks + block alignment waste), causing mkfs.ext4 to fail with "Could not allocate block" when the 4 GiB image ran out of space.

Changes:
- Add directoryStats() that returns both total size and file count in a single walk
- Add estimateExt4Overhead() that dynamically calculates ext4 metadata overhead based on file count (inode table + block alignment waste) and image size (journal + reserved blocks) using a two-pass approach
- Rewrite createExt4Image() to use dynamic overhead with structured logging
- Add comprehensive tests for directoryStats and estimateExt4Overhead

### DIFF
--- a/CubeMaster/pkg/templatecenter/template_image.go
+++ b/CubeMaster/pkg/templatecenter/template_image.go
@@ -1586,29 +1586,57 @@ func exportImageRootfs(ctx context.Context, source *resolvedSourceImage, workDir
 	return tarPath, nil
 }
 
+// maxExt4Retries is the maximum number of times to bump the image size when
+// mkfs.ext4 fails. Each retry doubles to the next power-of-2 GiB. Five
+// retries covers the jump from 1 GiB to 32 GiB, which is more than enough
+// for any realistic rootfs.
+const maxExt4Retries = 5
+
 func createExt4Image(ctx context.Context, rootfsDir, ext4Path string) error {
-	sizeBytes, err := directorySize(rootfsDir)
+	sizeBytes, fileCount, err := directoryStats(rootfsDir)
 	if err != nil {
 		return err
 	}
 
-	raw := sizeBytes + 256*1024*1024
 	const gib int64 = 1024 * 1024 * 1024
+
+	// Start with double the rootfs size; the retry loop will bump further if
+	// mkfs.ext4 needs more space (inode tables, journal, reserved blocks, etc.).
+	raw := sizeBytes * 2
 	if raw < gib {
 		raw = gib
 	}
 
+	// Round up to the next power-of-2 GiB.
 	gibs := (raw + gib - 1) / gib
 	pow := int64(1)
 	for pow < gibs {
 		pow <<= 1
 	}
 	imageSize := pow * gib
-	if err := runCommand(ctx, "", "truncate", "-s", strconv.FormatInt(imageSize, 10), ext4Path); err != nil {
-		return fmt.Errorf("truncate ext4 image failed: %w", err)
-	}
-	if err := runCommand(ctx, "", "mkfs.ext4", "-F", "-d", rootfsDir, ext4Path); err != nil {
-		return fmt.Errorf("mkfs.ext4 failed: %w", err)
+
+	for attempt := 0; attempt <= maxExt4Retries; attempt++ {
+		log.G(ctx).Infof("createExt4Image: rootfs=%.2f GiB (%d files), "+
+			"imageSize=%.2f GiB, attempt=%d/%d",
+			float64(sizeBytes)/float64(gib), fileCount,
+			float64(imageSize)/float64(gib),
+			attempt+1, maxExt4Retries+1)
+
+		if err := runCommand(ctx, "", "truncate", "-s", strconv.FormatInt(imageSize, 10), ext4Path); err != nil {
+			return fmt.Errorf("truncate ext4 image failed: %w", err)
+		}
+		if err := runCommand(ctx, "", "mkfs.ext4", "-F", "-d", rootfsDir, ext4Path); err == nil {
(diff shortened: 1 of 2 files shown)


fslongjin (changes requested):
I did an extra round of validation on this change, and I agree with the direction, but I do not think the current calculation is proven correct yet, so I am requesting changes for now.

The main concern is `estimateExt4Overhead()` in `CubeMaster/pkg/templatecenter/template_image.go`.

1. The inode table estimate is currently derived from `fileCount`, but the code is still invoking the default `mkfs.ext4 -F -d` behavior without passing `-N` or `-i`. In practice, the default inode count appears to be driven much more by the final filesystem size than by the file count. I verified this locally with `mkfs.ext4 -n -F` and saw:
   - a blank 4 GiB ext4 image creates 262144 inodes
   - a blank 8 GiB ext4 image creates 524288 inodes

   That means the current model can significantly underestimate inode table overhead for large rootfs images with relatively few files. The risk is not just that one breakdown item is off, but that `raw` stays below the next GiB boundary and the code still picks a 4 GiB image when the real ext4 defaults would push it past that threshold. In that case we could still reproduce the same `mkfs.ext4` space allocation failure this PR is trying to fix.

2. `directoryStats()` only counts regular files and ignores directory metadata entirely. Directory-heavy rootfs layouts will therefore still be systematically underestimated because directory inodes and directory blocks are not part of the model.

3. The new tests mostly validate what the current heuristic returns, but they do not validate that the heuristic is close to real `mkfs.ext4` defaults. I think we need at least one of the following before merging:
   - a comparison against `mkfs.ext4 -n`
   - a validation against a real generated ext4 image via `tune2fs`/`dumpe2fs`
   - a boundary test near the 4 GiB / 8 GiB threshold with a small file count
   - a directory-heavy test case

I also tried this against `ubuntu:24.04`, and for a small image the estimate is in roughly the right range. However, that does not cover the large near-threshold scenario this PR is intended to fix, so I do not think it is enough evidence yet.

My suggestion would be either:
- make the inode / journal / reserved calculations follow `mkfs.ext4` defaults more closely, or
- probe a candidate size with `mkfs.ext4 -n` and use that result to decide the final ext4 size.

Until we have that, I do not think this is safe to merge.

chenggui53:
@fslongjin  Thanks for the thorough review. You're right that the inode estimation model has fundamental issues — mkfs.ext4 allocates inodes based on filesystem size, not file count, and our heuristic can't reliably replicate that.

Instead of trying to perfect the estimation, I've taken a different approach: **add a retry loop that bumps to the next power-of-2 GiB when mkfs.ext4 fails**.

The key insight is: we don't need to guess exactly what mkfs.ext4 needs — mkfs.ext4 itself will tell us if the image is too small. The estimation (`estimateExt4Overhead`) is kept as a "good starting point" to minimize retries, but correctness no longer depends on it.

Changes:
- `createExt4Image` now retries up to 5 times, doubling the image size
  (next power-of-2 GiB) on each mkfs.ext4 failure
- This naturally handles ALL the edge cases you identified:
  - Inode count driven by filesystem size → larger image = more inodes
  - Directory metadata → more space accommodates it
  - Any other mkfs.ext4 overhead we haven't modeled
- Added structured logging with attempt count for debugging
- Added integration tests (`TestCreateExt4ImageSmallRootfs`,
  `TestCreateExt4ImageEmptyRootfs`) that exercise real mkfs.ext4

The retry approach is more robust than perfecting the estimation because it delegates the "is this space enough?" decision to mkfs.ext4 itself. At most 5 retries covers 1→32 GiB, which handles any realistic rootfs.

PTAL

fslongjin:
Thanks @chenggui53 for working on this fix! The ext4 metadata overhead issue you identified for file-heavy images is real, and I appreciate the analysis with the gravitational.teleport image (93,244 files).

However, I think the "retry with powers of 2" approach has some drawbacks:
- It can waste significant disk space (each retry doubles the image size, and the failed ext4 file from the previous attempt may persist on disk for some time before cleanup)
- It increases build latency (each retry involves a full `mkfs.ext4` run, which is expensive for large images)
- The power-of-2 alignment itself is overly aggressive — a 5 GiB image aligns to 8 GiB, wasting nearly 3 GiB of apparent size

I have addressed this same root cause in PR #472, using a one-shot approach instead of retries:
- **Triple-overhead model**: fixed overhead (256 MiB default) + percentage overhead (10% default) + per-file overhead (1 KiB/file), calculated upfront in a single `filepath.Walk`
- **256 MiB alignment** instead of power-of-2, which saves ~35-50% ext4 apparent size for most images
- **Configurable via env vars** (`CUBEMASTER_EXT4_OVERHEAD_PERCENT`, `CUBEMASTER_EXT4_FIXED_OVERHEAD_MIB`) so operators can tune for unusually file-dense images

For the teleport image with 93K files, PR #472 would calculate:
- Fixed: 256 MiB + Percentage: 10% × 3.71 GiB ≈ 380 MiB + Per-file: 93K × 1 KiB ≈ 91 MiB
- Total overhead: ~727 MiB → raw = 4.44 GiB → aligned to 256 MiB = **4.5 GiB** (one shot, no retry)

Closing this PR in favor of #472 which provides a more comprehensive solution. Thanks again for the contribution!

fslongjin:
Closed in favor of #472 which solves the ext4 sizing problem with a one-shot triple-overhead model instead of retries.

PR #384, opened 2026-05-28
fix(one-click): force 4K block size when building guest rootfs

chenhengqi:
mkfs.ext4 picks the block size based on filesystem size, so small guest images end up with 1K blocks. CubeShim boots the guest kernel with rootflags=dax, and the DAX path requires the ext4 block size to match the page size (4K on x86_64) — a 1K-block rootfs panics at boot time.

Pass -b 4096 explicitly in build_guest_image_artifacts. The downstream shrink_ext4_image already reads block size dynamically via dumpe2fs, and the 2 MiB pmem alignment it enforces is a multiple of 4096, so no other changes are needed.


ls-ggg:
LGTM

lisongqian (approved):
LGTM

PR #388, opened 2026-05-28
virtio-devices: vsock: RST vsocks on snapshot restore

up2wing:
Backport of upstream cloud-hypervisor PR #7958.

On snapshot restore the host-side Unix socket connections are gone, but the guest still believes its vsock connections are alive, which makes guest-side connections hang until they time out. Proactively queue an RST packet for every previously-recorded connection so the guest learns about the disconnection immediately, matching the virtio-vsock spec.

Save the active (local_port, peer_port) pairs into VsockState and queue MuxerRx::RstPkt for each of them when the device is restored. `#[serde(default)]` keeps old snapshots compatible, and the new VsockBackend trait methods carry default implementations so other backends (e.g. TestBackend) need no change.

Assisted-by: Anthropic:claude-opus-4-7

### DIFF
--- a/hypervisor/virtio-devices/src/vsock/device.rs
+++ b/hypervisor/virtio-devices/src/vsock/device.rs
@@ -362,6 +362,8 @@ pub struct Vsock<B: VsockBackend> {
 pub struct VsockState {
     pub avail_features: u64,
     pub acked_features: u64,
+    #[serde(default)]
+    pub connections: Vec<(u32, u32)>,
 }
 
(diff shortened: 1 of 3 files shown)


lisongqian:
Upstream commit id should be added to the commit message?

fslongjin (approved):

up2wing:
> Upstream commit id should be added to the commit message?

There is PR number in the commit comment, I suppose that's enough.

PR #389, opened 2026-05-28
feat(snapshot): add soft-dirty per-cycle incremental memory snapshots

fslongjin:
### Summary

This PR introduces a new memory-snapshot type — **soft-dirty** — across
the hypervisor and Cubelet, and switches `CommitSandbox` to use it.
Soft-dirty produces a true *per-commit delta* (only pages dirtied since
the previous commit), reflink-cloned on top of the previous snapshot's
memory file. The result is that repeated commits on the same sandbox
write **~95× less** to disk in steady state versus full snapshots,
without changing user-visible semantics.

The change has two layered parts:

1. **Hypervisor** (by @lisongqian ) — adds `SnapshotType::SoftDirty` (CLI:
   `--snapshot-type soft-dirty`) backed by `/proc/self/clear_refs`
   plus `/proc/self/pagemap` bit 55. Exists alongside the older
   `incremental` (pagemap_anon) path, which is *cumulative since
   restore* and so degrades on long-running VMs.
2. **Cubelet** (by @fslongjin ) — wires soft-dirty into `CommitSandbox`, reflink-cloning
   the previous snapshot's memory file as the base, and correctly
   maintains the binding chain across `commit → commit`,
   `commit → rollback → commit`, and missing-base scenarios.

### Why soft-dirty (and not the existing `incremental`)?

The existing `incremental` path uses `pagemap_anon`, which captures
**all anonymous pages dirtied since VM restore**. The set is
monotonically growing for the lifetime of the VM, so a long-lived
sandbox's "incremental" snapshot is no longer small. Soft-dirty resets
the bitmap after each snapshot, so commit N+1 contains only pages
written between commits N and N+1.

Combined with reflink-cloning the previous snapshot file as the base,
this gives O(actual writes since last commit) on-disk bytes per commit
instead of O(working set since restore).

### Hypervisor changes

* New `SnapshotType::SoftDirty` enum variant; CLI and JSON wire format
  accept `soft-dirty`.
* `hypervisor/vmm/src/soft_dirty.rs`: kernel-support probe (logs once
  per process and disables itself on failure), `clear_soft_dirty()`
  via `/proc/self/clear_refs`, dirty-range enumeration via
  `/proc/self/pagemap`. Self-disabling: any `clear_refs(4)` failure
  causes the next snapshot to fall back to `pagemap_anon`.
* `MemoryManager`:
  - On a soft-dirty snapshot, write only the dirty bytes onto the
    destination memory file and emit savings stats. The destination
    is expected to already contain the previous snapshot's bytes —
    Cubelet supplies that via reflink.
  - Sparse-base sanity: if no previous base file exists, the dirty
    set is written onto a fresh sparse file and a warning is logged
    so operators can spot the misuse.
* Documentation: `hypervisor/docs/snapshot_restore.md` gains a
  soft-dirty section and a comparison table against full +
  pagemap_anon.

#### 1. Reflink-cloned base memory in `CommitSandbox`

`CommitSandbox` now resolves the previous snapshot's memory file
through `cb.Labels[cube.master.runtime.snapshot.id]` (the *runtime*
binding, kept in sync below) and reflink-clones it as the destination
for the new snapshot, then asks Cloud Hypervisor for `--snapshot-type
soft-dirty`. The new helper `prepareCommitMemoryArtifact`
(`Cubelet/services/cubebox/snapshot_base_memory.go`) encapsulates
this base-resolution + clone step.

#### 2. Graceful fallback to `full` when the base is gone

If the previous snapshot was deleted, GC'd, or the binding label
is unset, `resolveBaseMemoryObject` returns
`ErrNoBaseMemoryForIncremental` and `prepareCommitMemoryArtifact`
falls back to a fresh empty memory volume + `--snapshot-type full`.
This preserves correctness for restored VMs, where passing an empty
base to `soft-dirty` would corrupt the snapshot. The fallback is
logged at WARN with the underlying reason so operators can spot
broken lineages.

#### 3. Binding-label update on every successful commit

`CommitSandbox` was previously **not** updating
`cb.Labels[cube.master.runtime.snapshot.id]` after a successful commit.
This was harmless for `pagemap_anon` (cumulative anyway) but a hidden
correctness bug for soft-dirty: the next commit would resolve the
*template's* memory file as base instead of the just-committed
snapshot, so any guest page dirtied between create and commit A and
not touched between A and B would silently revert to its template-era
content on a clone of B. After every successful commit we now stamp
the new snapshot id back into the binding label and persist it via
`SyncByID`, which makes the next `resolveBaseMemoryObject` pick the
right base.

This also handles the rollback chain
`T → A → B → rollback A → commit C` correctly: rollback already
resets the binding back to A, so commit C's base is A and its delta
is "writes since rollback". Without the binding update, C would
have inherited B's bytes instead.

#### 4. Tests

* `snapshot_incremental_test.go`: extended to cover the new
  `soft-dirty` type in `normalizeSnapshotType` and
  `buildCubeRuntimeSnapshotArgs`.
* `snapshot_base_memory_fallback_test.go`: new file covering
  - `ErrNoBaseMemoryForIncremental` is the canonical sentinel for
    fallback,
  - `resolveBaseSnapshotID` follows the commit chain across the
    `T → A → B → rollback A → C` scenario,
  - the runtime-binding label takes precedence over the
    create-time annotation when present.

### CubeShim changes

`CubeShim/shim/src/snapshot/cmd.rs` — extend `--snapshot-type` help
text to advertise `soft-dirty` alongside `full` / `incremental`.

### Test plan

Verified end-to-end on two distinct PVM hosts using the e2b
code-interpreter template image.

#### 1. Functional regression — does not break existing flows

Full snapshot/rollback/clone/run_code acceptance: PASS.

#### 2. Soft-dirty correctness — chain + rollback

Two scenarios exercised:

* **Scenario A**: `T → commit A → commit B → clone(B)` — verifies
  pre-A and pre-B markers (filesystem and Python kernel state)
  are both visible on a fresh clone of B. Catches the
  binding-update bug above.
* **Scenario B**: `T → A → B → rollback A → mutate → commit C →
  clone(C)` — the rollback-chain scenario. Verifies that
  clone(C) carries pre-A + post-rollback writes but **does
  not** see the rolled-back A→B writes, confirming C's
  reflink source is A (not T, not B).

Both scenarios pass. A 5×stability run produced 4/5 PASS; the
single failure was a `504 Gateway Time-out` on data-plane proxy
readiness during clone(C) — unrelated to soft-dirty correctness
(the snapshot itself was produced correctly; the new sandbox's
envd just lost the race with the proxy table) and reproduces on
the baseline acceptance pre-change.

#### 3. Disk-write reduction — the headline benefit

Measured *actual* per-commit written bytes via FIEMAP
`FIEMAP_EXTENT_SHARED` analysis: bytes in extents **without** the
SHARED flag are the only bytes this commit physically wrote;
SHARED extents are reflink pointers to the base file's blocks
(no fresh I/O).

5 consecutive commits on a 2 GB-RAM sandbox:

| Source | apparent | exclusive (actual writes) | shared (reflink, no I/O) |
|---|---|---|---|
| commit#1 | 2000 MB | 8.0 MB | 160.8 MB |
| commit#2 | 2000 MB | 1.7 MB | 167.1 MB |
| commit#3 | 2000 MB | 1.8 MB | 166.9 MB |
| commit#4 | 2000 MB | 1.7 MB | 167.1 MB |
| commit#5 | 2000 MB | 1.8 MB | 167.0 MB |
| **total** | 10000 MB | **14.9 MB** | 829.0 MB |

Baseline note: the apparent-size column overstates a full
snapshot's actual cost because the filesystem folds zero pages
into holes. The honest projected baseline is `n × allocated`
of the first commit's memory file = 5 × ~169 MB ≈ **844 MB**.
Soft-dirty actual writes: **14.9 MB**. **Disk-write saving:
98.2%**. Steady-state per-commit saving: ~95× (1.7 MB vs ~165 MB).

A second host produced 14.9 MB / 862.8 MB ≈ 98.3% — within
measurement noise.

CubeVmm logs match: `Soft-dirty snapshot: total=2097152000 bytes
(512000 pages), dirty=1769472 bytes (432 pages), savings=99.9%`.

#### 4. Unit tests

`go test ./Cubelet/services/cubebox/...` — passes for the new
tests (`TestErrNoBaseMemoryForIncrementalIsSentinel`,
`TestResolveBaseSnapshotIDFollowsCommitChain`,
`TestRuntimeBindingLabelOverridesCreateAnnotation`,
extended `TestBuildCubeRuntimeSnapshotArgsAlwaysIncludesSnapshotType`,
extended `TestNormalizeSnapshotTypeAcceptsKnownValues`).

### Operator-visible behaviour changes

* `CommitSandbox` now produces `--snapshot-type soft-dirty` by
  default (previously `incremental`). Existing snapshots are
  unaffected and remain restorable.
* When the previous snapshot has been deleted between commits,
  cubelet logs `CommitSandbox: base memory unavailable (...),
  falling back to full snapshot` at WARN, then performs a full
  snapshot. No request fails because of a missing base.
* Catalog files now carry `memory_kind=snapshot` for soft-dirty
  commits (vs `memory_kind=volume` for the fallback-to-full
  branch). This is the canonical signal for "did soft-dirty take
  effect".

### Notes for reviewers

* **Kernel requirement**: soft-dirty needs `CONFIG_MEM_SOFT_DIRTY=y`.
  Without it, hypervisor self-disables and falls back to
  `pagemap_anon` automatically (logs once per VMM process).
* **Filesystem requirement**: reflink. cubelet's cubecow store is
  already on a reflink-capable FS by design.
* **Forward compatibility**: `pagemap_anon` continues to work and
  is reachable via `--snapshot-type incremental`. Switching the
  default for `CommitSandbox` to soft-dirty is reversible — set
  the snapshot type back to `incremental` in the request to opt
  out per-commit.

### DIFF
--- a/CubeShim/shim/src/snapshot/cmd.rs
+++ b/CubeShim/shim/src/snapshot/cmd.rs
@@ -82,11 +82,14 @@ pub struct SnapshotArgs {
     )]
     pub vm_id: Option<String>,
 
-    /// Snapshot type: 'full' or 'incremental' (saves only CoW anonymous pages)
+    /// Snapshot type: 'full', 'incremental' (saves only CoW anonymous pages,
+    /// cumulative since restore) or 'soft-dirty' (true delta of pages
+    /// dirtied since the previous soft-dirty snapshot; falls back to
+    /// 'incremental' on kernels without CONFIG_MEM_SOFT_DIRTY).
     #[arg(
         long = "snapshot-type",
         value_name = "snapshot type",
-        help = "snapshot type: 'full' or 'incremental'",
+        help = "snapshot type: 'full', 'incremental' or 'soft-dirty'",
         default_value = "full"
     )]
     pub snapshot_type: String,
--- a/Cubelet/pkg/constants/const.go
+++ b/Cubelet/pkg/constants/const.go
@@ -187,6 +187,15 @@ const (
 	MasterAnnotationAppSnapshotTemplateID     = "cube.master.appsnapshot.template.id"
 	MasterAnnotationRuntimeSnapshotID         = "cube.master.runtime.snapshot.id"
 	MasterAnnotationRuntimeSnapshotAttachedAt = "cube.master.runtime.snapshot.attached_at"
+	// MasterAnnotationRuntimeRestoreSnapshotID tracks the snapshot id whose
+	// memory image the running VM was last *restored* from (set by Create
+	// for restore-from-snapshot path, and by Rollback). Unlike
+	// MasterAnnotationRuntimeSnapshotID, Commit does NOT bump this — Commit
+	// does not restart the VM, so the in-process pagemap_anon bitmap is
+	// still tracking "anon pages dirtied since the last restore", which
+	// matches the memory file recorded here.
+	MasterAnnotationRuntimeRestoreSnapshotID         = "cube.master.runtime.restore.snapshot.id"
+	MasterAnnotationRuntimeRestoreSnapshotAttachedAt = "cube.master.runtime.restore.snapshot.attached_at"
 
 	MasterAnnotationAppSnapshotVersion               = "cube.master.appsnapshot.version"
 	MasterAnnotationRootfsArtifactID                 = "cube.master.rootfs.artifact.id"
--- a/Cubelet/services/cubebox/appsnapshot.go
+++ b/Cubelet/services/cubebox/appsnapshot.go
@@ -552,6 +552,21 @@ const snapshotTypeFull = "full"
 // the destination file already contains (i.e. the reflink-cloned base).
 const snapshotTypeIncremental = "incremental"
 
+// snapshotTypeSoftDirty asks cube-runtime to write only the pages dirtied
+// since the previous soft-dirty snapshot (a true per-cycle delta) on top of
+// the destination memory file. The destination MUST already contain a valid
+// base image (the reflink-cloned previous snapshot's memory), otherwise pages
+// untouched-since-last-clear would read back as zero on restore. Cubelet
+// guarantees this precondition by reflink-cloning the binding base before
+// invoking cube-runtime; if the base cannot be resolved, the caller falls
+// back to snapshotTypeFull instead.
+//
+// The host kernel needs CONFIG_MEM_SOFT_DIRTY=y for soft-dirty to do
+// anything useful; on kernels without it the hypervisor silently downgrades
+// to the pagemap_anon (incremental) path, so this value is safe to send
+// unconditionally.
+const snapshotTypeSoftDirty = "soft-dirty"
+
 // normalizeSnapshotType defaults to snapshotTypeFull when an empty value is
 // supplied so callers that don't care (legacy code paths) keep producing full
 // snapshots.
@@ -559,6 +574,8 @@ func normalizeSnapshotType(snapshotType string) string {
 	switch strings.TrimSpace(strings.ToLower(snapshotType)) {
 	case snapshotTypeIncremental:
 		return snapshotTypeIncremental
+	case snapshotTypeSoftDirty:
+		return snapshotTypeSoftDirty
 	case "", snapshotTypeFull:
 		return snapshotTypeFull
 	default:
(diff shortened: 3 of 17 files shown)


xeonliu:
@lisongqian When will anonymous=0, soft-dirty=1 happen?

lisongqian:
> @lisongqian When will anonymous=0, soft-dirty=1 happen?

@xeonliu When the kernel enables `CONFIG_MEM_SOFT_DIRTY`, the soft-dirty bit of page always =1 if we don't exec `clear_refs`. So `anon=0,soft-dirty=1` means this page is shared and it's already in the previous snapshot file.

PR #400, opened 2026-05-29
hypervisor/vmm: drop sync_all() from all snapshot write paths

lisongqian:
Snapshot files live on local fs and are consumed via the page cache.
Remove it from pagemap-anon / soft-dirty / dirty-log / full paths since 
fsync is unnecessary and was dominating snap wall time.

Signed-off-by: Songqian Li <sionli@tencent.com>


fslongjin (approved):
LGTM

Issue #403, opened 2026-05-29
After calling `pause`, the underlying VM/task is already `PAUSED`, but Cube metadata remains stuck in `pausing`.


zhangningshitiancai-rgb:
## Summary

  After calling `pause`, the underlying VM/task is already `PAUSED`, but Cube metadata remains stuck in `pausing`. This creates a state mismatch
  and can block later delete operations.

  ## Environment

  - CubeSandbox version / commit: tested on 2026-05-29 from local `CubeSandbox-master`
  - Host OS and kernel version: Linux, server `root@47.105.122.2`
  - KVM info (`modinfo kvm`): not collected
  - Deployment mode: single-node
  - Relevant component: CubeMaster / Cubelet / CubeShim / CubeVmm

  ## Steps to Reproduce

  1. Create a sandbox from a READY template.
  2. Call the pause API:

     ```bash
     curl -X POST http://127.0.0.1:3000/sandboxes/{sandboxID}/pause

  3. Wait for the underlying VM pause to complete.
  4. Check status with:

     cubemastercli cubebox list
     ctr --address /data/cubelet/cubelet.sock --namespace default tasks ls

  ## Expected Behavior

  The sandbox should move from running to paused after pause completes.

  A later resume should move it back from paused to running.

  ## Actual Behavior

  The VM/task is paused successfully, but CubeMaster / Cubelet metadata stays in pausing.

  Observed output:

  cubemastercli cubebox list:

  sandbox_id                          status     host_id           create_at              pause_at
  50860503c73f4ce483ed9d0c17dffdff    pausing    172.17.217.233    2026-05-29 17:22:57    2026-05-29 17:24:03

  But containerd shows the task is actually PAUSED:

  ctr --address /data/cubelet/cubelet.sock --namespace default tasks ls:

  50860503c73f4ce483ed9d0c17dffdff    278949    PAUSED

  CubeMaster logs show the pause request timing out:

  2026-05-29 17:24:33

  Update fail:
  {"ret":{"ret_code":130595,"ret_msg":"rpc error: code = DeadlineExceeded desc = context deadline exceeded"}}

  CubeVmm logs show that the VM pause was actually executed:

  API request event: VmPauseToSnapshot(...)
  PAUSE_EVENT received, pausing epoll loop

  So the state becomes inconsistent:

  - CubeMaster / Cubelet metadata: pausing
  - containerd task: PAUSED
  - VMM: pause completed
  - delete may fail later with sandbox not in normal state

[...]

PR #404, opened 2026-05-29
fix(cubelet): converge pause/resume state on ttrpc errors and shim events

kinwin-ustc:
When task.Pause/Resume returned a ttrpc error (DeadlineExceeded, canceled, ttrpc closed), the cubelet in-memory view could stay stuck at PAUSING/PAUSED while cubeshim had already switched the VM state, leaving PausingAt lingering forever.  close issue #403 

update.go:
- Wrap task.Pause/Resume with explicit 30s timeouts so the calls cannot be stretched arbitrarily by a missing or blocked upstream ctx.
- On error, reconcile against the shim's real status using a fresh Background ctx (carrying over the namespace) and straighten the in-memory view to the truth, while still returning TaskPauseFailed/TaskResumeFailed upstream for alerting.

events.go:
- Subscribe to /tasks/paused and /tasks/resumed and handle them in convertEvent/handleEvent as an event-driven fallback.
- Idempotently converge PausedAt/PausingAt for all containers, skipping terminal or user-deleted sandboxes.

Assisted-by: Claude Opus 4.8

### DIFF
--- a/Cubelet/services/cubebox/events.go
+++ b/Cubelet/services/cubebox/events.go
@@ -77,6 +77,8 @@ func (em *eventMonitor) subscribe(subscriber events.Subscriber) {
 	filters := []string{
 		`topic=="/tasks/exit"`,
 		`topic=="/tasks/oom"`,
+		`topic=="/tasks/paused"`,
+		`topic=="/tasks/resumed"`,
 		`topic~="/images/"`,
 	}
 	em.ch, em.errCh = subscriber.Subscribe(em.ctx, filters...)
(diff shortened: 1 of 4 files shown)


fslongjin:
I addressed one review item because it can produce a real correctness issue after a resume that succeeds through the shim but is only observed via the `/tasks/resumed` event path.

Before this change, the resumed-event handler only cleared `PausedAt` and `PausingAt`, but did not invalidate the runtime snapshot bindings. That left Cubelet believing the VM was still bound to the pre-resume snapshot lineage, even though CubeShim had actually restored the VM from its internal pause snapshot. In that state, a later `CommitSandbox` could choose an incremental memory base that no longer matches the VM's real restore boundary.

The fix makes all successful resume convergence paths use the same helper:
- normal `UpdateWithResume` success
- resume RPC error reconciliation when shim status is already `RUNNING`
- `/tasks/resumed` event reconciliation

That keeps the state transition logic consistent and guarantees that an opaque restore always invalidates stale runtime/restore-base bindings before the next commit. I also added tests for both the shared helper and the resumed-event path.

fslongjin (approved):

PR #417, opened 2026-06-01
feat(one-click): prevent Docker from creating directories at bind-mount file paths

fslongjin:
Docker creates empty directories when bind-mount source file paths do not exist on the host, causing runtime failures on fresh deployments. Introduce a multi-layer defense:

- Add `ensure_not_directory_for_file`, `prepare_file_output`, and `ensure_bind_mount_file` guards to detect and remove empty directories that Docker may have created at expected file paths
- Add `render_template_atomic` using temp-file + mv for atomic config writes, eliminating the partial-write window
- Add `ExecStartPre` hooks to all systemd service units so config files are rendered before container startup
- Add `check_runtime_file_paths_not_directories` preflight in install.sh to catch directory-vs-file mismatches before deployment
- Use `flock` to serialize concurrent compose file rendering when mysql and redis prepare hooks run in parallel
- Add `wait_for_udp_port` and CoreDNS `ExecStartPost` check for UDP readiness verification
- Update quickcheck to validate bind-mount source files at runtime
- Add comprehensive runtime file safety tests


fslongjin:
**Re: finding #5** — this is intentional and correct.

The early return for `compute` role is by design:
- Compute nodes don't run Docker containers at all — both `cubelet` and `network-agent` are native binaries (`Type=forking` / `Type=simple`), so there's no Docker daemon to create empty directories at bind-mount paths.
- The five paths listed in `one_click_runtime_file_paths` are under `cubeproxy/`, `coredns/`, and `webui/` — directories that are only installed for the control role. A fresh compute node doesn't even have these parent directories.
- In an upgrade scenario (control → compute), stale directories wouldn't matter because the compute role doesn't use them and no Docker daemon is active to re-create them.

The function returns the set of paths *at risk*, and for compute nodes that set is empty — so the early return is semantically accurate.

PR #418, opened 2026-06-01
fix(python-sdk): align envd defaults and network policy

tinklone:
## Summary
- default envd process and filesystem requests to the root user for older envd compatibility
- document the Python SDK process/filesystem transport behavior and sandbox lifecycle helpers
- extend Python SDK tests for default user headers and network blocking behavior

## Tests
- `python3 -m compileall -q cubesandbox tests`
- `pytest -q tests` on cubesandbox002: 119 passed, 2 skipped

### DIFF
--- a/sdk/python/cubesandbox/sandbox.py
+++ b/sdk/python/cubesandbox/sandbox.py
@@ -101,7 +101,7 @@ class Sandbox:
             template: Template ID. Falls back to ``CUBE_TEMPLATE_ID`` env var.
             timeout: Sandbox TTL in seconds. Defaults to ``Config.timeout`` (300).
             env_vars: Environment variables injected into the sandbox.
-            metadata: Arbitrary key-value metadata (e.g. network-policy, hostdir-mount).
+            metadata: Arbitrary key-value metadata (e.g. network-policy, host-mount).
             config: SDK config. Uses default (env-based) config if omitted.
 
         Returns:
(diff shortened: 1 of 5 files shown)


sdk/python/cubesandbox/sandbox.py
```diff
@@ -122,9 +122,11 @@ def create(
         if metadata:
             payload["metadata"] = metadata
         if not allow_internet_access:
-            payload["allowInternetAccess"] = False
+            payload["allow_internet_access"] = False
```
chenhengqi, line 125: Is this change compatible with CubeAPI?

https://github.com/TencentCloud/CubeSandbox/blob/65ef720479e0a27eadd5375334745cd45c80cb9a/CubeAPI/src/cubemaster/mod.rs#L613-L619

tinklone: Yes, this is compatible with the current CubeAPI request model.

The `allow_internet_access` field here is the external `POST /sandboxes` request field accepted by CubeAPI's `NewSandbox` model. CubeAPI currently documents this as a known SDK snake_case quirk:

```rust
/// Field names match exactly what the E2B SDK sends.
/// Rule: ID abbreviations → uppercase (templateID, sandboxID, envVars, autoPause);
///       allow_internet_access is a known SDK snake_case quirk.
...
/// SDK sends this as snake_case (known quirk).
pub allow_internet_access: Option<bool>,
```

CubeAPI then maps this external field into the internal CubeMaster request through:

```rust
cubevs_context: build_cubevs_context(body.allow_internet_access, body.network.as_ref())
```

The linked `allowInternetAccess` serde rename is for the internal CubeMaster `CubeVSContext` payload, not the Python SDK's direct `/sandboxes` request body. So the mixed style is intentional: top-level `allow_internet_access` for the CubeAPI request, and nested `network.allowPublicTraffic` / `allowOut` / `denyOut` for the CubeMaster network policy fields.

Issue #74, opened 2026-04-24
[Feature Request] Support batch template builds via the E2B API


NoneWait:
## Problem / Motivation

是否能够通过e2b的接口去批量build template？
参考这里：https://e2b.dev/docs/template/quickstart

fslongjin:
Thanks for the suggestion. This is not currently supported.

We may need to do some further research on the specific design and implementation details before deciding whether and how to support it.

PR #422, opened 2026-06-02
feat(CubeAPI): support e2b V3 template-build protocol end-to-end

chengjoey:
Add V3 routes: POST /v3/templates, GET .../files/{hash}, POST /v2/.../builds/{bid}, GET .../status (handlers/templates_v3.rs)

feat: #74


CubeAPI/src/routes.rs
```diff
@@ -521,4 +551,42 @@ mod tests {
...
     }
-}
+
+    /// Regression: e2b Python SDK `Template.build()` calls `POST /v3/templates`
+    /// first; prior to the V3 routes we returned 404 with an empty body, which
+    /// surfaced to users as `BuildException: 404: b''`. After the fix, the
+    /// route exists and returns the V3 envelope.
+    #[tokio::test]
+    async fn v3_template_build_routes_are_reachable() {
+        let server = test_server();
```
fslongjin: This test currently does not compile because `test_server()` is async and returns a Future. Please await it before using the `TestServer` methods:

`let server = test_server().await;`

Right now `cargo test` fails with `no method named post/get found for opaque type impl Future<Output = TestServer>`, so this PR cannot pass the existing Rust test suite in its current state.

CubeAPI/src/services/templates.rs
```diff
@@ -263,27 +413,662 @@ impl TemplateService {
...
-    pub async fn get_template_build_logs(&self, build_id: &str) -> AppResult<serde_json::Value> {
+    /// `GET /templates/{tid}/files/{hash}` — file-cache probe.
+    ///
+    /// Until the in-cluster builder lands we don't actually consume uploaded
+    /// tarballs. We answer `present=true` so the SDK skips uploading; this is
+    /// safe because `from_image`-based builds (the only flow CubeMaster
+    /// currently supports) don't need the build context.
+    pub fn v3_get_file_upload(&self, _template_id: &str, _files_hash: &str) -> AppResult<crate::models::V3TemplateFileUpload> {
+        Ok(crate::models::V3TemplateFileUpload {
+            present: true,
```
fslongjin, line 772: Always returning `present: true` makes the SDK believe the build context already exists server-side, so it will skip uploading the tarball. That is only safe for pure `fromImage` builds.

For Dockerfile/steps-based V3 builds this silently drops the build context, and the later trigger step may dispatch a registry image reference that was never produced. I think unsupported context builds should fail explicitly here, or this endpoint should return a real upload URL and persist/consume the uploaded context.

CubeAPI/src/services/templates.rs
```diff
@@ -263,27 +413,662 @@ impl TemplateService {
...
+            .as_ref()
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty())
+        {
+            // Re-use an already-built CubeSandbox template as the base. We
+            // synthesise a CubeMaster reference of the form `cube://<tid>`,
+            // letting downstream callers resolve it. Adjust to your local
+            // convention if needed.
+            format!("cube://{}", parent)
+        } else if !ctx.image_ref.is_empty() {
```
fslongjin: This fallback uses `ctx.image_ref` as long as it is non-empty, but that value is populated during `POST /v3/templates`; it does not prove that the client has actually pushed the image.

If neither `fromImage` nor a supported `fromTemplate` source is provided, we should verify that the registry push has completed before dispatching CubeMaster. Otherwise the API accepts the build and moves the failure to a later image-pull stage, which makes the error much harder to diagnose.

CubeAPI/src/services/templates.rs
```diff
@@ -263,27 +413,662 @@ impl TemplateService {
...
+            .from_template
+            .as_ref()
+            .map(|s| s.trim().to_string())
+            .filter(|s| !s.is_empty())
+        {
+            // Re-use an already-built CubeSandbox template as the base. We
+            // synthesise a CubeMaster reference of the form `cube://<tid>`,
+            // letting downstream callers resolve it. Adjust to your local
+            // convention if needed.
+            format!("cube://{}", parent)
```
fslongjin: This introduces a `cube://<template>` source reference, but I do not see corresponding support in this PR for resolving that scheme downstream in CubeMaster/cubelet.

If the downstream stack does not already understand `cube://`, this path will look supported at the API layer but fail later during image resolution. Please either implement the resolution in this PR or reject `fromTemplate` with a clear `501 Not Implemented` until the path is actually wired end-to-end.

CubeAPI/src/services/templates.rs
```diff
@@ -263,27 +413,662 @@ impl TemplateService {
...
+        let logs: Vec<String> = internal
+            .logs
+            .iter()
+            .take(limit)
+            .cloned()
+            .collect();
+        let log_entries: Vec<V3BuildLogEntry> = logs
+            .iter()
+            .map(|line| V3BuildLogEntry {
+                timestamp: chrono::Utc::now(),
```
fslongjin: These timestamps are generated at status-poll time, not at log-write time. That means the same historical log line will receive a new timestamp every time the client polls.

`BuildContext` already stores `BuildLogLine { timestamp, line }`, so the V3 conversion should preserve that structured timestamp instead of assigning `Utc::now()` here.

CubeAPI/src/handlers/templates_v3.rs
```diff
@@ -0,0 +1,70 @@
...
+/// `from_image` references (no Dockerfile-from-context build yet).
+pub async fn v3_get_files_hash(
+    State(state): State<AppState>,
+    Path((template_id, hash)): Path<(String, String)>,
+) -> AppResult<impl IntoResponse> {
+    let resp = state
+        .services
+        .templates
+        .v3_get_file_upload(&template_id, &hash)?;
+    Ok((StatusCode::CREATED, Json(resp)))
```
fslongjin, line 61: This is a successful GET request, so `201 Created` is surprising. If this status code is required by a specific E2B SDK behavior, please document that compatibility requirement here; otherwise this should be `200 OK`.

CubeAPI/src/handlers/registry.rs
```diff
@@ -0,0 +1,240 @@
...
+        .registry_upstream
+        .as_deref()
+        .filter(|s| !s.is_empty())
+        .ok_or_else(registry_disabled)?
+        .to_string();
+
+    let method = request.method().clone();
+    let query = request.uri().query().unwrap_or("").to_string();
+    let headers = request.headers().clone();
+    let body = match axum::body::to_bytes(request.into_body(), 512 * 1024 * 1024).await {
```
fslongjin: Buffering the entire registry request body in CubeAPI is risky for Docker/OCI pushes. A single layer larger than 512 MiB will fail, and concurrent uploads can put substantial memory pressure on the API process.

For a registry reverse proxy, this should ideally stream the request body to the upstream registry instead of materializing the whole blob in memory.

CubeAPI/src/routes.rs
```diff
@@ -291,6 +313,14 @@ fn build_agenthub_routes(state: &AppState, auth_configured: bool) -> Router<AppS
     with_auth(routes, state, auth_configured)
 }
 
+fn build_registry_router(_state: &AppState) -> Router<AppState> {
+    use axum::routing::{any, get};
+    Router::new()
+        .route("/v2/", get(registry::ping))
+        .route("/v2", get(registry::ping))
+        .route("/v2/*path", any(registry::proxy))
```
fslongjin, line 369: The registry proxy is mounted without CubeAPI auth or rate limiting. I understand Docker registry auth is a different credential domain, but this implementation also does not validate the short-lived `RegistryCredential` issued by CubeAPI; it simply forwards requests to the upstream registry.

Given the one-click deployment points this at a default local registry, exposing CubeAPI could accidentally expose a public writable registry proxy. Please add credential validation/rate limiting here, or make the deployment contract require an authenticated upstream registry and document that clearly.

CubeAPI/src/routes.rs
```diff
@@ -170,6 +178,20 @@ fn build_template_routes(state: &AppState, auth_configured: bool) -> Router<AppS
     let routes = Router::new()
         .route("/templates", get(templates::list_templates))
         .route("/templates", post(templates::create_template))
+        // ── E2B V3 protocol (real SDK contract) ───────────────
+        // SDK calls these in order: POST /v3/templates → GET files/{hash}
+        // → POST /v2/.../builds/{bid} → GET .../status. Routes are mounted
+        // at the same level as /templates so both `/v3/...` and `/v2/...`
+        // segments are reachable.
+        .route("/v3/templates", post(templates_v3::v3_create_template))
```
fslongjin, line 203: Template build routes are currently protected by auth only, while sandbox routes use `with_auth_and_rate_limit`.

The new V3 endpoints can allocate in-memory build contexts and trigger expensive CubeMaster template-build pipelines, so they should also be rate-limited or governed by a dedicated build quota.

CubeAPI/src/services/builds.rs
```diff
@@ -0,0 +1,189 @@
...
+
+#[derive(Debug, Clone)]
+pub struct BuildLogLine {
+    pub timestamp: DateTime<Utc>,
+    pub line: String,
+}
+
+/// Thread-safe, in-process build registry.
+#[derive(Clone, Default)]
+pub struct BuildRegistry {
```
fslongjin, line 190: The in-memory build registry has no eviction policy. Completed, failed, and abandoned builds remain in the `DashMap` indefinitely, including the original create request, credentials, status fields, logs, and the duplicate pair index.

For a long-running CubeAPI process this can become an unbounded memory leak. Please add a TTL cleanup path for terminal builds and/or a maximum registry size.

CubeAPI/src/handlers/registry.rs
```diff
@@ -0,0 +1,240 @@
...
+        if let Some(parsed) = parse_manifest_path(&normalized) {
+            // tag carries either the buildID (preferred) or a digest. Pull the
+            // build context by tag first, then fall back to no-op.
+            if !parsed.tag.starts_with("sha256:") {
+                tracing::info!(
+                    repo = %parsed.repo,
+                    tag = %parsed.tag,
+                    "manifest pushed; marking build as image-pushed"
+                );
+                state.services.templates.mark_image_pushed(&parsed.tag);
```
fslongjin: This marks a build as image-pushed based only on the manifest tag. It does not verify that the repository path matches the expected `<repo_prefix>/<templateID>` for the build context.

The build IDs are UUID-like and hard to guess, but the proxy should still validate both repo and tag against the stored build context before advancing the build state.

CubeAPI/src/services/templates.rs
```diff
@@ -263,27 +413,662 @@ impl TemplateService {
...
+    ///
+    ///   1. `body.from_image`  — the standard E2B flow, e.g.
+    ///                            `python:3.11-slim`.
+    ///   2. The image already pushed to the bundled registry under
+    ///      `<repo>/<templateID>:<buildID>` (when the OCI Distribution path
+    ///      was used).
+    ///   3. `body.from_template` — copy from another known CubeSandbox
+    ///      template (resolved via CubeMaster `get_template`).
+    ///
+    /// `start_cmd` becomes container `args`; `ready_cmd` becomes a Probe.Exec.
```
fslongjin: Several comments still say `readyCmd` is translated into `Probe.Exec`, but the actual implementation parses a loopback URL and creates an HTTP GET probe because CubeMaster/cubelet do not support exec probes here.

Please update the comments/API model docs to say `Probe.HttpGet`; otherwise future maintainers may assume exec-style ready commands are supported end-to-end.

fslongjin (commented):
I reviewed the V3 template-build implementation with a focus on correctness, architecture, and workaround risk. The main blockers are the current test compilation failure, the fake file-cache hit for unsupported context builds, and the registry proxy/auth model.

CubeAPI/src/services/templates.rs
chengjoey: done

CubeAPI/src/routes.rs
chengjoey: done

CubeAPI/src/handlers/registry.rs
chengjoey: done, `mark_image_pushed` add repo arg, cross-checks the manifest repo

CubeAPI/src/handlers/templates_v3.rs
chengjoey: `201` is intentional., ref: https://github.com/e2b-dev/infra/blob/db88eee0fd5df4a5c90e544faa5c7b44c6719b51/packages/api/internal/handlers/template_layer_files_upload.go#L71

CubeAPI/src/handlers/registry.rs
chengjoey: done

CubeAPI/src/services/templates.rs
chengjoey: done

CubeAPI/src/services/builds.rs
chengjoey: `BuildRegistry` gains TTL + size-cap + background GC

CubeAPI/src/routes.rs
chengjoey: V3 build-pipeline routes moved to `with_auth_and_rate_limit`.

CubeAPI/src/services/templates.rs
chengjoey: done, Now the displayed return is `501`

CubeAPI/src/services/templates.rs
chengjoey: done， `fromImage` will succeed, Dockerfile/steps-based will point to 501

CubeAPI/src/services/templates.rs
chengjoey: done

fslongjin:
Hello, and thank you for your contribution. After careful review, we believe that the E2B template building functionality provided by this PR introduces additional complexity for a low-level system like Cube, given that it differs from E2B's commercial service in certain ways. The corresponding functionality can also be achieved by packaging OCI images through `docker build`, which is a more deterministic approach. Therefore, I prefer not to introduce the E2B template building protocol into Cube, and I will close this PR for now. Thank you again for your contribution.

PR #428, opened 2026-06-02
network-agent: disable GRO on bond slaves, not just the upper device

chenhengqi:
disableGRO() previously issued a single ETHTOOL_SGRO ioctl against cfg.EthName. For bond uplinks this only flips the feature flag on the bond netdev — GRO actually happens in the NAPI poll of the underlying slaves, so packets still get coalesced before reaching cubevs/tap and the original goal (preserve MTU-sized frames into the guest) is not met.

Resolve the configured interface, recursively walk into bond slaves and vlan parents, and call ethtool SGRO on every underlying netdev as well as the device itself (to keep the upper-layer feature flag consistent). For anything else (physical NIC, virtio-net, etc.) we fall through the default branch and just operate on the device itself, matching the previous behavior.

A per-slave failure is logged at Warn and the first error is returned, mirroring the existing Warn-only handling at the call site so a single bad NIC does not block service startup.

### DIFF
--- a/network-agent/internal/service/netdevice.go
+++ b/network-agent/internal/service/netdevice.go
@@ -68,6 +68,101 @@ type tapDevice struct {
 }
 
 func disableGRO(ifName string) error {
+	link, err := netlinkLinkByName(ifName)
+	if err != nil {
+		return fmt.Errorf("lookup link %s: %w", ifName, err)
+	}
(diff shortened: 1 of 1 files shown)


chenhengqi:
cc @jamesxia20181001 @pangsq

PR #452, opened 2026-06-03
fix(one-click): delay host DNS switchover until coredns is ready

fslongjin:
## Summary

- delay the NetworkManager DNS switchover until CoreDNS is actually listening, so first-install image pulls do not depend on an incomplete local DNS chain
- keep one non-reserved upstream resolver in `/etc/resolv.conf` as a fallback instead of making host DNS single-homed
- centralize reserved nameserver filtering and reuse it in CoreDNS startup and DNS route teardown to prevent proxy or stub addresses from being reused as upstream resolvers

### DIFF
--- a/deploy/one-click/scripts/systemd/common.sh
+++ b/deploy/one-click/scripts/systemd/common.sh
@@ -72,6 +72,22 @@ ensure_executable() {
   [[ -x "${path}" ]] || die "required executable not found: ${path}"
 }
 
+is_reserved_nameserver() {
+  local nameserver="${1:-}"
+  shift || true
+
+  [[ -n "${nameserver}" ]] || return 0
+  [[ "${nameserver}" == 127.* ]] && return 0
+  [[ "${nameserver}" == "::1" ]] && return 0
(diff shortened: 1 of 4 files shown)


fslongjin:
Thanks for the review. A few clarifications from our side:

1. The `rg` dependency is already enforced at runtime inside `wait_for_udp_port()`, and one-click install already ensures ripgrep is present via `install_ripgrep()`, so this is not a functional gap in the NetworkManager path.

2. The `systemd-resolved` path intentionally remains out of scope for this fix. The deadlock addressed in this PR is specific to the NetworkManager path rewriting `/etc/resolv.conf`, which Docker reads directly during the first image pull. The resolved path does not create the same host-level DNS cutover failure mode, so we kept this patch focused on the blocking first-install deadlock. We can revisit symmetry in a follow-up if needed.

3. Treating an empty nameserver as reserved/skip-worthy is intentional and preserves the previous `is_stub_nameserver()` semantics used by the callers.

4. The metadata sourcing/ordering observations are fair, but they match the previous behavior and are not regressions introduced by this PR.

Given that, we do not see any blocking issue for the current fix set.

PR #455, opened 2026-06-04
fix(CubeMaster): expire node health after stale nodemeta heartbeat

### LINKED ISSUE #454: [Bug Report] CubeMaster keeps nodemeta nodes healthy after heartbeats stop reporting
## Summary

On the nodemeta path, CubeMaster can continue reporting a node as healthy after `cubelet` stops sending heartbeats.

The last reported `Ready=True` fact is kept, but the control plane does not always expire the node's effective health based on heartbeat freshness. This can leave stale nodes visible as healthy and still eligible for scheduling decisions unless operators manually intervene.

## Environment

- CubeSandbox version / commit: `v0.3.0`
- Host OS and kernel version: `Rocky Linux 9.6 (Blue Onyx), kernel 5.14.0-570.17.1.el9_6.x86_64`
- KVM info (`modinfo kvm`): `kvm module from Rocky Linux 9.6 kernel package, signer "Rocky kernel signing key"`
- Deployment mode: `cluster`
- Relevant component: `CubeMaster / Cubelet / nodemeta`

## Steps to Reproduce

1. Deploy one control node and at least one compute node with the nodemeta path enabled.
2. Confirm the node is healthy:

   ```bash
   cubemastercli node list
   curl -s "http://127.0.0.1:8089/internal/meta/nodes/<node-id>" | jq
   ```

3. Stop `cubelet` on the target node so heartbeat reporting stops.
4. Wait longer than the effective heartbeat expiry window.
5. Query the node again:

   ```bash
   cubemastercli node list
   curl -s "http://127.0.0.1:8089/internal/meta/nodes/<node-id>" | jq
   curl -s "http://127.0.0.1:8089/internal/node?requestID=test&host_id=<node-id>" | jq
   ```

## Expected Behavior

Node health should be derived from current heartbeat freshness, not only from the last reported `Ready` value.

After heartbeat reporting stops and the heartbeat becomes stale:

- the node should automatically become `healthy=false`
- the API should expose a clear unhealthy reason
- scheduler prefilter should exclude the stale node

## Actual Behavior

The node can remain healthy after heartbeat reporting stops because the control plane still reflects the last reported `Ready=True` fact instead of expiring the node's effective health.

In practice, operators can work around this by manually posting `Ready=False`, but heartbeat expiry should not require manual intervention.

Example manual workaround:

```bash
curl -s -X POST "http://127.0.0.1:8089/internal/meta/nodes/<node-id>/status" \
  -H "Content-Type: application/json" \
  -d '{
    "conditions": [
      {
        "type": "Ready",
        "status": "False",
        "reason": "NodeShutdown",
        "message": "manual test"
      }
    ]
  }'
```

## Additional Context

This issue is scoped only to the nodemeta-based heartbeat health model.

The intended fix is:

- preserve the last reported node fact internally
- derive current `healthy` from `Ready` + heartbeat freshness
- make `healthy=false` explicit in API responses
- expose an unhealthy reason for stale heartbeat vs reported not ready
- make scheduler prefilter drop stale nodes instead of only logging timeout

zyl1121:
## Motivation

On the nodemeta path, CubeMaster can keep a node healthy after `cubelet` stops reporting heartbeats because the last reported `Ready=True` fact is reused as the current health result.

This makes stale nodes look healthy in control-plane reads and can leave them eligible for scheduling decisions longer than expected.

## What this PR changes

- derive current node health from the last reported readiness fact plus heartbeat freshness
- make nodemeta reads return the current effective `healthy` result instead of a stale cached value
- make `healthy=false` explicit in API output
- expose `unhealthy_reason` for `HeartbeatExpired` and `ReportedNotReady`
- make `localcache` read paths refresh health from cached facts without mutating shared cached nodes
- apply the same health refresh logic to `GetNodesByIp`
- make prefilter skip stale metadata / metric nodes instead of only logging timeout

## Testing

```bash
go test -vet=off ./pkg/localcache ./pkg/nodemeta ./pkg/selector/prefilter
```

Manual validation:

- stale heartbeat is returned as `healthy=false`
- `unhealthy_reason` distinguishes `HeartbeatExpired` and `ReportedNotReady`
- `/internal/meta/nodes/{id}` and `/internal/node` reflect the same health semantics

## Documentation

No standalone documentation update is included in this PR. However, this change does affect the JSON shape of node responses: `healthy=false` is now emitted explicitly instead of being omitted.

## Related issue

Fixes #454

### DIFF
--- a/CubeMaster/pkg/base/node/node.go
+++ b/CubeMaster/pkg/base/node/node.go
@@ -92,7 +92,12 @@ func (n *Node) Clone() *Node {
 	if n == nil {
 		return nil
 	}
+	// Clone provides a best-effort read-side snapshot. Mutable counters such
+	// as LocalCreateNum are refreshed via atomic loads after the structural
+	// copy so cloned read models stay aligned with the write path.
+	localCreateNum := atomic.LoadInt64(&n.LocalCreateNum)
 	cloned := *n
+	cloned.LocalCreateNum = localCreateNum
 	if n.VirtualNodeQuotaArray != nil {
 		cloned.VirtualNodeQuotaArray = append([]int64(nil), n.VirtualNodeQuotaArray...)
 	}
--- a/CubeMaster/pkg/base/node/node.go
+++ b/CubeMaster/pkg/base/node/node.go
@@ -54,7 +54,11 @@ type Node struct {
 
 	MetaDataUpdateAt time.Time `json:"MetaDataUpdateAt,omitempty"`
 
-	Healthy bool `json:"Healthy,omitempty"`
+	ReportedReady bool `json:"-"`
+
+	Healthy bool `json:"Healthy"`
+
+	UnhealthyReason string `json:"UnhealthyReason,omitempty"`
 
 	Score float64 `json:"Score,omitempty"`
 
@@ -82,6 +86,17 @@ type Node struct {
 
 	LocalCreateNum int64 `json:"LocalCreateNum,omitempty"`
 	NicQueues      int64 `json:"nic_queues,omitempty"`
+}
+
+func (n *Node) Clone() *Node {
+	if n == nil {
+		return nil
+	}
+	cloned := *n
+	if n.VirtualNodeQuotaArray != nil {
+		cloned.VirtualNodeQuotaArray = append([]int64(nil), n.VirtualNodeQuotaArray...)
+	}
+	return &cloned
 }
 
 func (n *Node) ID() string {
--- a/CubeMaster/pkg/base/nodehealth/health.go
+++ b/CubeMaster/pkg/base/nodehealth/health.go
@@ -20,6 +20,10 @@ type Status struct {
 	UnhealthyReason string
 }
 
+func MetadataTimeout(syncMetaDataInterval time.Duration) time.Duration {
+	return syncMetaDataInterval + 10*time.Second
+}
+
 func ReadyConditionTrue(conditions []corev1.NodeCondition) bool {
 	for _, cond := range conditions {
 		if cond.Type == corev1.NodeReady {
--- a/CubeMaster/pkg/base/nodehealth/health.go
+++ b/CubeMaster/pkg/base/nodehealth/health.go
@@ -1,0 +1,44 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package nodehealth
+
+import (
+	"time"
+
+	corev1 "k8s.io/api/core/v1"
+)
+
+const (
+	ReasonReportedNotReady = "ReportedNotReady"
+	ReasonHeartbeatExpired = "HeartbeatExpired"
+)
+
+type Status struct {
+	Healthy         bool
+	UnhealthyReason string
+}
+
+func ReadyConditionTrue(conditions []corev1.NodeCondition) bool {
+	for _, cond := range conditions {
+		if cond.Type == corev1.NodeReady {
+			return cond.Status == corev1.ConditionTrue
+		}
+	}
+	return false
+}
+
+func EvaluateFromFacts(reportedReady bool, heartbeatTime, now time.Time, timeout time.Duration) Status {
+	if heartbeatTime.IsZero() || now.Sub(heartbeatTime) > timeout {
+		return Status{Healthy: false, UnhealthyReason: ReasonHeartbeatExpired}
+	}
+	if !reportedReady {
+		return Status{Healthy: false, UnhealthyReason: ReasonReportedNotReady}
+	}
+	return Status{Healthy: true}
+}
+
+func Evaluate(conditions []corev1.NodeCondition, heartbeatTime, now time.Time, timeout time.Duration) Status {
+	return EvaluateFromFacts(ReadyConditionTrue(conditions), heartbeatTime, now, timeout)
+}
--- a/CubeMaster/pkg/base/nodehealth/health.go
+++ b/CubeMaster/pkg/base/nodehealth/health.go
@@ -1,7 +1,3 @@
-// Copyright (c) 2026 Tencent Inc.
-// SPDX-License-Identifier: Apache-2.0
-//
-
 package nodehealth
 
 import (
--- a/CubeMaster/pkg/nodemeta/service.go
+++ b/CubeMaster/pkg/nodemeta/service.go
@@ -194,6 +194,7 @@ func RegisterNode(ctx context.Context, req *RegisterNodeRequest) (*NodeSnapshot,
 	snap.QuotaMemMB = req.QuotaMemMB
 	snap.CreateConcurrentNum = req.CreateConcurrentNum
 	snap.MaxMvmNum = req.MaxMvmNum
+	applyCurrentHealth(snap, time.Now())
 	global.mu.Unlock()
 	syncLocalcache(snap)
 	return cloneSnapshot(snap), nil
@@ -409,7 +410,6 @@ func toSchedulerNode(snap *NodeSnapshot) *node.Node {
 	if snap == nil {
 		return nil
 	}
-	status := currentHealthStatus(snap, time.Now())
 	quotaCPU := snap.QuotaCPU
 	if quotaCPU == 0 {
 		quotaCPU = snap.Allocatable.MilliCPU
@@ -439,8 +439,8 @@ func toSchedulerNode(snap *NodeSnapshot) *node.Node {
 		InstanceType:        instanceType,
 		HostStatus:          constants.HostStatusRunning,
 		ReportedReady:       snap.ReportedReady,
-		Healthy:             status.Healthy,
-		UnhealthyReason:     status.UnhealthyReason,
+		Healthy:             snap.Healthy,
+		UnhealthyReason:     snap.UnhealthyReason,
 		CreateConcurrentNum: snap.CreateConcurrentNum,
 		MaxMvmLimit:         snap.MaxMvmNum,
 		MetaDataUpdateAt:    snap.HeartbeatTime,
--- a/CubeMaster/pkg/nodemeta/service.go
+++ b/CubeMaster/pkg/nodemeta/service.go
@@ -386,7 +386,7 @@ func (s *service) reload() error {
 }
 
 func healthTimeout() time.Duration {
-	return config.GetConfig().Common.SyncMetaDataInterval + 10*time.Second
+	return nodehealth.MetadataTimeout(config.GetConfig().Common.SyncMetaDataInterval)
 }
 
 func currentHealthStatus(snap *NodeSnapshot, now time.Time) nodehealth.Status {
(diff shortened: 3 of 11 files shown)


CubeMaster/pkg/base/nodehealth/health.go
```diff
@@ -0,0 +1,48 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
```
fslongjin: should not add this line

CubeMaster/pkg/selector/prefilter/prefilter_test.go
```diff
@@ -0,0 +1,113 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
```
fslongjin: should not add this line

zyl1121: > should not add this line

@fslongjin Removed the extra header comments in a follow-up commit. Thanks for catching that.

CubeMaster/pkg/selector/prefilter/prefilter.go
```diff
@@ -77,17 +78,20 @@ func (l *prefilter) Select(selCtx *selctx.SelectorCtx) (node.NodeList, error) {
...
-			}).Fatalf("%s MetricLocalUpdate timeout,lastupdate:%v", n.IP, n.MetricLocalUpdateAt)
+			}).Warnf("%s MetricLocalUpdate timeout,lastupdate:%v", n.IP, n.MetricLocalUpdateAt)
+			continue
 		}
 		if time.Since(n.MetaDataUpdateAt) > metaDataUpdateAtTimeout {
 			log.G(selCtx.Ctx).WithFields(map[string]any{
 				"CalleeCluster": n.ClusterLabel,
-			}).Fatalf("%s MetaDataUpdate timeout,lastupdate:%v", n.IP, n.MetaDataUpdateAt)
+			}).Warnf("%s MetaDataUpdate timeout,lastupdate:%v", n.IP, n.MetaDataUpdateAt)
+			continue
```
fslongjin: 🔁 **Redundant heartbeat-timeout check**

`GetHealthyNodesByInstanceType` (line 40) already clones every node with `cloneNodeWithCurrentHealth`, which calls `nodehealth.EvaluateFromFacts` using the same `MetadataTimeout`. A node whose heartbeat has expired will already have `Healthy=false` and will have been excluded from the candidate set before this loop runs.

This second `MetaDataUpdateAt` guard therefore duplicates what the health evaluation already does. Over time, keeping two separate timeout calculations risks the two checks drifting (e.g. someone changes how `cloneNodeWithCurrentHealth` computes the timeout but forgets to update this prefilter copy, or vice versa).

**Suggestion:** Rely solely on the `n.Healthy` check above (line 49). That gives a single source of truth for heartbeat-liveness decisions.

CubeMaster/pkg/base/node/node.go
```diff
@@ -84,6 +88,17 @@ type Node struct {
...
+func (n *Node) Clone() *Node {
+	if n == nil {
+		return nil
+	}
+	cloned := *n
+	if n.VirtualNodeQuotaArray != nil {
+		cloned.VirtualNodeQuotaArray = append([]int64(nil), n.VirtualNodeQuotaArray...)
+	}
+	return &cloned
+}
```
fslongjin, line 105: ⚠️ **Non-atomic read of `LocalCreateNum`**

The struct copy `cloned := *n` reads every field including `LocalCreateNum` (an `int64`), which is accessed concurrently via `atomic.AddInt64` (in `LocalCreateNumIncrBy`) and `atomic.LoadInt64` (in `LocalCreateConcurrentLimit`).

Per the Go memory model this is a data race and would be flagged by `go test -race`. In practice on amd64 it is safe (aligned 8-byte reads are atomic at the hardware level), but on 32-bit platforms (ARM) this could tear.

If this project only targets amd64, a brief comment acknowledging the deliberate non-atomic copy would help future readers (and suppress race-detector noise). Otherwise, consider using `atomic.LoadInt64(&n.LocalCreateNum)` for the copy.

CubeMaster/pkg/nodemeta/service.go
```diff
@@ -419,7 +438,9 @@ func toSchedulerNode(snap *NodeSnapshot) *node.Node {
 		OssClusterLabel:     snap.ClusterLabel,
```
fslongjin, line 438: 🔄 **Duplicate health evaluation with `UpdateNodeStatus` line 239**

`UpdateNodeStatus` already calls `applyCurrentHealth(snap, time.Now())` (line 239, position 46 in this diff), which runs `currentHealthStatus` internally. `syncLocalcache` on line 241 then calls `toSchedulerNode`, which calls `currentHealthStatus(snap, time.Now())` again here.

The two `time.Now()` calls are microseconds apart, so the result is effectively identical — but it is redundant work and creates a maintenance risk: if someone later extends the health-decision logic in one call site but not the other, the two representations of the same snapshot's health will diverge.

**Suggestion:** Have `applyCurrentHealth` return the `nodehealth.Status` it computed (e.g. `func applyCurrentHealth(...) nodehealth.Status`), and let `toSchedulerNode` consume it instead of re-deriving from scratch. This keeps a single evaluation per heartbeat update.

fslongjin (commented):
Thanks for this PR — the facts-vs-evaluation separation (`ReportedReady` vs `Healthy`) and the centralized `nodehealth` package are clean architectural improvements. I left a few inline comments on areas that could be tightened up.

---

### Summary of inline comments

1. **`prefilter.go`** — Redundant `MetaDataUpdateAt` timeout check; `GetHealthyNodesByInstanceType` already expires stale heartbeats via `cloneNodeWithCurrentHealth`.

2. **`node.go:Clone()`** — Struct copy reads `LocalCreateNum` non-atomically (data race per Go memory model).

3. **`service.go`** — Duplicate health evaluation in `UpdateNodeStatus` (line 239) and `toSchedulerNode` (line 412); two `time.Now()` calls on the same snapshot.

4. **`node.go:Healthy` JSON tag** — `omitempty` removed; API contract change worth calling out explicitly.

5. **`service.go:RegisterNode`** — Returns `cloneSnapshot` without applying current health, inconsistent with `GetNode`/`ListNodes`/`UpdateNodeStatus`.

CubeMaster/pkg/base/node/node.go
```diff
@@ -54,7 +54,11 @@ type Node struct {
...
 	MetaDataUpdateAt time.Time `json:"MetaDataUpdateAt,omitempty"`
 
-	Healthy bool `json:"Healthy,omitempty"`
+	ReportedReady bool `json:"-"`
+
+	Healthy bool `json:"Healthy"`
+
+	UnhealthyReason string `json:"UnhealthyReason,omitempty"`
 
 	Score float64 `json:"Score,omitempty"`
```
fslongjin, line 64: 📢 **API contract change: `omitempty` removed**

The `Healthy` field changed from `json:"Healthy,omitempty"` to `json:"Healthy"`, meaning `"healthy":false` is now always present in JSON output instead of being omitted. This is deliberate (the nodemeta test confirms it), but it is a wire-format change that existing API consumers may depend on.

Consider calling this out in the PR description and the next changelog entry so integrators know to expect the new field shape.

CubeMaster/pkg/nodemeta/service.go
```diff
@@ -464,6 +485,12 @@ func cloneSnapshot(in *NodeSnapshot) *NodeSnapshot {
 	return &out
 }
 
+func cloneSnapshotWithCurrentHealth(in *NodeSnapshot, now time.Time) *NodeSnapshot {
+	out := cloneSnapshot(in)
+	applyCurrentHealth(out, now)
+	return out
+}
+
```
fslongjin, line 493: ❓ **Inconsistency: `RegisterNode` doesn't use this helper**

`cloneSnapshotWithCurrentHealth` is now used by `GetNode` (line 300) and `ListNodes` (line 310), and `UpdateNodeStatus` calls `applyCurrentHealth` inline (line 239). But `RegisterNode` (line 199, unchanged) still returns a plain `cloneSnapshot(snap)` without calling `applyCurrentHealth`.

The caller gets `Healthy=false` (zero value) with an empty `UnhealthyReason` — rather than `UnhealthyReason="HeartbeatExpired"` which would be the consistent answer for a node that has never sent a heartbeat yet.

**Should `RegisterNode` also use `cloneSnapshotWithCurrentHealth` for consistency?**

zyl1121:
> Thanks for this PR — the facts-vs-evaluation separation (`ReportedReady` vs `Healthy`) and the centralized `nodehealth` package are clean architectural improvements. I left a few inline comments on areas that could be tightened up.
> 
> ### Summary of inline comments
> 1. **`prefilter.go`** — Redundant `MetaDataUpdateAt` timeout check; `GetHealthyNodesByInstanceType` already expires stale heartbeats via `cloneNodeWithCurrentHealth`.
> 2. **`node.go:Clone()`** — Struct copy reads `LocalCreateNum` non-atomically (data race per Go memory model).
> 3. **`service.go`** — Duplicate health evaluation in `UpdateNodeStatus` (line 239) and `toSchedulerNode` (line 412); two `time.Now()` calls on the same snapshot.
> 4. **`node.go:Healthy` JSON tag** — `omitempty` removed; API contract change worth calling out explicitly.
> 5. **`service.go:RegisterNode`** — Returns `cloneSnapshot` without applying current health, inconsistent with `GetNode`/`ListNodes`/`UpdateNodeStatus`.

Thanks for the detailed review. I addressed the issues you found in follow-up commit, and I also updated the PR description to explicitly mention that `healthy=false` is now emitted in JSON output.

fslongjin (approved):

fslongjin:
Hi @zyl1121, thanks for this PR!

The DCO check is currently failing because all 4 commits are missing the `Signed-off-by` trailer. The CubeSandbox project requires every commit to carry a DCO sign-off.

Could you please:

1. **Amend each commit** to add your `Signed-off-by` line (e.g., `Signed-off-by: zhengyilei <zhengyilei@myhexin.com>`)
2. **Force push** the updated branch

You can do this with an interactive rebase:

```bash
git rebase -i HEAD~4
# mark each commit as 'reword', and append:
# Signed-off-by: zhengyilei <zhengyilei@myhexin.com>
```

Or more conveniently use `git rebase --signoff`:

```bash
git rebase --signoff HEAD~4
git push --force-with-lease
```

Once all commits have the sign-off, the DCO check should pass. Thanks!

zyl1121:
> Hi @zyl1121, thanks for this PR!
> 
> The DCO check is currently failing because all 4 commits are missing the `Signed-off-by` trailer. The CubeSandbox project requires every commit to carry a DCO sign-off.
> 
> Could you please:
> 
> 1. **Amend each commit** to add your `Signed-off-by` line (e.g., `Signed-off-by: zhengyilei <zhengyilei@myhexin.com>`)
> 2. **Force push** the updated branch
> 
> You can do this with an interactive rebase:
> 
> ```shell
> git rebase -i HEAD~4
> # mark each commit as 'reword', and append:
> # Signed-off-by: zhengyilei <zhengyilei@myhexin.com>
> ```
> 
> Or more conveniently use `git rebase --signoff`:
> 
> ```shell
> git rebase --signoff HEAD~4
> git push --force-with-lease
> ```
> 
> Once all commits have the sign-off, the DCO check should pass. Thanks!

Fixed and force-pushed the branch with signed-off commits. Thanks!

PR #460, opened 2026-06-04
workaround(virtiofs): set rw virtiofs cache mode to always

lisongqian:
Change the cache mode of rwVirtioCfg in generateEmptyVirtiofsDevices
from VirtiofsCacheNone to VirtiofsCacheAlways. The ro channel keeps
VirtiofsCacheNone unchanged.

Signed-off-by: Songqian Li <sionli@tencent.com>


kinwin-ustc:
It may cause inconsistent file views inside and outside the sandbox, and should not be merged

Issue #233, opened 2026-05-13
[Bug Report] POST /sandboxes/<id>/refreshes returns 500 instead of 501 Not Implemented


tinklone:
## Description

`POST /sandboxes/<id>/refreshes` returns HTTP 500 Internal Server Error for any valid running sandbox. This endpoint is not yet implemented but the route exists, so it should return **501 Not Implemented** instead of a misleading 500.

SDK calls to `sandbox.keep_alive()` get an opaque 500 error with no indication that the feature is simply not available yet.

## Reproduction

```bash
# Create sandbox
sid=$(curl -s -X POST http://localhost:3000/sandboxes \
  -H 'content-type: application/json' \
  -d '{"templateID":"e2b-code-interpreter"}' | jq -r '.sandboxID')

# Refresh / keep-alive
curl -sv -X POST "http://localhost:3000/sandboxes/$sid/refreshes" \
  -H 'content-type: application/json' \
  -d '{"duration": 60}'
# Actual: HTTP 500
# Expected: HTTP 501 (until implemented), then HTTP 200/204

curl -X DELETE "http://localhost:3000/sandboxes/$sid"
```

## Expected Behavior

- **Short term**: Return HTTP 501 Not Implemented with a clear message (e.g., `{"code":501,"message":"refresh/keep-alive is not yet implemented"}`)
- **Long term**: Implement the feature and return HTTP 200/204

## Component

CubeAPI `handlers/sandboxes.rs` refresh handler

## Severity

P2 — Unimplemented feature returning misleading 500

PR #483, opened 2026-06-06
feat(sandbox): add absolute TTL reaper with SetTimeout/Refresh endpoints and dashboard EndAt

chengjoey:
add `EndAt` for sandbox

image

image

feat: #233 #232

### DIFF
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -77,13 +77,14 @@ pub struct NewSandbox {
     #[serde(rename = "templateID")]
     pub template_id: String,
 
-    /// Sandbox lifetime in seconds. `None` (field omitted) means "no
-    /// automatic destroy" — the sandbox lives until killed explicitly
-    /// via DELETE /sandboxes/{id} or until SetTimeout/Refresh installs
-    /// a deadline. `Some(0)` is treated the same as `None`.
+    /// Sandbox lifetime in seconds. Defaults to 15 if the field is
+    /// omitted, matching the historical behavior every official SDK
+    /// relies on. `0` means "no automatic destroy" — the sandbox lives
+    /// until killed explicitly via DELETE /sandboxes/{id} or until
+    /// SetTimeout / Refresh installs a new deadline.
     #[validate(range(min = 0))]
-    #[serde(default)]
-    pub timeout: Option<i32>,
+    #[serde(default = "default_timeout")]
+    pub timeout: i32,
 
     #[serde(rename = "autoPause", default)]
     pub auto_pause: bool,
@@ -377,11 +378,17 @@ fn default_log_limit() -> i32 {
 
 // ─── Sandbox — timeout / refresh ──────────────────────────────────────────
 
-/// Request body for POST /sandboxes/{id}/timeout
+/// Request body for POST /sandboxes/{id}/timeout.
+///
+/// Field name is `timeout` (seconds, absolute). Earlier drafts also
+/// accepted a `duration` alias for cross-SDK compatibility, but this
+/// invited callers to confuse SetTimeout (absolute reset) with the
+/// /refreshes endpoint (which still uses `duration`). The two
+/// endpoints now keep their natural field names so the on-wire shape
+/// communicates the semantics.
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct SetTimeoutRequest {
     #[validate(range(min = 0))]
-    #[serde(alias = "duration")]
     pub timeout: i32,
 }
 
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -77,9 +77,13 @@ pub struct NewSandbox {
     #[serde(rename = "templateID")]
     pub template_id: String,
 
+    /// Sandbox lifetime in seconds. `None` (field omitted) means "no
+    /// automatic destroy" — the sandbox lives until killed explicitly
+    /// via DELETE /sandboxes/{id} or until SetTimeout/Refresh installs
+    /// a deadline. `Some(0)` is treated the same as `None`.
     #[validate(range(min = 0))]
-    #[serde(default = "default_timeout")]
-    pub timeout: i32,
+    #[serde(default)]
+    pub timeout: Option<i32>,
 
     #[serde(rename = "autoPause", default)]
     pub auto_pause: bool,
@@ -160,8 +164,10 @@ pub struct ListedSandbox {
     pub client_id: String,
     #[serde(rename = "startedAt")]
     pub started_at: DateTime<Utc>,
-    #[serde(rename = "endAt")]
-    pub end_at: DateTime<Utc>,
+    /// Absolute deadline when the TTL reaper will destroy this sandbox.
+    /// `None` means the sandbox has no automatic-destroy schedule.
+    #[serde(rename = "endAt", skip_serializing_if = "Option::is_none")]
+    pub end_at: Option<DateTime<Utc>>,
     #[serde(rename = "cpuCount")]
     pub cpu_count: i32,
     #[serde(rename = "memoryMB")]
@@ -190,8 +196,10 @@ pub struct SandboxDetail {
     pub client_id: String,
     #[serde(rename = "startedAt")]
     pub started_at: DateTime<Utc>,
-    #[serde(rename = "endAt")]
-    pub end_at: DateTime<Utc>,
+    /// Absolute deadline when the TTL reaper will destroy this sandbox.
+    /// `None` means the sandbox has no automatic-destroy schedule.
+    #[serde(rename = "endAt", skip_serializing_if = "Option::is_none")]
+    pub end_at: Option<DateTime<Utc>>,
     #[serde(rename = "envdVersion")]
     pub envd_version: String,
     #[serde(rename = "envdAccessToken", skip_serializing_if = "Option::is_none")]
@@ -373,13 +381,15 @@ fn default_log_limit() -> i32 {
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct SetTimeoutRequest {
     #[validate(range(min = 0))]
+    #[serde(alias = "duration")]
     pub timeout: i32,
 }
 
 /// Request body for POST /sandboxes/{id}/refreshes
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct RefreshRequest {
-    #[validate(range(min = 0, max = 3600))]
+    #[validate(range(min = 0, max = 86400))]
+    #[serde(default, alias = "timeout")]
     pub duration: Option<i32>,
 }
 
--- a/CubeMaster/pkg/service/httpservice/cube/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/httpservice/cube/sandbox_timeout.go
@@ -1,0 +1,69 @@
+// Copyright (c) 2024 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package cube
+
+import (
+	"context"
+	"net/http"
+
+	"github.com/google/uuid"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/api/services/cubebox/v1"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/log"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/utils"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+	"github.com/tencentcloud/CubeSandbox/cubelog"
+)
+
+func handleSandboxTimeoutAction(_ http.ResponseWriter, r *http.Request, rt *CubeLog.RequestTrace) interface{} {
+	req := &types.SandboxTimeoutRequest{}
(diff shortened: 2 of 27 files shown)


kinwin-ustc:
some comments from AI：

This PR turns "absolute sandbox TTL reaping" from concept into reality, touching the full stack from CubeAPI (Rust) → CubeMaster (Go) → Dashboard. The overall direction is correct, but there are a few issues worth attention.

---

## 🔴 High-Priority Issues

### 1. traceRedis nil-pointer panic — Incomplete Fix

The PR fixes the `traceRedis` panic that occurs in background goroutines (no trace info), changing it to:

```go
var baseRt *CubeLog.RequestTrace
if rt := CubeLog.GetTraceInfo(ctx); rt != nil {
    baseRt = rt.DeepCopy()
} else {
    baseRt = &CubeLog.RequestTrace{}
}
```

But looking at the current mainline code:

`redis_cache.go` L434-L447

```go
func traceRedis(ctx context.Context, action, redisOp, key string, start time.Time, err error) {
	cost := time.Since(start)
	baseRt := CubeLog.GetTraceInfo(ctx).DeepCopy()
	baseRt.Callee = constants.Redis
	baseRt.Action = action
	baseRt.CalleeAction = redisOp
	baseRt.InstanceID = key
	baseRt.Cost = cost
	baseRt.RetCode = int64(errorcode.ErrorCode_Success)
	if err != nil {
		baseRt.RetCode = int64(errorcode.ErrorCode_DBError)
	}
	CubeLog.Trace(baseRt)
}
```

**Problem**: `traceRedis` is called by multiple functions such as `setByPassProsyToRedis` and `deleteKeyFromRedis`. After the reaper goroutine calls `ListExpiredSandboxIDs` to perform a SCAN, it inevitably calls `DeleteSandboxProxyMap`→`deleteKeyFromRedis`→`traceRedis`. This means the reaper's context must always carry this fix. It is better to put the nil check at the entry of `traceRedis` rather than doing a workaround only at certain call sites; otherwise, future background tasks will still trigger the same panic.

### 2. ListExpiredSandboxIDs — Performance Risk of Full SCAN

```go
func ListExpiredSandboxIDs(ctx context.Context, now time.Time) ([]string, error) {
    keys, err := l.scanByPassProxyKeys(ctx) // SCAN bypass_host_proxy:*
    // iterate each key → HGETALL → parse EndAt
}
```

**Problems**:

- Redis SCAN + per-key HGETALL is very slow at large scale (tens of thousands of sandboxes) and consumes a lot of Redis read connections.
- Every reaper tick does a full scan, with O(N) complexity.

**Suggestions**:

- Use a Redis Sorted Set (e.g. `zset:sandbox_ttl`, score = EndAt nanoseconds). The reaper only needs `ZRANGEBYSCORE ... -inf {now}` to fetch expired items in O(logN + K).
- Or `ZADD` to the TTL zset when calling `SetSandboxProxyMap`, and `ZREM` when calling `DeleteSandboxProxyMap`.
- If the architecture won't change in the short term, at least add a per-tick cap (e.g. process at most 100 items each time) to avoid the reaper blocking too long in a single tick.

### 3. Duplicate Destruction with Concurrent Reapers Across Multiple Masters

The PR starts `sandbox.StartTimeoutReaper(ctx)` in `main.go`. If multiple CubeMaster instances are deployed (a common HA deployment), all instances run the reaper in parallel, which may cause:

- Multiple `DestroySandbox` calls against the same sandbox
- A flood of "sandbox not found" errors in the logs
- Extra useless load on Cubelet

**Suggestions**:

- Use a Redis distributed lock (e.g. `SET reaper_lock NX EX 30`) to ensure only one master runs the reaper at a time
- Or use a leader election mechanism so a single master runs the reaper
- At minimum: perform an idempotency check before delete (only delete when the sandbox is found)

### 4. EndAt Write Logic in setByPassProsyToRedis

The PR changes it to always write the EndAt field:

```go
fieldValues = append(fieldValues, "EndAt", byPassProsy.EndAt)
```

**Problems**:

- For sandboxes without a TTL (EndAt = ""), it writes an empty string that overwrites the existing value. If some other code path previously set EndAt (e.g. the Refresh endpoint), and then an update path calls `SetSandboxProxyMap` again without EndAt, the existing TTL gets cleared.
- `getByPassProsyFromRedis` does not read EndAt! The current code only reads HostIP/CreatedAt/SandboxIP, and dumps the remaining fields into the `ContainerToHostPorts` map. This means EndAt is wrongly treated as a port mapping and stored as `ContainerToHostPorts["EndAt"] = "1717699200000000000"`.

**Fix suggestion**: Add explicit extraction of EndAt in `getByPassProsyFromRedis` (the same way CreatedAt/SandboxIP are read):

```go
if endAt, ok := mapvalues["EndAt"]; ok {
    nodeIdIp.EndAt = endAt
    delete(mapvalues, "EndAt")
}
```

---

## 🟡 Medium-Priority Issues

### 5. RefreshRequest.duration Upper Bound Relaxed from 3600 to 86400

Current code:

`mod.rs` L382-L384

```rust
pub struct RefreshRequest {
    #[validate(range(min = 0, max = 3600))]
    pub duration: Option<i32>,
}
```

The PR changes it to `max = 86400`. Needs confirmation:

- Does the business really need a single 24-hour refresh?
- Should CubeMaster also do a secondary validation, to prevent someone who bypasses CubeAPI and calls CubeMaster directly from setting an unbounded TTL?

### 6. SetTimeoutRequest Adds #[serde(alias = "duration")]

```rust
#[serde(alias = "duration")]
pub timeout: i32,
```

**Risk**: SetTimeout (set an absolute TTL) and Refresh (relative extension) are two semantically distinct APIs. If SetTimeout also accepts the `duration` field name, callers may confuse the parameter meanings of the two endpoints. Recommend emphasizing the difference in the docs, or removing this alias.

### 7. NewSandbox.timeout Changed from i32 to Option<i32> — Backward Compatibility

```rust
// before
#[serde(default = "default_timeout")]
pub timeout: i32,   // default 15

// after
#[serde(default)]
pub timeout: Option<i32>,   // None = never expires
```

**Problem**: `default_timeout` (15 seconds) is removed. When an old client does not send the timeout field, it previously got a default of 15 seconds; now it becomes None (never expires). This is a semantic breaking change.

**Suggestions**:

- Either keep `default = "default_timeout_option"` so the default is `Some(15)`
- Or confirm this is the intended behavior (never expires) and clearly note it in the changelog

### 8. end_at Changed from DateTime<Utc> to Option<DateTime<Utc>> — SDK Compatibility

```rust
// ListedSandbox, SandboxDetail
pub end_at: Option<DateTime<Utc>>,
```

Adding `#[serde(skip_serializing_if = "Option::is_none")]` means sandboxes without a TTL no longer have the endAt field in the JSON response. Existing SDKs that assume endAt always exists will fail to parse.

**Suggestions**:

- Check the Go SDK and Python SDK model definitions to ensure `end_at` is optional
- Note this is an incompatible change in the API docs

---

## 🟢 Positive Highlights

### 9. deserialize_end_at_nanos — Robust Deserialization

Supports multiple input formats (nanosecond numbers, numeric strings, RFC3339, null, 0, empty string) with full test coverage. Very practical for cases where CubeMaster returns inconsistent formats.

### 10. traceRedis nil Fix Is in the Right Direction

Although it is recommended to move it to the function entry, discovering and fixing the background-goroutine nil panic reflects good test-coverage awareness.

### 11. timeout: body.timeout.filter(|t| *t > 0) — Elegant Zero-Value Filtering

Avoids passing 0 to CubeMaster being misinterpreted as "set a 0-second TTL".

---

## 📋 Review Summary

| Severity | Issue | Suggestion |
| --- | --- | --- |
| 🔴 Critical | traceRedis nil check is only local, not at the function entry | Move the nil guard to the traceRedis entry |
| 🔴 Critical | Full SCAN does not scale | Switch to a Sorted Set or add a batch cap |
| 🔴 Critical | Concurrent reapers across multiple masters | Add a distributed lock or leader election |
| 🔴 Critical | getByPassProsyFromRedis does not parse EndAt | Explicitly read EndAt and delete it from the map |
| 🟡 Medium | timeout default changed from 15 to None (breaking) | Confirm intent or keep the default |
| 🟡 Medium | SetTimeoutRequest alias "duration" is confusing | Remove the alias or emphasize it in docs |
| 🟡 Medium | end_at optionality impacts SDK compatibility | Check and update SDK models |
| 🟡 Medium | refresh duration max=86400 lacks business rationale | Add the reason in the PR description |
| 🟢 Good | deserialize_end_at_nanos multi-format support + tests | ✅ |
| 🟢 Good | `filter(\|t\| *t > 0)` elegant zero-value filtering | ✅ |

Overall, the feature design direction is correct, but the **missing EndAt read in Redis** (which causes EndAt to be wrongly grouped into port mappings) and **reaper concurrency safety** are the key issues that must be resolved before merging.

fslongjin (changes requested):
Thanks for this PR. Per the prior reviews, there are several critical issues that need to be addressed before merging:

1. **`ListExpiredSandboxIDs` full SCAN does not scale** — Redis SCAN + per-key HGETALL is O(N) and expensive at scale. Consider using a Sorted Set (`ZRANGEBYSCORE`) or adding a per-tick batch cap.

2. **Concurrent reapers across multiple masters** — In HA deployments, all CubeMaster instances run the reaper simultaneously, causing duplicate destruction and log noise. A distributed lock or leader election is needed.

3. **`Sync: false` + EndAt cleared before destroy** — In `destroyExpiredSandbox`, `proxy.EndAt = ""` is written to Redis *before* `DestroySandbox` is called. If the async destroy fails silently, the sandbox is leaked with no retry.

4. **No graceful drain on shutdown** — `runTimeoutReaper` abandons in-flight `destroyExpiredSandbox` goroutines on context cancellation. Add a `sync.WaitGroup` to drain them.

5. **`reapOnce` semaphore drain can block forever** — The `for i := 0; i < cap(sem); i++ { sem <- struct{}{} }` loop has no cancellation path if a goroutine hangs.

6. **`timeout` default changed from 15 to None** — This is a breaking API change for clients relying on the default. Either keep `Some(15)` or document clearly in the changelog.

7. **`SetTimeoutRequest` alias `"duration"` is confusing** — The two endpoints (SetTimeout and Refresh) have different semantics; sharing field names invites misuse.

The feature direction is solid and the multi-format `deserialize_end_at_nanos` is well done. Please address the critical items above before merging.

chengjoey:
> Thanks for this PR. Per the prior reviews, there are several critical issues that need to be addressed before merging:
> 
> 1. **`ListExpiredSandboxIDs` full SCAN does not scale** — Redis SCAN + per-key HGETALL is O(N) and expensive at scale. Consider using a Sorted Set (`ZRANGEBYSCORE`) or adding a per-tick batch cap.
> 2. **Concurrent reapers across multiple masters** — In HA deployments, all CubeMaster instances run the reaper simultaneously, causing duplicate destruction and log noise. A distributed lock or leader election is needed.
> 3. **`Sync: false` + EndAt cleared before destroy** — In `destroyExpiredSandbox`, `proxy.EndAt = ""` is written to Redis _before_ `DestroySandbox` is called. If the async destroy fails silently, the sandbox is leaked with no retry.
> 4. **No graceful drain on shutdown** — `runTimeoutReaper` abandons in-flight `destroyExpiredSandbox` goroutines on context cancellation. Add a `sync.WaitGroup` to drain them.
> 5. **`reapOnce` semaphore drain can block forever** — The `for i := 0; i < cap(sem); i++ { sem <- struct{}{} }` loop has no cancellation path if a goroutine hangs.
> 6. **`timeout` default changed from 15 to None** — This is a breaking API change for clients relying on the default. Either keep `Some(15)` or document clearly in the changelog.
> 7. **`SetTimeoutRequest` alias `"duration"` is confusing** — The two endpoints (SetTimeout and Refresh) have different semantics; sharing field names invites misuse.
> 
> The feature direction is solid and the multi-format `deserialize_end_at_nanos` is well done. Please address the critical items above before merging.

@fslongjin @kinwin-ustc thanks for reveiw, The following are the modifications corresponding to the review content:

1. Changed to `ZRANGEBYSCORE`, destroy some sandboxes in batches each time, with a default batch size of 256.

2. done, a distributed lock added

3. Used Sync: true to call DestroySandbox, so that callCubelet calls DeleteSandboxProxyMap after the cubelet actually succeeds in destroying, and DeleteSandboxProxyMap will synchronously ZREM the TTL index.If the destroy fails, it will return directly, and EndAt will still be retained, so the reaper can still get this sandboxID to retry in the next tick. This completely eliminates the issue of "destroy failure → always missed";

4. done, `sync.WaitGroup` added

5. Used per-tick localWG.Wait(): a pure counting signal, not dependent on channel backpressure.

At the same time, the cancel path using select { case <-ctx.Done(): goto drain; case sem <- struct{}{}: } in the dispatch loop is still used to ensure that new work is stopped immediately when ctx is canceled, and the dispatched work is waited for by localWG.

6. keep `Some(15)`

7. done, removed the alias `duration`

CubeAPI/src/models/mod.rs
```diff
@@ -77,6 +77,11 @@ pub struct NewSandbox {
     #[serde(rename = "templateID")]
     pub template_id: String,
 
+    /// Sandbox lifetime in seconds. Defaults to 15 if the field is
+    /// omitted, matching the historical behavior every official SDK
```
chenhengqi, line 81: What is this `historical behavior`?

CubeMaster/pkg/service/httpservice/cube/sandbox_timeout.go
```diff
@@ -0,0 +1,69 @@
+// Copyright (c) 2024 Tencent Inc.
```
chenhengqi, line 1: It's year 2026 now.

chenhengqi (changes requested):
I think this so called `absolute TTL` should be reworded as `deadline` instead (like `context.WithDeadline()`).

PR #484, opened 2026-06-06
fix(sdk): make Sandbox.create(env_vars) visible to commands.run

xiaojunxiang2023:
## Summary

Fix `Sandbox.create(env_vars={...})` so injected variables are visible to subsequent `commands.run(...)` calls, matching the SDK field description: *Environment variables injected into the sandbox.*

## Problem

```python
sb = Sandbox.create(env_vars={"MY_TOKEN": "abc123"}, config=cfg)

# expected: abc123, got: empty
print(sb.commands.run("echo MY_TOKEN的值为: $MY_TOKEN").stdout)
```
image

Sandbox.create(env_vars={...}) `env_vars` were not reliably available inside the sandbox for login-shell commands.

## Solution

- **CubeAPI**: map SDK `envVars` into the create sandbox payload
- **CubeMaster**: merge create-time envs with template envs; keep template `command`/`args` for env-only placeholders
- **Cubelet**: inject create-time env exports into the guest so `commands.run` (`bash -l -c`) can read them

## Test plan

- [x] CubeAPI: `build_containers_from_env_vars_maps_sdk_payload`
- [x] CubeMaster: `TestApplyTemplateToContainerCreateTimeEnvOverridesTemplateDefaults`
- [x] CubeMaster: `TestApplyTemplateToContainerPreservesExplicitCommandAndArgs`
- [x] Cubelet: `TestGenCreateEnvProfileExportsCreateTimeVars`
- [x] Cubelet: `TestCreateNetfilesAddsCreateEnvProfile`
- [x] Cubelet: `TestEnsureCreateEnvProfileRefreshesFromLatestEnvs`
- [x] Python SDK unit: `TestCreate::test_create_sends_env_vars`
- [x] Python SDK e2e: `tests/test_env_injection_e2e.py`

## fix then deploy
image

---
Signed-off-by: xiaojunxiang <xiaojunxiang@kingsoft.com>


kinwin-ustc:
You missed DCO line in your commit,  you can use git commit --amend -s to add DCO line to your commit

.github/workflows/release-one-click.yml
fslongjin, line 1: Why change this？

fslongjin (changes requested):
Your PR accidently reverts some changes in `master`. It seems you may have made a mistake during the code merge. Please review the PR again and make sure it only contains the necessary changes.

.github/workflows/release-one-click.yml
xiaojunxiang2023: Sorry, I accidentally committed some irrelevant files from a local temporary build. They have been reverted.

xiaojunxiang2023:
> Your PR accidently reverts some changes in `master`. It seems you may have made a mistake during the code merge. Please review the PR again and make sure it only contains the necessary changes.

Sorry, I accidentally committed some irrelevant files from a local temporary build. They have been reverted.

fslongjin (changes requested):
Thanks for working on this and for adding the test coverage. However, I do not think we should merge this PR in its current form.

The current implementation injects create-time environment variables by writing `/etc/profile.d/99-cube-create-env.sh`, so that later `commands.run(...)` calls using a login shell can read them. This only covers part of the problem: it mainly affects commands launched later through the SDK command API. It does not truly refresh the runtime environment for existing processes inside the sandbox, nor for child processes that those in-sandbox processes may start themselves. In other words, this does not solve environment synchronization at the sandbox/runtime layer after snapshot restore.

The technical approach also feels too hacky and introduces undesirable coupling. It ties environment variable semantics to the guest rootfs, shell profile behavior, login-shell startup, and special snapshot restore handling. That creates several risks:

- `env_vars` may contain tokens or secrets; writing them into profile/rootfs files expands their lifetime and creates plaintext-at-rest exposure.
- The profile `export` statements may override per-command environment variables passed via `commands.run(..., env=...)`, which would break the expected precedence semantics.
- Environment variable keys are interpolated directly into a shell script, so they would require strict validation to avoid shell parsing or injection issues.
- The implementation adds coupling across normal start, restore, netfile, hostdir, and rootfs mutation paths, increasing maintenance cost and failure surface.
- The behavior depends on `/bin/bash -l` and `/etc/profile.d`, so it is not a general solution for non-login shells, non-bash entrypoints, or other process models inside the sandbox.

I think the better direction is to treat `env_vars` as sandbox runtime metadata. Cubelet should synchronize that metadata into the guest runtime through a guest agent/envd init or resume-sync mechanism after create/restore. Then the agent/exec layer should merge environment variables when creating new processes, with a clear precedence such as:

```text
image env < template env < sandbox create env < per-command env
```

That approach matches a VM snapshot sandbox architecture much better and avoids persisting runtime configuration into rootfs/profile files.

So I am requesting changes here. The API/Master-side propagation of `envVars` may be a useful part of the solution, but I do not think the Cubelet profile/rootfs/bind-mount injection approach should be merged.

xiaojunxiang2023:
> Thanks for working on this and for adding the test coverage. However, I do not think we should merge this PR in its current form.
> 
> The current implementation injects create-time environment variables by writing `/etc/profile.d/99-cube-create-env.sh`, so that later `commands.run(...)` calls using a login shell can read them. This only covers part of the problem: it mainly affects commands launched later through the SDK command API. It does not truly refresh the runtime environment for existing processes inside the sandbox, nor for child processes that those in-sandbox processes may start themselves. In other words, this does not solve environment synchronization at the sandbox/runtime layer after snapshot restore.
> 
> The technical approach also feels too hacky and introduces undesirable coupling. It ties environment variable semantics to the guest rootfs, shell profile behavior, login-shell startup, and special snapshot restore handling. That creates several risks:
> 
> * `env_vars` may contain tokens or secrets; writing them into profile/rootfs files expands their lifetime and creates plaintext-at-rest exposure.
> * The profile `export` statements may override per-command environment variables passed via `commands.run(..., env=...)`, which would break the expected precedence semantics.
> * Environment variable keys are interpolated directly into a shell script, so they would require strict validation to avoid shell parsing or injection issues.
> * The implementation adds coupling across normal start, restore, netfile, hostdir, and rootfs mutation paths, increasing maintenance cost and failure surface.
> * The behavior depends on `/bin/bash -l` and `/etc/profile.d`, so it is not a general solution for non-login shells, non-bash entrypoints, or other process models inside the sandbox.
> 
> I think the better direction is to treat `env_vars` as sandbox runtime metadata. Cubelet should synchronize that metadata into the guest runtime through a guest agent/envd init or resume-sync mechanism after create/restore. Then the agent/exec layer should merge environment variables when creating new processes, with a clear precedence such as:
> 
> ```
> image env < template env < sandbox create env < per-command env
> ```
> 
> That approach matches a VM snapshot sandbox architecture much better and avoids persisting runtime configuration into rootfs/profile files.
> 
> So I am requesting changes here. The API/Master-side propagation of `envVars` may be a useful part of the solution, but I do not think the Cubelet profile/rootfs/bind-mount injection approach should be merged.

OK, I've opened a new PR [#549 ] to fix it again.

xiaojunxiang2023:
> Thanks for working on this and for adding the test coverage. However, I do not think we should merge this PR in its current form.
> 
> The current implementation injects create-time environment variables by writing `/etc/profile.d/99-cube-create-env.sh`, so that later `commands.run(...)` calls using a login shell can read them. This only covers part of the problem: it mainly affects commands launched later through the SDK command API. It does not truly refresh the runtime environment for existing processes inside the sandbox, nor for child processes that those in-sandbox processes may start themselves. In other words, this does not solve environment synchronization at the sandbox/runtime layer after snapshot restore.
> 
> The technical approach also feels too hacky and introduces undesirable coupling. It ties environment variable semantics to the guest rootfs, shell profile behavior, login-shell startup, and special snapshot restore handling. That creates several risks:
> 
> * `env_vars` may contain tokens or secrets; writing them into profile/rootfs files expands their lifetime and creates plaintext-at-rest exposure.
> * The profile `export` statements may override per-command environment variables passed via `commands.run(..., env=...)`, which would break the expected precedence semantics.
> * Environment variable keys are interpolated directly into a shell script, so they would require strict validation to avoid shell parsing or injection issues.
> * The implementation adds coupling across normal start, restore, netfile, hostdir, and rootfs mutation paths, increasing maintenance cost and failure surface.
> * The behavior depends on `/bin/bash -l` and `/etc/profile.d`, so it is not a general solution for non-login shells, non-bash entrypoints, or other process models inside the sandbox.
> 
> I think the better direction is to treat `env_vars` as sandbox runtime metadata. Cubelet should synchronize that metadata into the guest runtime through a guest agent/envd init or resume-sync mechanism after create/restore. Then the agent/exec layer should merge environment variables when creating new processes, with a clear precedence such as:
> 
> ```
> image env < template env < sandbox create env < per-command env
> ```
> 
> That approach matches a VM snapshot sandbox architecture much better and avoids persisting runtime configuration into rootfs/profile files.
> 
> So I am requesting changes here. The API/Master-side propagation of `envVars` may be a useful part of the solution, but I do not think the Cubelet profile/rootfs/bind-mount injection approach should be merged.

Alright, I have submitted a new PR[#554 ] to address this issue again.

PR #488, opened 2026-06-07
fix(agent): handle cgroup v2 manager creation

fslongjin:
## Summary

- Fix cgroup v2 container cgroup creation in the guest agent by using the cgroups-rs native v2 creation path instead of passing the v1 controller allowlist.
- Attach container processes through `cgroup.procs` in cgroup v2 mode while preserving the existing v1 `add_task` behavior.
- Return cgroup creation errors with context instead of panicking on `unwrap()`.

## Test Plan

- `make agent`
- Manually smoke-tested container creation with cgroup v2 enabled and verified the guest mounts `/sys/fs/cgroup` as `cgroup2`.

Assisted-by: Cursor:GPT-5.5


ls-ggg:
LGTM

PR #490, opened 2026-06-08
fix(cubelet): do not auto-allow DNS when internet is disabled

### LINKED ISSUE #408: # [Question] Why are DNS servers auto-whitelisted in allowOut when AllowInternetAccess=false?
## Description

When creating a sandbox with `allow_internet_access=False` via the E2B SDK, we expect all outbound network access to be blocked. Testing confirms that TCP connections and DNS queries to external servers (e.g., `dig @8.8.8.8`) are correctly blocked. However, DNS resolution using the system's default DNS server (from resolv.conf) still works.

```python
sandbox = Sandbox.create(
    template=os.environ["E2B_TEMPLATE_ID"],
    allow_internet_access=False,
)
```

| Test Command | Expected | Actual |
|-------------|----------|--------|
| `curl https://chat.baidu.com/` | Timeout/Refused | Timeout ✅ |
| `dig @8.8.8.8 chat.baidu.com` | Timeout/Refused | Timeout ✅ |
| `dig chat.baidu.com` (using resolv.conf DNS) | Timeout/Refused | **Resolves successfully** ❌ |
| `dig +tcp chat.baidu.com` | Timeout/Refused | **Resolves successfully** ❌ |

## Investigation

By dumping the eBPF maps on the compute node with `bpftool map dump`, we found:

- **deny_out**: `0.0.0.0/0` (deny all outbound) — as expected
- **allow_out**: `119.29.29.29/32` (default DNS server) — **DNS auto-whitelisted**

Since the eBPF policy evaluation order is `allow_out > deny_out > default allow`, DNS traffic matches the `allow_out` entry for the DNS server and bypasses the deny-all rule in `deny_out`.

### Code Path

The DNS server auto-whitelist happens in Cubelet's network plugin during TAP device creation:

**`Cubelet/network/plugin_tap.go`**:

```
Line 446: cubeVSContextBeforeDNS := buildNetworkAgentCubeVSContext(request)
Line 447: resolvedDNSServers, err := localnetfile.ResolveEffectiveDNSServers(request)
Line 451: cubeVSContext, dnsAllowOutCIDRs := mergeDNSAllowOutCIDRs(cubeVSContextBeforeDNS, resolvedDNSServers)
```

The key decision function `shouldAppendDNSAllowOut` (lines 595-603):

```go
func shouldAppendDNSAllowOut(ctx *networkagentclient.CubeVSContext) bool {
    if ctx == nil {
        return false
    }
    if ctx.AllowInternetAccess != nil && !*ctx.AllowInternetAccess {
        return true  // When AllowInternetAccess=false, trigger DNS whitelist
    }
    return len(ctx.AllowOut) > 0 || len(ctx.DenyOut) > 0
}
```

## Questions

1. **Design Rationale**: What is the original motivation for auto-whitelisting DNS servers in `allowOut` when `AllowInternetAccess=false`? Was this driven by specific user scenarios or failure cases (e.g., applications crashing due to DNS resolution failures)?

2. **Impact of Disabling DNS Access**: If we modify the code so that `AllowInternetAccess=false` also blocks DNS access (i.e., does not auto-whitelist DNS servers), what known issues would this cause? For example:
   - Does the SDK's health check mechanism depend on DNS?
   - Would sandbox internal services (e.g., envd) malfunction due to DNS failures?
   - Are there other internal components that rely on DNS connectivity from the sandbox?

3. **Recommended Approach for Full Outbound Blocking**: For security scenarios that require completely blocking all outbound traffic including DNS, what is the recommended approach? Possible options we've considered:
   - Passing an empty DNS server list via template configuration
   - Adding an explicit control field to CubeVSContext (e.g., `skipDNSAllowOut`)
   - Other approaches

## Environment

- Verified via `bpftool map dump` on compute node
- Default DNS server: `119.29.29.29` (hardcoded in `Cubelet/pkg/container/netfile/netfile.go:37`)

YangYuS8:
## Summary

- Stop auto-appending sandbox DNS servers to `allow_out` when `AllowInternetAccess=false`
- Preserve DNS auto-allow behavior for explicit allow/deny network policies when internet access is not explicitly disabled
- Add test coverage for disabled-internet behavior

Fixes #408

## Checks

- `gofmt -w Cubelet/network/plugin_tap.go Cubelet/network/plugin_tap_create_test.go`
- `go test ./network` from the `Cubelet` module
- `git diff --check`

Assisted-by: opencode:OpenAI/GPT-5.5

### DIFF
--- a/Cubelet/network/plugin_tap.go
+++ b/Cubelet/network/plugin_tap.go
@@ -275,12 +275,12 @@ type Config struct {
 
 	RootPath string `toml:"root_path"`
 
-	EnableNetworkAgent          bool             `toml:"enable_network_agent"`
-	NetworkAgentEndpoint        string           `toml:"network_agent_endpoint"`
-	NetworkAgentTapSocket       string           `toml:"network_agent_tap_socket"`
-	NetworkAgentInitTimeout     tomlext.Duration `toml:"network_agent_init_timeout"`
-	NetworkAgentRetryInterval   tomlext.Duration `toml:"network_agent_retry_interval"`
-	NetworkAgentTapFDTimeout    tomlext.Duration `toml:"network_agent_tap_fd_timeout"`
+	EnableNetworkAgent        bool             `toml:"enable_network_agent"`
+	NetworkAgentEndpoint      string           `toml:"network_agent_endpoint"`
+	NetworkAgentTapSocket     string           `toml:"network_agent_tap_socket"`
+	NetworkAgentInitTimeout   tomlext.Duration `toml:"network_agent_init_timeout"`
+	NetworkAgentRetryInterval tomlext.Duration `toml:"network_agent_retry_interval"`
+	NetworkAgentTapFDTimeout  tomlext.Duration `toml:"network_agent_tap_fd_timeout"`
 
 	ReconcileInterval tomlext.Duration `toml:"reconcile_interval"`
 }
@@ -621,7 +621,7 @@ func shouldAppendDNSAllowOut(ctx *networkagentclient.CubeVSContext) bool {
 		return false
 	}
 	if ctx.AllowInternetAccess != nil && !*ctx.AllowInternetAccess {
-		return true
+		return false
 	}
 	return len(ctx.AllowOut) > 0 || len(ctx.DenyOut) > 0
 }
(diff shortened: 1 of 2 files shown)


fslongjin (approved):
LGTM

PR #492, opened 2026-06-08
feat(CubeMaster): support daemonless template image export via skopeo/umoci

lkml-likexu:
feat(CubeMaster): support daemonless template image export via skopeo/umoci

Building template images previously required a Docker daemon on the
cubemaster node to pull source images and export their root filesystem.
This adds an alternative, daemonless path that uses skopeo and umoci when
both are available, and falls back to Docker otherwise so existing
deployments keep working unchanged.

Running without a privileged Docker daemon reduces the node's attack
surface and runtime dependencies, and makes template builds easier to run
in restricted or rootless environments. The export strategy is chosen once
when the source image is resolved, so preparation and export stay
consistent.

Signed-off-by: Like Xu <likexu@tencent.com>

### DIFF
--- a/CubeMaster/docker/Dockerfile
+++ b/CubeMaster/docker/Dockerfile
@@ -22,7 +22,9 @@ RUN DEBIAN_FRONTEND=noninteractive apt-get install -y \
 	telnet \
 	tzdata \
 	vim \
-	wget
+	wget \
+	skopeo \
+	umoci
 
 RUN cp /usr/share/zoneinfo/Asia/Shanghai /etc/localtime
 RUN update-ca-certificates
--- a/CubeMaster/pkg/templatecenter/template_image.go
+++ b/CubeMaster/pkg/templatecenter/template_image.go
@@ -8,6 +8,7 @@ import (
 	"bytes"
 	"context"
 	"crypto/sha256"
+	"encoding/base64"
 	"encoding/hex"
 	"encoding/json"
 	"errors"
@@ -17,6 +18,7 @@ import (
 	"os"
 	"os/exec"
 	"path/filepath"
+	"regexp"
 	"sort"
 	"strconv"
 	"strings"
@@ -96,6 +98,7 @@ var deleteRootfsArtifactRecord = func(ctx context.Context, artifactID string) er
 	return store.db.WithContext(ctx).Unscoped().Table(constants.RootfsArtifactTableName).
 		Where("artifact_id = ?", artifactID).Delete(&models.RootfsArtifact{}).Error
 }
+var executableLookPath = exec.LookPath
 
 var ErrNoFailedTemplateReplicas = errors.New("no failed template replicas matched redo request")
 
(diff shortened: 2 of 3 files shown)


lkml-likexu:
Round-1 Fixed:

All six issues are fixed, the build passes, and the full templatecenter test package is green. Here's a summary of each fix in CubeMaster/pkg/templatecenter/template_image.go:

(1) Argument injection / CWE-88 in validateImageRef Added a strict character whitelist (^[A-Za-z0-9._:/@-]+$) so any reference containing whitespace or other argument delimiters is rejected — closing the registry.example.com/image --authfile /etc/shadow injection. The empty-ref and leading-dash checks are kept. Added "regexp" to imports.

(2) Missing doc comment on dockerlessExportImageRootfs Added a doc comment describing the flow (skopeo copy → umoci unpack → rename rootfs) and explaining why the scratch dir is placed under destRootfsDir's parent (same-filesystem rename to avoid EXDEV full copies).

(3) Duplicated parsing logic Extracted splitImageRef(imageRef) (name, tag string) for the shared prefix/digest/colon logic; imageTagFromRef and imageNameWithoutTagDigest now delegate to it with identical behavior.

(4) Nil-pointer guard in exportImageRootfs Added an explicit if source == nil check before dereferencing source.localRef, matching the defensive style of estimateImageSizeFromInspect.

(5) Noisy NOCC annotation Removed // NOCC:Path Traversal() from defer os.RemoveAll(workDir), since workDir comes from os.MkdirTemp (not a user-controlled path). The remaining annotations stay where dynamic values are genuinely involved.

(6) Test gap Added TestSplitImageRef (table-driven, covering no-colon, port-without-tag, port-with-tag, tag+digest after docker://, digest-without-tag, bare name+tag — also asserting imageTagFromRef/imageNameWithoutTagDigest) and TestValidateImageRef (valid refs plus injection attempts).

lkml-likexu:
Round-2 Fixed:

Fixed all three locations where json.Marshal errors were silently discarded. Build and the full templatecenter test package pass, no lint issues.

Changes in CubeMaster/pkg/templatecenter/template_image.go:

prepareDockerlessSourceImage (skopeo path, ~L1469): now returns fmt.Errorf("marshal image config: %w", err) and calls cleanup() first to release the temporary skopeo auth file.
prepareLocalSourceImage (redo/local docker path, ~L626): returns the wrapped error.
prepareDockerSourceImage (docker path, ~L1533): returns the wrapped error; the function's existing deferred cleanup still removes dockerConfigDir because the early return happens before removeDockerConfigDir = false.
All three now propagate the marshal failure instead of writing a zero-value config into the artifact record.

fslongjin (approved):
LGTM

fslongjin:
One concern with the new dockerless (`skopeo` + `umoci`) path is peak disk usage during snapshot/artifact creation.

The current docker path has an important optimization: when Phase 2 loop-mount streaming is available, it can stream `docker export` directly into the ext4 image, which avoids materializing a full intermediate rootfs on disk. In other words, the artifact-building side is close to “final ext4 only”.

The dockerless path does not have that property. It currently does:

1. `skopeo copy` into an on-disk OCI layout
2. `umoci unpack` into an on-disk bundle/rootfs
3. `mkfs.ext4 -d <rootfs>` to build the final ext4

So even though the code deletes the OCI layout early after unpack, the daemonless path still needs an unpacked rootfs on disk before ext4 creation. During the ext4 build stage, peak usage is roughly `rootfs + ext4`, which for larger images is typically around ~2.1x of the unpacked rootfs size (plus fixed overhead). That is materially higher than the optimized docker streaming path.

So I do think this change can reintroduce the “high peak disk usage while creating snapshots” problem that the docker path was specifically optimized to reduce. At minimum, I think the PR should call out this trade-off explicitly and explain whether it is acceptable for the intended deployment environments.

lkml-likexu:
@fslongjin Loop-mount requires higher privileges than running Docker does. In that scenario, docker-streaming and umoci raw unpack (to loop device) are essentially equivalent in terms of I/O resource consumption — in other words, umoci-streaming is technically feasible.

But in deployment environments that lack CAP_SYS_ADMIN, we can use neither docker-in-docker nor loop_mount. There, the current OCI-toolchain approach and its trade-off are acceptable — though the performance call-out is still warranted.

PR #502, opened 2026-06-09
feat(affinity): support custom node affinity selector via annotation

devincd:
## Summary

Add `com.nodeaffinity.selector` annotation to allow specifying arbitrary `NodeSelectorRequirement` objects when creating a sandbox, enabling custom node affinity rules beyond the predefined cluster label and instance type selectors.

Also add `NodeLabels` field to `node.Node` and merge it into the `Labels()` method, so that labels reported during node registration are visible to the scheduler for affinity matching.

### Custom node affinity selector annotation

A new annotation `com.nodeaffinity.selector` accepts a JSON array of `NodeSelectorRequirement` objects. Each object has:

- `key` (string) — the node label key
- `operator` (string) — one of `In`, `NotIn`, `Exists`, `DoesNotExist`, `Gt`, `Lt`
- `values` (array of strings, optional) — required for `In` and `NotIn`, must be omitted for `Exists` and `DoesNotExist`

The parsed requirements are appended to the `matchExpressions` in `constructNodeAffinity`, alongside existing cluster label and instance type selectors. If the annotation value is invalid JSON, the error is logged and the request proceeds without the custom selector (graceful degradation).

The `NodeSelectorRequirement` struct is unchanged — `Values` remains `map[string]any`. The conversion from JSON array to map happens at the parsing boundary via `valuesSliceToMap`.

### Node labels propagation

Previously, `node.Node.Labels()` only returned 6 hardcoded affinity keys derived from struct fields (`Zone`, `ClusterLabel`, `CPUType`, `MemorySize`, `CPUCores`, `InstanceType`). Custom labels reported by cubelet during node registration were stored in `NodeSnapshot.Labels` and persisted to MySQL, but dropped during the `toSchedulerNode` conversion — making them invisible to the scheduler.

This PR adds a `NodeLabels` field to `node.Node` and propagates `snap.Labels` through `toSchedulerNode` and the localcache update path (`updateNodeFromMetaData`). The `Labels()` method now merges `NodeLabels` on top of the hardcoded keys, so custom labels (e.g. `kubernetes.io/hostname`) become available for affinity matching.

### Pin sandbox to a specific host

```bash
curl -sS -X POST "http://localhost:8089/cube/sandbox" \
  -H "Content-Type: application/json" \
  -d '{
    "RequestID": "pin-host-001",
    "instance_type": "cubebox",
    "timeout": 60,
    "containers": [],
    "annotations": {
[...]

PR #504, opened 2026-06-09
feat(affinity): support custom node affinity selector via annotation

devincd:
Add com.nodeaffinity.selector annotation to allow specifying arbitrary NodeSelectorRequirements when creating a sandbox. The annotation value is a JSON array of {key, operator, values} objects, supporting all six operators (In, NotIn, Exists, DoesNotExist, Gt, Lt).

Also add NodeLabels field to node.Node and merge it into Labels() so that labels reported during node registration are visible to the scheduler for affinity matching. Deep-copy NodeLabels in Clone() to prevent data races.

Validate that In/NotIn operators have non-empty values and Exists/DoesNotExist have empty values, returning MasterParamsError on invalid input. Consolidate the repeated req.Annotations nil checks in constructNodeAffinity into a single block.

Supersedes #502 (previous PR was from a dirty branch with unrelated upstream commits; this is a clean single-commit PR from feat/node-affinity-selector).

### DIFF
--- a/CubeMaster/pkg/base/node/node.go
+++ b/CubeMaster/pkg/base/node/node.go
@@ -8,6 +8,7 @@ package node
 import (
 	"fmt"
 	"sort"
+	"sync"
 	"sync/atomic"
 	"time"
 
@@ -88,7 +89,19 @@ type Node struct {
 	NicQueues      int64 `json:"nic_queues,omitempty"`
 
 	NodeLabels map[string]string `json:"NodeLabels,omitempty"`
+
+	labelsCache *nodeLabelsCacheStore
 }
+
+type nodeLabelsCache struct {
+	labels map[string]string
+}
+
+type nodeLabelsCacheStore struct {
+	cache atomic.Pointer[nodeLabelsCache]
+}
+
+var nodeLabelsCacheInitMu sync.Mutex
 
 func (n *Node) Clone() *Node {
 	if n == nil {
@@ -100,6 +113,7 @@ func (n *Node) Clone() *Node {
 	localCreateNum := atomic.LoadInt64(&n.LocalCreateNum)
 	cloned := *n
 	cloned.LocalCreateNum = localCreateNum
+	cloned.labelsCache = nil
 	if n.VirtualNodeQuotaArray != nil {
 		cloned.VirtualNodeQuotaArray = append([]int64(nil), n.VirtualNodeQuotaArray...)
 	}
@@ -110,6 +124,18 @@ func (n *Node) Clone() *Node {
 		}
 	}
 	return &cloned
+}
+
+func (n *Node) labelsCacheStore() *nodeLabelsCacheStore {
+	if n.labelsCache != nil {
+		return n.labelsCache
+	}
+	nodeLabelsCacheInitMu.Lock()
+	defer nodeLabelsCacheInitMu.Unlock()
+	if n.labelsCache == nil {
+		n.labelsCache = &nodeLabelsCacheStore{}
+	}
+	return n.labelsCache
 }
 
 func (n *Node) ID() string {
@@ -126,9 +152,14 @@ func (n *Node) LocalCreateNumIncrBy(i int64) int64 {
 }
 
 func (n *Node) Labels() map[string]string {
+	cacheStore := n.labelsCacheStore()
+	if cache := cacheStore.cache.Load(); cache != nil {
+		return cache.labels
+	}
+
 	// Canonical affinity keys derived from Node struct fields always take
 	// priority over node-reported labels, so they are written last.
-	labels := make(map[string]string)
+	labels := make(map[string]string, len(n.NodeLabels)+6)
 	for k, v := range n.NodeLabels {
 		labels[k] = v
 	}
@@ -138,7 +169,20 @@ func (n *Node) Labels() map[string]string {
 	labels[constants.AffinityKeyMemorySize] = fmt.Sprintf("%dMi", n.QuotaMem)
 	labels[constants.AffinityKeyCPUCores] = fmt.Sprintf("%dm", n.QuotaCpu)
 	labels[constants.AffinityKeyInstanceType] = n.InstanceType
+	cache := &nodeLabelsCache{labels: labels}
+	if cacheStore.cache.CompareAndSwap(nil, cache) {
+		return labels
+	}
+	if cache := cacheStore.cache.Load(); cache != nil {
+		return cache.labels
+	}
 	return labels
+}
+
+func (n *Node) InvalidateLabelsCache() {
+	if n.labelsCache != nil {
+		n.labelsCache.cache.Store(nil)
+	}
 }
 
 type NodeList []*Node
--- a/CubeMaster/pkg/base/node/node.go
+++ b/CubeMaster/pkg/base/node/node.go
@@ -86,6 +86,8 @@ type Node struct {
 
 	LocalCreateNum int64 `json:"LocalCreateNum,omitempty"`
 	NicQueues      int64 `json:"nic_queues,omitempty"`
+
+	NodeLabels map[string]string `json:"NodeLabels,omitempty"`
 }
 
 func (n *Node) Clone() *Node {
@@ -100,6 +102,12 @@ func (n *Node) Clone() *Node {
 	cloned.LocalCreateNum = localCreateNum
 	if n.VirtualNodeQuotaArray != nil {
 		cloned.VirtualNodeQuotaArray = append([]int64(nil), n.VirtualNodeQuotaArray...)
+	}
+	if n.NodeLabels != nil {
+		cloned.NodeLabels = make(map[string]string, len(n.NodeLabels))
+		for k, v := range n.NodeLabels {
+			cloned.NodeLabels[k] = v
+		}
 	}
 	return &cloned
 }
@@ -118,7 +126,12 @@ func (n *Node) LocalCreateNumIncrBy(i int64) int64 {
 }
 
 func (n *Node) Labels() map[string]string {
+	// Canonical affinity keys derived from Node struct fields always take
+	// priority over node-reported labels, so they are written last.
 	labels := make(map[string]string)
+	for k, v := range n.NodeLabels {
+		labels[k] = v
+	}
 	labels[constants.AffinityKeyZone] = n.Zone
 	labels[constants.AffinityKeyClusterID] = n.ClusterLabel
 	labels[constants.AffinityKeyCPUType] = n.CPUType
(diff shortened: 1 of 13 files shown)


devincd:
Thanks @cubesandboxbot for the thorough review. All four issues have been addressed in the latest push (d79b920):

  1. Duplicated switch prevents Gt/Lt from working — Removed the redundant second switch r.Operator block. The first switch had no default case, causing Gt/Lt to silently fall through into the second switch which rejected them as
  "unsupported operator". Consolidated into a single switch with all six operators (In, NotIn, Exists, DoesNotExist, Gt, Lt) covered.
  2. Dead code — redundant second switch — The second switch block (lines 184-195) duplicated the exact same In/NotIn/Exists/DoesNotExist validation already performed by the first switch. Since matching operators never reach the second
  switch, those branches were dead code. Removed entirely.
  3. Missing validation for Gt/Lt operators — Added validation that Gt and Lt operators require exactly one value in the values array, consistent with the Kubernetes NodeSelectorRequirement spec. Per k8s semantics, no key restriction is
  applied — Gt/Lt work with any label key.
  4. Shared map reference instead of deep copy — updateNodeFromMetaData was assigning n.NodeLabels directly (map pointer), meaning the cached node and the source shared the same underlying map. This could cause race conditions if the
  source map is modified after assignment. Fixed with a deep copy, matching the defensive pattern already used in Clone().

CubeMaster/pkg/base/node/node.go
```diff
@@ -125,6 +133,9 @@ func (n *Node) Labels() map[string]string {
 	labels[constants.AffinityKeyMemorySize] = fmt.Sprintf("%dMi", n.QuotaMem)
 	labels[constants.AffinityKeyCPUCores] = fmt.Sprintf("%dm", n.QuotaCpu)
 	labels[constants.AffinityKeyInstanceType] = n.InstanceType
+	for k, v := range n.NodeLabels {
```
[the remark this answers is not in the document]
devincd: Thanks for the suggestion. While the per-call map allocation is a valid concern in theory, in practice Labels() only creates a small map (6 hardcoded keys + entries from NodeLabels), and the scheduler typically evaluates a few dozen candidate nodes per cycle. The allocation overhead is negligible at this scale compared to the actual scheduling logic (prefilter matching, scoring, etc.).

Adding an atomic.Value cache would introduce complexity: every metadata update path (updateNodeFromMetaData, Clone, etc.) would need to invalidate the cache correctly under concurrency. The risk of subtle bugs from stale cache outweighs the marginal GC benefit at the current scale.

If scheduling throughput becomes a measurable bottleneck in the future (e.g., hundreds of nodes evaluated per cycle at high QPS), we can revisit this with proper benchmarking. For now, I'd prefer to keep the current simple approach.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
```diff
@@ -139,3 +151,63 @@ func isLargeCpucores(ctx context.Context, req *types.CreateCubeSandboxReq, large
...
+type nodeSelectorRequirementJSON struct {
+	Key      string                      `json:"key"`
+	Operator affinity.NodeSelectorOperator `json:"operator"`
+	Values   []string                    `json:"values,omitempty"`
+}
+
+func parseNodeAffinitySelector(selectorJSON string) ([]affinity.NodeSelectorRequirement, error) {
+	var raw []nodeSelectorRequirementJSON
+	if err := json.Unmarshal([]byte(selectorJSON), &raw); err != nil {
+		return nil, err
```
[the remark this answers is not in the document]
devincd: Fixed as requested.

CubeMaster/pkg/nodemeta/service.go
```diff
@@ -444,6 +444,7 @@ func toSchedulerNode(snap *NodeSnapshot) *node.Node {
 		CreateConcurrentNum: snap.CreateConcurrentNum,
 		MaxMvmLimit:         snap.MaxMvmNum,
 		MetaDataUpdateAt:    snap.HeartbeatTime,
+		NodeLabels:          snap.Labels,
```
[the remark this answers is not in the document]
devincd: Fixed as requested.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
```diff
@@ -102,10 +106,18 @@ func constructNodeAffinity(ctx context.Context, req *types.CreateCubeSandboxReq)
 			}
 			matchExpressions = append(matchExpressions, requiredV)
 		}
+
+		if selectorJSON, ok := req.Annotations[constants.AnnotationsNodeAffinitySelector]; ok && selectorJSON != "" {
+			parsed, err := parseNodeAffinitySelector(selectorJSON)
+			if err != nil {
+				return nil, fmt.Errorf("parseNodeAffinitySelector: %w", err)
```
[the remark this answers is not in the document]
devincd: Fixed as requested.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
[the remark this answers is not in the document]
devincd: Fixed as requested.

CubeMaster/pkg/localcache/node_cache.go
```diff
@@ -311,6 +311,14 @@ func (l *local) updateNodeFromMetaData(n *node.Node) error {
 		old.CPUType = n.CPUType
 		old.InstanceType = n.InstanceType
 		old.OssClusterLabel = n.OssClusterLabel
+		if n.NodeLabels != nil {
+			old.NodeLabels = make(map[string]string, len(n.NodeLabels))
```
fslongjin: This publishes the new map to the cached node before the map is fully populated. Scheduler read paths can clone/read the same cached node concurrently, and `Node.Clone()` iterates `NodeLabels` without holding this lock. That can race with the writes below and may panic with concurrent map iteration/write. Please build the replacement map in a local variable first, fill it completely, and then assign it to `old.NodeLabels` once.

CubeMaster/pkg/base/node/node.go
fslongjin: Merging `NodeLabels` after the built-in scheduler labels lets node-reported labels override canonical affinity keys such as cluster ID, instance type, CPU, and memory. That can change existing `com.nodeaffinity.cluster.label`, `com.nodeaffinity.instancetype`, and large-size affinity behavior. Please either prevent/skip reserved key overrides or explicitly define and test that external labels have higher priority.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
```diff
@@ -139,3 +151,63 @@ func isLargeCpucores(ctx context.Context, req *types.CreateCubeSandboxReq, large
...
+		switch r.Operator {
+		case affinity.NodeSelectorOpIn, affinity.NodeSelectorOpNotIn:
+			if len(r.Values) == 0 {
+				return nil, fmt.Errorf("operator %q requires non-empty values for key %q", r.Operator, r.Key)
+			}
+		case affinity.NodeSelectorOpExists, affinity.NodeSelectorOpDoesNotExist:
+			if len(r.Values) != 0 {
+				return nil, fmt.Errorf("operator %q requires empty values for key %q", r.Operator, r.Key)
+			}
+		case affinity.NodeSelectorOpGt, affinity.NodeSelectorOpLt:
```
fslongjin, line 216: The parser accepts `Gt`/`Lt` for any key as long as there is exactly one value, but the matcher currently only evaluates these operators for the built-in CPU and memory affinity keys; other keys always return false. This makes some valid-looking annotations silently unschedulable. Please either implement the intended numeric comparison for arbitrary labels or reject unsupported `Gt`/`Lt` keys during parsing with a clear parameter error.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
```diff
@@ -139,3 +151,63 @@ func isLargeCpucores(ctx context.Context, req *types.CreateCubeSandboxReq, large
...
+// converted to map[string]any to fit affinity.NodeSelectorRequirement.
+type nodeSelectorRequirementJSON struct {
+	Key      string                      `json:"key"`
+	Operator affinity.NodeSelectorOperator `json:"operator"`
+	Values   []string                    `json:"values,omitempty"`
+}
+
+func parseNodeAffinitySelector(selectorJSON string) ([]affinity.NodeSelectorRequirement, error) {
+	var raw []nodeSelectorRequirementJSON
+		if err := json.Unmarshal([]byte(selectorJSON), &raw); err != nil {
```
fslongjin: This file is not gofmt-formatted: this block is over-indented, the struct field alignment above is off, and the file is missing a trailing newline. Please run `gofmt` on the changed Go files before merging.

fslongjin (commented):
I left a few inline comments focused on scheduler correctness and concurrency risks.

CubeMaster/pkg/localcache/node_cache.go
devincd: Fixed as requested.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
devincd: Fixed as requested.

CubeMaster/pkg/base/node/node.go
devincd: Thanks for raising this. It has been addressed in the latest push (9905f6f). The implementation now writes NodeLabels first, then the canonical affinity keys (Zone, ClusterID, CPUType, MemorySize, CPUCores, InstanceType) last. Since map writes overwrite existing entries, the built-in scheduler keys always take priority over any node-reported labels with the same key. This guarantees that com.nodeaffinity.cluster.label, com.nodeaffinity.instancetype, and large-size affinity behavior cannot be altered by external labels. No reserved-key set is needed — the write ordering alone ensures the correct precedence.

CubeMaster/pkg/service/httpservice/cube/affinityutil.go
devincd: It has been addressed in the latest push (9905f6f). The parser now rejects Gt/Lt operators for keys other than kubernetes.io/memory-size and kubernetes.io/cpu-cores with a clear parameter error:

operator "Gt" is only supported for keys "kubernetes.io/memory-size" and "kubernetes.io/cpu-cores", got "custom-key"
  
This prevents silently unschedulable annotations. Supporting numeric comparison for arbitrary labels would require extending the Match() function to parse label values as quantities for any key, which is a larger change. I'd prefer to keep the current approach (whitelist the supported keys) for now and revisit general-purpose Gt/Lt if a concrete use case arises.

PR #509, opened 2026-06-10
Enable build on aarch64

JunHe77:
This PR enables build on aarch64 paltform. Tested with both `make all` and ./deploy/one-click/build-release-bundle-builder.sh.


lkml-likexu:
@JunHe77 Thanks for your contributions. 

Could you please provide more test environments and the results of the corresponding test cases?

May I ask which aarch64 hardware and host kernel versions you have actually tested
the CubeSandbox use cases on — for example, the quickstart in the examples?

Could you help check whether this [PR](https://github.com/TencentCloud/CubeSandbox/pull/256/) is actually helpful for aarch64 support?

For instance, in this PR we are still missing (at least):

- An aarch64 mvm vmlinux builder
- Documentation updates for developers and deployers

If the default page size on the arm machine is not 4KB,  will Cloud Hypervisor (CH) still work properly?

Besides, being able to run CubeSandbox on Apple M-series chips would make cube and ARM much more popular.

lkml-likexu (commented):
For commit "docker/Dockerfile.builder: support multi-arch builder image"

Overall the arch-mapping is mostly correct (Go via `dpkg --print-architecture`, protoc `aarch_64`, rustup musl target, libseccomp `--host`, openssl multiarch dir all map correctly). But there is one confirmed regression that defeats part of the patch's own goal, plus a few smaller items.

### High — OpenSSL lib-dir override is now dead on both arches
`docker/Dockerfile.builder:28-29`

```28:29:docker/Dockerfile.builder
    UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR=/usr/lib/${HOST_ARCH}-linux-gnu \
    UNKNOWN_LINUX_MUSL_OPENSSL_LIB_DIR=/usr/lib/${HOST_ARCH}-linux-gnu \
```

`openssl-sys` only reads env vars prefixed with the **full uppercased target triple** (e.g. `X86_64_UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR`, `AARCH64_UNKNOWN_LINUX_MUSL_OPENSSL_LIB_DIR`) or the unprefixed `OPENSSL_LIB_DIR`. The renamed `UNKNOWN_LINUX_GNU_OPENSSL_LIB_DIR` matches no target, so it's silently never read — on **both** x86_64 and aarch64. The original block already listed all four arch-prefixed vars (so it was already arch-agnostic), and `hypervisor/resources/Dockerfile:87-90` still uses the correct names. **Fix:** revert to the four arch-prefixed vars (no `HOST_ARCH` interpolation needed — they already covered both arches).

### Medium — World-writable Rust toolchain
`docker/Dockerfile.builder:133`

```bash
chmod -R a+rwX /usr/local/rustup
```

`a+rwX` makes every toolchain binary world-writable (tampering surface / least-privilege violation). `a+rX` is sufficient for compilation; the write bit is only needed for runtime `rustup install/update`. Impact is bounded (ephemeral `--rm` container, non-root `--user`), but recommend dropping the `w` bit unless runtime rustup mutation is intended.

### Low — `HOST_ARCH` default can silently produce a broken mixed-arch image
`HOST_ARCH` defaults to `x86_64`, but the Go download independently uses `$(dpkg --print-architecture)`. A bare `docker build` (no `--build-arg`) on an arm host — which is how CI invokes it (`build-builder-image.yml` doesn't pass `HOST_ARCH`) — would fetch an arm64 Go but x86_64 rustup/libseccomp/openssl. Only the `make builder-image` path auto-detects the arch. Consider sourcing a single arch via buildx `TARGETARCH`, or at least documenting that direct `docker build` on non-x86_64 requires `--build-arg HOST_ARCH=`.

### Low — The aarch64 path has zero CI coverage
All workflows run on x86_64 only (no QEMU/`linux/arm64`/arm runners). Every new `HOST_ARCH=aarch64` branch is unexercised, so the patch's primary purpose is unvalidated, and the OpenSSL rename above wouldn't be caught by any existing check. Worth an arm64 build job (QEMU + buildx) building the image with `HOST_ARCH=aarch64` and at least one musl Rust component.

### Low — Stale docs
- `CONTRIBUTING.md:36` hardcodes the `x86_64-unknown-linux-musl` target prerequisite, now arch-dependent.
- `docker/README.md` is the natural place to document the new `HOST_ARCH` arg / multi-arch capability.

### Dismissed (raised by reviewers, but not real issues)
- Removing the blank line between the two `RUN`s does **not** merge layers (continuation depends on trailing `\`).
- The `chmod -R` does **not** create a duplicate/oversized layer — it's in the same `RUN` as the rustup install.
- Unverified downloads (go/protoc/libseccomp/rustup) and unquoted `HOST_ARCH` interpolation are pre-existing / trusted-input, not introduced or worsened here.

**Recommendation:** address the High OpenSSL finding before merge; the Medium `chmod` is a quick safe improvement; the rest are optional follow-ups.

docker/Dockerfile.builder
```diff
@@ -59,6 +59,7 @@ RUN apt-get update -o Acquire::Retries=3 \
         git \
         git-lfs \
         jq \
+        libbpf-dev \
```
lkml-likexu: Remove the duplicate `libbpf-dev`

CubeNet/cubevs/localgw_x86_bpfel.o
chenhengqi, line 1: Please add `.gitignore` entries for these auto-generated files.

chenhengqi (approved):
Networking-related changes looks good to me. Thanks.

dev-env/run_vm.sh
```diff
@@ -136,9 +137,16 @@ else
   log_info "Clean shutdown: in another terminal run ./login.sh, then poweroff in the guest (do not Ctrl+a x — abrupt QEMU exit)"
 fi
 
+case "${TARGET_ARCH}" in
+  "x86_64") VM_MACHINE='q35'; BIOS_PARAM='';;
+  "aarch64") VM_MACHINE=virt; BIOS_PARAM="-bios $(find /usr/share/ -name QEMU_EFI.fd)";;
```
lkml-likexu: - **[major] `run_vm.sh` firmware discovery is fragile:**
  ```sh
  "aarch64") VM_MACHINE=virt; BIOS_PARAM="-bios $(find /usr/share/ -name QEMU_EFI.fd)";;
  ```
  - If `QEMU_EFI.fd` is **not installed**, the command substitution is empty and you get a bare `-bios` with no path (QEMU error, or it consumes the next arg).
  - If `find` returns **multiple matches**, the unquoted expansion produces multiple tokens and a malformed `-bios`.
  - `find` failures are swallowed under `set -euo pipefail` (it's a command substitution). Recommend resolving the firmware path explicitly, validating it's a single existing file, and failing with a clear message otherwise (e.g. `AAVMF`/`edk2-aarch64` paths vary by distro).

dev-env/run_vm.sh
```diff
@@ -136,9 +137,16 @@ else
...
+  *) log_error "Not supported yet";;
```
lkml-likexu: - **[minor] Unsupported-arch case doesn't abort:**
  ```sh
  *) log_error "Not supported yet";;
  ```
  If `log_error` doesn't `exit`, execution continues with an empty `VM_MACHINE`, producing `-machine ,accel=kvm`. Add an explicit `exit 1`.

deploy/one-click/build-release-bundle.sh
```diff
@@ -28,7 +28,7 @@ CUBE_PROXY_SOURCE_DIR="${ONE_CLICK_CUBE_PROXY_SOURCE_DIR:-${ROOT_DIR}/CubeProxy}
 CUBE_EGRESS_SOURCE_DIR="${ONE_CLICK_CUBE_EGRESS_SOURCE_DIR:-${ROOT_DIR}/CubeEgress}"
 WEB_SOURCE_DIR="${ONE_CLICK_WEB_SOURCE_DIR:-${ROOT_DIR}/web}"
 WEB_DIST_OVERRIDE="${ONE_CLICK_WEB_DIST_DIR:-}"
-MKCERT_BIN_ASSET="${ONE_CLICK_MKCERT_BIN:-${SCRIPT_DIR}/assets/bin/mkcert}"
+MKCERT_BIN_ASSET="${ONE_CLICK_MKCERT_BIN:-${SCRIPT_DIR}/assets/bin/mkcert-v1.4.4-linux-$(dpkg --print-architecture)}"
```
lkml-likexu: - [minor] `build-release-bundle.sh` selects the asset via `$(dpkg --print-architecture)` (→ `amd64`/`arm64`, matching the filenames). `dpkg` is Debian/Ubuntu-only; this is fine inside the Ubuntu builder but will break if the script is ever run on a non-dpkg host. A `uname -m` mapping would be more portable.

configs/kernel-oc9.x86_64.config
lkml-likexu, line 1: - **[major] The rename breaks existing references to `configs/kernel-oc9.config`.** That path is still used by CI workflows that are not updated in this patch:
  - `.github/workflows/build-vmlinux.yml` — lines 6 (`paths:` trigger), 34 (`sha256sum configs/kernel-oc9.config`), 82 (`cp configs/kernel-oc9.config "${KERNEL_SRC}/.config"`)
  - `.github/workflows/release-one-click.yml` — line 136 (`sha256sum configs/kernel-oc9.config`)

  After this rename those steps will fail (file not found) / silently stop triggering. The rename must be accompanied by arch-aware updates to these workflows (and any Makefiles/scripts) that pick `kernel-oc9.x86_64.config` vs `kernel-oc9.aarch64.config`. Please grep the tree for `kernel-oc9.config` and fix every consumer in this commit.

configs/kernel-oc9.aarch64.config
```diff
@@ -0,0 +1,8653 @@
+#
+# Automatically generated file; DO NOT EDIT.
+# Linux/arm64 6.6.119 Kernel Configuration
+#
+CONFIG_CC_VERSION_TEXT="gcc (Ubuntu 12.3.0-1ubuntu1~22.04.3) 12.3.0"
```
lkml-likexu: This is hilarious. I suggest you actually install and try out the oc9 distribution.

agent/.cargo/config.toml
```diff
@@ -11,7 +11,7 @@ rustflags = [ "-C", "link-arg=-lpthread" ]
 [target.aarch64-unknown-linux-musl]
 ## Only setting linker with `aarch64-linux-musl-gcc`, the
```
lkml-likexu: - **[minor] Stale comment.** In `agent/.cargo/config.toml` the comment directly above the changed line still reads:
  ```
  ## Only setting linker with `aarch64-linux-musl-gcc`, the
  ## `rust-agent` could be totally statically linked.
  ```
  but the value is now `musl-gcc`. Update the comment so it doesn't contradict the code.

CubeShim/shim/src/snapshot/mod.rs
```diff
@@ -230,7 +230,10 @@ impl Snapshot {
     fn launch_vmm(&mut self) -> CResult<()> {
         //launch
         cube_hypervisor::set_runtime_seccomp_rules(vec![
+            #[cfg(target_arch = "x86_64")]
             (libc::SYS_mkdir, vec![]),
+            #[cfg(target_arch = "aarch64")]
+            (libc::SYS_mkdirat, vec![]),
             (libc::SYS_getsockopt, vec![]),
             (libc::SYS_setsockopt, vec![]),
         ]);
```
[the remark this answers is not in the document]
JunHe77: Updated,

JunHe77:
> @JunHe77 Thanks for your contributions.
> 
> Could you please provide more test environments and the results of the corresponding test cases?
> 
> May I ask which aarch64 hardware and host kernel versions you have actually tested the CubeSandbox use cases on — for example, the quickstart in the examples?
> 
> Could you help check whether this [PR](https://github.com/TencentCloud/CubeSandbox/pull/256/) is actually helpful for aarch64 support?
> 
> For instance, in this PR we are still missing (at least):
> 
> * An aarch64 mvm vmlinux builder
> * Documentation updates for developers and deployers
> 
> If the default page size on the arm machine is not 4KB, will Cloud Hypervisor (CH) still work properly?
> 
> Besides, being able to run CubeSandbox on Apple M-series chips would make cube and ARM much more popular.
Hi @lkml-likexu , thanks for reviewing this PR.
I tested it with Arm N2 and V2 platforms followed self-build-deploy.md, bare-metal-deploy.md. All works as expected.
Arm vmlinux (Image) builder and doc updated are added (782ce2154df588c74d7238ef725a95d8dd8f9b9f, 4a0039aa3bf364f4f513354c6b1b9f95a67baa51). Please review. Thanks.

lkml-likexu:
@JunHe77  To fix Hypervisor Tests (x86-64) regression,

Could you review this diff to verify if it introduces any regressions on ARM?
```
diff --git a/hypervisor/test_infra/src/lib.rs b/hypervisor/test_infra/src/lib.rs
index 49e0e17..e0878a3 100644
--- a/hypervisor/test_infra/src/lib.rs
+++ b/hypervisor/test_infra/src/lib.rs
@@ -15,7 +15,7 @@ use std::io::{Read, Write};
 use std::net::TcpListener;
 use std::net::TcpStream;
 use std::os::unix::fs::PermissionsExt;
-use std::os::unix::io::{AsRawFd, FromRawFd};
+use std::os::unix::io::{AsRawFd, FromRawFd, RawFd};
 use std::path::Path;
 use std::process::{Child, Command, ExitStatus, Output, Stdio};
 use std::str::FromStr;
@@ -755,6 +755,49 @@ pub fn kill_child(child: &mut Child) {
 
 pub const PIPE_SIZE: i32 = 32 << 20;
 
+// Smallest pipe size we are willing to fall back to (1 MiB). Below this the
+// child can block on a full buffer while emitting verbose logs, so we keep a
+// generous floor.
+const MIN_PIPE_SIZE: i32 = 1 << 20;
+
+/// Enlarge a pipe's kernel buffer on a best-effort basis.
+///
+/// `F_SETPIPE_SZ` rounds the request up and asks the kernel for a contiguous
+/// `pipe_buffer` array. For large sizes (e.g. 32 MiB => ~320 KiB, order-7) this
+/// is a high-order allocation that can fail with `ENOMEM` under memory
+/// fragmentation/pressure, which is common when `common_parallel` spawns many
+/// hypervisor children at once (each wanting a big stdout *and* stderr pipe).
+///
+/// Rather than aborting the whole test on a transient allocation failure, we
+/// retry with progressively smaller sizes down to `MIN_PIPE_SIZE`, returning
+/// the size the kernel actually granted. Only a failure at the floor is fatal.
+fn set_pipe_size_best_effort(fd: RawFd) -> io::Result<i32> {
+    let mut desired = PIPE_SIZE;
+    loop {
+        let actual = unsafe { libc::fcntl(fd, libc::F_SETPIPE_SZ, desired) };
+        if actual != -1 {
+            return Ok(actual);
+        }
+
+        let err = io::Error::last_os_error();
+        // Only back off for allocation-related failures; anything else (e.g.
+        // EPERM, EINVAL) won't be cured by a smaller size.
+        let retryable = matches!(
+            err.raw_os_error(),
+            Some(libc::ENOMEM) | Some(libc::EAGAIN)
+        );
+        if !retryable || desired <= MIN_PIPE_SIZE {
+            return Err(err);
+        }
+
+        let next = (desired / 2).max(MIN_PIPE_SIZE);
+        eprintln!(
+            "warning: F_SETPIPE_SZ({desired}) failed ({err}); retrying with {next} bytes"
+        );
+        desired = next;
+    }
+}
+
 static NEXT_VM_ID: Lazy<Mutex<u8>> = Lazy::new(|| Mutex::new(1));
 
 pub struct Guest {
@@ -1296,17 +1339,11 @@ impl<'a> GuestCommand<'a> {
                 .unwrap();
 
             let fd = child.stdout.as_ref().unwrap().as_raw_fd();
-            let pipesize = unsafe { libc::fcntl(fd, libc::F_SETPIPE_SZ, PIPE_SIZE) };
-            if pipesize == -1 {
-                return Err(io::Error::last_os_error());
-            }
+            let pipesize = set_pipe_size_best_effort(fd)?;
             let fd = child.stderr.as_ref().unwrap().as_raw_fd();
-            let pipesize1 = unsafe { libc::fcntl(fd, libc::F_SETPIPE_SZ, PIPE_SIZE) };
-            if pipesize1 == -1 {
-                return Err(io::Error::last_os_error());
-            }
+            let pipesize1 = set_pipe_size_best_effort(fd)?;
 
-            if pipesize >= PIPE_SIZE && pipesize1 >= PIPE_SIZE {
+            if pipesize >= MIN_PIPE_SIZE && pipesize1 >= MIN_PIPE_SIZE {
                 Ok(child)
             } else {
                 Err(std::io::Error::new(

```

deploy/one-click/build-release-bundle.sh
lkml-likexu: This issue still exists and affects users of the RPM platform when building packages:

```
[one-click] packaging one-click release bundle on host with prebuilt artifacts
/data/likexu/CubeSandbox/deploy/one-click/build-release-bundle.sh: line 31: dpkg: command not found
```

lkml-likexu:
@JunHe77 Pls note, the "deploy/one-click/build-release-bundle-builder.sh" does not work on oc9:
```
[one-click] packaging one-click release bundle on host with prebuilt artifacts
/data/likexu/CubeSandbox/deploy/one-click/build-release-bundle.sh: line 31: dpkg: command not found
```

You may need this fix:

```
diff --git a/deploy/one-click/build-release-bundle.sh b/deploy/one-click/build-release-bundle.sh
index 1b17f37..1ec440a 100755
--- a/deploy/one-click/build-release-bundle.sh
+++ b/deploy/one-click/build-release-bundle.sh
@@ -28,7 +28,7 @@ CUBE_PROXY_SOURCE_DIR="${ONE_CLICK_CUBE_PROXY_SOURCE_DIR:-${ROOT_DIR}/CubeProxy}
 CUBE_EGRESS_SOURCE_DIR="${ONE_CLICK_CUBE_EGRESS_SOURCE_DIR:-${ROOT_DIR}/CubeEgress}"
 WEB_SOURCE_DIR="${ONE_CLICK_WEB_SOURCE_DIR:-${ROOT_DIR}/web}"
 WEB_DIST_OVERRIDE="${ONE_CLICK_WEB_DIST_DIR:-}"
-MKCERT_BIN_ASSET="${ONE_CLICK_MKCERT_BIN:-${SCRIPT_DIR}/assets/bin/mkcert-v1.4.4-linux-$(dpkg --print-architecture)}"
+MKCERT_BIN_ASSET="${ONE_CLICK_MKCERT_BIN:-${SCRIPT_DIR}/assets/bin/mkcert-v1.4.4-linux-$(dpkg_arch)}"
 CUBE_KERNEL_VMLINUX="${ONE_CLICK_CUBE_KERNEL_VMLINUX:-${RAW_ARTIFACTS_DIR}/vmlinux}"
 KERNEL_ARTIFACT_ZIP="${WORK_ROOT}/cube-kernel-scf.zip"
 
diff --git a/deploy/one-click/lib/common.sh b/deploy/one-click/lib/common.sh
index 15b9ec1..902ff72 100755
--- a/deploy/one-click/lib/common.sh
+++ b/deploy/one-click/lib/common.sh
@@ -40,6 +40,20 @@ require_cmd() {
   command -v "${cmd}" >/dev/null 2>&1 || die "required command not found: ${cmd}"
 }
 
+# Print the Debian/dpkg-style architecture name (amd64/arm64) without depending
+# on `dpkg`, which is absent on RPM-based hosts (OpenCloudOS, RHEL, ...). Bundle
+# assets (e.g. mkcert-v1.4.4-linux-amd64) follow the dpkg naming convention, so
+# we map the kernel machine name from `uname -m` to it.
+dpkg_arch() {
+  local machine
+  machine="$(uname -m)"
+  case "${machine}" in
+    x86_64 | amd64) printf 'amd64\n' ;;
+    aarch64 | arm64) printf 'arm64\n' ;;
+    *) die "unsupported architecture: ${machine}" ;;
+  esac
+}
+
 _version_trim_leading_zeroes() {
   local value="$1"
   value="${value#${value%%[!0]*}}"

```

scripts/build-kernel.sh
lkml-likexu, line 1: @JunHe77  For arm64, we need arch/arm64/boot/Image, otherwise:

```
cubemastercli run fail: template tpl-326ea3ded4de4273b9fb3d96 creation failed: failed to run container tpl-326ea3ded4de4273b9fb3d96_0: failed to create shim task: Others("Other: Create sandbox failed:Boot vm failed:Cannot load the UEFI binary in memory: UefiTooBig, status: launched")
```

CubeNet/cubevs/localgw_x86_bpfel.o
JunHe77: Thanks for pointing this out. Done.

JunHe77:
Updated kernel config for aarch64 (cd6ca7efa81c5cd7570cf4e1944df7955d5e5b37), fixed dpkg issue on fedora-like system (84b3ca3b38f57d7831e6f6a492f6d77ec6898cc2).

docker/Dockerfile.builder
lkml-likexu: @JunHe77 This issue still stand.

deploy/guest-image/Dockerfile
```diff
@@ -1,4 +1,4 @@
-FROM cube-sandbox-image.tencentcloudcr.com/opensource/tencentos4-minimal:4.4-v20250331
+FROM tencentos/tencentos4-minimal:4.4-v20250331
```
lkml-likexu: Let's try "tencentos/tencentos4-minimal:latest".

deploy/one-click/env.example
```diff
@@ -73,6 +73,27 @@ ONE_CLICK_LOG_DIR=/var/log/cube-sandbox-one-click
...
 CUBE_SANDBOX_SYSTEMD_RUNTIME_DIR=/run/cube-sandbox-systemd
 
+# ---- Container image overrides ----
+# Canonical override names for the upstream container images used by the
+# bundled dependency services. These are consumed by deploy/one-click/scripts:
+#   COREDNS_IMAGE  -> scripts/one-click/up-dns.sh + scripts/systemd/coredns-start.sh
+#   MYSQL_IMAGE    -> scripts/one-click/up-support.sh
+#   REDIS_IMAGE    -> scripts/one-click/up-support.sh
+#   WEB_UI_IMAGE   -> scripts/one-click/up-webui.sh
+#   CUBE_PROXY_BASE_IMAGE -> scripts/one-click/up-cube-proxy.sh (FROM base image
```
lkml-likexu, line 80: Variable names should be consistent to avoid misleading users:

```
+# Canonical override variables for the upstream container images used by the
+# bundled dependency services. Set THESE names (left column); the consuming
+# scripts read them and fall back to the defaults below:
+#   CUBE_PROXY_COREDNS_IMAGE -> scripts/one-click/up-dns.sh + scripts/systemd/coredns-start.sh
+#   CUBE_SANDBOX_MYSQL_IMAGE -> scripts/one-click/up-support.sh
+#   CUBE_SANDBOX_REDIS_IMAGE -> scripts/one-click/up-support.sh
+#   WEB_UI_IMAGE             -> scripts/one-click/up-webui.sh
+#   CUBE_PROXY_BASE_IMAGE    -> scripts/one-click/up-cube-proxy.sh (FROM base image
+#                              that CubeProxy/Dockerfile is built on; passed as the
+#                              CUBE_PROXY_BASE_IMAGE docker build arg)
```

lkml-likexu:
One more thing, we may need docs/blog/posts/2026-07-01-cubesandbox-perf-benchmark-arm64.md

deploy/one-click/env.example
JunHe77: Updated.

configs/kernel-oc9.x86_64.config
JunHe77: Updated.

scripts/build-kernel.sh
JunHe77: Now both x86_64 and aarch64 use `make guest-kernel` to build image, and target image is stored as _output/kernel/$(uname -m)/vmlinux

deploy/guest-image/Dockerfile
JunHe77: Updated.

agent/.cargo/config.toml
JunHe77: Updated.

configs/kernel-oc9.aarch64.config
JunHe77: Re-generated.

deploy/one-click/build-release-bundle.sh
JunHe77: Updated.

dev-env/run_vm.sh
JunHe77: Updated.

dev-env/run_vm.sh
JunHe77: Updated.

docker/Dockerfile.builder
JunHe77: Updated.

chenhengqi:
The subject line for the 4th commit is wrong:
```
cubevs: remove arch specific toolchain dependency
```
Should be `cubecow`.

JunHe77:
Good catch. Updated. Thanks.

agent/src/rpc.rs
```diff
@@ -1829,15 +1831,58 @@ pub fn start(s: Arc<Mutex<Sandbox>>, server_address: &str) -> Result<TtrpcServer
...
+
+        unsafe {
+            ioport.write(data);
+        }
     }
-    let mut ioport = x86_64::instructions::port::Port::new(port);
 
-    unsafe {
-        ioport.write(data);
+    #[cfg(target_arch = "aarch64")]
```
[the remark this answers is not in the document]
lkml-likexu: I will take care of this later.

agent/src/rpc.rs
```diff
@@ -1829,15 +1831,58 @@ pub fn start(s: Arc<Mutex<Sandbox>>, server_address: &str) -> Result<TtrpcServer
...
+    {
+        const SYS_CTRL_MMIO_ADDR: libc::off_t = 0x0903_0000;
```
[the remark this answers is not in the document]
lkml-likexu: I will take care of this later.

fslongjin (approved):

chenhengqi:
Merged, thanks.

lkml-likexu:
@JunHe77 Any status update, especially for guest kernel config ?

PR #513, opened 2026-06-10
feat(CubeAPI): add container image build for the cube-api service

lkml-likexu:
The cube-api server had no packaging story, so deploying it required building and shipping the binary by hand and made it hard to run the service consistently across environments.

Add a container build for cube-api that produces a small, self-contained runtime image suitable for one-click and orchestrated deployments. The image follows the project's security and operability expectations so it can be rolled out without extra manual setup, while keeping the build context lean.

### DIFF
--- a/CubeAPI/.dockerignore
+++ b/CubeAPI/.dockerignore
@@ -1,0 +1,15 @@
+# Build artifacts — never needed in the image build context.
+target/
+
+# VCS / editor noise
+.git/
+.gitignore
+*.swp
+
+# Local env / secrets must not leak into the build context
+.env
+.env.*
+
+# Docker files themselves
+Dockerfile
+.dockerignore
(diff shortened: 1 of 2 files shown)


lkml-likexu:
Round-1 Fiexed:

What changed and how each item maps:

- (1)/(6) Missing build.rs — added COPY build.rs ./ between the manifest and source copies.
- (2) Missing --locked — both cargo build invocations now use --release --locked.
- (3) musl mismatch — resolved by going the explicit-musl route: rustup target add x86_64-unknown-linux-musl + --target x86_64-unknown-linux-musl, so the installed musl-dev headers are actually linked (and the runtime copy path was updated to target/x86_64-unknown-linux-musl/release/cube-api).
- (4) Caching — adopted the two-phase dummy-main pattern, plus --mount=type=cache,target=/usr/local/cargo/registry for BuildKit crate caching on both build steps.
- (5) Redundant || exit 1 — simplified the HEALTHCHECK to CMD curl -fsS http://127.0.0.1:3000/health.

fslongjin (approved):

PR #514, opened 2026-06-10
feat(deploy): support external MySQL/Redis in one-click installer

lkml-likexu:
The one-click installer always provisioned MySQL and Redis as bundled local containers, which forced every deployment to run its own database and cache even when a managed or shared instance was already available.

Add an opt-in path so operators can point CubeSandbox at an existing MySQL and/or Redis server via .env. When configured, the installer wires every service and helper to the external endpoint and stops managing the corresponding local container, while deployments that leave it unset keep the existing bundled behavior unchanged.

This makes one-click suitable for environments with centrally operated data stores and reduces redundant local resource usage. Documentation (en + zh) and env.example are updated to describe the new variables.

Note: this feature is also required for standalone deployment.

### DIFF
--- a/deploy/one-click/README.md
+++ b/deploy/one-click/README.md
@@ -250,6 +250,38 @@ During installation, runtime files are prepared in this directory and the follow
 - `mysql:8.0`
 - `redis:7-alpine`
 
+### Using an external MySQL / Redis
+
+To point CubeSandbox at an existing MySQL/Redis server instead of the bundled
+local containers, set the following in `.env` before running `install.sh`
+(see `env.example`):
+
+```bash
+# External MySQL (any subset of the credential fields may be overridden)
+CUBE_EXTERNAL_MYSQL_HOST=10.0.0.20
+CUBE_EXTERNAL_MYSQL_PORT=3306
+CUBE_EXTERNAL_MYSQL_USER=cube
+CUBE_EXTERNAL_MYSQL_PASSWORD=cube_pass
+CUBE_EXTERNAL_MYSQL_DB=cube_mvp
+
+# External Redis
+CUBE_EXTERNAL_REDIS_HOST=10.0.0.21
+CUBE_EXTERNAL_REDIS_PORT=6379
+CUBE_EXTERNAL_REDIS_PASSWORD=ceuhvu123
+```
+
+When `CUBE_EXTERNAL_MYSQL_HOST` (and/or `CUBE_EXTERNAL_REDIS_HOST`) is set, `install.sh`:
+
+- patches `CubeMaster/conf.yaml` with the external MySQL/Redis endpoint;
+- writes `DATABASE_URL` (CubeAPI) and `CUBE_PROXY_REDIS_*` (cube proxy) to `.one-click.env` so every service consumes the external endpoint;
+- masks the corresponding `cube-sandbox-mysql.service` / `cube-sandbox-redis.service` so the local container is never started; and
+- makes `quickcheck.sh` and `up-support.sh` skip lifecycle management of the now-external dependency. (`down-support.sh` has no external-dep awareness and still issues a `docker compose down`, but this is a harmless no-op because the local containers were never started for the external dependency.)
+
+The external MySQL must already grant the configured user access to the target
+database; CubeMaster runs its own schema migrations on first start, and the
+single-node seed rows are applied through a host-side `mysql` client (install
+the MySQL client package on the control node when seeding an external DB).
+
 `cube proxy` and its DNS resolution are mandatory capabilities in one-click. The following two values in `.env` must remain `1`:
 
 ```bash
--- a/deploy/one-click/README_zh.md
+++ b/deploy/one-click/README_zh.md
@@ -239,6 +239,34 @@ MySQL/Redis 依赖默认会部署到：
 - `mysql:8.0`
 - `redis:7-alpine`
 
+### 使用外部 MySQL / Redis
+
+如果希望使用已有的 MySQL/Redis 服务器，而不是内置的本地容器，可在执行
+`install.sh` 之前在 `.env` 中设置以下变量（参见 `env.example`）：
+
+```bash
+# 外部 MySQL（凭据字段可按需覆盖）
+CUBE_EXTERNAL_MYSQL_HOST=10.0.0.20
+CUBE_EXTERNAL_MYSQL_PORT=3306
+CUBE_EXTERNAL_MYSQL_USER=cube
+CUBE_EXTERNAL_MYSQL_PASSWORD=cube_pass
+CUBE_EXTERNAL_MYSQL_DB=cube_mvp
+
+# 外部 Redis
+CUBE_EXTERNAL_REDIS_HOST=10.0.0.21
+CUBE_EXTERNAL_REDIS_PORT=6379
+CUBE_EXTERNAL_REDIS_PASSWORD=ceuhvu123
+```
+
+当设置了 `CUBE_EXTERNAL_MYSQL_HOST`（和/或 `CUBE_EXTERNAL_REDIS_HOST`）时，`install.sh` 会：
+
+- 用外部 MySQL/Redis 地址改写 `CubeMaster/conf.yaml`；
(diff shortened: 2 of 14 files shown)


lkml-likexu:
Round-1 Fixed:

All five issues are fixed and the scripts pass syntax checks. Here's a summary of what I changed.

- (1) HIGH — Plaintext MySQL password on command line (up-with-deps.sh)
Replaced "-p${CUBE_EXTERNAL_MYSQL_PASSWORD}" with a temporary 0600 defaults file passed via --defaults-extra-file, so the password is never visible in /proc/<pid>/cmdline. The temp file is cleaned up via an EXIT trap (and removed immediately after use).

```
up-with-deps.sh
Lines 68-90
  MYSQL_CNF="$(mktemp)"
  trap 'rm -f "${MYSQL_CNF}"' EXIT
  chmod 600 "${MYSQL_CNF}"
  cat > "${MYSQL_CNF}" <<EOF
[client]
password="${CUBE_EXTERNAL_MYSQL_PASSWORD}"
EOF
  sed ... | mysql --defaults-extra-file="${MYSQL_CNF}" ...
```

- (2) MEDIUM — Weak default credentials silently accepted (install.sh)
Added warn_default_external_credentials(), called during preflight, which logs a warning when an external MySQL/Redis host is configured but the password is still the published default (cube_pass / ceuhvu123).

- (3) MEDIUM — Unsanitized values in sed substitutions (install.sh)
Added a shared escape_sed() helper to lib/common.sh (escapes |, \, & for the | delimiter, mirroring up-support.sh). patch_cubemaster_external_deps() now escapes all user-supplied values before substitution, and ensure_file "${cfg}" was hoisted out of both branches to run once (early-returning when neither external host is set).

- (4) MEDIUM — Credentials in world-readable env file (install.sh)
Added chmod 600 "${RUNTIME_ENV_FILE}" immediately after the file is created, before any secrets are written. (upsert_env_kv uses mktemp + mv, which preserves 0600, so subsequent writes stay restricted.)

- (5) SUGGESTION — Early connectivity check (install.sh)
Added check_external_deps_preflight() (run right after check_install_preflight) that does a best-effort mysqladmin ping (via a 0600 defaults file) and redis-cli ping with a short, configurable timeout (ONE_CLICK_EXTERNAL_DEP_TIMEOUT, default 5s). It fails fast with a clear message on misconfiguration, and skips gracefully if the client binary isn't installed.

Both bash -n syntax checks pass.

One note: in the MySQL defaults files I quoted the password value (password="...") rather than leaving it bare as in the review snippet, since quoting correctly handles passwords containing spaces or #.

lkml-likexu:
Round-2 Fixed:

(1) DATABASE_URL built without URL-encoding (install.sh)
Added a byte-wise, C-locale urlencode() helper to lib/common.sh (encodes everything except RFC 3986 unreserved chars) and applied it to the user/password before assembling the connection string:

```
install.sh
Lines 839-846
  upsert_env_kv "${RUNTIME_ENV_FILE}" "CUBE_EXTERNAL_MYSQL_DB" "${CUBE_EXTERNAL_MYSQL_DB}"
  # Percent-encode the userinfo components so passwords/users containing URL
  # metacharacters (@, :, /, %, ...) cannot corrupt the connection string.
  database_url_user="$(urlencode "${CUBE_EXTERNAL_MYSQL_USER}")"
  database_url_pass="$(urlencode "${CUBE_EXTERNAL_MYSQL_PASSWORD}")"
  upsert_env_kv "${RUNTIME_ENV_FILE}" "DATABASE_URL" "mysql://${database_url_user}:${database_url_pass}@${CUBE_EXTERNAL_MYSQL_HOST}:${CUBE_EXTERNAL_MYSQL_PORT}/${CUBE_EXTERNAL_MYSQL_DB}"
fi
```
Verified: a password cube@123 now becomes cube%40123, producing mysql://cube:cube%40123@host:3306/db instead of the malformed URL.

(2) escape_sed doesn't handle newlines (lib/common.sh)
escape_sed now strips \n/\r (via tr -d '\n\r') before escaping the |, \, and & characters, neutralizing any sed-command injection an embedded newline could otherwise cause:

```
common.sh
Lines 108-110
escape_sed() {
  printf '%s' "$1" | tr -d '\n\r' | sed 's/[|\\&]/\\&/g'
}
```
Verified: escape_sed $'ab\ncd|ef' now yields abcd\|ef (newline removed, delimiter escaped).

lkml-likexu:
Round-3 Fixed:

All five issues are fixed. Here's a summary of the changes:

- deploy/one-click/install.sh: Redis connect timeout — Replaced -t "${connect_timeout}" with --connect-timeout "${connect_timeout}" in the Redis preflight so an unreachable host fails fast instead of falling back to TCP's ~127s connect timeout.

- g flag removal — Dropped the g flag from all six external-dep sed expressions (4 MySQL, 2 Redis) and added a comment noting each line has exactly one match and the distinct key prefixes (addr:/user:/pwd:/db_name:/nodes:/password:) keep replacements bound to the correct field.

- Temp file hardening — Added umask 077 around the mktemp for the MySQL credentials file (closing the create→chmod 600 race window) and a trap 'rm -f ...' EXIT handler (cleared with trap - EXIT on both success and die paths) so the plaintext password is cleaned up even on abrupt termination — mirroring up-with-deps.sh.

- DATABASE_URL encoding — Now urlencodes host, port, and database name in addition to user/password, so URI-special characters in any component can't corrupt the connection string.

deploy/one-click/README.md

- Docs correction — The claim now states only quickcheck.sh and up-support.sh skip lifecycle management, and clarifies that down-support.sh lacks external-dep awareness but its docker compose down is a harmless no-op since the local containers were never started.

lkml-likexu:
Round-4 Fixed:

- install.sh (sed patterns coupled to defaults) — Both the MySQL and Redis sed blocks now match only on the key prefix (addr: ".*", user: ".*", pwd: ".*", db_name: ".*", nodes: ".*", password: ".*"). They no longer break if conf.yaml defaults are regenerated. I verified against the actual conf.yaml that these keys are partitioned cleanly between the MySQL sections (ossdb_config/instance_db_config) and the redis* sections, so no cross-field clobbering occurs, and the comments document this.

- README_zh.md (inaccurate down-* claim) — Replaced the incorrect statement with the nuance from the English README: only quickcheck.sh and up-support.sh skip external-dep lifecycle management, while down-support.sh still issues a harmless no-op docker compose down.

- escape_sed doesn't escape " — Extended the lib/common.sh helper to also escape " (s/[|\\&"]/\\&/g), with an updated comment explaining it's needed because the only caller embeds the result in double-quoted sed replacement strings. This helper is only used by install.sh's patch_cubemaster_external_deps (the other scripts define their own /-delimiter variants), so the change is safe.

lkml-likexu:
Round-5 Fixed:

#### (1) deploy/one-click/scripts/one-click/up-with-deps.sh — sed injection via CUBE_SANDBOX_NODE_IP
The old ${CUBE_SANDBOX_NODE_IP//\//\\/} only escaped the / delimiter, leaving & and \ unescaped. Note that the common.sh sourced by this script (scripts/one-click/common.sh) does not define escape_sed() — that helper only exists in lib/common.sh / scripts/systemd/common.sh, and sibling runtime scripts (up-support.sh, up-cube-proxy.sh) each define their own local copy. I followed that established local-definition pattern:

Added a local escape_sed() (printf '%s' "$1" | sed 's/[\/&]/\\&/g') which escapes \, /, and &.
Computed CUBE_SANDBOX_NODE_IP_SED once after the value is validated, and used it in both the external-MySQL sed | mysql path and the internal sed | docker exec path.

####  (2) deploy/one-click/install.sh — Redis preflight didn't verify auth
redis-cli PING returns PONG even with a wrong/extraneous password when the server has no requirepass, so misconfigured credentials passed silently. Now, when a password is configured, the check issues AUTH and requires an OK reply (which also confirms reachability); when no password is set it falls back to PING. The password is passed via stdin (-x AUTH) with --no-auth-warning, keeping it off /proc/<pid>/cmdline and stderr — consistent with the security-conscious MySQL handling above it.

fslongjin (approved):

PR #515, opened 2026-06-10
fix(cubelet): register chi image_forward_latency histogram

chenggui53:
The chi_cubebox_image_forward_latency_milliseconds histogram was created via prometheus.NewHistogram but never added to the docker/go-metrics namespace, so metrics.Register(ns) never exposed it on /metrics. Even when image forwards occurred and Observe() was called, the metric was invisible to scraping.

Add the missing ns.Add(imageForwardLatency), matching the pattern in internal/cube/server/images/metrics.go.

### DIFF
--- a/Cubelet/plugins/chi/chics/metrics.go
+++ b/Cubelet/plugins/chi/chics/metrics.go
@@ -38,6 +38,7 @@ func init() {
 			Buckets:   prometheus.ExponentialBuckets(1, 10, 8),
 		},
 	)
+	ns.Add(imageForwardLatency)
 	metrics.Register(ns)
 }
 


fslongjin (approved):
LGTM. Single-line fix to register the image_forward_latency histogram that was created but never added to the metrics namespace. Without this, the metric would be silently invisible on /metrics despite Observe() being called.

Verified:
- The histogram is created via prometheus.NewHistogram (bare creation, unlike the counter/gauge which use ns.NewLabeledCounter/ns.NewGauge that auto-register)
- The fix follows the pattern used in internal/cube/server/images/metrics.go
- No other consumers of this metric in the repo — it's exposed for external Prometheus scraping

Approved.

PR #516, opened 2026-06-10
fix(one-click): extend sandbox CIDR preflight to default config and host resolvers

zyl1121:
## Motivation

While troubleshooting a local network issue, I found that the effective sandbox CIDR can still conflict with the host network even when `ip route` does not make the conflict obvious.

In particular, the default Cubelet sandbox CIDR can overlap with host DNS resolver addresses such as `192.168.0.1`. When that happens, host DNS traffic may be routed through the sandbox network instead of the real LAN path, breaking host-side DNS resolution and causing operations such as image pulls to fail.

CubeSandbox `v0.3.1` already added preflight validation for `CUBE_SANDBOX_NETWORK_CIDR` when users explicitly override the CIDR. This leaves the default-CIDR path uncovered, because the installer previously skipped conflict validation when `CUBE_SANDBOX_NETWORK_CIDR` was not set.

This PR extends that protection to the remaining gap:

- validate the effective sandbox CIDR even when the user does not set `CUBE_SANDBOX_NETWORK_CIDR`
- detect conflicts not only from host interfaces and routes, but also from host resolver nameservers

## What Changed

- extract the packaged one-click bundle earlier so `install.sh` can read the packaged Cubelet `config.toml` and fail during preflight before stopping or replacing the existing deployment
- when `CUBE_SANDBOX_NETWORK_CIDR` is unset, read the default CIDR from the packaged Cubelet network plugin config and run the same preflight validation instead of skipping it
- extend CIDR conflict detection to check common host resolver sources:
  - `/run/systemd/resolve/resolv.conf`
  - `/run/systemd/resolve/stub-resolv.conf`
  - `/run/NetworkManager/no-stub-resolv.conf`
  - `/var/run/NetworkManager/no-stub-resolv.conf`
  - `/run/resolvconf/resolv.conf`
  - `/etc/resolvconf/run/resolv.conf`
  - `/etc/resolv.conf`
- avoid reporting known Cube-owned leftover devices such as `cube-dev` and `z<ip>` as host LAN conflicts during reinstall / recovery paths

## Design Note

This PR currently excludes known Cube-owned leftover virtual devices such as `cube-dev` and `z<ip>` from the host conflict check.

This is intended to keep reinstall / recovery paths usable when stale Cube-owned devices are present, while still detecting conflicts with the real host LAN, routes, and resolver nameservers.

If you prefer to keep explicit manual cleanup for old Cube virtual devices, I can drop this part and keep the PR focused only on:

- validating the packaged default CIDR when no override is set
- checking resolver nameservers in addition to interfaces / routes
- running that validation before the installer stops the existing deployment

## Validation

Real machine validation:

- confirmed that `install.sh` now fails before stopping the existing deployment when the packaged default CIDR `192.168.0.0/18` overlaps with host resolver nameservers such as `192.168.0.1`
- confirmed that the installer continues past preflight when the effective sandbox CIDR does not overlap with host interfaces, routes, or resolver nameservers
- confirmed that resolver overlap is now rejected during preflight instead of surfacing later as host-side DNS failures, such as image pull failures on the host machine
- confirmed the conflict is reported from the real resolver files, for example:
  - `/run/NetworkManager/no-stub-resolv.conf`
  - `/var/run/NetworkManager/no-stub-resolv.conf`
  - `/etc/resolv.conf`

Focused shell test:

```bash
bash deploy/one-click/tests/test_cidr_preflight.sh
```

This test covers:

- host interface overlap
- host route overlap
- resolver nameserver overlap
- avoiding false positives from known Cube-owned leftover devices
- reading the default CIDR from packaged Cubelet config
- ensuring preflight runs before the script stops the existing deployment

## Scope

This PR is limited to one-click CIDR preflight behavior. It does not change the CIDR format validation, mask validation, or alignment checks.

The existing `CUBE_SANDBOX_NETWORK_CIDR_SKIP_CONFLICT_CHECK` bypass remains unchanged.

### DIFF
--- a/deploy/one-click/install.sh
+++ b/deploy/one-click/install.sh
@@ -544,11 +544,29 @@ else
   log "primary network interface not detected; keeping packaged Cubelet eth_name"
 fi
 
-# Validate cubevs CIDR from env var (if set)
+PACKAGE_TAR="${ONE_CLICK_PACKAGE_TAR:-${SCRIPT_DIR}/assets/package/sandbox-package.tar.gz}"
+WORK_DIR="$(mktemp -d)"
+trap 'rm -rf "${WORK_DIR}"' EXIT
+
+require_cmd tar
+ensure_file "${PACKAGE_TAR}"
+
+log "extracting package ${PACKAGE_TAR}"
+tar -xzf "${PACKAGE_TAR}" -C "${WORK_DIR}"
+PKG_ROOT="${WORK_DIR}/sandbox-package"
+ensure_dir "${PKG_ROOT}"
+
+# Validate the effective cubevs CIDR before installing packages or replacing
+# the existing deployment. When the env var is unset, use the packaged Cubelet
+# config value so the default CIDR is checked too.
+CUBELET_PACKAGE_CONFIG="${PKG_ROOT}/Cubelet/config/config.toml"
 CUBE_SANDBOX_NETWORK_CIDR="${CUBE_SANDBOX_NETWORK_CIDR:-}"
 if [[ -n "${CUBE_SANDBOX_NETWORK_CIDR}" ]]; then
-  check_cidr_preflight "${CUBE_SANDBOX_NETWORK_CIDR}"
+  check_cidr_preflight "${CUBE_SANDBOX_NETWORK_CIDR}" "CUBE_SANDBOX_NETWORK_CIDR"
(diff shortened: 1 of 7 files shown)


zyl1121:
Closing this as superseded by #586, which already includes the resolver and default-CIDR coverage from this PR.

PR #518, opened 2026-06-11
Introduce security proxy

chenhengqi:
Add CubeEgress, an OpenResty-based egress gateway that intercepts sandbox outbound traffic via TPROXY and provides:

- Credential injection: per-sandbox secrets are attached to outbound requests at the proxy, so user code never handles raw credentials.
- Domain filtering: policy-driven allow/deny lists gate which destinations a sandbox may reach.
- Access auditing: structured logs of every egress request, with optional body redaction, for downstream review.

Wire CubeMaster, network-agent and Cubelet to provision per-template CA material, push policy, and route sandbox traffic through the proxy. Ship systemd units, deployment scripts and user-facing docs.


network-agent/internal/cubeegress/client_test.go
```diff
@@ -0,0 +1,215 @@
+// Copyright (c) 2024 Tencent Inc.
```
fslongjin: 2024-》2026

PR #526, opened 2026-06-11
fix(cubelet): add DNS server IP to AllowOut when any DNS rule exists

FakeLearne:


fslongjin:
Hi @FakeLearne, thanks for the contribution. However, this change removes the check that was intentionally added in #490 (fix: do not auto-allow DNS when internet is disabled).

IP allowlisting should be **explicit**, not implicit. Automatically appending DNS server IPs to `AllowOut` when any DNS rule exists — even when internet access is explicitly disabled — creates a security risk: it silently opens a network path that the user did not explicitly authorize, bypassing the intent of `AllowInternetAccess: false`.

PR #490 explicitly addressed this security concern by blocking implicit DNS allow-out when internet is disabled. This PR would revert that fix.

Because of this, I'm closing this PR. Feel free to reopen if you'd like to discuss the approach further.

fslongjin (approved):
LGTM

PR #534, opened 2026-06-12
feat(pvm): rename PVM kernel LOCALVERSION to a descriptive scheme

lkml-likexu:
The previous LOCALVERSION carried a redundant distro segment that made the resulting kernel version string longer and harder to read than necessary. Simplify it to a cleaner, more descriptive scheme that still conveys the target OS and the PVM host/guest role, so users can identify the running kernel at a glance.

Update the deploy configs together with the user-facing guides and blog references so the documented version string stays consistent with what the build actually produces, avoiding confusion during install and verification.

Requested-by: qingqlin <qingqlin@tencent.com>


docs/zh/blog/posts/2026-06-03-cubesandbox-perf-benchmark-pvm.md
fslongjin, line 1: do not change this blog

fslongjin (approved):

PR #543, opened 2026-06-12
fix(network): ignore allow_public_traffic for egress

FakeLearne:


sdk/python/cubesandbox/sandbox.py
```diff
@@ -134,14 +134,9 @@ def create(
...
                 network.get("deny_out"),
-                default_deny_all=(
-                    not allow_internet_access
-                    or network.get("allow_public_traffic") is False
-                ),
+                default_deny_all=not allow_internet_access,
             )
             net: dict = {}
-            if "allow_public_traffic" in network:
-                net["allowPublicTraffic"] = network["allow_public_traffic"]
```
chenhengqi, line 144: Let's keep it as it is. Just ignore it on server side.

sdk/python/tests/test_sandbox.py
```diff
@@ -141,11 +141,11 @@ def test_create_allow_internet_access_true_not_in_payload(self):
         body = m.call_args.kwargs["json"]
         assert "allow_internet_access" not in body
 
-    def test_create_network_allow_public_traffic(self):
+    def test_create_network_ignores_allow_public_traffic(self):
```
chenhengqi: We can drop all these `allow_public_traffic` test cases.

PR #546, opened 2026-06-13
feat(template): expose latest job id in template list across clients

xiaojunxiang2023:
## Summary

Template list/detail responses now include each template's **latest create/rebuild job id** (`job_id` / `jobID`). This lets operators look up build status without guessing ids after `tpl ls`, Web UI browsing, or SDK listing.

Fixes the gap where `cubemastercli tpl ls` showed building template ids but not the corresponding job id needed for `tpl status --job-id` / `tpl watch --job-id`.

## Changes

| Layer | Field | Behavior |
|-------|-------|----------|
| CubeMaster | `job_id` | Populated from the latest `TemplateImageJob` per template |
| cubemastercli | `JOB_ID` column | Added to `tpl ls` / `tpl ls -o wide`; `tpl info` prints `job_id` when present |
| CubeAPI | `jobID` | Added to `GET /templates` and `GET /templates/:id` |
| Web UI | `jobID` | Shown on template cards; detail page uses it to auto-poll in-progress builds |
| Python SDK | `TemplateInfo.job_id` | Parsed from list/detail responses |
| Go SDK | `TemplateInfo.JobID` | Parsed from list/detail responses |

**Semantics:** returns the **latest** job id for the template (including `READY` / `FAILED`), not only in-progress jobs. Templates with no job history (e.g. legacy entries) omit the field.

## Example

```bash
cubemastercli tpl ls
# TEMPLATE_ID  STATUS   JOB_ID              CREATED_AT  IMAGE_INFO
# tpl-abc...   RUNNING  job-xyz...          ...         nginx:latest@sha256:...

cubemastercli tpl status --job-id job-xyz...
cubemastercli tpl watch --job-id job-xyz...
```

## Deploy and Verify
image
image


fslongjin:
Thanks for this proposal @xiaojunxiang2023 — exposing the latest job ID in template list responses is a great idea. It closes a real UX gap where operators had to guess the right `--job-id` after running `tpl ls`.

However, it looks like this PR now has conflicts with `master`, likely because #580 (merged June 22) introduced a TUI-based interface for `tpl ls` / `tpl watch` / `tpl status`, which significantly changed `CubeMaster/cmd/cubemastercli/commands/cubebox/template.go` — the same file your CLI changes touch.

Could you please rebase this branch onto the latest `master` and resolve the conflicts? In particular, the `tpl ls` output is now rendered through the TUI rather than the old `tabwriter` table, so the `JOB_ID` column integration there may need to be rethought to fit the new TUI layout. (The API, SDK, and Web UI parts of this PR should be unaffected and are still very welcome!)

Happy to take another look once it's rebased. Thanks for the contribution! 🙏

xiaojunxiang2023:
> Thanks for this proposal @xiaojunxiang2023 — exposing the latest job ID in template list responses is a great idea. It closes a real UX gap where operators had to guess the right `--job-id` after running `tpl ls`.
> 
> However, it looks like this PR now has conflicts with `master`, likely because #580 (merged June 22) introduced a TUI-based interface for `tpl ls` / `tpl watch` / `tpl status`, which significantly changed `CubeMaster/cmd/cubemastercli/commands/cubebox/template.go` — the same file your CLI changes touch.
> 
> Could you please rebase this branch onto the latest `master` and resolve the conflicts? In particular, the `tpl ls` output is now rendered through the TUI rather than the old `tabwriter` table, so the `JOB_ID` column integration there may need to be rethought to fit the new TUI layout. (The API, SDK, and Web UI parts of this PR should be unaffected and are still very welcome!)
> 
> Happy to take another look once it's rebased. Thanks for the contribution! 🙏

Thanks for the review! I've rebased onto the latest master and resolved the conflicts in template.go, httpservice/cube/template.go, and store.go.

For tpl ls, I kept the JOB_ID column in the existing tabwriter output — on current master, the TUI changes from #580 apply to watch/status, while ls still uses the table layout, so this should fit without extra rework. API, SDK, and Web UI parts are unchanged.

Ready for another look whenever you have time.

CubeAPI/src/models/mod.rs
```diff
@@ -556,6 +556,9 @@ pub struct TemplateSummary {
     pub created_at: Option<String>,
     #[serde(rename = "imageInfo", skip_serializing_if = "Option::is_none")]
     pub image_info: Option<String>,
+    /// Latest create/rebuild job id for the template.
+    #[serde(rename = "jobID", skip_serializing_if = "Option::is_none")]
+    pub job_id: Option<String>,
```
fslongjin, line 561: why `jobID` here and `job_id` above?

xiaojunxiang2023: @fslongjin Hi~, because job_id in cubemaster/mod.rs is the CubeMaster client DTO — it deserializes CubeMaster's snake_case job_id.
jobID in models/mod.rs is the public CubeAPI JSON field, same pattern as templateID / imageInfo / lastError.
Rust field stays job_id in both layers; only the external wire format uses camelCase.

PR #554, opened 2026-06-13
feat(sdk): inject Sandbox.create env_vars into the guest via envd /init

xiaojunxiang2023:
## Summary

`Sandbox.create(env_vars=...)` values were accepted by the control plane but never
reached the guest runtime, so `commands.run` / `run_code` could not see them.

This PR is a follow-up to #484. It does **not** use `/etc/profile.d`, rootfs
injection, bind-mount, or Cubelet guest-IP routing (all rejected in #484).

After create succeeds, the **Python and Go SDKs** call envd's native `POST /init`
over the **same CubeProxy data-plane channel** that `commands.run` already uses.
envd stores the variables as global defaults and merges them on every
`Process/Start`:

```
template env  <  create env (/init)  <  per-command env
```

**Scope:** SDK only (`sdk/python`, `sdk/go`). No CubeAPI / CubeMaster / Cubelet / envd changes.

---

## Response to #484 feedback

| #484 concern | This PR |
|--------------|---------|
| `/etc/profile.d` / rootfs writes | Not used |
| Secrets persisted to disk or template snapshots | Not used |
| Cubelet → guest IP injection | Not used |
| Runtime injection via envd | `POST /init` (E2B-compatible) |
| Merge at process execution time | envd default-env merge on `Process/Start` |

---

## Why SDK-only (not CubeAPI / Cubelet)

External clients can only reach **CubeAPI** and **CubeProxy**.

The data-plane path to envd is already used by `commands.run`:

```
SDK  →  CubeProxy (proxy_node_ip : proxy_port)
         Host: <envd-port>-<sandbox-id>.<domain>
      →  envd
```

Calling `/init` on that same path needs no new deployment config (reuse
`CUBE_PROXY_NODE_IP` / `proxy_node_ip`). It avoids broken server-side paths such as
CubeAPI calling `http://127.0.0.1/init`, or Cubelet dialing guest IP directly.

---

## Deploy and Verify
```bash
PYTHONPATH=/home/xjx/fix_env/CubeSandbox/sdk/python python3 env_vars_verify.py
```

env_vars_verify.py:
```python
from cubesandbox import Config, Sandbox

CFG = Config(
    api_url="http://10.69.66.75:8301",
    proxy_node_ip="10.69.66.75",
    proxy_port=8302,
    timeout=600,
)

TEMPLATE = "tpl-openclaw-1000m-1024mb-20260612091229"
ENV_KEY = "TEST1"
ENV_VAL = "hello-from-create-env"

print("[1/3] create sandbox + env_vars")
sb = Sandbox.create(
    template=TEMPLATE,
    env_vars={ENV_KEY: ENV_VAL},
    config=CFG,
)
print(f"sandbox_id={sb.sandbox_id}")

print("[2/3] commands.run echo env")
r = sb.commands.run(f'echo "${ENV_KEY}"')
print(f"stdout={r.stdout.strip()!r} exit={r.exit_code}")
if r.exit_code != 0 or r.stdout.strip() != ENV_VAL:
    raise SystemExit(f"FAIL want {ENV_VAL!r}")

print("[3/3] kill")
sb.kill()
print("PASS")
```
image

**Pass criteria:** last line is `PASS`; step 2 prints
`stdout='hello-from-create-env' exit=0`.

## Unit tests

```bash
cd sdk/python && pytest tests/test_sandbox.py -q
cd sdk/go && go test ./... -count=1
```

---

## Related

Follow-up to #484. Supersedes the rejected guest rootfs / Cubelet injection approach.

### DIFF
--- a/sdk/python/cubesandbox/sandbox.py
+++ b/sdk/python/cubesandbox/sandbox.py
@@ -3,12 +3,15 @@
 
 from __future__ import annotations
 
+import random
+import time
+from datetime import datetime, timezone
 from typing import Any, Callable, Dict
 
 import httpx
 import requests
 
-from ._commands import CommandResult, Commands
+from ._commands import CommandResult, Commands, ENVD_PORT
 from ._config import Config
 from ._exceptions import ApiError, AuthenticationError, CubeSandboxError, SandboxNotFoundError, TemplateNotFoundError
 from ._filesystem import Filesystem
@@ -18,6 +21,10 @@ from ._stream import _parse_line
 from ._transport import build_client
 
 JUPYTER_PORT = 49999
+ENVD_INIT_MAX_ATTEMPTS = 5
+ENVD_INIT_RETRY_BASE_SECS = 0.8
+ENVD_INIT_RETRY_JITTER_SECS = 0.4
+ENVD_INIT_REQ_TIMEOUT_SECS = 10.0
 
 
 def _check_response(resp: requests.Response) -> None:
@@ -156,7 +163,10 @@ class Sandbox:
         resp = s.post(f"{cfg.api_url}/sandboxes", json=payload,
                       headers={"Content-Type": "application/json"})
         _check_response(resp)
-        return cls(resp.json(), config=cfg)
+        sandbox = cls(resp.json(), config=cfg)
+        if env_vars:
+            sandbox._init_env_vars(env_vars)
+        return sandbox
 
     @classmethod
     def connect(cls, sandbox_id: str, *, config: Config | None = None) -> "Sandbox":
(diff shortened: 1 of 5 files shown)


fslongjin:
@xiaojunxiang2023 Thanks for taking the time to work on this PR! I've reviewed both #554 and #566 carefully — they're solving the same problem: `Sandbox.create(env_vars=...)` not reaching the guest runtime. I've decided to go with #566 (CubeAPI server-side), and have patched several of the good ideas from #554 into #566. Here's the reasoning.

## Why #566 (server-side) over #554 (SDK-side)

**1. Aligns with E2B upstream semantics**

I traced the [E2B infra source](https://github.com/e2b-dev/infra/blob/main/packages/envd/internal/api/init.go) (`packages/envd/internal/api/init.go` + `packages/orchestrator/pkg/server/sandboxes.go`). In E2B's architecture, `POST /init` is called by the **Orchestrator (server-side)** after the VM boots — the SDK never calls it directly. The E2B SDK only sends `envs` as part of the `POST /sandboxes` request body, and the server handles the actual initialization.

```
E2B:    SDK → API Service → Orchestrator → POST /init → envd
#566:   SDK → CubeAPI → POST /sandbox/{id}/49983/init → envd  (server-side)
#554:   SDK → POST /sandboxes → SDK then calls POST /init → envd  (client-side)
```

**2. Transparent to all SDKs**

Putting `/init` on the server means Python, Go, TypeScript, the E2B SDK, and even plain curl calls all get correct env vars. SDK-side requires reimplementing this in every SDK, and external users calling CubeSandbox via `e2b_code_interpreter` would still not get env vars injected.

**3. Separation of concerns**

The SDK shouldn't need to know envd's port (49983), the `/init` API format, or CubeProxy's routing rules — those are infrastructure-layer details. The SDK's role should be limited to putting `envVars` in the `POST /sandboxes` body, just like the E2B SDK does.

## What I adopted from #554

Your PR had several well-designed aspects that I've patched into #566:

- **Jittered retry** (5 attempts, 800ms base delay + jitter): envd may not be ready immediately after sandbox creation, and a single 10s timeout isn't robust enough. #566 was originally single-attempt best-effort; it now does hard failure after retries are exhausted, with best-effort sandbox cleanup.
- **Hard failure strategy**: silently lost env vars are worse than a slightly slower creation — a user passes `env_vars={"API_KEY": "sk-xxx"}` expecting it to work, but it doesn't, and that's very hard to debug.

Full comparison and fix plan:
- [docs/pr-554-vs-pr-566-comparison.md](https://github.com/TencentCloud/CubeSandbox/blob/fix/cubeapi-envd-init/docs/pr-554-vs-pr-566-comparison.md)
- [docs/pr-566-fix-plan.md](https://github.com/TencentCloud/CubeSandbox/blob/fix/cubeapi-envd-init/docs/pr-566-fix-plan.md)

So I'd suggest closing #554 and consolidating all changes in #566. Thanks again for the contribution — the jittered retry and hard failure design in particular made #566 more robust.

sdk/python/cubesandbox/sandbox.py
```diff
@@ -667,3 +677,64 @@ def _build_session(self) -> requests.Session:
...
+        if access_token:
+            headers["X-Access-Token"] = access_token
+
+        url = f"http://{self.get_host(ENVD_PORT)}/init"
+        body = {
+            "envVars": env_vars,
+            "timestamp": datetime.now(timezone.utc).isoformat(),
+        }
+
+        # The proxy route to a freshly created sandbox may settle a moment after
```
chenhengqi, line 711: Do you have any clues on this?

PR #556, opened 2026-06-14
fix(cube_egress_ca): seed CA bundle for distroless/scratch rootfs

kinwin-ustc:
Bake previously hard-failed on images that ship neither a ca-bundle file nor any anchor directory (distroless / scratch), because it only appended to existing trust stores and wrote zero targets.

Now Bake seeds a fresh bundle at etc/ssl/certs/ca-certificates.crt with the CubeEgress root when nothing else matches and the CA is not already present (Result.Seeded). This is safe under the egress-MITM model: workloads only need to trust the CubeEgress root, not the public Mozilla set. Seeding is gated so re-bakes stay idempotent and never clobber an image's own roots.

Relax the caller hard-error accordingly, log seeded=%t, and update the WithCubeCA doc. Tests updated to assert seeding + idempotency.

Assisted-by: CodeBuddy
Signed-off-by: Feng Jin<ronyjin@tencent.com>


fslongjin (approved):
LGTM

PR #562, opened 2026-06-14
fix(cube_egress_ca): suppress default x509v3 extensions to prevent duplicate extensions

fslongjin:
On Ubuntu 20.04 (OpenSSL 1.1.1f), the system openssl.cnf [v3_ca] section auto-adds basicConstraints and subjectKeyIdentifier when generating a self-signed CA certificate via "openssl req -x509". The explicit -addext flags in cube-egress-prepare.sh and gen-ca.sh then append the same extensions a second time, producing a certificate with duplicate x509v3 extensions that Go's crypto/x509 parser rejects per RFC 5280.

Fix: supply a minimal OpenSSL config file via -config that has no x509_extensions directive, suppressing system defaults so only the explicitly listed -addext extensions appear in the output certificate.

This only manifests on distributions whose default openssl.cnf includes x509_extensions=v3_ca with pre-populated basicConstraints and subjectKeyIdentifier (e.g. Ubuntu 20.04). OpenSSL 3.x-based distros are not affected.

Verified on both Ubuntu 20.04 (OpenSSL 1.1.1f) and OpenSSL 3.6.0: before fix, duplicate extensions on 1.1.1f; after fix, each extension appears exactly once on both platforms.

### DIFF
--- a/CubeEgress/gen-ca.sh
+++ b/CubeEgress/gen-ca.sh
@@ -10,12 +10,26 @@ CRT_FILE="ca.crt"
 openssl ecparam -name "$CURVE" -genkey -noout -out "$KEY_FILE"
 chmod 600 "$KEY_FILE"
 
+# Minimal config to suppress the system openssl.cnf's default x509v3
+# extensions (specifically [v3_ca]'s basicConstraints + subjectKeyIdentifier).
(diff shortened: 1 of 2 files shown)


chenhengqi (approved):
LGTM. Tested on BMI5 + OpenCloud OS 9.

zhouxianping (approved):

PR #576, opened 2026-06-16
feat(network-agent): make host port allocator range configurable

dushulin:
Expose host_port_min / host_port_max in the cubelet TOML so the port pool can be sized to the cluster instead of the hard-coded 20000-29999. With the default 10k ports and 4 exposed ports per sandbox, a single node caps at ~2500 sandboxes regardless of the cpu/mem quotas. Letting the operator widen the range removes that ceiling without code changes.

Defaults stay 20000-29999, so existing deployments keep the same behavior. Validation rejects min > max and min < 1024.

### DIFF
--- a/network-agent/cmd/network-agent/main.go
+++ b/network-agent/cmd/network-agent/main.go
@@ -246,7 +246,7 @@ func validateService(svc service.Service) error {
 
 func summarizeConfig(cfg service.Config) string {
 	return fmt.Sprintf(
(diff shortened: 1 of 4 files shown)


chenhengqi (changes requested):
The bot has some comments, please address.

> a single node caps at ~2500 sandboxes regardless of the cpu/mem quotas

Please share your use case where 2500 sandboxes is not enough for a single node.

chenhengqi:
The PR is stale, closed. Please reopen if this is a real production issue.

dushulin:
> The PR is stale, closed. Please reopen if this is a real production issue.
Indeed, it’s not appropriate to assign too many sandboxes to a single node. With the current range of port numbers, if each sandbox uses 4 ports, a density of 2500 sandboxes per node can be achieved, which meets our business requirements.

PR #577, opened 2026-06-16
[experimental]CubeProxy: Disable buffering for PTY streams

chenhengqi:
This is suboptimal but should address the PTY use cases.


chenhengqi:
Close in favor of #647.

PR #581, opened 2026-06-16
fix(CubeMaster): preserve template network rules and resource defaults

zyl1121:
## Motivation

The template / sandbox request materialization path currently has a few correctness gaps where accepted or configured fields are not preserved consistently.

The most user-visible issue is in template creation from image: `cube_network_config.rules` can be submitted successfully, but the stored template `create_request` drops the `rules` array while keeping the rest of `cube_network_config`.

This PR fixes that confirmed regression and also tightens two closely related resource-handling cases in the same request materialization path:

* preserve the first validation error when both CPU and memory exceed configured limits
* preserve template CPU / memory defaults when applying them to a request whose `Resources` field is initially nil

## Reproduction

I first observed the network rules regression on a one-click deployment running `v0.4.0`.

Steps:

1. Create a template from image with both `allowInternetAccess` and an egress `rules` entry inside `cube_network_config`.
2. Wait for the template image job to succeed.
3. Inspect the stored template request:

```text
GET /cube/template?template_id=...&include_request=true
```

Before this fix, the stored `create_request.cube_network_config` keeps:

```json
{
  "allowInternetAccess": true
}
```

but silently drops the submitted `rules` array.

That means the request is accepted, but the resulting template no longer carries the intended egress rule set forward into later sandbox creation.

## What This PR Changes

* Preserve `cube_network_config.rules` when cloning a template create request from `CreateTemplateFromImageReq` into `CreateCubeSandboxReq`.

  * `pkg/templatecenter/template_request.go`
  * adds deep-copy helpers for egress rules

* Preserve the first resource validation error when both CPU and memory exceed configured limits.

  * `pkg/service/sandbox/util.go`
  * prevents the later memory validation check from overwriting an earlier CPU validation error

* Preserve template CPU / memory defaults when merging template resources into a request with `Resources == nil`.

  * `pkg/service/httpservice/cube/cubeboxutil.go`
  * makes `applyTemplateResources()` return the effective `*types.Resource`, and callers keep the returned pointer

* Add focused regression tests for all three cases.

## Validation

Manual validation on the deployed cluster:

* Confirmed that `POST /cube/template/from-image` accepts `cube_network_config.rules`, but `GET /cube/template?...&include_request=true` drops `rules` while keeping `allowInternetAccess`.
* Confirmed that `POST /cube/sandbox` with `cpu=101, mem=2000Mi` is rejected with a CPU validation error.
* Confirmed that `POST /cube/sandbox` with both `cpu=101` and `mem=301Gi` returns the later memory validation error, which matches the validation-order issue fixed by this PR.

Focused regression tests on this branch:

```bash
cd CubeMaster

go test -vet=off ./pkg/templatecenter -run TestGenerateTemplateCreateRequestClonesCubeNetworkRules -count=1
go test -vet=off ./pkg/service/sandbox -run 'TestGetReqResourceRejectsCPUOverflowWhenMemIsValid|TestGetReqResourceRejectsCPUOverflowBeforeMemOverflow' -count=1
go test -vet=off ./pkg/service/httpservice/cube -run TestApplyTemplateToContainerSetsResourcesWhenRequestResourceIsNil -count=1
```

The focused `pkg/service/sandbox` tests cover both the `cpu overflow + mem valid` case and the `cpu overflow + mem overflow` case.

## Scope

This PR is limited to request correctness in the template / sandbox materialization path.

It does not change scheduler limits, template policy semantics, or introduce any new API fields.

The main behavior change is that already accepted request fields and already configured template resource defaults are now preserved consistently when requests are cloned, validated, and merged.

### DIFF
--- a/CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
+++ b/CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
@@ -264,7 +264,7 @@ func mergeCubeNetworkConfigs(templateCfg *types.CubeNetworkConfig, requestCfg *t
 		return cloneCubeNetworkConfig(templateCfg)
 	}
 
-	out := cloneCubeNetworkConfig(templateCfg)
+	out := cloneCubeNetworkConfigBase(templateCfg)
 	if requestCfg.AllowInternetAccess != nil {
 		allowInternetAccess := *requestCfg.AllowInternetAccess
 		out.AllowInternetAccess = &allowInternetAccess
@@ -285,19 +285,29 @@ func mergeCubeNetworkConfigs(templateCfg *types.CubeNetworkConfig, requestCfg *t
 		out.DenyOut = appendUniqueCIDRs(out.DenyOut, requestCfg.DenyOut)
 	}
 	if len(requestCfg.Rules) > 0 {
-		out.Rules = mergeEgressRules(out.Rules, requestCfg.Rules)
+		out.Rules = mergeEgressRules(templateCfg.Rules, requestCfg.Rules)
+	} else {
+		out.Rules = cloneEgressRules(templateCfg.Rules)
 	}
 	return out
 }
 
 func cloneCubeNetworkConfig(in *types.CubeNetworkConfig) *types.CubeNetworkConfig {
+	out := cloneCubeNetworkConfigBase(in)
+	if out == nil {
+		return nil
+	}
+	out.Rules = cloneEgressRules(in.Rules)
+	return out
+}
+
+func cloneCubeNetworkConfigBase(in *types.CubeNetworkConfig) *types.CubeNetworkConfig {
 	if in == nil {
 		return nil
 	}
 	out := &types.CubeNetworkConfig{
 		AllowOut: append([]string(nil), in.AllowOut...),
 		DenyOut:  append([]string(nil), in.DenyOut...),
-		Rules:    cloneEgressRules(in.Rules),
 	}
 	if in.AllowInternetAccess != nil {
 		allowInternetAccess := *in.AllowInternetAccess
@@ -310,32 +320,29 @@ func cloneCubeNetworkConfig(in *types.CubeNetworkConfig) *types.CubeNetworkConfi
 	return out
 }
 
-// mergeEgressRules combines template + request rules. Rules sharing the same
-// Name are overridden by the request side; otherwise request rules are
-// appended after template rules to preserve first-match-wins ordering with
-// the template's policy taking precedence on overlap.
+// mergeEgressRules combines template + request rules. Because egress rules are
+// first-match-wins, per-sandbox/request rules must come before template rules.
 func mergeEgressRules(base []*types.EgressRule, extra []*types.EgressRule) []*types.EgressRule {
 	if len(extra) == 0 {
-		return base
+		return cloneEgressRules(base)
 	}
-	indexByName := make(map[string]int, len(base))
-	out := cloneEgressRules(base)
-	for i, r := range out {
-		if r != nil {
-			indexByName[r.Name] = i
-		}
+	if len(base) == 0 {
+		return cloneEgressRules(extra)
 	}
+
+	out := make([]*types.EgressRule, 0, len(extra)+len(base))
 	for _, r := range extra {
 		if r == nil {
 			continue
 		}
-		cloned := cloneEgressRule(r)
-		if idx, ok := indexByName[r.Name]; ok {
-			out[idx] = cloned
+		out = append(out, cloneEgressRule(r))
+	}
+
+	for _, r := range base {
+		if r == nil {
 			continue
 		}
-		indexByName[r.Name] = len(out)
-		out = append(out, cloned)
+		out = append(out, cloneEgressRule(r))
 	}
 	return out
 }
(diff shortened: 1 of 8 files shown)


CubeMaster/pkg/service/sandbox/util.go
```diff
@@ -110,7 +110,7 @@ func getReqResource(req *types.CreateCubeSandboxReq) (cpu, mem resource.Quantity
 			err = ret.Errorf(errorcode.ErrorCode_MasterParamsError, "request Resources cpu[%dm] is invalid",
 				cpu.MilliValue())
 		}
-		if mem.Cmp(config.GetConfig().Scheduler.MaxMvmMemoryRes()) >= 0 {
+		if err == nil && mem.Cmp(config.GetConfig().Scheduler.MaxMvmMemoryRes()) >= 0 {
```
chenhengqi: Let's return early instead of nil check here.

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
```diff
@@ -114,10 +114,7 @@ func applyTemplateToContainer(ctr *types.Container, templateCtr *types.Container
 	}
 	applyTemplateImageSpec(templateCtr.Image, ctr.Image)
 
-	if ctr.Resources == nil {
-		ctr.Resources = &types.Resource{}
-	}
-	applyTemplateResources(templateCtr.Resources, ctr.Resources)
+	ctr.Resources = applyTemplateResources(templateCtr.Resources, ctr.Resources)
```
chenhengqi: It seems like changes to this file is unnecessary.

CubeMaster/pkg/templatecenter/template_request.go
```diff
@@ -134,6 +134,7 @@ func cloneCubeNetworkConfig(in *types.CubeNetworkConfig) *types.CubeNetworkConfi
 	out := &types.CubeNetworkConfig{
 		AllowOut: append([]string(nil), in.AllowOut...),
 		DenyOut:  append([]string(nil), in.DenyOut...),
+		Rules:    cloneEgressRules(in.Rules),
```
chenhengqi, line 137: Currently, there is no way to pass egress rules when create template.

zyl1121: > Currently, there is no way to pass egress rules when create template.

I think this is true for `cubemastercli` today, but not for the HTTP API.

`CreateTemplateFromImageReq` already includes `cube_network_config.rules`, and the `/cube/template/from-image` handler unmarshals the request body directly into that type.

I reproduced this on a `v0.4.0` one-click deployment by sending `cube_network_config.rules` in the raw `POST` body. The request was accepted, `allowInternetAccess` was preserved, but `rules` was missing from the stored template `create_request`.

So the issue here seems to be that raw API callers can already pass `rules`, but the template request clone path drops them.

If the intended scope is to only support fields currently exposed by `cubemastercli`, I can split the `rules` preservation into a follow-up and keep this PR focused on other fixes.

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
zyl1121: > It seems like changes to this file is unnecessary.

Good point. I added this as defensive hardening for the case where the destination `Resources` is nil, but the current call sites already handle allocation. To keep this PR focused, I will drop this change.

CubeMaster/pkg/service/sandbox/util.go
zyl1121: > Let's return early instead of nil check here.

Done, I updated this in a new commit to return early.

CubeMaster/pkg/templatecenter/template_request.go
chenhengqi: Good point. I think we should also revise `mergeEgressRules()`. The per sandbox rules should prepend to templates rules.

zyl1121: > Good point. I think we should also revise `mergeEgressRules()`. The per sandbox rules should prepend to templates rules.

Updated this part.

`mergeEgressRules()` now prepends per-sandbox/request rules before template rules, so first-match-wins evaluation gives request-side rules precedence.

I also added focused tests for name-based overrides, no-conflict merge order, empty-side handling, and nil rule entries.

chenhengqi:
@zyl1121  Please resolve conflicts and squash your commits.

zyl1121:
@chenhengqi Done. The merge conflict is resolved, and the commits have been squashed into one.

docs/guide/network-policy.md
```diff
@@ -194,7 +194,7 @@ If the template also carries a `CubeNetworkConfig`, CubeMaster merges template c
 - `allowInternetAccess`: request value overrides the template value when explicitly set.
 - `allowOut`: request entries are appended to template entries and deduplicated by string.
 - `denyOut`: request entries are appended to template entries and deduplicated by string.
-- `rules`: merged by `name`. A request rule with the same name overrides the template rule; rules with new names are appended after template rules, preserving first-match-wins order.
+- `rules`: merged by `name`. Request rules are placed before template rules. Template rules with the same `name` are dropped, so request rules take precedence under first-match-wins.
```
chenhengqi: I think we don't need to drop template rules.

zyl1121: > I think we don't need to drop template rules.

Updated.

I removed the name-based dropping logic in `mergeEgressRules`.
Now request rules are still placed before template rules, but same-named template rules are kept in the merged list and rely on first-match-wins ordering instead of being removed.

I also updated the network policy docs to match that behavior.

chenhengqi (approved):
LGTM.

fslongjin (approved):

PR #582, opened 2026-06-16
feat(agenthub): persist assistant state and auth settings

tinklone:
## Summary
- persist AgentHub assistant auth/settings and OpenClaw runtime state
- add AgentHub auth guard/login flow and settings UI
- add snapshot/template persistence metadata and recovery-related API handling
- update one-click Web UI reverse proxy config for the new auth/API paths


CubeAPI/src/crypto.rs
```diff
@@ -0,0 +1,110 @@
...
+use aes_gcm::{
+    aead::{Aead, KeyInit},
+    Aes256Gcm, Key, Nonce,
+};
+use base64::{engine::general_purpose::STANDARD as BASE64, Engine as _};
+
+const ENC_PREFIX: &str = "enc:v1:";
+const NONCE_LEN: usize = 12;
+/// 32-byte development fallback key. Override with AGENTHUB_SECRET_KEY.
+const DEFAULT_DEV_KEY: &[u8; 32] = b"0123456789abcdef0123456789abcdef";
```
fslongjin, line 26: ⚠️ **Security**: This hardcoded dev key is compiled into the binary and used when `AGENTHUB_SECRET_KEY` is not set. Consider logging a prominent startup warning when the fallback key is active, so operators don't accidentally run production instances with the dev key.

Suggested code:
```rust
if std::env::var("AGENTHUB_SECRET_KEY").is_err() {
    tracing::warn!("AGENTHUB_SECRET_KEY not set — using INSECURE development key");
}
```

CubeAPI/src/crypto.rs
```diff
@@ -0,0 +1,110 @@
...
+/// Returns whether a stored value is in the encrypted envelope format.
+pub fn is_encrypted(stored: &str) -> bool {
+    stored.starts_with(ENC_PREFIX)
+}
+
+/// Decrypts an encrypted value, or returns the input unchanged when it is not
+/// in the encrypted format (legacy plaintext rows) or cannot be decrypted.
+pub fn decrypt_or_passthrough(stored: &str) -> String {
+    if is_encrypted(stored) {
+        decrypt_secret(stored).unwrap_or_else(|_| stored.to_string())
```
fslongjin: ⚠️ **Silent failure**: `unwrap_or_else(|_| stored.to_string())` returns the raw `enc:v1:...` ciphertext when decryption fails (wrong key, data corruption). This means the LLM API key gets set to the literal encrypted string, causing AI features to break silently with no error logged.

Consider at least logging: `tracing::warn!("Failed to decrypt secret, returning as-is")` or propagating the error with `Result`.

CubeAPI/src/handlers/auth.rs
```diff
@@ -0,0 +1,198 @@
...
+use axum::{extract::State, http::HeaderMap, http::StatusCode, response::IntoResponse, Json};
+use serde::{Deserialize, Serialize};
+
+use crate::{
+    error::{AppError, AppResult},
+    state::AppState,
+};
+
+const SESSION_HEADER: &str = "x-session-token";
+const SESSION_TTL_SECS: i64 = 7 * 24 * 60 * 60;
```
fslongjin: ♻️ **Design**: 7-day session TTL with no sliding expiration is quite permissive. Consider reducing to 24 hours OR implementing sliding expiration (extend TTL on each `/auth/session` validation) so active users stay logged in while idle sessions expire promptly.

CubeAPI/src/handlers/auth.rs
```diff
@@ -0,0 +1,198 @@
...
+    };
+
+    let stored = store
+        .get_user_password(username)
+        .await
+        .map_err(|e| AppError::Internal(anyhow::anyhow!("failed to read user: {}", e)))?;
+    match stored {
+        Some(expected) if crate::crypto::verify_password(&expected, &body.old_password) => {}
+        Some(_) => {
+            return Err(AppError::Unauthorized(
```
fslongjin: 🔐 **User enumeration risk**: This returns `"current password is incorrect"` on line 184, but line 187 returns `"user {} not found"` — two different error messages. An attacker can enumerate valid usernames by observing the error type.

The `/auth/login` endpoint (line 89) correctly uses a unified `"invalid credentials"` message. This endpoint should follow the same pattern, e.g.: `"current password is incorrect or user not found"`.

fslongjin (changes requested):
## PR #582 Code Review

### Critical Issues Found
Detailed inline comments below. Key findings:

1. 🐛 **Bug**: WeCom bot secret is returned encrypted (not decrypted) in API response — causes double-encryption and breaks WeChat bot integration
2. 🛡️ **Security**: Auth routes (`/auth/login`, `/auth/change-password`) have no rate limiting — vulnerable to brute force
3. 🧪 **Missing tests**: `crypto.rs` and `auth.rs` have zero test coverage
4. 🔐 **User enumeration**: `change_password` returns different errors for "wrong password" vs "user not found"
5. ♻️ **Session TTL**: 7-day session with no sliding expiration

See inline comments for detailed analysis and fix suggestions.

web/src/components/AuthGuard.tsx
```diff
@@ -0,0 +1,51 @@
...
+  useEffect(() => {
+    let cancelled = false;
+    authApi
+      .session()
+      .then((res) => {
+        if (cancelled) return;
+        setState(!res.authRequired || res.authenticated ? 'allowed' : 'guest');
+      })
+      .catch(() => {
+        // If the session check itself fails (e.g. backend unreachable), fail
```
fslongjin: ⚠️ **Security consideration**: When `/auth/session` is unreachable, the guard "fails open" granting access to all routes. While intentional (avoid locking users out), a brief backend restart would bypass auth entirely. Consider caching the last-known auth state in `sessionStorage` as a fallback: if the session check fails, consult the cached state before allowing access.

CubeAPI/src/routes.rs
```diff
@@ -83,12 +83,23 @@ fn build_e2b_snapshot_long_router(state: &AppState, auth_configured: bool) -> Ro
...
         .route("/health", get(health::health))
+        .merge(build_auth_routes())
         .merge(build_sandbox_routes(state, auth_configured))
         .merge(build_template_routes(state, auth_configured))
         .merge(build_cluster_routes(state, auth_configured))
         .merge(build_agenthub_routes(state, auth_configured))
 }
 
+/// WebUI login routes. These are intentionally left unauthenticated (like
+/// `/health`) so the login/session flow itself is always reachable.
```
fslongjin, line 94: 🛡️ **Missing rate limiting**: Auth routes (`/auth/login`, `/auth/change-password`) bypass both authentication and rate limiting middleware. While bypassing auth is intentional (users must be able to login), rate limiting should still apply to prevent brute-force attacks. Consider adding a lightweight rate-limit layer specifically for these auth routes.

CubeAPI/src/db.rs
```diff
@@ -1044,9 +1469,14 @@ impl AgentHubInstanceRecord {
...
-                self.gateway_token,
+                gateway_token,
             ),
             env_url: crate::handlers::agenthub::sandbox_url(8080, &self.sandbox_id, &self.domain),
+            persistence_mode: self.persistence_mode,
+            rootfs_source_type: self.rootfs_source_type,
+            rootfs_source_id: self.rootfs_source_id,
+            openclaw_persist_id: self.openclaw_persist_id,
+            openclaw_state_path: self.openclaw_state_path,
             wecom_config: match (self.wecom_bot_id, self.wecom_bot_secret) {
```
fslongjin, line 1480: 🐛 **Related to wecom_secret encryption**: `into_response()` returns the raw encrypted `wecom_bot_secret` (stored as `enc:v1:...`) without decrypting it. When instance records are serialized for list/get agent endpoints, the encrypted blob is exposed to the frontend. Consider decrypting here via `crate::crypto::decrypt_or_passthrough`, or create a dedicated API response method that handles decryption.

fslongjin (commented):
### Additional finding: WeCom bot secret returned encrypted in `get_agent_wecom_config`

**File**: `CubeAPI/src/handlers/agenthub.rs` around line 3020 (function is unchanged in this PR so couldn't leave inline comment)

```rust
let config = match (record.wecom_bot_id.clone(), record.wecom_bot_secret.clone()) {
    (Some(bot_id), Some(bot_secret)) if !bot_id.is_empty() && !bot_secret.is_empty() => {
        Some(AgentWeComConfig { bot_id, bot_secret })  // ← bot_secret is still encrypted!
    }
    _ => read_openclaw_wecom_config(...).await.unwrap_or(None),
};
```

`record.wecom_bot_secret` is stored as `enc:v1:...` (encrypted via `crate::crypto::encrypt_secret` in `db.rs:L523` and `db.rs:L611`), but the API returns it without decryption.

**Impact**:
- Frontend displays the encrypted ciphertext as the bot secret
- If resubmitted without modification → double-encryption (`enc:v1:enc:v1:...`)
- Encrypted string is passed to `configure_openclaw()` as the actual WeChat secret → breaks integration

**Fix**:
```rust
let bot_secret = crate::crypto::decrypt_or_passthrough(&bot_secret);
```

fslongjin (approved):

PR #586, opened 2026-06-17
fix(one-click): distinguish same-CIDR reinstall from CIDR change and use systemd to stop services

fslongjin:
## Summary

The install-time CIDR pre-flight flagged Cube's own residual `cube-dev` interface and `z*` TAP devices as a host-network conflict, so reinstalling with the same `CUBE_SANDBOX_NETWORK_CIDR` failed spuriously. Components are now systemd-managed (`install.sh` enables `cube-sandbox-{control,compute}.target`, services carry `PartOf=` back-links), while `down-with-deps.sh` is the pre-systemd legacy stop path (pidfile-based, ineffective against `Restart=on-failure` services).

## Changes

**common.sh — `_check_cidr_conflict`:**
- Records `cube-dev`'s CIDR; skips `cube-dev` and `z*` TAPs in **both** interface and route checks (previously the route check only skipped `dev cube-dev`).
- At the tail decides: same network+mask → reuse (log only); overlapping different network → `die` with guidance; non-overlapping → allow.
- The `die` message now tells users to `systemctl stop 'cube-sandbox-*.target'` instead of the legacy `down-with-deps.sh`. The glob matches the role's target on both control and compute nodes; `PartOf=` propagates the stop to every service.

**docs — CIDR conflict troubleshooting page (en/zh):**
- Rewrite around the two current scenarios: same-CIDR reinstall reuses `cube-dev`, changing the CIDR to an overlapping range is blocked with a deterministic-reset hint.
- Swap stop/start commands to `systemctl stop/start 'cube-sandbox-*.target'`.
- Unify the `z*` TAP cleanup command across the page (was two different commands).

## Why `systemctl stop 'cube-sandbox-*.target'` is correct

Verified on a live control/all-in-one node (`systemctl is-active` + `PartOf=` inspection):
- 14 `cube-sandbox-*` services active, `cube-sandbox-control.target` enabled.
- `cubelet`/`cubemaster`/`network-agent` all `PartOf=cube-sandbox-control.target` → `systemctl stop` the target propagates to every service via `PartOf=`.
- `network-agent` is an `exec` process under systemd with `Restart=on-failure`, not a nohup pidfile — so the legacy `down-with-deps.sh` (`stop_by_pidfile`) is a no-op against it.

## Behavior verified on live node

Ran the patched `check_cidr_preflight` against a host with 10 residual `z*` TAPs and `cube-dev 192.168.0.1/18`:

| Scenario | CIDR | Result |
|---|---|---|
| Same CIDR reinstall | `192.168.0.0/18` | ✅ `reusing existing cube-dev network` → exit 0 |
| Change to overlapping | `192.168.0.0/17` | ✅ `die` with systemd stop guidance → exit != 0 |
| Change to non-overlapping | `10.5.0.0/16` | ✅ `preflight OK` → exit 0 |

## Checklist

- [x] Bilingual docs (en/zh) updated
- [x] Verified on a live node (no conflict false-positive on same-CIDR reinstall)
- [x] No local paths / secrets / real node IPs in the commit


zyl1121:
Hi, I noticed this overlaps with #516, especially around skipping Cube-owned leftover devices such as `cube-dev` and `z<ip>` during reinstall / recovery paths.

This PR seems to handle the same-CIDR reinstall / CIDR-change behavior more explicitly, and the systemd stop guidance also looks cleaner.

One thing worth considering is how we want to de-duplicate the two PRs. #516 still has a couple of useful coverage points that are not mainly about Cube-owned leftover devices:

- validating the effective CIDR even when `CUBE_SANDBOX_NETWORK_CIDR` is unset, by reading the packaged Cubelet config
- checking host DNS resolver nameservers from resolv.conf-style files, not only interfaces and routes

I think the default-CIDR case is worth covering as well, because from the user's perspective the effective sandbox CIDR can conflict with the host network regardless of whether it comes from an explicit override or the packaged default config.

The resolver case was the actual issue I hit: the default sandbox CIDR overlapped with a host DNS resolver address such as `192.168.0.1`. This was not obvious from `ip route`, but later caused host-side DNS / image pull failures.

Maybe we can use this PR as the main one and fold in the remaining resolver/default-CIDR coverage from #516. Alternatively, if you prefer to keep the Cube-owned leftover handling here, I can re-scope #516 to remove that part and keep it focused only on the resolver/default-CIDR checks.

fslongjin:
Thanks for pointing this out. I agree #516 covers a useful remaining gap beyond the Cube-owned leftover device handling.

I have updated this PR to fold in the default-CIDR and resolver nameserver coverage from #516: when `CUBE_SANDBOX_NETWORK_CIDR` is unset, the one-click preflight now checks the fixed default `192.168.0.0/18`, and `_check_cidr_conflict` also checks host resolver nameservers. I kept the more explicit `cube-dev` behavior here: same-CIDR reinstall reuses the existing network, while changing to an overlapping CIDR is still rejected with the reset guidance.

I also added focused tests for resolver overlap, resolver path de-duplication, normal host interface overlap, and the `cube-dev` same-CIDR / CIDR-change cases.

Credit added in the commit with `Co-authored-by: zhengyilei <zheng_yilei@qq.com>`, and the commit message references #516.

zyl1121:
Thanks, this looks good to me.

I checked the updated changes and the remaining resolver/default-CIDR coverage from #516 is now included here, with focused tests for the resolver and cube-dev cases. The same-CIDR reinstall / CIDR-change behavior is also clearer in this PR.

One minor thought: the default CIDR check currently hardcodes `192.168.0.0/18`, which is fine for the current packaged default. Reading it from the packaged Cubelet config would avoid possible drift if the default changes later, but I do not think this block the PR.

Once this PR is merged, I will close #516 to avoid duplicate work.

fslongjin:
> Thanks, this looks good to me.
> 
> I checked the updated changes and the remaining resolver/default-CIDR coverage from #516 is now included here, with focused tests for the resolver and cube-dev cases. The same-CIDR reinstall / CIDR-change behavior is also clearer in this PR.
> 
> One minor thought: the default CIDR check currently hardcodes `192.168.0.0/18`, which is fine for the current packaged default. Reading it from the packaged Cubelet config would avoid possible drift if the default changes later, but I do not think this block the PR.
> 
> Once this PR is merged, I will close #516 to avoid duplicate work.

The hardcoding of `192.168.0.0/18` here is intentional, to simplify the implementation. This network segment serves as a default value and will not change in the future.

zyl1121:
> The hardcoding of `192.168.0.0/18` here is intentional, to simplify the implementation. This network segment serves as a default value and will not change in the future.

Got it, that makes sense to me.

PR #594, opened 2026-06-17
fix(CubeMaster): populate tpl info created_at and image_info

zyl1121:
## Motivation

Follow-up to #592.

`tpl info` was updated to expose `display_name`, `created_at`, and `image_info` in the API/CLI response, but real-cluster verification showed that `templatecenter.GetTemplateInfo()` still did not populate the backing `created_at` and `image_info` fields for persisted templates.

As a result, the response/CLI layer was ready, but the data path behind `tpl info` was still incomplete.

## What Changed

* Populate `CreatedAt` from the persisted template definition in `GetTemplateInfo()`.
* Populate `ImageInfo` from the stored request and latest image job, matching the existing `ListTemplates()` behavior.
* Add a focused regression test for the `GetTemplateInfo()` backing-field path.

This makes `tpl info` consistent with `tpl list` for `created_at` and `image_info`.

## Validation

* Real-cluster verification:
  * before the fix, `/cube/template` and `cubemastercli tpl info` still returned empty / omitted values for `created_at` and `image_info`
  * after the fix, both surfaces returned the expected values for the same template
* Added a focused regression test for `GetTemplateInfo()` backing-field population.

### DIFF
--- a/CubeMaster/pkg/templatecenter/store.go
+++ b/CubeMaster/pkg/templatecenter/store.go
@@ -696,6 +696,11 @@ func GetTemplateInfo(ctx context.Context, templateID string) (*TemplateInfo, err
 	}
 	info := templateInfoFromDefinition(*def)
 	out := &info
+	out.CreatedAt = formatUTCRFC3339(def.CreatedAt)
+	out.ImageInfo = extractImageInfoFromRequestJSON(def.RequestJSON)
+	if latestJob, jobErr := getLatestTemplateImageJobByTemplateID(ctx, templateID); jobErr == nil && latestJob != nil {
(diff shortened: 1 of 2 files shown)


zyl1121:
@kinwin-ustc FYI, this is a follow-up to #592.

#592 updated the API/CLI response layer for `tpl info` and should not introduce side effects, but real-cluster verification showed that `created_at` and `image_info` were still empty for persisted templates because `GetTemplateInfo()` did not populate the backing fields yet.

This PR completes the underlying data path. With both changes together, `tpl info` can now show `created_at` and `image_info` as expected.

Sorry I missed this backing-field gap in #592.

PR #609, opened 2026-06-22
refactor(redis): unify Redis key naming and drop read/write pool separation

fslongjin:
- Add centralized redis key definitions in pkg/base/rediskey
- Unify CubeMaster Redis to single pool, drop read/write/metadata separation
- Update CubeProxy Lua scripts to use new key scheme
- Add migration script for existing Redis data
- Add bilingual Redis key specification docs (zh/en)
- Update configs for single-node deployment


chenhengqi (commented):
I have landed #613, I think this PR need refresh.

PR #612, opened 2026-06-22
feat(template): support custom display name for templates

kinwin-ustc:
## Summary

Adds an optional, human-readable **display name** for templates that flows
end-to-end — from creation to listing/detail — across the CLI, REST API, and
WebUI. The name is purely cosmetic: it does **not** replace the auto-generated
`template_id` used to create sandboxes; tools fall back to `template_id` when no
name is set.

Backed by the existing `display_name` column on `template_definitions` (already
used by the snapshot flow), so no schema migration is required.

### Backend — CubeMaster (Go)
- `CreateTemplateFromImageReq` gains a `DisplayName` field.
- New `ensureTemplateDefinitionWithOptions` persists `DisplayName` via
  `definitionCreateOptions`; `image_job_runner` passes it through on create.
- `templateSummary` / list / create responses now return `display_name`.
- `cubemastercli`: new `--name` flag on `tpl create-from-image`, and a new
  `NAME` column in `tpl list` (shows `-` when unset).

### Backend — CubeAPI (Rust)
- Thread `display_name` through `models`, the cubemaster client, and the
  template service.
- Expose `displayName` in `TemplateSummary` / `TemplateDetail`, and accept it in
  the create request body.

### API contract
- `openapi.yml`: add `displayName` to `TemplateSummary` and `TemplateDetail`.

### Frontend — WebUI
- Optional **Template Name** field in the create form.
- Show the name in the template list cards and the detail page hero header
  (with `template_id` demoted to a subtitle).
- zh/en i18n strings added.

### Docs
- Document `--name` / `display_name` in the template-building guides (both
  English and 中文).

## How it works

WebUI / CLI ──displayName──► CubeAPI ──display_name──► CubeMaster ──► DB │ list / detail ◄── displayName ◄── display_name ◄──────────────────┘

## Testing
- `cargo` lints: clean (0 errors) for CubeAPI.
- Go: existing branch tests (`TestPrintTemplateSummaryIncludesOptionalMetadata`,
  `TestGetTemplateIncludesDisplayMetadata`) match the implementation.
- WebUI: TypeScript lints clean.
- Pre-existing Linux-only build symbols in `pkg/templatecenter/image/` are
  unrelated to this change (only surface on darwin tooling).

## Usage

```bash
  --image registry.example.com/my-app:latest \
  --writable-layer-size 2G --expose-port 8080 --probe 8080 --probe-path / \
  --name "My Dev Environment"


Assisted-by: CodeBuddy
Signed-off-by: Feng Jin(ronyjin@tencent.com)

### CODE REVIEW

File: web/src/api/client.ts [hunk 1]
```diff
@@ -120,6 +121,7 @@ function mapSandboxDetail(dto: SandboxDetailDto): SandboxDetail {
 function mapTemplateSummary(dto: TemplateSummaryDto): TemplateSummary {
   return {
     templateID: dto.templateID,
+    displayName: (dto as unknown as { displayName?: string }).displayName ?? null,
```
fslongjin, line 124: `TemplateSummaryDto` is `components['schemas']['TemplateSummary']`, and the regenerated `schema.ts` now types `displayName?: string | null` on both DTOs. So the `as unknown as` cast is redundant here and actually weakens type safety — it narrows `string | null` down to `string | undefined`.

Suggested simplification:
```ts
displayName: dto.displayName ?? null,
```
(same applies to `mapTemplateDetail` at L139)

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -786,6 +788,7 @@ var TemplateCreateFromImageCommand = cli.Command{
 			Request:        &types.Request{RequestID: uuid.New().String()},
 			SourceImageRef: c.String("image"),
 			// TemplateID is auto-generated by normalizeTemplateImageRequest.
+			DisplayName:        c.String("name"),
```
fslongjin, line 791: Two notes on this line:

1. `c.String("name")` is persisted raw, so `--name "   "` would store whitespace. The WebUI/Rust entry points normalize whitespace to empty (`.trim()` / `non_empty_option`), so the CLI diverges. Consider `strings.TrimSpace(c.String("name"))` so all entry points agree.
2. The `--name` → `req.DisplayName` → `definitionCreateOptions{DisplayName}` path isn't covered by a test. The `newCreateFromImageContext` helper already exists in `template_test.go` — a small assertion that `--name X` yields `req.DisplayName == "X"` would guard this.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -1086,19 +1089,23 @@ var TemplateListCommand = cli.Command{
...
 		w := tabwriter.NewWriter(os.Stdout, 4, 8, 4, ' ', 0)
-		tabHeader := "TEMPLATE_ID\tSTATUS\tCREATED_AT\tIMAGE_INFO"
+		tabHeader := "TEMPLATE_ID\tNAME\tSTATUS\tCREATED_AT\tIMAGE_INFO"
 		if wideOutput {
-			tabHeader = "TEMPLATE_ID\tSTATUS\tLAST_ERROR\tCREATED_AT\tIMAGE_INFO"
+			tabHeader = "TEMPLATE_ID\tNAME\tSTATUS\tLAST_ERROR\tCREATED_AT\tIMAGE_INFO"
 		}
 		fmt.Fprintln(w, tabHeader)
 		for _, item := range rsp.Data {
+			name := item.DisplayName
```
fslongjin, line 1098: The list output was restructured here — new `NAME` column, and the `fmt.Fprintf` calls went from 4/5 to 5/6 args. `go vet` confirms the arg counts are correct, but there's no test guarding the tabular layout. A table-driven test asserting the header (`TEMPLATE_ID\tNAME\t...`), a row with a name, a row showing `-` when unset, and the wide-mode column order would catch future regressions in this formatting.

CubeMaster/pkg/templatecenter/image_job_runner.go
```diff
@@ -144,7 +144,7 @@ func runTemplateImageJob(ctx context.Context, jobID string, req *types.CreateTem
 		})
 		return
 	}
-	if _, err := ensureTemplateDefinition(ctx, req.TemplateID, storedReq, generatedReq.InstanceType, constants.GetAppSnapshotVersion(generatedReq.Annotations)); err != nil {
+	if _, err := ensureTemplateDefinitionWithOptions(ctx, req.TemplateID, storedReq, generatedReq.InstanceType, constants.GetAppSnapshotVersion(generatedReq.Annotations), definitionCreateOptions{DisplayName: req.DisplayName}); err != nil {
```
fslongjin, line 147: Heads-up on the retry/rebuild path: `ensureTemplateDefinitionWithOptions` short-circuits when the definition already exists (it returns without updating), so a build that fails after the definition is created and is then retried with a *different* `--name` will keep the originally-stored name. This is consistent with how sibling fields (`Kind`, `OriginSandboxID`, …) behave, so it's not a regression — but it means `display_name` is effectively write-once at create time. Worth a one-line note in the docs that the name can't be changed post-create.

CubeAPI/src/models/mod.rs
```diff
@@ -566,6 +570,9 @@ pub struct CreateTemplateRequest {
     pub template_id: String,
     #[serde(rename = "instanceType", default)]
     pub instance_type: Option<String>,
+    /// Optional human-friendly display name for the template.
+    #[serde(rename = "displayName", default)]
+    pub display_name: Option<String>,
```
fslongjin, line 575: The DB column is `varchar(256)` (see `0002_v0_2_2_to_head.sql`), but nothing caps the length on client or server. An over-long name errors on strict MySQL or silently truncates otherwise. Since `image` right below already uses `#[validate(length(min = 1))]`, consider `#[validate(length(max = 128))]` here (plus a matching `maxLength` on the WebUI `<Input>`) so it fails fast with a clear error instead of a DB-level failure.

CubeAPI/src/services/templates.rs
```diff
@@ -512,6 +515,7 @@ mod tests {
         CreateTemplateRequest {
             template_id: String::new(),
             instance_type: Some("cubebox".to_string()),
+            display_name: Some("My Template".to_string()),
```
fslongjin, line 518: `sample_request()` now sets `display_name`, but no test asserts that `create_template` actually propagates it to the outgoing `CreateTemplateFromImageReq.display_name` via `non_empty_option` — including the empty/whitespace → `None` normalization. A mock-client test covering both the set case and the blank case would verify the Rust-side threading.

Issue #600, opened 2026-06-22
[Bug Report] the ca.crt file  can not fecth success on compute node


DHclly:
## Summary

I do the action by this tutorial   https://cubesandbox.com/zh/guide/multi-node-deploy.html

but the log tell me can not get the ca file 

<img width="1582" height="731" alt="Image" src="https://github.com/user-attachments/assets/a8243b2c-6e8f-447c-81c3-e8618466fda4" />

## Environment

- CubeSandbox version / commit: v0.4.0
- Host OS and kernel version: RockyLinux 9.6
- KVM info (modinfo kvm): filename: /lib/modules/5.14.0-570.17.1.el9_6.x86_64/kernel/arch/x86/kvm/kvm.ko.xz
- Deployment mode: multi-node
- Relevant component: CubeMaster

## Steps to Reproduce

just do by the link : https://cubesandbox.com/zh/guide/multi-node-deploy.html

## Expected Behavior

no error

## Actual Behavior

<img width="1582" height="731" alt="Image" src="https://github.com/user-attachments/assets/a8243b2c-6e8f-447c-81c3-e8618466fda4" />

<img width="1582" height="731" alt="Image" src="https://github.com/user-attachments/assets/0a4d34c7-2b9b-42de-9f24-d6c7dfcf0a1c" />

I have to manual copy the ca file to the compute node ， then it works

## Additional Context

I ask the gpt , I found  the ca Router is not register:

<img width="1498" height="840" alt="Image" src="https://github.com/user-attachments/assets/0b77b08e-95d2-4939-9280-9b9af5ad2905" />

should add route in server.go:

```
cubeGroup.HandleFunc(cube.CADownloadActionPrefix+"{filename}", cube.HttpHandler).Methods(http.MethodGet, http.MethodHead)
```

PR #614, opened 2026-06-22
fix(cube-egress): refresh compute CA from master

FakeLearne:
Register the CubeMaster CA download route so compute nodes can fetch CubeEgress CA materials via /cube/ca/{filename}. (issue #600 )

Refresh compute-node CA materials from the master on every prepare run, validate the downloaded cert/key pair before installing, and keep control nodes on their local generate-if-missing behavior.


chenhengqi (approved):
LGTM.

PR #615, opened 2026-06-22
Feat/webCaseCenter

liciazhu:
add cube web case center and mod web sandbox detail


CubeAPI/src/examples/registry.rs
```diff
@@ -0,0 +1,374 @@
...
+
+// ─── Scenario registry ────────────────────────────────────────────────────────
+
+/// Return the full static scenario registry.
+///
+/// `Box::leak` materialises the `Vec` exactly once at process start; subsequent
+/// calls return the same `&'static` slice, so the front-end gets a stable
+/// ordering without the borrow checker complaining about temporaries.
+pub fn scenario_registry() -> &'static [ScenarioSpec] {
+    Box::leak(Box::new(vec![
```
fslongjin: `scenario_registry()` currently allocates and leaks a new `Vec<ScenarioSpec>` on every call, even though the comment says the registry is materialized once. Endpoints such as list, source, and run can call this repeatedly, so a long-lived API process will grow memory over time. Please store the vector behind `OnceLock`/`LazyLock` and return the initialized slice, ideally with a regression test that repeated calls reuse the same registry.

CubeAPI/src/services/exec.rs
```diff
@@ -0,0 +1,423 @@
...
+        &self,
+        sandbox_id: &str,
+        domain: &str,
+        req: &ExecCodeRequest,
+    ) -> AppResult<ExecCodeResponse> {
+        let timeout_ms = (req.timeout_secs.unwrap_or(30).clamp(1, 300) as u64) * 1000;
+        let start = Instant::now();
+
+        let output = match req.language.as_str() {
+            "python" => self.exec_python(sandbox_id, domain, &req.code).await?,
```
fslongjin: `timeout_secs` is only used after the Jupyter/envd request has already returned; the service still awaits `exec_python`/`exec_bash` without a real deadline. A hung or slow sandbox request can keep the handler occupied, and larger user timeouts will still be cut off by the outer 30s route timeout. Please wrap the actual execution future in `tokio::time::timeout` and align the route timeout with the allowed request budget.

CubeAPI/src/services/examples.rs
```diff
@@ -0,0 +1,548 @@
...
+            cmd.env("CUBE_PROXY_NODE_IP", proxy_ip);
+        }
+        if let Some(proxy_port) = self.cube_proxy_port_http {
+            cmd.env("CUBE_PROXY_PORT_HTTP", proxy_port.to_string());
+        }
+        cmd.env("CUBE_SANDBOX_DOMAIN", &self.sandbox_domain);
+
+        let start = Instant::now();
+        let max_secs = sc.timeout_secs.unwrap_or(120);
+        let run_result = timeout(Duration::from_secs(max_secs), cmd.output()).await;
```
fslongjin, line 336: Timing out `cmd.output()` drops the future but does not explicitly kill the spawned process or its process tree, so long-running Python/Go/Chromium children can continue after the API has returned. This path also installs `requirements.txt` during a request, mutating the server's Python environment and using a source-tree stamp file without concurrency control. Please spawn and kill/wait the child on timeout, and move dependency installation to build/deploy time or an isolated, locked cache that fails closed.

CubeAPI/src/routes.rs
```diff
@@ -377,6 +417,35 @@ fn apply_http_layers(router: Router<AppState>, timeout: Duration) -> Router<AppS
...
+        routes
+    }
+}
+
+/// Long-budget route for `POST /examples/run`.  Mounted on its own router so
+/// the 30 s default HTTP timeout does not kill long-running example scripts.
+fn build_examples_run_routes(state: &AppState, auth_configured: bool) -> Router<AppState> {
+    let routes = Router::new().route("/examples/run", post(examples::run_example));
+    if auth_configured {
+        routes.layer(middleware::from_fn_with_state(state.clone(), unified_auth))
```
fslongjin: `POST /examples/run` is moved to a long-timeout router with auth only, but it no longer gets the rate limiting used by other resource-creating or execution endpoints. Because this route starts local processes and can create sandboxes or browser sessions for up to several minutes, missing rate limits increases abuse and resource-exhaustion risk. Please apply the existing auth+rate-limit layer here, or add a dedicated concurrency/rate budget for example runs.

web/src/data/exampleScenarios.ts
```diff
@@ -0,0 +1,641 @@
...
+  return t;
+}
+
+// ─── Scenario registry ──────────────────────────────────────────────────────
+//
+// 8 scenarios — none of them AI / LLM. The Rust handler enforces the same
+// `hidden: true` filter so even an attacker who knows the IDs cannot reach
+// these scenarios through the HTTP API.
+
+export const EXAMPLE_SCENARIOS: ExampleScenario[] = [
```
fslongjin, line 362: This adds a second hand-maintained example registry in the frontend while the backend registry also defines scenario IDs, files, categories, store items, visibility, and run metadata. The two sources can drift, producing UI entries that cannot run or backend examples with stale labels/topology in the UI. Please make the backend `/examples` response include the UI metadata, or generate both Rust and TS registries from a shared data source.

CubeAPI/src/handlers/cube/exec_code.rs
```diff
@@ -0,0 +1,78 @@
...
+        (status = 200, description = "Code execution result", body = crate::models::ExecCodeResponse),
+        (status = 400, description = "Invalid request", body = crate::models::ApiError),
+        (status = 404, description = "Sandbox not found", body = crate::models::ApiError),
+        (status = 500, description = "Unexpected backend error", body = crate::models::ApiError)
+    )
+)]
+pub async fn exec_code(
+    State(state): State<AppState>,
+    Path(sandbox_id): Path<String>,
+    Json(body): Json<ExecCodeRequest>,
```
fslongjin, line 42: `ExecCodeRequest` derives `Validate`, but this handler only extracts `Json(body)` and never calls `body.validate()` before executing the request. As a result, the documented `code` length constraint is not enforced and the API can accept payloads larger than the model contract suggests. Please validate here and map failures to a 400 response, or use a shared validated-JSON extractor.

examples/cube-bench/cube-bench
fslongjin, line 1: This PR adds the compiled `examples/cube-bench/cube-bench` binary alongside the Go source. Committing the build artifact substantially increases PR size, is platform/architecture-specific, and makes supply-chain review noisier. Please remove the binary and add an ignore rule for this executable or for example build outputs.

CubeAPI/src/services/examples.rs
```diff
@@ -0,0 +1,548 @@
...
+                    "example run complete"
+                );
+
+                Ok(RunExampleResponse {
+                    stdout,
+                    stderr,
+                    exit_code,
+                    success,
+                    elapsed_ms,
+                    steps: Vec::new(),
```
fslongjin: The run response now exposes `steps`, and the frontend adds a `StepTimeline`, but the service always returns `steps: Vec::new()` and the page does not render the timeline from backend data. This leaves a partially implemented API/UI surface that users cannot observe and future maintainers may assume is complete. Please either generate and render real step logs, or remove the unused field/component/comment until the feature is wired through.

examples/code-sandbox-quickstart/exec_code.py
```diff
@@ -14,4 +16,5 @@
 """
 
 with Sandbox.create(template=template_id) as sandbox:
+
```
fslongjin: This added blank line contains trailing whitespace, so `git diff --check` still fails for the PR. Please remove the stray spaces (and any other diff-check formatting issues) to avoid CI or review noise.

web/src/pages/SandboxCases.tsx
```diff
@@ -0,0 +1,1232 @@
...
+                </div>
+              </Card>
+            </>
+          )}
+        </div>
+      </div>
+    </div>
+  );
+}
+
```
fslongjin: `git diff --check` also reports this as a new blank line at EOF. Please remove the extra trailing blank line so the PR is clean for whitespace checks.

fslongjin (approved):

PR #624, opened 2026-06-23
feat(lifecycle): kill path, set_timeout/refresh, and stable endAt

chengjoey:
feat: #233 #232 

implement lifecycle:kill base on #613 

### Summary

Bring sandbox lifecycle management to parity with e2b: on top of the existing `auto_pause` flow, add a **timeout-then-kill** default path, implement the previously-stubbed **`set_timeout` / `refresh`** APIs, and surface a deterministic **`end_at`** timestamp across the entire stack.

The change spans four layers — CubeMaster (Go), CubeProxy sidecar (Go), CubeProxy data plane (Lua), CubeAPI (Rust) — plus documentation.

### Motivation

The previous implementation only handled `lifecycle.on_timeout="pause"`:
- Sandboxes **without** `auto_pause` were never reaped on idle timeout and held resources indefinitely.
- `SandboxTimeoutRequest` / `SandboxRefreshRequest` in CubeAPI were marked `❌ not implemented`, so `sandbox.set_timeout()` / `sandbox.refresh()` from the SDK were silent no-ops.
- `GetSandboxResponse.end_at` was always `None`, forcing SDK clients to guess the real expiry.

This PR closes those three gaps in one go, aligning with e2b's `on_timeout="kill"` default and the `set_timeout(duration)` API surface.

#### 1. New timeout → kill path (the new default)

| Component | Change |
|---|---|
| `CubeProxy/sidecar/sweeper` | Stop short-circuiting on `!AutoPause`; dispatch to `tryPause` or `tryKill` based on `auto_pause` when idle threshold is exceeded |
| `cubemasterclient.Kill` | New `DELETE /cube/sandbox` call with `kill_reason=timeout`, reusing the SETNX → push → RPC → finalise pattern from `Pause` |
| `sandbox_state.lua` | Gate recognises `killing` / `killed` states and returns `410 Gone` (retcode `310410`) so SDK clients stop retrying immediately |
| `CubeMaster/sandbox_remove` | `DestroyReq` carries `KillReason`, written into the `cube.master.instance.kill_reason` annotation for audit/billing |
| Terminal-state fast path | Existing fast path extended to `killing` / `killed` to avoid RPC churn after a kill is in flight |

#### 2. New `set_timeout` / `refresh` endpoints (CubeMaster)

| Route | Input | Behaviour |
|---|---|---|
| `POST /cube/sandbox/timeout` | `sandboxID`, `timeout` (seconds) |  set `TimeoutSeconds = timeout`, `EndAt = now + timeout*1000` |
| `POST /cube/sandbox/refresh` | `sandboxID`, `duration` (seconds) | Equivalent to `set_timeout` (matches e2b convergence) |

### DIFF
--- a/CubeMaster/pkg/service/httpservice/cube/cube.go
+++ b/CubeMaster/pkg/service/httpservice/cube/cube.go
@@ -25,6 +25,8 @@ const (
 	SandboxInfoAction              = "/sandbox/info"
 	SandboxExecAction              = "/sandbox/exec"
 	SandboxUpdateAction            = "/sandbox/update"
+	SandboxTimeoutAction           = "/sandbox/timeout"
+	SandboxRefreshAction           = "/sandbox/refresh"
 	SandboxCommitAction            = "/sandbox/commit"
 	SandboxRollbackAction          = "/sandbox/rollback"
 	SnapshotAction                 = "/snapshot"
@@ -80,6 +82,10 @@ func HttpHandler(w http.ResponseWriter, r *http.Request) {
 		rsp = handleSandboxLogsAction(w, r, rt)
 	case r.URL.Path == actionURI(SandboxUpdateAction):
 		rsp = handleUpdateAction(w, r, rt)
+	case r.URL.Path == actionURI(SandboxTimeoutAction):
+		rsp = handleSandboxTimeoutAction(w, r, rt)
+	case r.URL.Path == actionURI(SandboxRefreshAction):
+		rsp = handleSandboxRefreshAction(w, r, rt)
 	case r.URL.Path == actionURI(SandboxCommitAction):
 		rsp = handleSandboxCommitAction(w, r, rt)
 	case r.URL.Path == actionURI(SandboxRollbackAction):
--- a/CubeMaster/pkg/service/sandbox/sandbox_remove.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_remove.go
@@ -44,6 +44,11 @@ func DestroySandbox(ctx context.Context, req *types.DeleteCubeSandboxReq) (rsp *
 		destroyReq.Annotations = make(map[string]string)
 		destroyReq.Annotations[constants.CubeAnnotationsInsType] = req.InstanceType
 	}
+	reason := req.KillReason
+	if reason == "" {
+		reason = "request"
+	}
+	destroyReq.Annotations[constants.CubeAnnotationsKillReason] = reason
 	collectMemoryOption(req, destroyReq)
 	if config.GetConfig().Common.CubeDestroyCheckFilter {
 
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -1,0 +1,126 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package sandbox
+
+import (
+	"context"
+	"time"
+
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/log"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/utils"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/localcache"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+)
+
+// SetTimeout implements POST /cube/sandbox/timeout. It is the master-side
+// counterpart of CubeAPI's SetTimeoutRequest -> SetTimeoutResponse.
+func SetTimeout(ctx context.Context, req *types.SetTimeoutRequest) (rsp *types.SetTimeoutRes) {
+	rsp = &types.SetTimeoutRes{
+		RequestID: req.RequestID,
+		SandboxID: req.SandboxID,
+		Ret: &types.Ret{
+			RetCode: int(errorcode.ErrorCode_Success),
+			RetMsg:  errorcode.ErrorCode_Success.String(),
+		},
+	}
+	defer func() {
+		logger := log.G(ctx).WithFields(map[string]interface{}{
+			"RequestId": req.RequestID,
+			"RetCode":   int64(rsp.Ret.RetCode),
+		})
+		logger.Infof("SetTimeout:%+v", utils.InterfaceToString(req))
+		if rsp.Ret.RetCode != int(errorcode.ErrorCode_Success) {
+			logger.Errorf("SetTimeout fail:%+v", utils.InterfaceToString(rsp))
+		}
+	}()
+
+	if req.SandboxID == "" {
+		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
+		rsp.Ret.RetMsg = "should provide sandboxID"
+		return
+	}
+	if req.Timeout <= 0 {
+		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
+		rsp.Ret.RetMsg = "timeout must be positive (seconds)"
+		return
+	}
+
+	if !sandboxExists(ctx, req.SandboxID) {
+		rsp.Ret.RetCode = int(errorcode.ErrorCode_NotFound)
+		rsp.Ret.RetMsg = "sandbox not found"
+		return
+	}
+
+	endAt := refreshTimeoutMeta(ctx, req.SandboxID, int(req.Timeout))
+	rsp.EndAt = endAt
+	return
+}
+
+// Refresh implements POST /cube/sandbox/refresh. Semantically identical to
+// SetTimeout: refresh(d) rebases the idle clock and sets TimeoutSeconds = d.
+func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.RefreshSandboxRes) {
+	rsp = &types.RefreshSandboxRes{
+		RequestID: req.RequestID,
+		SandboxID: req.SandboxID,
+		Ret: &types.Ret{
+			RetCode: int(errorcode.ErrorCode_Success),
+			RetMsg:  errorcode.ErrorCode_Success.String(),
+		},
+	}
+	defer func() {
+		logger := log.G(ctx).WithFields(map[string]interface{}{
+			"RequestId": req.RequestID,
+			"RetCode":   int64(rsp.Ret.RetCode),
+		})
+		logger.Infof("RefreshSandbox:%+v", utils.InterfaceToString(req))
+		if rsp.Ret.RetCode != int(errorcode.ErrorCode_Success) {
+			logger.Errorf("RefreshSandbox fail:%+v", utils.InterfaceToString(rsp))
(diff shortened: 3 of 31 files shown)


chenhengqi (commented):
Please update your commit message and PR summary. I don't have a whole picture what this PR trying to achieve.

I presume we don't have to touch CubeProxy for killing sandboxes.

chengjoey:
> Please update your commit message and PR summary. I don't have a whole picture what this PR trying to achieve.
> 
> I presume we don't have to touch CubeProxy for killing sandboxes.

done

kinwin-ustc:
Our semantics differ slightly from E2B. E2B timeout starts from sandbox startup, while ours starts from no request. In the automatic pause scenario, I think idle timeout is more reasonable, but from the perspective of how long an E2B sandbox can survive, should we follow our auto_puase semantics?

chengjoey:
> Our semantics differ slightly from E2B. E2B timeout starts from sandbox startup, while ours starts from no request. In the automatic pause scenario, I think idle timeout is more reasonable, but from the perspective of how long an E2B sandbox can survive, should we follow our auto_puase semantics?

- **e2b's model**: `EndTime` is an **absolute deadline** = `StartTime + timeout`. The evictor compares `now` to `EndTime`; activity does **not** reset it. The e2b *SDK clients* (not the server) call `keep_alive` on every `run_code` / `commands.run` to push `EndTime` forward — so "activity extends life" is implemented client-side via explicit RPCs. They also have a hard wall-clock ceiling `MaxInstanceLength` (24h Pro / 1h Base) on top of that.

- **Our model**: server-side idle reaper. `baseline = max(LastActiveMs, CreatedAt)`, `idleFor = now - baseline`. CubeProxy sidecar reports `last_active` on every HTTP request, so the server itself observes activity and rebases the idle clock without any client-side cooperation.
 
 I think **keep idle-based semantics** is better:

1. **It's the right primitive for `on_timeout=pause`** — auto-pause means "free resources when idle"; if we switched to absolute-deadline, a sandbox actively serving requests would get paused mid-flight unless the SDK keeps pushing `keep_alive`. That regression isn't worth it.

2. **We already invested in server-side activity tracking** (sidecar `last_active` stream). E2b had to use absolute-deadline because their server doesn't see request traffic; we do, so we can give a strictly better experience.

What I think we **should** adopt from e2b is something orthogonal: a **`MaxInstanceLength`** hard ceiling that bounds the worst case independent of idle activity (so an always-busy sandbox doesn't live forever).

https://github.com/e2b-dev/infra/blob/ecd13b04d8aead46e7afe386538d5cb10ef22b40/packages/api/internal/sandbox/sandboxtypes/sandbox.go#L91

docs/zh/guide/lifecycle.md
```diff
@@ -141,7 +141,7 @@ sandbox = Sandbox.create(
 - 通过 SDK 调用：`sandbox.run_code(...)`、`sandbox.commands.run(...)`、`sandbox.files.read(...)` / `write(...)`。
 - 通过 HTTP 直连沙箱内的服务（例如 `getHost()` 返回的 URL）。
 
-未配置 `auto_pause` / 不传 `lifecycle` 的沙箱完全保留旧行为：空闲超时直接销毁。
+未配置 `auto_pause` / 不传 `lifecycle` 的沙箱默认行为是 `on_timeout="kill"`：空闲超过 `timeout` 秒后，平台（CubeProxy sidecar 的 sweeper）会主动销毁该沙箱并广播 `killing` → `killed` 状态，CubeProxy 会对后续请求立即返回 410 Gone。这与 e2b `lifecycle.on_timeout="kill"` 语义一致。如果完全不希望被自动回收，请在创建时把 `timeout` 设得足够大、或主动在客户端发心跳调用刷新 idle 计时。
```
chenhengqi: These user-facing documents only need to describe the behavior and do not need to expose implementation details(`Cubeproxy` etc).

CubeMaster/pkg/service/httpservice/cube/cube.go
```diff
@@ -80,6 +82,10 @@ func HttpHandler(w http.ResponseWriter, r *http.Request) {
 		rsp = handleSandboxLogsAction(w, r, rt)
 	case r.URL.Path == actionURI(SandboxUpdateAction):
 		rsp = handleUpdateAction(w, r, rt)
+	case r.URL.Path == actionURI(SandboxTimeoutAction):
```
chenhengqi, line 85: We should employ some web framework here. cc @fslongjin

CubeProxy/sidecar/internal/sweeper/sweeper.go
```diff
@@ -141,34 +140,53 @@ func (s *Sweeper) sweepOnce(ctx context.Context) {
...
+			// Nothing to do. "pausing" / "killing" mean a peer (or our own
+			// previous invocation) is mid-flight; let it finish.
 			continue
 		}
 
-		s.o.Log.Info("idle threshold exceeded; pausing",
-			zap.String("sandbox_id", e.Meta.SandboxID),
-			zap.Duration("idle_for", idleFor),
-			zap.Int("timeout_seconds", e.Meta.TimeoutSeconds))
+		if e.Meta.AutoPause {
```
chenhengqi: Should we use a switch here?

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -0,0 +1,133 @@
...
+			"RequestId": req.RequestID,
+			"RetCode":   int64(rsp.Ret.RetCode),
+		})
+		logger.Infof("SetTimeout:%+v", utils.InterfaceToString(req))
+		if rsp.Ret.RetCode != int(errorcode.ErrorCode_Success) {
+			logger.Errorf("SetTimeout fail:%+v", utils.InterfaceToString(rsp))
+		}
+	}()
+
+	if req.SandboxID == "" {
```
chenhengqi, line 40: I think this kind of check should reside in `CubeMaster/pkg/service/httpservice/cube/sandbox_timeout.go‎`.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -0,0 +1,133 @@
...
+		rsp.Ret.RetMsg = "should provide sandboxID"
+		return
+	}
+	if req.Timeout <= 0 {
+		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
+		rsp.Ret.RetMsg = "timeout must be positive (seconds)"
+		return
+	}
+	if req.InstanceType == "" {
+		req.InstanceType = cubebox.InstanceType_cubebox.String()
```
chenhengqi: Same here. The caller already set default value?

chenhengqi (commented):
I left some comments, but this PR looks good overall.

chenhengqi:
One more request, please add an example for this feature. Thanks.

docs/zh/guide/lifecycle.md
chengjoey: done

CubeProxy/sidecar/internal/sweeper/sweeper.go
chengjoey: done

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
chengjoey: Here, I did not choose to move it to `cube/sandbox_timeout.go‎` because I observed that in `sandbox/sandbox_*.go`, the logic is used as a fallback, while `httpservice/cube` only performs basic validations at the HTTP layer, just like `sandbox_update`.

https://github.com/TencentCloud/CubeSandbox/blob/08e5309ad76df6b19d40f667568e8374ecf93e51/CubeMaster/pkg/service/sandbox/sandbox_update.go#L39-L48

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
chengjoey: done

chengjoey:
> One more request, please add an example for this feature. Thanks.

done, `auto-kill.py` added

CubeMaster/pkg/service/httpservice/cube/cube.go
fslongjin: I'll refactor these codes later.

examples/code-sandbox-quickstart/README.md
```diff
@@ -125,6 +125,7 @@ hello cube
 | `read.py` | `sandbox.files.read()` — read a file from the sandbox filesystem |
 | `pause.py` | `sandbox.pause()` / `sandbox.connect()` — snapshot and restore |
 | `auto_resume.py` | `lifecycle={"on_timeout": "pause", "auto_resume": True}` — let the platform pause idle sandboxes and resume them on the next request |
+| `auto_kill.py` | `lifecycle={"on_timeout": "kill"}` — let the platform tear down idle sandboxes (the default — destruction is final, no snapshot kept) |
```
chenhengqi: `no snapshot kept`?

chenhengqi:
cc @zhouxianping

CubeAPI/src/services/sandboxes.rs
```diff
@@ -991,6 +991,7 @@ mod tests {
             distribution_scope: None,
             auto_pause: false,
             auto_resume: false,
+            distribution_scope: None,
```
chenhengqi: duplicated?

CubeMaster/pkg/service/sandbox/sandbox_remove.go
```diff
@@ -44,6 +44,14 @@ func DestroySandbox(ctx context.Context, req *types.DeleteCubeSandboxReq) (rsp *
 		destroyReq.Annotations = make(map[string]string)
 		destroyReq.Annotations[constants.CubeAnnotationsInsType] = req.InstanceType
 	}
+	if destroyReq.Annotations == nil {
```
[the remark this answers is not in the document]
chenhengqi: This is a valid comment, please address.

examples/code-sandbox-quickstart/README.md
chengjoey: timeout-kill in sidecar call CubeMaster.Kill(sid, KillReasonTimeout), this shares the same destruction path as the explicit kill() function

This does not refer to a template snapshot.

`lifecycle.md `line 98 say `always wins and discards the snapshot`

perhaps it would be more reasonable to say "destruction is irreversible, the sandbox cannot be resumed"?

CubeMaster/pkg/service/sandbox/sandbox_remove.go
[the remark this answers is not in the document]
chengjoey: done

examples/code-sandbox-quickstart/README.md
chenhengqi: Yes, please rephrase it. Let's don't mention snapshot.

CubeAPI/src/services/sandboxes.rs
chengjoey: Yes, there was duplicated, it has been deleted.

examples/code-sandbox-quickstart/README.md
chengjoey: done

zhouxianping (approved):

chenhengqi:
Merged, thanks.

PR #626, opened 2026-06-23
feat(Cubemaster): opt-in envd injection for image-based templates

0717dn6hyc8-byte:
## Summary

This PR introduces an optional `envd` injection path for templates created from OCI images.

When enabled, `cubemastercli` will select an `envd` binary from the CLI's local `--envd-path` parameter or from the default binary embedded in `cubemastercli` during build. The selected binary is uploaded to CubeMaster as part of a `create-from-image` multipart request. CubeMaster verifies the uploaded bytes, writes the binary to the template root filesystem before sealing the ext4 artifact.

By default, this injection feature is disabled to maintain the existing behavior for the community.

## Motivation

Some CubeSandbox users need an envd process to be available automatically in restored sandboxes, so developer tooling can attach without requiring image authors to modify every base image manually.

### 1. CLI opt-in controls

| Component | Change |
| --- | --- |
| `cubemastercli template create-from-image` | Adds `--enable-inject-envd` to explicitly enable `envd` injection. |
| `cubemastercli template create-from-image` | Adds `--envd-path` as a CLI-local path to the `envd` binary to upload. The path is read only by `cubemastercli` and is not sent to CubeMaster. |
| `cubemastercli` | If `--enable-inject-envd` is set without `--envd-path`, uses the embedded default `envd` binary when available. |
| `cubemastercli` | If no embedded default is available and `--envd-path` is omitted, fails before submitting the request. |
| `cubemastercli` | Rejects `--envd-path` unless `--enable-inject-envd` is also set. |

### 2. Build-time embedded envd

| Component | Change |
| --- | --- |
| `CubeMaster/Makefile` | Supports `make cubemastercli ENVD_LOCAL_PATH=/path/to/envd` to y with an embedded default `envd`. |
| `CubeMaster/Makefile` | Validates that `ENVD_LOCAL_PATH` points to a non-empty ELF binary no larger than 16 MiB. |
| `CubeMaster/Makefile` | Uses the `embed_envd` build tag only when `ENVD_LOCAL_PATH` is provided, and removes the temporary embedded asset after the build. |
| `deploy/one-click/build-release-bundle-builder.sh` | Supports `ENVD_LOCAL_PATH=/path/to/envd ./deploy/one-click/build-release-bundle-builder.sh` so one-click release bundles can package a `cubemastercli` with embedded `envd`. |

### 3. Multipart envd upload

| Component | Behavior |
| --- | --- |
| `cubemastercli` | Uses the existing JSON request path when `envd` injection is disabled. |
| `cubemastercli` | Uses `multipart/form-data` when `--enable-inject-envd` is set. |
| `cubemastercli` | Sends the serialized `CreateTemplateFromImageReq` in the `request` field and the selected binary in the `envd` file field. |
| `CubeMaster HTTP service` | Parses multipart `create-from-image` requests and constructs a request-scoped `envd` payload from the uploaded file. |

### 4. Template rootfs injection

| Component | Behavior |
| --- | --- |
| `templatecenter` | Writes the uploaded `envd` bytes into `/usr/local/bin/envd` inside tefore sealing the ext4 artifact. |
| `templatecenter` | Injection only happens when `cube.master.inject_envd=true` is present in template container overrides and an uploaded payload is available. |
| `templatecenter` | Missing, empty, oversized, or non-ELF `envd` payload fails the templg an incomplete artifact. |
| `templatecenter` | Does not read `envd` from CubeMaster host paths. |

Uploaded `envd` binaries must be:

- non-empty
- ELF binaries
- no larger than 16 MiB
- compatible with the target template rootfs OS and CPU architecture

### 5. Artifact fingerprinting

| Component | Behavior |
| --- | --- |
| `templatecenter` | Includes `sha256(uploaded bytes)` in the template spec fingerprint when `envd` injection is enabled. |
| `templatecenter` | Leaves fingerprints unchanged for templates that do not enable `envd` injection. |

This prevents reusing artifacts baked with an older or different `envd` binary after the selected `envd` version changes.

### 6. Runtime entrypoint wrapping

| Component | Behavior |
| --- | --- |
| `CubeMaster sandbox service` | When a template carries `cube.master.inject_envd=true`, wraps the main container command. |
| `CubeMaster sandbox service` | Starts `/usr/local/bin/envd` in the background and then `exec`s the original command. |
| `CubeMaster sandbox service` | Adds the default `envd` exposed port. |
| `CubeMaster sandbox service` | Skips wrapping if the command is already wrapped. |

### 7. Documentation

Updated documentation for the new client-supplied upload flow:

- `docs/guide/tutorials/template-from-image.md`
- `docs/zh/guide/tutorials/template-from-image.md`
- `deploy/one-click/README.md`
- `deploy/one-click/README_zh.md`
- `deploy/one-click/env.example`

Assisted-by: Claude Code:ppio/pa/gpt-5.5


fslongjin:
I want to understand the motivation behind this PR and under what scenarios it is necessary. I feel that this goes against common usage—since it's built from a container image, why would we need to additionally inject an envd? This will also affect the container's startup commands and such. I find the design of this PR very strange.

0717dn6hyc8-byte:
> I want to understand the motivation behind this PR and under what scenarios it is necessary. I feel that this goes against common usage—since it's built from a container image, why would we need to additionally inject an envd? This will also affect the container's startup commands and such. I find the design of this PR very strange.

For example, in development-environment / code-sandbox scenarios, we want to quickly build interactive development sandboxes from existing base images. `envd` provides the access layer for the development environment, but many base images do not have it preinstalled. Maintaining a large number of derived images only to integrate `envd` would add unnecessary cost and complexity.

Therefore, this PR injects `envd` during template rootfs creation. This allows those base images to gain a consistent development runtime capability without modifying their Dockerfiles or rebuilding the source images.

fslongjin:
> > I want to understand the motivation behind this PR and under what scenarios it is necessary. I feel that this goes against common usage—since it's built from a container image, why would we need to additionally inject an envd? This will also affect the container's startup commands and such. I find the design of this PR very strange.
> 
> For example, in development-environment / code-sandbox scenarios, we want to quickly build interactive development sandboxes from existing base images. `envd` provides the access layer for the development environment, but many base images do not have it preinstalled. Maintaining a large number of derived images only to integrate `envd` would add unnecessary cost and complexity.
> 
> Therefore, this PR injects `envd` during template rootfs creation. This allows those base images to gain a consistent development runtime capability without modifying their Dockerfiles or rebuilding the source images.

Indeed, you make a good point. From that perspective, automatically injecting envd is indeed a great idea! Thank you for your PR. I'll study the code in your PR over the next couple of days.

kinwin-ustc:
1. You should add your DCO; 

2. Please modify the CubeBot review comments;

3. Please add corresponding documentation for the newly added CubeMasterCLI parameters.

0717dn6hyc8-byte:
> 1. You should add your DCO;
> 2. Please modify the CubeBot review comments;
> 3. Please add corresponding documentation for the newly added CubeMasterCLI parameters.

Thanks for the review. I have resolved these problems:

- Added DCO sign-off to all commits.
- Addressed the CubeBot review comments, including envd path consistency, fingerprint error handling, redundant envd reads, misleading log wording, and the unused error return.
- Added documentation for the new CubeMasterCLI parameters in both English and Chinese:
  - `docs/guide/tutorials/template-from-image.md`
  - `docs/zh/guide/tutorials/template-from-image.md`

Please take another look when convenient.

kinwin-ustc:
There is a conflict between the current code and the main branch. You need to rebase your code based on the latest main branch and resubmit it

CubeMaster/pkg/service/sandbox/types/types.go
```diff
@@ -540,6 +521,7 @@ type CreateTemplateFromImageReq struct {
 	ExposedPorts       []int32             `json:"exposed_ports,omitempty"`
 	DistributionScope  []string            `json:"distribution_scope,omitempty"`
 	ContainerOverrides *ContainerOverrides `json:"container_overrides,omitempty"`
+	EnvdPath           string              `json:"envd_path,omitempty"`
```
kinwin-ustc: Why expose this field in the HTTP API?

CubeMaster/pkg/service/sandbox/util.go
```diff
@@ -387,6 +387,10 @@ func checkAndGetContainerAnnotation(c *cubebox.ContainerConfig, cnt *types.Conta
 			c.Annotations[k] = v
 		}
 
+		if strings.HasPrefix(k, "cube.container.") {
+			c.Annotations[k] = v
+		}
```
kinwin-ustc: What is the relationship between this and envd inject?

CubeMaster/pkg/service/sandbox/types/types.go
0717dn6hyc8-byte: > Why expose this field in the HTTP API?

The --envd-path flag was added to let users specify the host-side envd binary path.

Your comment reminded me that this is not necessary, because the envd source can already be configured by a host-side environment variable. So I removed both envd_path and --envd-path to keep the change smaller.

CubeMaster/pkg/service/sandbox/util.go
0717dn6hyc8-byte: > What is the relationship between this and envd inject?

This was an unrelated cleanup for container annotation passthrough. It is not required for envd injection.

I have removed it to avoid mixing unrelated changes into this PR.

0717dn6hyc8-byte:
> > > I want to understand the motivation behind this PR and under what scenarios it is necessary. I feel that this goes against common usage—since it's built from a container image, why would we need to additionally inject an envd? This will also affect the container's startup commands and such. I find the design of this PR very strange.
> > 
> > 
> > For example, in development-environment / code-sandbox scenarios, we want to quickly build interactive development sandboxes from existing base images. `envd` provides the access layer for the development environment, but many base images do not have it preinstalled. Maintaining a large number of derived images only to integrate `envd` would add unnecessary cost and complexity.
> > Therefore, this PR injects `envd` during template rootfs creation. This allows those base images to gain a consistent development runtime capability without modifying their Dockerfiles or rebuilding the source images.
> 
> Indeed, you make a good point. From that perspective, automatically injecting envd is indeed a great idea! Thank you for your PR. I'll study the code in your PR over the next couple of days.

I have addressed the feedback from cubesandboxbot and the reviewers and pushed the latest changes.

Could you please review the updated PR again when you have time? If everything looks good, I would appreciate it if you could consider merging it into the main branch.

Please let me know if any further changes are needed. Thank you for your time and support!

0717dn6hyc8-byte:
To strengthen security, we now add basic validation before injecting the host-side envd binary: only accepting regular files, disallowing symbolic links, prohibiting global writes, and performing an ELF magic check.

CubeMaster/pkg/templatecenter/envd_inject.go
```diff
@@ -0,0 +1,127 @@
...
+)
+
+const (
+	envdHostDirDefault    = "/usr/local/share/cubesandbox-envd"
+	envdHostDirEnv        = "CUBE_MASTER_ENVD_HOST_DIR"
+	envdBinaryName        = "envd"
+	envdInjectionFileMode = 0o755
+	envdInjectionDirMode  = 0o755
+)
+
```
fslongjin, line 24: ### Design concern: host-fixed path vs client-supplied binary

Hard-coding `/usr/local/share/cubesandbox-envd` (plus `CUBE_MASTER_ENVD_HOST_DIR`) makes CubeMaster’s **host filesystem** the supply chain for guest rootfs content. That has a few concrete problems:

1. **Trust / security boundary** — enabling `--enable-inject-envd` means whatever is present on the Master node at build time gets baked into every matching template. Operators must carefully provision that directory on every Master that can run image jobs; a mis-deployed / swapped / unexpectedly world-reachable binary becomes guest rootfs content. The ELF / non-symlink / non-world-writable checks reduce some local FS footguns, but they do not change the trust model: Master still opens and injects **its own** host file.
2. **Flexibility** — there is no per-build way to choose which envd binary to inject without mutating shared host state. Different teams / versions need either separate Master configs or careful file swaps (and fingerprinting then depends on whoever last wrote that path).
3. **Multi-Master / HA** — which node’s `/usr/local/share/...` is authoritative? Image jobs are not guaranteed to always land on a node that was provisioned identically.

I understand `EnvdPath` / `--envd-path` was previously removed (see discussion on `CreateTemplateFromImageReq`) because exposing a **Master-local path** over the HTTP API is the wrong abstraction — clients should not tell Master which of *its* files to open. Agreed.

The better alternative is still client-driven, but with a different payload shape:

- **`cubemastercli`** accepts a *local* `--envd-path` (CLI-side only).
- CLI **uploads the binary bytes** (or a content-addressed staging blob + digest) as part of `create-from-image`.
- CubeMaster validates the uploaded bytes (size limit + ELF + mode when written into rootfs), injects from the **request-scoped payload**, and fingerprints `sha256(uploaded bytes)` as today.
- Drop the host-directory convention entirely for the opt-in path (or keep it only as an internal/dev fallback, not the supported contract).

That keeps the earlier lesson (“don’t expose Master host paths in the API”) while avoiding the new risk of baking Master-environment files into templates, and makes per-build version selection trivial.

Happy to discuss upload transport (multipart field vs short-lived blob upload + digest ref) if this direction sounds right.

fslongjin (changes requested):
request for changes

fslongjin:
Thanks for the latest update. The client-supplied envd upload approach looks much better to me, and I don't have any further concerns about the overall implementation.

There are just two things left to clean up before we merge this PR:

1. The Chinese documentation is still describing the previous host-side approach. It currently says that CubeMaster reads envd from `CUBE_MASTER_ENVD_HOST_DIR` or `/usr/local/share/cubesandbox-envd/envd`, which no longer matches the latest implementation.

   Could you please update the Chinese guide to match the English one? In particular, it should explain that `--envd-path` is a local path on the machine running `cubemastercli`, that the binary is uploaded to CubeMaster through the multipart request, and that the embedded envd binary is used as the fallback when available. Please also mention the 16 MiB limit and any necessary binary compatibility requirements.

2. The PR is currently still conflicting with the latest `master`, and the DCO check is failing because the latest commit is missing its sign-off. Please rebase onto the latest `master`, resolve the conflict in `CubeMaster/pkg/templatecenter/artifact_build.go`, and add the required sign-off to the latest commit.

The other build, formatting, and documentation checks are passing. Once these two items are addressed, please ping me and I'll take another look. Thanks again for working through the previous feedback.

0717dn6hyc8-byte:
> Thanks for the latest update. The client-supplied envd upload approach looks much better to me, and I don't have any further concerns about the overall implementation.
> 
> There are just two things left to clean up before we merge this PR:
> 
> 1. The Chinese documentation is still describing the previous host-side approach. It currently says that CubeMaster reads envd from `CUBE_MASTER_ENVD_HOST_DIR` or `/usr/local/share/cubesandbox-envd/envd`, which no longer matches the latest implementation.
>    Could you please update the Chinese guide to match the English one? In particular, it should explain that `--envd-path` is a local path on the machine running `cubemastercli`, that the binary is uploaded to CubeMaster through the multipart request, and that the embedded envd binary is used as the fallback when available. Please also mention the 16 MiB limit and any necessary binary compatibility requirements.
> 2. The PR is currently still conflicting with the latest `master`, and the DCO check is failing because the latest commit is missing its sign-off. Please rebase onto the latest `master`, resolve the conflict in `CubeMaster/pkg/templatecenter/artifact_build.go`, and add the required sign-off to the latest commit.
> 
> The other build, formatting, and documentation checks are passing. Once these two items are addressed, please ping me and I'll take another look. Thanks again for working through the previous feedback.

Thanks for the review. I’ve updated the Chinese template-from-image guide to match the English documentation. It now covers the CLI-local `--envd-path`, multipart upload flow, embedded envd fallback, the 16 MiB limit, and the ELF / OS / architecture compatibility requirements.

I’ve also resolved the conflict , and amended the latest commit with the required DCO sign-off.

I updated the PR description as well to reflect the final client-supplied envd upload design instead of the previous host-side path approach.

As a follow-up, I’d suggest maintaining an official `envd` binary artifact for release builds. That would give `ENVD_LOCAL_PATH` a canonical source for embedding the default envd into `cubemastercli`, while still allowing local development builds to use custom envd binaries.

0717dn6hyc8-byte:
Hi @fslongjin , could you please take another look at this PR?
I've updated it based on your previous feedback.
Thanks!

CubeMaster/pkg/templatecenter/envd_inject.go
```diff
@@ -0,0 +1,92 @@
...
+}
+
+func prepareEnvdInjectionPayload(req *types.CreateTemplateFromImageReq) (*EnvdInjectionPayload, error) {
+	if !ShouldInjectEnvdIntoTemplate(req) {
+		return nil, nil
+	}
+	return nil, fmt.Errorf("envd-inject: uploaded envd payload is required; use multipart create-from-image upload")
+}
+
+func injectEnvdIntoRootfs(ctx context.Context, rootfsDir string, req *types.CreateTemplateFromImageReq) (string, error) {
```
kinwin-ustc: Dead code, never called?

CubeMaster/pkg/base/constants/constants.go
```diff
@@ -86,6 +86,11 @@ const (
 
 	CubeAnnotationsVirtiofsCache = "cube.master.virtiofs.cache"
 
+	CubeAnnotationsInjectEnvd       = "cube.master.inject_envd"
+	CubeAnnotationsInjectEnvdOptIn  = "true"
+	CubeAnnotationsEnvdSidecarImage = "cube.master.envd_sidecar_image"
```
kinwin-ustc: What is this for?

CubeMaster/pkg/templatecenter/redo.go
```diff
@@ -248,7 +248,7 @@ func runRedoTemplateImageJob(ctx context.Context, jobID string, req *types.RedoT
 		if source.Cleanup != nil {
 			defer source.Cleanup(ctx)
 		}
-		artifact, _, _, err = ensureRootfsArtifact(ctx, &workingReq, source, downloadBaseURL)
+		artifact, _, _, err = ensureRootfsArtifact(ctx, &workingReq, source, downloadBaseURL, nil)
```
kinwin-ustc, line 255: Have you tested using redo for templates that support envd-enable?

CubeMaster/pkg/service/sandbox/envd_sidecar.go
```diff
@@ -0,0 +1,71 @@
...
+	}
+	mainContainer := req.Containers[0]
+	if isEnvdEntrypointWrapped(mainContainer) {
+		log.G(ctx).Infof("envd: container %q entrypoint already wrapped, skipping", mainContainer.Name)
+		return
+	}
+	wrapEntrypointForEnvd(mainContainer)
+	appendExposedPortIfMissing(out, envdDefaultPort)
+	log.G(ctx).Infof("envd: wrapped entrypoint of container %q (envd path=%s, pre-baked at build time)", mainContainer.Name, constants.CubeEnvdInImagePath)
+	return
```
kinwin-ustc: Redundant

CubeMaster/cmd/cubemastercli/commands/cubebox/envd_upload.go
```diff
@@ -0,0 +1,111 @@
...
+	"mime/multipart"
+	"os"
+	"strings"
+
+	jsoniter "github.com/json-iterator/go"
+	embeddedenvd "github.com/tencentcloud/CubeSandbox/CubeMaster/cmd/cubemastercli/internal/envd"
+	"github.com/urfave/cli"
+)
+
+const maxEnvdUploadBytes = 16 * 1024 * 1024
```
kinwin-ustc: There are multiple definitions of the same constants, can they be merged into the constants package?

kinwin-ustc (changes requested):
Have you verified the redo path for envd injection?

CubeMaster/pkg/templatecenter/envd_inject.go
0717dn6hyc8-byte: Thanks for the reminder. I've removed them and kept injectEnvdPayloadIntoRootfs as the single entrypoint.

CubeMaster/pkg/base/constants/constants.go
0717dn6hyc8-byte: `cube.master.envd_sidecar_image` is a temporary compatibility key for the old per-request override path. I've removed it.

CubeMaster/pkg/templatecenter/redo.go
0717dn6hyc8-byte: Thanks for raising this. I reproduced the redo path for an envd-enabled template whose original build failed in BUILDING_EXT4. Redo works when it can reuse an existing artifact, but if it needs to rebuild the artifact it currently cannot proceed without the original envd payload. I added an explicit failure branch for that rebuild path with a clearer error message and covered it with a unit test.

CubeMaster/pkg/service/sandbox/envd_sidecar.go
0717dn6hyc8-byte: I removed the redundant `return`, and there is no behavior change.

CubeMaster/cmd/cubemastercli/commands/cubebox/envd_upload.go
0717dn6hyc8-byte: Thanks for pointing that out. I moved the shared envd payload size limit into constants.MaxEnvdPayloadBytes and reused it across the CLI upload path, multipart parsing, and rootfs injection.

docs/zh/dev/index.md
```diff
@@ -7,6 +7,7 @@
 ## 约定
 
 - [Redis Key 命名规范](./redis-key-spec)——所有服务在共享 Redis 实例上必须遵循的统一命名空间：命名格式、归属划分、已注册 Key 清单、TTL 策略，以及各服务的 key 构造模块。
+- [redo / envd 测试设计](./redo-envd-test-design)——在模板启用 envd 时，redo 路径需要覆盖的测试用例，包括重建、复用与失败预期。
```
kinwin-ustc: Where is this document?

docs/dev/index.md
```diff
@@ -7,6 +7,7 @@ If you are *using* CubeSandbox (deploying it, building templates, calling the AP
 ## Conventions
 
 - [Redis Key Convention](./redis-key-spec) — the unified namespace every service must use for the shared Redis instance: naming format, scope ownership, the registered key catalog, TTL policy, and the per-service key-builder modules.
+- [Redo / envd Test Design](./redo-envd-test-design) — test cases for redo behavior when templates are created with envd enabled, including rebuild, reuse, and failure expectations.
```
kinwin-ustc: where is this document？

kinwin-ustc (commented):
Have you tested this PR in a real production environment?

docs/dev/index.md
0717dn6hyc8-byte: Thanks for the reminder. That redo / envd test design page was only a local scratch note and never intended to be part of the repo, so I removed it from the docs index in both languages.

0717dn6hyc8-byte:
> Have you tested this PR in a real production environment?

Thanks for checking. I haven’t tested this in a real production environment; I verified it in our test environment. The validation covered both the envd-injection and non-injection template creation paths, as well as the follow-up sandbox creation and command execution flow.

fslongjin (approved):

CubeMaster/pkg/templatecenter/artifact_build.go
```diff
@@ -251,14 +258,17 @@ func restoreRootfsArtifact(ctx context.Context, artifactID string) error {
...
 	var caBakeResult cube_egress_ca.Result
 	opts := image.BuildOptions{ArtifactID: record.ArtifactID}
 	// Bake the CubeEgress root CA into the rootfs while it's still a mutable
 	// host-side directory, before mkfs.ext4 freezes the layout. See
 	// design/cube-egress-ca-bake.md for the contract.
 	opts.PostRootfsExport = func(ctx context.Context, rootfsDir string) error {
 		var err error
+		if _, err = injectEnvdPayloadIntoRootfs(ctx, rootfsDir, envdPayload); err != nil {
+			return err
+		}
```
kinwin-ustc, line 271: After injecting envd, the memory used by envd in memory is no longer needed. This memory can be released in advance to avoid putting too much memory pressure on cubemaster.

kinwin-ustc (commented):
After handling this issue, we can merge this PR

CubeMaster/pkg/templatecenter/artifact_build.go
0717dn6hyc8-byte: Addressed, thanks.

0717dn6hyc8-byte:
@kinwin-ustc @fslongjin Thanks for the review. I’ve addressed the requested change by releasing the in-memory envd payload immediately after it is injected into the rootfs.

If there are no further blocking issues, could you please help review and merge this PR when convenient? Since the upstream master branch is moving quickly, I may otherwise need to keep rebasing this PR frequently.

kinwin-ustc (approved):

fslongjin (approved):

PR #629, opened 2026-06-23
Add terraform tencentcloud support

lkml-likexu:
Add a Terraform-based deployer that stands up a **clustered** CubeSandbox on
Tencent Cloud, complementing the existing single-machine `install.sh`.
- A managed **TKE** control plane runs `cubemaster` / `cube-api` / `cube-proxy`
  / `cube-webui`, backed by cloud **MySQL + Redis**, with one or more **CVM PVM**
  compute nodes inside a private VPC reached through an SSH jumpserver (port 443).
- `create.sh` is the single entry point and `destroy.sh` tears everything down.
  Both run directly from an extracted release bundle and can work fully offline
  from the local bundle (with an online-install fallback).
- Provisioning is **phased, fail-fast and re-runnable**: network → TCR → CVMs →
  image build/push → MySQL/Redis → TKE + addons → health checks → compute setup.
  A partial failure can be fixed and resumed by simply re-running `create.sh`.
- The deployer is shipped at the top level of the bundle (and embedded in
  `sandbox-package`), wired into the release-bundle build, and documented in both
  the English and Chinese one-click READMEs.
- Adds a `terraform-validate` GitHub Actions workflow (fmt + validate +
  shellcheck) for `deploy/one-click/terraform/**`.

How to test:
```
sh build-release-bundle-builder.sh
tar xvzf cube-sandbox-one-click.tar.gz
cd cube-sandbox-one-click/terraform/tencentcloud
cat > .env <<EOF
# https://console.cloud.tencent.com/cam/capi
TENCENTCLOUD_SECRET_ID='your_secret'
TENCENTCLOUD_SECRET_KEY='your_key'
EOF
./create.sh
```
Tested .env demo:
```
TENCENTCLOUD_REGION='ap-shanghai'
TENCENTCLOUD_VPC_NAME='cubesandbox-terraform-vpc'
TENCENTCLOUD_AVAILABILITY_ZONE='ap-shanghai-5'
TENCENTCLOUD_JUMPSERVER_AVAILABILITY_ZONE='ap-shanghai-5'
TENCENTCLOUD_COMPUTE_AVAILABILITY_ZONE='ap-shanghai-5'
TENCENTCLOUD_TKE_WORKER_AVAILABILITY_ZONE='ap-shanghai-5'
TENCENTCLOUD_IMAGE_NAME='OpenCloudOS Server 9'
TENCENTCLOUD_JUMPSERVER_INSTANCE_TYPE='S5.MEDIUM4'
TENCENTCLOUD_COMPUTE_INSTANCE_TYPE='C3.LARGE8'
TENCENTCLOUD_COMPUTE_NODE_COUNT='2'
TENCENTCLOUD_MYSQL_PASSWORD='CubeSandbox123!'
TENCENTCLOUD_REDIS_PASSWORD='ceuhvu123'
TENCENTCLOUD_CUBE_DB='cube_mvp'
TENCENTCLOUD_CUBE_USER='cube'
TENCENTCLOUD_CUBE_PASSWORD='cube_pass'
TENCENTCLOUD_CUBE_IMAGE_TAG='latest'
TENCENTCLOUD_TKE_CLUSTER_VERSION='1.34.1'
TENCENTCLOUD_TKE_NODE_COUNT='2'
TENCENTCLOUD_VERBOSE='1'
```

You may try different CVMs in different regions.

I expect to uncover all sorts of defects and issues while testing the cluster version of Cubesandbox. 
There’s a lot to explore here when it comes to contributing to open source - let’s get started.


deploy/one-click/CubeAPI/Dockerfile
```diff
@@ -0,0 +1,52 @@
...
+#
+#   docker build -f CubeAPI/Dockerfile -t cube-api:TAG CubeAPI/
+#
+# See terraform/tencentcloud/build_images.sh.
+#
+# The prebuilt binary is dynamically linked against glibc, so the runtime image
+# must be glibc-based (ubuntu), not musl-based (alpine). On alpine the kernel
+# cannot find the ELF interpreter and exec fails with "no such file or
+# directory" even though the binary exists.
+FROM ubuntu:25.04
```
fslongjin: It's better to use 24.04 or any other LTS version

deploy/one-click/CubeMaster/Dockerfile
```diff
@@ -0,0 +1,98 @@
...
+#   docker build -f CubeMaster/Dockerfile -t cubemaster:TAG CubeMaster/
+#
+# See terraform/tencentcloud/build_images.sh.
+#
+# The prebuilt cubemaster binary is a default `go build` (cgo-enabled) artifact
+# dynamically linked against glibc, so the runtime image must be glibc-based
+# (ubuntu), not musl-based (alpine): on alpine the kernel cannot find the ELF
+# interpreter and exec fails with "no such file or directory" even though the
+# binary exists.
+FROM ubuntu:25.04
```
fslongjin: +1

deploy/one-click/terraform/tencentcloud/tke-addons.tf
```diff
@@ -0,0 +1,805 @@
...
+      req_template_conf = {
+        whitelist_req_tag = {
+          WorkingDir  = true
+          RLimit      = true
+          DnsConfig   = true
+          HostAliases = true
+          Poststop    = true
+          Prestop     = true
+        }
+        cube_box_req_template = "{\"volumes\":[{\"name\":\"tmp\",\"volume_source\":{\"empty_dir\":{\"medium\":0}}}],\"containers\":[{\"name\":\"cubebox-default\",\"envs\":[{\"key\":\"TZ\",\"value\":\"Asia/Shanghai\"},{\"key\":\"TERM\",\"value\":\"xterm\"}],\"volume_mounts\":[{\"name\":\"tmp\",\"container_path\":\"/\"}],\"security_context\":{\"privileged\":true,\"readonly_rootfs\":false,\"no_new_privs\":false}}],\"network_type\":\"tap\",\"cubevs_context\":{\"allowInternetAccess\":true,\"denyOut\":[\"10.0.0.0/8\",\"100.64.0.0/10\",\"172.16.0.0/12\",\"192.168.0.0/18\"]}}"
```
fslongjin: This template uses `cubevs_context`, but CubeMaster deserializes the request into `cube_network_config` / `CreateCubeSandboxReq.CubeNetworkConfig`. Go's JSON decoder will ignore the unknown field, so the default `allowInternetAccess` / `denyOut` policy configured here will not be applied in the TKE deployment. Please switch this to `cube_network_config` and add a regression test that deserializes this template and asserts the network config is populated.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
```diff
@@ -0,0 +1,805 @@
...
+  # four images to that TCR. Override with var.image_registry / var.image_namespace
+  # (e.g. to reuse a public image source).
+  image_registry    = var.image_registry != "" ? var.image_registry : "${tencentcloud_tcr_instance.demo.name}.tencentcloudcr.com"
+  image_namespace   = var.image_namespace != "" ? var.image_namespace : tencentcloud_tcr_namespace.demo.name
+  cube_master_image = "${local.image_registry}/${local.image_namespace}/cubemaster:${var.image_tag}"
+  cube_api_image    = "${local.image_registry}/${local.image_namespace}/cube-api:${var.image_tag}"
+  cube_proxy_image  = "${local.image_registry}/${local.image_namespace}/cubeproxy:${var.image_tag}"
+  cube_webui_image  = "${local.image_registry}/${local.image_namespace}/cube-webui:${var.image_tag}"
+
+  cube_db       = "cube_mvp"
```
fslongjin: `create.sh` and `env.example` expose `TENCENTCLOUD_CUBE_DB` / `TENCENTCLOUD_CUBE_USER`, but the Terraform addon config still hard-codes `cube_mvp` / `cube`. If a user enters non-default values, the saved environment and later checks can drift from what the control plane actually uses. Please either wire these through Terraform as real variables everywhere, or remove the unsupported prompts/env examples so the deployer does not imply customization works.

deploy/one-click/terraform/tencentcloud/create.sh
```diff
@@ -0,0 +1,4815 @@
...
+		echo -e "${GREEN}✓ cube-proxy TLS certificates are ready: ${cert_dir}${NC}"
+		return 0
+	fi
+
+	local js_pub_ip ssh_key
+	js_pub_ip=$(terraform output -raw jumpserver_public_ip 2>/dev/null || echo "")
+	ssh_key="${TENCENTCLOUD_SSH_PRIVATE_KEY_PATH:-$SSH_PRI_KEY}"
+	if [ -z "$js_pub_ip" ]; then
+		echo -e "${YELLOW}⚠ jumpserver unavailable, skipping cube-proxy certificate generation${NC}"
+		return 0
```
fslongjin: This failure path continues even though `cube-proxy` later requires `cube.app+3.pem` and `cube.app+3-key.pem` to be present. In `TENCENTCLOUD_BUILD_IMAGES=0`, weak-network, or incomplete local-bundle deployments, this can leave Terraform without the `cubeproxy-certs` ConfigMap or start `cube-proxy` with missing TLS files, causing apply failures or Pod CrashLoop. Please fail fast before deploying addons unless both required cert files exist, or explicitly support and validate a BYO-certificate path.

.github/workflows/terraform-validate.yml
```diff
@@ -0,0 +1,41 @@
+name: Terraform Deployer Validate
+
+on:
+  pull_request:
+    types: [opened, synchronize, reopened]
+    paths:
+      - 'deploy/one-click/terraform/**'
```
fslongjin, line 7: The Terraform validation workflow only triggers on `deploy/one-click/terraform/**`, but this PR's deployer also depends on `deploy/one-click/build-release-bundle.sh` and the component Dockerfile/package contexts. Changes there can break bundle layout or image builds without running this validation. Please extend the path filters and consider adding a lightweight package-layout dry run so quickstart/package deployment regressions are covered.

fslongjin (commented):
Inline review comments from the deep review of PR #629.

deploy/one-click/terraform/tencentcloud/create.sh
```diff
@@ -0,0 +1,4815 @@
...
+	# stateful resources that exist in the cloud but are missing from the
+	# local state, so the applies below don't collide with them
+	# ("... already exists") or work off stale attributes. Best-effort and
+	# a no-op on a first creation (empty state / no jumpserver yet).
+	ss_sync_state || true
+	_sync_compute_config_from_state
+
+	# ============================================================
+	# Sequential, fail-fast provisioning. Each step below is a synchronous
+	# terraform apply (restricted to that step's resources via -target) that must
```
fslongjin, line 4574: `create.sh` now owns interaction, Terraform orchestration, state recovery, image build/push, certificate generation, health checks, and compute-node initialization. That is acceptable for an initial deployer, but the script is becoming the deployment state machine. Please consider splitting the high-risk phases into smaller sourced modules, or at least adding scripted coverage for the phase transitions and rerun paths so future changes do not regress one-click deployment recovery.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
```diff
@@ -0,0 +1,805 @@
...
+      redis = {
+        nodes        = "${tencentcloud_redis_instance.redis.ip}:6379"
+        password     = var.redis_password
+        db_no        = 0
+        max_idle     = 8
+        max_active   = 32
+        idle_timeout = 30
+        max_retry    = 2
+      }
+      redis_read = {
```
fslongjin: Please verify that CubeMaster actually consumes `redis_read` / `redis_write`. I do not see matching fields in the current CubeMaster config structure, so these entries appear to be dead configuration. If they are not supported, removing them would avoid implying read/write Redis separation that the control plane ignores; otherwise the server-side config type should be updated and covered by a config parsing test.

deploy/one-click/terraform/tencentcloud/create.sh
```diff
@@ -0,0 +1,4815 @@
...
+#   TENCENTCLOUD_REDEPLOY_TKE_DEPLOY     recreate the four addon Deployments (1)
+#   TENCENTCLOUD_REDEPLOY_TKE_SERVICE    recreate the addon Services/CLBs (0)
+#   TENCENTCLOUD_ALLOW_INSECURE_DEFAULTS allow the built-in demo passwords on a non-interactive run (0)
+#   TENCENTCLOUD_LOCAL_BUNDLE            path to cube-sandbox-one-click-*.tar.gz (auto-detected inside a bundle)
+#   TENCENTCLOUD_PVM_KERNEL_VMLINUX      path to vmlinux-pvm (only if the bundle ships none)
+#   TENCENTCLOUD_PVM_KERNEL_RPM_URL      PVM kernel RPM URL (OpenCloudOS mirror default)
+#   TENCENTCLOUD_SSH_PORT                SSH port on compute nodes through the jumpserver (22)
+#   TENCENTCLOUD_SSH_PRIVATE_KEY_PATH    SSH private key (default ./.ssh/id_rsa, auto-generated)
+#   TENCENTCLOUD_SSH_PUBLIC_KEY_PATH     SSH public key  (default ./.ssh/id_rsa.pub)
+#   TENCENTCLOUD_CUBE_{MASTER,API,PROXY,WEBUI}_IMAGE  per-component image ref overrides
```
fslongjin: This advertises per-component image overrides, but the addon Terraform only composes images from `image_registry` / `image_namespace` / `image_tag`, and `create.sh` only exports `TF_VAR_image_tag`. In `TENCENTCLOUD_BUILD_IMAGES=0` or bring-your-own-image scenarios, users may set these variables and still get the default composed images. Please either wire per-component image refs through Terraform, or remove this documented override.

fslongjin (commented):
Additional inline comments from the responsibility/redundancy/correctness review section.

deploy/one-click/terraform/tencentcloud/create.sh
```diff
@@ -0,0 +1,4815 @@
...
+
+	# ============================================================
+	# Sequential, fail-fast provisioning. Each step below is a synchronous
+	# terraform apply (restricted to that step's resources via -target) that must
+	# finish before the next one starts; any failure aborts the whole deployment.
+	# The TKE cluster + addons are created LAST (Step 6) so the kubernetes provider
+	# only connects once the API Server exists — until then create_tke /
+	# deploy_tke_addons stay OFF for every base apply.
+	# ============================================================
+	export TF_VAR_create_tke=false
```
fslongjin: The phased `terraform apply -target` flow depends on `TF_VAR_create_tke` and `TF_VAR_deploy_tke_addons` being exactly right across first runs and reruns. A stale or wrong phase flag can hide count-gated resources from Terraform or make the Kubernetes provider connect at the wrong time. Please add a lightweight dry-run/integration check for rerun scenarios, especially after partial failure, so this one-click path remains recoverable.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
lkml-likexu: cubevs_context → cube_network_config + regression test

tke-addons.tf: the cube_box_req_template egress policy now uses cube_network_config, the only key CubeMaster deserializes (CreateCubeSandboxReq.CubeNetworkConfig).
Added CubeMaster/pkg/service/httpservice/cube/cubebox_req_template_test.go: deserializes the exact deployment template and asserts CubeNetworkConfig is populated (allowInternetAccess=true, the 4 denyOut CIDRs), plus a negative test locking in that the legacy cubevs_context key is silently dropped.
Also fixed the identical dead-config bug in configs/single-node/cubemaster.yaml (this same one-click bundle ships it as CubeMaster/conf.yaml).

deploy/one-click/terraform/tencentcloud/tke-addons.tf
lkml-likexu: cube_db / cube_user wired through Terraform — added validated cube_db/cube_user variables (variables.tf), used them in the MySQL account/privilege/CREATE DATABASE (main.tf), the cube-master ConfigMap (tke-addons.tf), and exported TF_VAR_cube_db/TF_VAR_cube_user plus the RESET_DB drop in create.sh. Now a custom DB/user stays consistent end to end instead of drifting against hard-coded cube_mvp/cube.

deploy/one-click/terraform/tencentcloud/create.sh
lkml-likexu: cube-proxy cert fail-fast — prepare_cubeproxy_certs now returns non-zero (instead of silently return 0) when it can't generate certs and no BYO certs exist; added _require_cubeproxy_certs and made the call site abort before deploying addons unless both cube.app+3.pem/cube.app+3-key.pem are present. The BYO path is documented.

.github/workflows/terraform-validate.yml
lkml-likexu: terraform-validate.yml — extended path filters to build-release-bundle*.sh, the component contexts (CubeAPI/, CubeMaster/, webui/, cubeproxy/, CubeProxy/Dockerfile, lib/, tests/), and added a lightweight package-layout dry run (tests/test_package_layout.sh) that checks Dockerfile presence and image-name parity between build_images.sh and tke-addons.tf.

deploy/one-click/terraform/tencentcloud/create.sh
lkml-likexu: phase/rerun state machine modularized + covered — extracted the high-risk phase decisions into a pure, sourced module lib-phases.sh (phase flags, no-scale-down compute count, cluster reuse, stale-k8s pruning), wired create.sh to use them, and added tests/test_phase_flags.sh exercising first-run/rerun/partial-failure transitions with no cloud. Both run in CI.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
lkml-likexu: dead redis_read / redis_write removed — CubeMaster only consumes a single redis pool, so those blocks were removed from tke-addons.tf.

deploy/one-click/terraform/tencentcloud/create.sh
lkml-likexu: unsupported per-component image overrides removed — dropped the TENCENTCLOUD_CUBE_{MASTER,API,PROXY,WEBUI}_IMAGE header doc and dead assignments from create.sh; the deployer now only documents the supported shared-tag override.

PR #634, opened 2026-06-24
feat(CubeAPI): inject sandbox env_vars on sandbox creation

xiongxz:
## Summary

The SDK's `envVars` field was silently dropped at sandbox creation: `CreateSandboxRequest.containers` was hard-coded to `vec![]`, so user-supplied env vars never reached the container. The downstream path (CubeMaster template merge + Cubelet OCI builder) already supports per-container envs — only the CubeAPI wiring was missing. This wires `NewSandbox.env_vars` into the create request.

## Scope

This PR puts create-time env vars into the container's PID-1 OCI `Process.Env`, so they are inherited by processes that start **after those vars are present** — cold-boot / rootfs-only paths, services started after a snapshot restore, and non-envd agents/services that read `os.environ` at startup. It does **not** affect processes already running at template-snapshot time: a normal Template is booted, runtime/service processes start, and memory is snapshotted; sandbox creation restores from that snapshot rather than re-running the entrypoint, so those restored processes keep their snapshot-time env and do not re-read the per-sandbox vars.

It also does **not** cover envd-backed command execution (`commands.run` / `run_code`). envd's `process.Start` (`e2b-dev/infra` `packages/envd`, `internal/services/process/handler/handler.go`) builds each command's env from an in-process `defaults.EnvVars` map plus the per-request `envs`, taking **only `PATH`** from its inherited environment; vars baked into PID-1's env therefore never reach an envd-spawned command. `defaults.EnvVars` is populated solely by envd's `/init` endpoint, which is what **#566** wires up. The two PRs are complementary layers:

- **#566** — envd-backed command execution (`commands.run` / `run_code`); post-create envd `/init`, snapshot-restore-safe.
- **This PR** — lower-level container env propagation for processes actually started after the env vars are present (cold-boot / rootfs-only paths, post-restore-started services, non-envd agents).

## Changes

- **`CubeAPI/src/services/sandboxes.rs`**: new `build_env_container` converts `Option<EnvVars>` into a single-container `ContainerSpec` carrying the vars on `envs`, replacing the previous `containers: vec![]`.
- The container's `image` and `command`/`args` are intentionally left empty:
  - CubeMaster's `applyTemplateToContainer` fills the image from the resolved template.
  - Cubelet's `command.WithProcessArgs` falls back to the image `Entrypoint`/`Cmd`, so `cube-entrypoint.sh` still runs and envd still starts.
- With no env vars, an empty list is returned — prior behaviour is unchanged.
- Validation: keys must be non-empty, whitespace-free, with no `=` and no control characters (`is_valid_env_key`); values must be non-empty, not whitespace-only, and NUL-free; reserved names (`LD_PRELOAD`/`LD_AUDIT`/`LD_LIBRARY_PATH`/`GCONV_PATH`/`GCONV_MODULES`) are dropped. All drops are logged via `tracing::warn!`.
- Tests cover the empty-map, populated, all-dropped early-return, reserved+whitespace+invalid filtering, and `is_valid_env_key` cases.

## How it works

`POST /sandboxes` (`envVars`) → `build_env_container` (validate + filter) → `ContainerSpec { envs }` → CubeMaster template merge (`ctr.Envs = append(ctr.Envs, templateCtr.Envs...)`; the template envs are a copy of the image ENV, appended *after* the user envs) → Cubelet `env.GenOpt` feeds first the image ENV, then the container envs, into containerd's `oci.WithEnv`. `oci.WithEnv` dedupes by key keeping the last value, writing the result into the OCI `Process.Env` (PID 1), inherited by envd and all user commands.

## Safety

- **Image-owned vars** consumed by `cube-entrypoint.sh` (`ENVD_PORT`, `ENVD_BIN`, `ENVD_LOG_FILE`, `ENVD_EXTRA_ARGS`) and `PATH` are protected: the template's `Envs` (a full copy of the image ENV) is appended *after* the user's, so the template value wins on key conflicts via containerd's `oci.WithEnv` last-value-wins dedup (the call site is `Cubelet/pkg/container/env/env.go`). A user cannot override them through env vars.
- **Non-image dangerous vars** (`LD_PRELOAD`/`LD_AUDIT`/`LD_LIBRARY_PATH`/`GCONV_PATH`/`GCONV_MODULES`) are *not* covered by that protection (they are not in the image ENV, so no template copy shadows them), so they are dropped by the reserved-name check. These inject code via the dynamic linker or glibc's gconv mechanism. The check is intentionally case-sensitive (glibc's dynamic linker is case-sensitive). The container is privileged and the owner has root, so this is API-boundary defence-in-depth rather than a hard isolation boundary.
- `HOSTNAME` and `CUBE_CONTAINER_ID_*` are injected by the runtime/Cubelet and cannot be overridden through this path.
- `envd_access_token` is a vestigial field (always `None`), so this change does not affect token auth.

## Known limitation

When a user-supplied env var shares a key with an image/template ENV, the template value wins (due to the append order in `applyTemplateToContainer` plus `oci.WithEnv` last-value-wins). This does not fully match E2B's "user wins" semantics, but only affects keys that collide with image-owned vars; business env vars (the common case, e.g. `OPENAI_API_KEY`, `SESSION_ID`) inject correctly. Fully aligning with E2B requires flipping the CubeMaster append order — to be addressed in a follow-up.

### DIFF
--- a/CubeAPI/src/services/sandboxes.rs
+++ b/CubeAPI/src/services/sandboxes.rs
@@ -715,19 +715,47 @@ pub(crate) fn build_cube_network_config(
 ///   - Empty `command`/`args` fall back to the image `Entrypoint`/`Cmd` inside
 ///     Cubelet's `command.WithProcessArgs`, so the entrypoint (cube-entrypoint.sh)
 ///     still runs and envd still starts.
-/// With no env vars we return an empty list, preserving the prior behaviour
-/// where `containers` was an empty `vec![]`.
+/// Env vars are validated before being forwarded:
+///   - Entries with an empty key or empty value are dropped (they carry no
+///     usable value and only add noise).
+///   - Reserved names a sandbox owner must not set are dropped. `LD_PRELOAD`
+///     and `LD_AUDIT` would inject a shared library into every dynamically
+///     linked process in the sandbox (including envd), so they are blocked at
+///     the API boundary as defence-in-depth.
+///
+/// With no env vars left after validation we return an empty list, preserving
+/// the prior behaviour where `containers` was an empty `vec![]`.
 pub(crate) fn build_env_container(env_vars: Option<EnvVars>) -> Vec<ContainerSpec> {
     let Some(vars) = env_vars else {
         return Vec::new();
     };
-    if vars.is_empty() {
-        return Vec::new();
+    // Reserved env var names a sandbox owner must not set via the API.
+    const RESERVED_ENV_KEYS: &[&str] = &["LD_PRELOAD", "LD_AUDIT"];
+
+    // Warn about reserved names we are about to drop so a misconfigured caller
+    // is not silently surprised. Empty key/value entries are dropped quietly.
+    let dropped: Vec<&String> = vars
+        .keys()
+        .filter(|k| RESERVED_ENV_KEYS.contains(&k.as_str()))
+        .collect();
+    if !dropped.is_empty() {
+        tracing::warn!(
+            dropped = ?dropped,
+            "dropping reserved env var names from sandbox create request; \
+             LD_PRELOAD/LD_AUDIT can inject a library into every process and are not allowed"
+        );
     }
-    let envs = vars
+
+    let envs: Vec<EnvVar> = vars
         .into_iter()
+        .filter(|(key, value)| {
+            !key.is_empty() && !value.is_empty() && !RESERVED_ENV_KEYS.contains(&key.as_str())
+        })
         .map(|(key, value)| EnvVar { key, value })
         .collect();
+    if envs.is_empty() {
+        return Vec::new();
+    }
     vec![ContainerSpec {
         name: None,
         image: ImageSpec {
@@ -821,6 +849,26 @@ mod tests {
             .collect();
         assert_eq!(lookup.get("API_KEY").copied(), Some("sk-123"));
         assert_eq!(lookup.get("DEBUG").copied(), Some("true"));
+    }
+
+    #[test]
+    fn env_container_drops_reserved_and_empty_entries() {
+        // Reserved names (LD_PRELOAD/LD_AUDIT) and empty key/value entries are
+        // filtered out; only the valid entry survives.
+        let vars = HashMap::from([
+            ("API_KEY".to_string(), "sk-123".to_string()),
+            ("LD_PRELOAD".to_string(), "/tmp/x.so".to_string()),
+            ("LD_AUDIT".to_string(), "/tmp/y.so".to_string()),
+            ("".to_string(), "no-key".to_string()),
+            ("EMPTY".to_string(), "".to_string()),
+        ]);
+
+        let containers = build_env_container(Some(vars));
+        assert_eq!(containers.len(), 1);
+        let envs = containers[0].envs.as_ref().expect("envs should be set");
+        assert_eq!(envs.len(), 1);
+        assert_eq!(envs[0].key, "API_KEY");
+        assert_eq!(envs[0].value, "sk-123");
     }
 
     #[test]
@@ -1104,7 +1152,7 @@ mod tests {
     /// covers each meaningful combination.
     #[test]
     fn lifecycle_object_translates_to_cubemaster_bools() {
-        use crate::models::{NewSandbox, SandboxLifecycleConfig, SandboxOnTimeout};
+        use crate::models::{NewSandbox, SandboxOnTimeout};
 
         // Helper that mimics services::create_sandbox's lifecycle decoding.
         fn translate(body: &NewSandbox) -> (bool, bool) {
--- a/CubeAPI/src/services/sandboxes.rs
+++ b/CubeAPI/src/services/sandboxes.rs
@@ -10,16 +10,17 @@ use super::validate_allow_out_domains_require_deny_all;
 use crate::{
     constants::ENVD_VERSION,
     cubemaster::{
-        datetime_from_unix_nanos, extract_template_id, CreateSandboxRequest, CubeEgressRule,
-        CubeEgressRuleAction, CubeEgressRuleInject, CubeEgressRuleMatch, CubeMasterClient,
-        CubeMasterError, CubeNetworkConfig, DeleteSandboxRequest, ListSandboxRequest, SandboxInfo,
-        SandboxLogsRequest, SandboxRefreshRequest, SandboxStatus, SandboxTimeoutRequest,
-        SandboxUpdateRequest,
+        datetime_from_unix_nanos, extract_template_id, ContainerSpec, CreateSandboxRequest,
+        CubeEgressRule, CubeEgressRuleAction, CubeEgressRuleInject, CubeEgressRuleMatch,
+        CubeMasterClient, CubeMasterError, CubeNetworkConfig, DeleteSandboxRequest, EnvVar,
+        ImageSpec, ListSandboxRequest, SandboxInfo, SandboxLogsRequest, SandboxRefreshRequest,
+        SandboxStatus, SandboxTimeoutRequest, SandboxUpdateRequest,
     },
     error::{AppError, AppResult},
     models::{
-        EgressRule, LogLevel as ModelLogLevel, NewSandbox, Sandbox, SandboxDetail, SandboxLog,
-        SandboxLogEntry, SandboxLogs, SandboxLogsV2Response, SandboxNetworkConfig, SandboxState,
+        EgressRule, EnvVars, LogLevel as ModelLogLevel, NewSandbox, Sandbox, SandboxDetail,
+        SandboxLog, SandboxLogEntry, SandboxLogs, SandboxLogsV2Response, SandboxNetworkConfig,
+        SandboxState,
     },
 };
 
@@ -161,7 +162,7 @@ impl SandboxService {
             labels,
             distribution_scope: body.distribution_scope,
             volumes: None,
-            containers: vec![],
+            containers: build_env_container(body.env_vars),
             exposed_ports: vec![],
             network_type: Some("tap".to_string()),
             cube_network_config,
@@ -704,6 +705,49 @@ pub(crate) fn build_cube_network_config(
     }))
 }
 
+/// Build the single-container payload that carries the user-supplied env vars
+/// down to CubeMaster.
+///
+/// Each sandbox is modelled as one container. The env vars are attached to that
+/// container's `envs` and every other field is left empty:
+///   - The empty `image` is overwritten with the template image by CubeMaster's
+///     `applyTemplateToContainer`, after which Cubelet resolves the image config.
+///   - Empty `command`/`args` fall back to the image `Entrypoint`/`Cmd` inside
+///     Cubelet's `command.WithProcessArgs`, so the entrypoint (cube-entrypoint.sh)
+///     still runs and envd still starts.
+/// With no env vars we return an empty list, preserving the prior behaviour
+/// where `containers` was an empty `vec![]`.
+pub(crate) fn build_env_container(env_vars: Option<EnvVars>) -> Vec<ContainerSpec> {
+    let Some(vars) = env_vars else {
+        return Vec::new();
+    };
+    if vars.is_empty() {
+        return Vec::new();
+    }
+    let envs = vars
+        .into_iter()
+        .map(|(key, value)| EnvVar { key, value })
+        .collect();
+    vec![ContainerSpec {
+        name: None,
+        image: ImageSpec {
+            image: String::new(),
+            storage_media: None,
+        },
+        command: None,
+        args: None,
+        working_dir: None,
+        resources: None,
+        envs: Some(envs),
+        volume_mounts: None,
+        dns_config: None,
+        r_limit: None,
+        security_context: None,
+        probe: None,
+        annotations: None,
+    }]
+}
+
 fn map_egress_rule(rule: &EgressRule) -> CubeEgressRule {
     CubeEgressRule {
         name: rule.name.clone(),
@@ -734,12 +778,50 @@ fn map_egress_rule(rule: &EgressRule) -> CubeEgressRule {
 mod tests {
     use std::collections::HashMap;
 
-    use super::{build_cube_network_config, filter_by_metadata, from_cubemaster_info};
+    use super::{
+        build_cube_network_config, build_env_container, filter_by_metadata, from_cubemaster_info,
+    };
     use crate::cubemaster::{CreateSandboxRequest, ListSandboxResponse, SandboxInfo};
     use crate::models::{
         EgressRule, EgressRuleAction, EgressRuleInject, EgressRuleMatch, SandboxNetworkConfig,
         SandboxState,
     };
+
+    #[test]
+    fn env_container_is_empty_when_no_env_vars() {
+        // No env vars at all (the SDK omits the field) and an explicitly empty
+        // map both keep the prior `containers: vec![]` behaviour.
+        assert!(build_env_container(None).is_empty());
+        assert!(build_env_container(Some(HashMap::new())).is_empty());
+    }
+
+    #[test]
+    fn env_container_wraps_each_var_as_an_envvar() {
+        let vars = HashMap::from([
+            ("API_KEY".to_string(), "sk-123".to_string()),
+            ("DEBUG".to_string(), "true".to_string()),
+        ]);
+
+        let containers = build_env_container(Some(vars));
+        assert_eq!(containers.len(), 1);
+
+        let container = &containers[0];
+        // The image is intentionally empty: CubeMaster's template merge fills it
+        // from the resolved template. command/args stay unset so Cubelet falls
+        // back to the image entrypoint.
+        assert!(container.image.image.is_empty());
+        assert!(container.command.is_none());
+        assert!(container.args.is_none());
+
+        let envs = container.envs.as_ref().expect("envs should be set");
+        assert_eq!(envs.len(), 2);
+        let lookup: HashMap<&str, &str> = envs
+            .iter()
+            .map(|e| (e.key.as_str(), e.value.as_str()))
+            .collect();
+        assert_eq!(lookup.get("API_KEY").copied(), Some("sk-123"));
+        assert_eq!(lookup.get("DEBUG").copied(), Some("true"));
+    }
 
     #[test]
     fn metadata_filter_matches_all_pairs() {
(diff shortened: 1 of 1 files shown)


xiongxz:
All CI checks are passing now. Could a maintainer please review when you have a chance? Thanks.

kinwin-ustc:
Should we perform some validation on the incoming environment variables? For example, empty values, or key environment variables that customers cannot customize, such as LD_PRELOAD

xiongxz:
@kinwin-ustc Thanks for the catch — added validation in `build_env_container` (65addd5):

- **Empty values/keys**: dropped (no effect, only noise).
- **Reserved names**: `LD_PRELOAD` and `LD_AUDIT` are dropped, with a `warn!` logged when any are. These are *not* image-owned, so the template-merge protection does not cover them — they would inject a shared library into every dynamically linked process in the sandbox (including envd).

On scope: I considered a broader reserved set (`LD_LIBRARY_PATH`, `HTTP(S)_PROXY`/`ALL_PROXY`, `PYTHONPATH`/`NODE_PATH`/...) but kept it to the pure-injection vectors that have no legitimate use. The others all have strong legitimate uses, and the container is privileged (the owner has root), so this is API-boundary defence-in-depth rather than a hard isolation boundary. Happy to expand the list if you'd prefer.

The image-owned vars consumed by `cube-entrypoint.sh` (`ENVD_*`) and the Cubelet-injected `HOSTNAME`/`CUBE_CONTAINER_ID_*` remain protected downstream (template env + last-write-wins), so they don't need to be listed here.

xiongxz:
Thanks for the review pass — went through all five and acted on each:

**#2 (double traversal) — adopted.** Merged into a single `filter_map` pass with `dropped_reserved`/`dropped_invalid` collected as side effects; no second iteration of the map.

**#3 (key/value validation) — adopted, scoped to correctness not POSIX.** Added `is_valid_env_key`: rejects empty, `=`-containing, and control-char keys. The `=` case is the real correctness bug — downstream serialises as `KEY=VALUE` and `strings.Cut` (in containerd's `oci.WithEnv`/`replaceOrAppendEnvValues`) splits on the first `=`, so a key like `FOO=BAR` would silently become key `FOO`; a value-less entry there is even treated as *unset*. Whitespace-only values are now dropped too (doc already promised this). I deliberately did **not** enforce `[A-Za-z_][A-Za-z0-9_]*`: real keys use lowercase, dashes, dots and digits (`app.db.url`, `npm_config_foo`), and a strict allow-list would break legitimate callers for no security gain in a privileged+root container. No length cap for the same reason — happy to add `MAX_ENV_VAR_VALUE_LENGTH` if you'd like a hard ceiling, but it's policy not correctness.

**#4 (tests) — mostly adopted.** Added: whitespace-only value dropped, invalid-key (`FOO=BAR`, `\n`, `\0`) dropped, full `ContainerSpec` field assertions (all the previously-unchecked `None` fields), and an `is_valid_env_key` unit test. Skipped the `tracing::warn!` capture test — there's no in-tree subscriber harness for it, the boilerplate outweighs the value, and the drop *behaviour* it logs is already covered. Also skipped the serialization round-trip: that exercises serde config on `CreateSandboxRequest`, not `build_env_container`, so it belongs elsewhere if we want it.

**#5 (docs + E2B priority) — adopted, and you're right that the priority difference deserves more than a footnote.** Two things here:

1. **`replaceOrAppendEnvValues` correction.** Good catch — it's never *called* in this repo; it's an internal of containerd's `oci.WithEnv`. Rewrote the doc comment and PR body to cite the actual in-repo path: `Cubelet/pkg/container/env/env.go` builds `[image.Env...] ++ [container.Envs...]` and feeds it to `oci.WithEnv`, whose dedup is last-value-wins.

2. **The priority is the reverse of E2B, and here's the full chain.** E2B's contract is "user wins" on a key collision; CubeSandbox is the opposite — **the image/template value wins**:
   - `CubeMaster` `cubeboxutil.go:126` appends the template envs *after* the user envs: `ctr.Envs = append(ctr.Envs, templateCtr.Envs...)`, and the template `Envs` is a copy of the image ENV (`getCubeboxReqTemplate`). So the array is `[user...] ++ [image-ENV-copy...]`.
   - `Cubelet` `env.go` then does `oci.WithEnv(image.Env)` followed by `oci.WithEnv(containerEnvs)`.
   - `oci.WithEnv` keeps the **last** value per key, so the image-ENV copy (appended last) overrides the user value on any collision.

   **Impact is narrow**: only a user var whose key collides with an image-baked ENV (`ENVD_*`, `PATH`, or anything the template image sets) is silently overridden. Business vars (`OPENAI_API_KEY`, `SESSION_ID`, …) inject correctly because they don't collide. **It's also load-bearing for safety** — this reversal is exactly why a user can't override `ENVD_PORT`/`ENVD_BIN`/`PATH` through env vars.

   Fully matching E2B ("user wins") means flipping the append order in `cubeboxutil.go:126` to `append(slices.Clone(templateCtr.Envs), ctr.Envs...)`. But that removes the merge protection from `ENVD_*`/`HOSTNAME`/`CUBE_CONTAINER_ID_*`, so the deny-list would have to grow to cover them, and the template isn't cached (fresh unmarshal per request), so there's no aliasing risk. That's a larger change than this PR's wiring fix, so I've recorded it as a follow-up rather than bundling it here.

**#1 (expand the deny-list) — partially adopted.** The container is privileged and the owner already has root, so this list is API-boundary defence-in-depth, not a hard boundary — it stays scoped to pure-injection vectors with no legitimate use. Working through the suggestions:
- `LD_LIBRARY_PATH` — **added**. Genuinely the closest peer to `LD_PRELOAD`: it can hijack the dynamic linker's search path to load a trojan `libc`/other lib into every dynamically linked process (including envd). Same no-legitimate-use-that-isn't-already-better-served-directly class as the others.
- `BASH_ENV`/`ENV`/`IFS` — shell-injection vectors, but the owner has root and already runs arbitrary code; these only add surprise against a non-root sub-process. Defence-in-depth the root owner bypasses; kept off to avoid breaking legitimate shell tooling.
- `GODEBUG`/`GOMEMLIMIT`/`GOGC` — pushing back. These inject into the container's PID 1 and the *user's own code*, not into Cubelet (Cubelet is a separate process on the host). A sandbox user running a Go program legitimately sets these; blocking them breaks valid use. envd itself is Rust, not Go. `GLIBC_TUNABLES` likewise mostly affects the user's own glibc programs.
- Interpreter load-path vars (`PYTHONPATH`/`NODE_PATH`/`PERL5OPT`/`RUBYLIB`/`JAVA_TOOL_OPTIONS`…) — same reasoning: strong legitimate use as standard runtime configuration, and root bypasses them.
- On allow-list vs deny-list long-term: agreed a deny-list is reactive; a documented allow-pattern is a reasonable follow-up once we've seen what keys real callers send.

The reserved set is now `LD_PRELOAD`, `LD_AUDIT`, `LD_LIBRARY_PATH`.

xiongxz:
Thanks for the second pass — went through all six:

**GCONV_PATH / GCONV_MODULES missing — added.** Right call; same class as `LD_PRELOAD`. gconv modules are `.so`s loaded on `iconv_open()`, so `GCONV_PATH` (and `GCONV_MODULES`, which can reference one by absolute path) are pure code-injection vectors with no legitimate use in a sandbox. Reserved set is now `LD_PRELOAD`, `LD_AUDIT`, `LD_LIBRARY_PATH`, `GCONV_PATH`, `GCONV_MODULES`.

**`" LD_PRELOAD"` whitespace bypass — fixed.** You noted ld.so is whitespace-sensitive so it's inert on Linux today, but you're right that a future downstream `TrimSpace` would reopen it. `is_valid_env_key` now rejects any key with leading/trailing whitespace (and whitespace-only keys), so a padded reserved name is dropped as invalid before the exact-match check ever runs. We reject rather than trim, so valid input is never silently mutated.

**Whitespace-only keys / value control chars — keys fixed; values, NUL only.** Whitespace-only keys are now rejected (same change as above). For values, I reject only a NUL byte — it truncates the `KEY=VALUE` C string at execve. `\n`/`\r` are forwarded verbatim: they're user data, JSON/protobuf escape them for transport, and the runtime treats them as ordinary bytes. Rejecting them would break legitimate multiline values for no correctness gain.

**Early-return test gap — added.** `env_container_returns_empty_when_all_entries_dropped` covers the path where a non-empty map filters down to nothing.

**Case-sensitivity — documented, intentionally kept.** Since ld.so is case-sensitive, `ld_preload` is inert and the exact match is correct; added a doc line saying so. On completeness (`PYTHONPATH`/`PERL5LIB`/`BASH_ENV`/`GIO_EXTRA_MODULES`/…): same reasoning as before — strong legitimate use as runtime config, and the privileged+root owner bypasses them, so they stay out. The list stays scoped to pure injection vectors with no legitimate use.

**Brittle full-field assertion — simplified.** Fair point (and I'll note it partly contradicts the earlier ask for full-field assertions). Trimmed it to the contract this function owns: one container, `image` left empty, `envs` populated with the survivors. Dropped the exhaustive `None` list so it won't break on an unrelated `ContainerSpec` field.

zyl1121:
This looks very close to #566, but the injection point is different.

My understanding is that this PR injects create-time env vars into the container / OCI env, while #566 focused on envd initialization for envd-backed command execution. My earlier concern with the container-level approach was that if `envd` is already started on the template-based startup path, `commands.run` may still not see the injected envs even though later-started processes do.

Did you happen to verify this specifically on a real cluster for the post-create execution paths? In particular, after `Sandbox.create(envs=...)` returns, can `commands.run(...)` immediately read those env vars? If you also checked whether it affects `run_code`, that result would be very helpful too.

xiongxz:
Thanks for the careful read — and you're right to flag this. I traced it into the envd source (`e2b-dev/infra` `packages/envd`, which is what `/usr/bin/envd` is built from per `docker/Dockerfile.cube-base`) and confirmed the concern: **`commands.run` / `run_code` will not see these vars through this PR alone.**

envd's `process.Start` (`internal/services/process/handler/handler.go:147-170`) builds each command's env from `defaults.EnvVars` (an in-process global) plus the per-request `envs`, taking **only `PATH`** from its inherited environment. So vars placed in the container's PID-1 OCI env — which is what this PR does — never reach an envd-spawned command. `defaults.EnvVars` is populated solely by envd's `/init` endpoint (`internal/api/init.go:189-194`), which is precisely the mechanism #566 wires up. For the envd-backed execution path, #566 is the correct fix and I don't intend this PR to cover it.

Where this PR does carry weight — and where #566 can't help — is **processes started independently in the sandbox**: a separate agent or service with its own process and port that the template image launches, rather than something envd spawns per-command. Those inherit the container's PID-1 environment normally, so the create-time vars are visible to them from the moment the container starts. #566's `/init` only feeds envd's internal `defaults.EnvVars`, which is consumed solely by envd's own `process.Start` and does not propagate to processes started outside envd. So the two are genuinely complementary, at different layers:

- **#566** — envd-backed command execution (`commands.run` / `run_code`).
- **This PR** — container-wide env for anything that inherits PID-1's environment (template-launched services, sidecar agents, processes that read `os.environ` at startup).

One concrete benefit worth calling out: the common agent pattern is to read config from env at process startup (12-factor). Because this PR bakes the vars into PID-1's env at creation time, such a process sees them immediately on start — there's no post-create injection step and no race window between container start and a later `/init`. Today we work around the absence of this by exposing an `/init`-style endpoint on the agent's own port and pushing envs after the sandbox is up; this PR removes that adaptation.

Happy to add a short "Scope" note to the PR description making the `commands.run` limitation explicit and pointing at #566, if that's useful.

zyl1121:
Thanks for the clarification. I agree these two PRs are complementary, but I still have one concern about the wording / scope.

The key nuance is CubeSandbox's Template lifecycle: a normal Template is not just a rootfs image, it goes through Boot & Snapshot first. The VM is cold-booted, runtime/service processes may already be started, and then the memory/state is snapshotted. Later sandbox creation restores from that snapshot instead of freshly starting the template entrypoint again. Because of that, processes already alive at template snapshot time usually will not re-read create-time OCI Process.Env during sandbox creation, so the effect may not match the expectation that all template services are newly started and will pick up those env vars.

So my understanding is: #566 fixes the immediate envd-backed execution path by initializing envd's runtime env state after sandbox creation, while #634 improves the lower-level container env propagation path for cases where processes are actually started after those env vars are present, such as cold-boot/rootfs-only paths, services started after restore, or future non-envd agents/services. That split makes sense to me; maybe it would be worth considering narrowing the “template-launched services” wording in the PR description a bit, to avoid implying that already-running snapshot-restored services will also re-read these env vars.

xiongxz:
Good catch — that's a real nuance and I'd over-stated it. The snapshot lifecycle is exactly as you describe: a normal Template is booted, runtime/service processes start, and memory is snapshotted; sandbox creation restores from that snapshot rather than re-running the entrypoint, so processes already alive at snapshot time keep their snapshot-time env and won't re-read the per-sandbox vars.

I've narrowed the Scope wording in the PR description to match the split you laid out:

- **#566** — envd-backed execution (`commands.run` / `run_code`); post-create envd `/init`, snapshot-restore-safe.
- **This PR** — lower-level container env propagation for processes that actually start *after* the env vars are present (cold-boot / rootfs-only paths, services started after a snapshot restore, non-envd agents/services), with an explicit note that snapshot-restored already-running processes don't re-read them.

Thanks — that keeps the description from implying more than it delivers.

xiongxz:
Hi @kinwin-ustc — gentle nudge to take another look when you have a moment.

Your earlier point on validation is now in place: env-var keys must be non-empty, whitespace-free, `=`-free, and control-char-free, values are NUL-free, and a reserved-name list (`LD_PRELOAD` / `LD_AUDIT` / `LD_LIBRARY_PATH` / `GCONV_PATH` / `GCONV_MODULES`) drops the pure code-injection vectors — all drops are logged via `tracing::warn!`. The two rounds of bot review and @zyl1121's `commands.run` point have been addressed too, and the description now has an explicit **Scope** section making clear what this PR covers (PID-1 container env, inherited by processes started after the vars are present) vs. what's left to #566 (envd-backed `commands.run` / `run_code`).

All CI checks are green and tests pass. Is there anything else you'd like added or adjusted before this is ready to merge? Happy to handle whatever's outstanding.

kinwin-ustc:
We already have a method to inject environment variables when creating templates, which can truly inject environment variables into all programs in the sandbox, including programs pre-launched during snapshot creation, as well as programs launched after the sandbox is established. This has a wider applicability than the method in this PR. I would like to understand the real demand scenarios covered by this PR.

xiongxz:
Thanks — you're right that the template-time path (`Template.build(envs=...)` / `--env`, baked into the snapshot via `container_overrides.envs` in `template_request.go`) is broader: it reaches both pre-launched and post-create processes.

The gap this PR targets is **per-sandbox** env vars that differ on each `Sandbox.create()` — `OPENAI_API_KEY`, `SESSION_ID`, per-tenant config. Template envs are frozen per template (identical for every sandbox), so they can't carry per-instance values; standing up a separate template per value isn't viable (cold-boot + snapshot + replication each time).

Is there an existing per-sandbox path I've missed? `Sandbox.create(env_vars=...)` exists in the SDK but is currently a no-op — CubeAPI drops `envVars` (`containers: vec![]`), which is exactly what this PR wires up. If the template mechanism or another path truly covers per-sandbox envs, we'll happily close this.

kinwin-ustc:
For each sandbox's sanbox.create env vars，  from the perspective of compatibility with E2B, I think PR #566 is more suitable because all subsequent operations through the E2B SDK can utilize this environment variable. For non-envd scenarios (custom images), the available injection methods for each sandbox may include the following: 1) Implement a custom service with a similar /init interface to envd; 2) Pass specific variables through host-mount in a file system manner; 3) For API-KEY, we can make use of our [Cube Egress credential hosting capabilities](https://github.com/TencentCloud/CubeSandbox/blob/master/docs/guide/security-proxy.md) where the key is completely invisible within the sandbox and is automatically replaced after exiting the sandbox.

xiongxz:
Fair point on credentials — for API keys Cube Egress is the better tool (the key never touches the sandbox), and we'll adopt it for that case.

General per-sandbox config is a different need, though. A project's `.env` is rarely all secrets — it's mostly non-sensitive runtime config: `APP_ENV`, `LOG_LEVEL`, feature flags, an internal endpoint URL, `SESSION_ID`, a model name, retry/timeout tuning. None of those are credentials, so routing them through a credential-hosting egress proxy is the wrong fit; and pushing every value into a custom `/init` service or a mounted file works, but it's plumbing every image has to build and maintain. The `/init` path is exactly what we run today, and this PR exists to drop that workaround.

So I'd ask you to weigh whether a zero-adaptation per-sandbox env path fits the project's intent. The SDK already exposes `Sandbox.create(env_vars=...)` as an E2B-compatible surface — it's just a no-op today because CubeAPI drops it. #634 isn't a new mechanism; it completes a contract the project has already chosen to offer, and for non-secret per-sandbox config it's the natural channel.

ls-ggg:
@xiongxz Thanks for your PR! The core reason why Cube currently doesn't support setting env variables during sandbox creation is that sandboxes are started based on templates. At that point, PID 1 inside the sandbox has already started, making it impossible to update its environment variables from the system level.

The approach in PR #566 is feasible, but it needs some optimization. I've already replied with my thoughts under PR #566, so we can continue the discussion over there

zyl1121:
@ls-ggg I noticed a specific gap today, so it might be a bit early to close this PR.

#566 covers the envd-backed path. For example, E2B `commands.run` can see create-time env vars after envd is initialized. However, in my test, the new process started by `cubecli exec` still cannot see them. I believe it's because `cubecli exec` communicates with `containerd` via its socket and ultimately uses `task.exec`, so it bypasses envd and its `/init`.

Because of that, I still think the two fixes are complementary. #566 handles envd-backed execution. This PR or an equivalent lower-level fix is still needed for non-envd exec paths such as `cubecli exec`. The exact implementation can be discussed, but the requirement seems real.

kinwin-ustc:
I believe there should be a solution that addresses a broader range of issues, not only those in the envd environment but also those in the OS environment. Ideally, it should also resolve issues with already started programs in template snapshots. Perhaps we need to provide a `cube_envd` or similar agent, offering an API compatible with envd. However, `/init` not only injects environment variables into envd itself but also into the bash startup configuration, thus affecting programs launched via bash. Furthermore, we can handle buffering issues for various service types within this agent, avoiding the need for wordarounds on `cube-proxy`.

zyl1121:
> Perhaps we need to provide a cube_envd or similar agent, offering an API compatible with envd.

That makes sense to me. A Cube-owned `cube_envd` sounds like a better long-term direction.

PR #638, opened 2026-06-25
ci: add workflow to publish Python SDK to PyPI on v* tag

wbzdssm:
## Summary

Add a GitHub Actions workflow to automatically publish the Python SDK to PyPI when a `v*` tag is pushed.

### Behavior
- Triggers on `v*` tag pushes
- Checks if `sdk/python/` has changed since the last tag
- If no changes → skips publish (avoids duplicate releases)
- If changed → builds and publishes to PyPI using `CUBE_PYPI_TOKEN` secret

### Files changed
- `.github/workflows/publish-python-sdk.yml` (new)

### DIFF
--- a/.github/workflows/publish-python-sdk.yml
+++ b/.github/workflows/publish-python-sdk.yml
@@ -1,0 +1,68 @@
+name: Publish Python SDK to PyPI
+
+on:
+  push:
+    tags:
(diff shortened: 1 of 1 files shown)


kinwin-ustc:
I think only officially released tags should trigger automatic PyPI publishing. tags for testing or debugging purposes should not

PR #639, opened 2026-06-25
network: Restrict public access via per-sandbox traffic token

chenhengqi:
Sandboxes created with network.allow_public_traffic=false now receive a per-sandbox traffic_access_token; CubeProxy rejects inbound requests that don't carry it in either e2b-traffic-access-token (E2B-compatible) or cube-traffic-access-token (native alias). Defaults unchanged.

* CubeMaster: SandboxProxyMap gains AllowPublicTraffic / TrafficAccessToken; setProxyToRedis() mints a UUID token after Redis acks the write. mergeCubeNetworkConfigs now propagates the flag so the per-create override wins over the template's default.
* CubeProxy: enforce_traffic_token() in sandbox_backend.lua gates both the cold path and the cache-hit path; 403 has empty body; token values are never logged.
* CubeAPI: allowPublicTraffic threaded into the wire request, traffic_access_token surfaced on the create response.
* Python SDK: Sandbox.traffic_access_token property (None for unrestricted sandboxes; delivered exactly once at create time).
* Docs/examples: new Networking sidebar group (Network Policy renamed from Egress Network Policy, Security Proxy moved in, Restrict Public Access added, en + zh). New restrict_public_access.py TUI demo.

Verified on a PVM host with the all-in-one test image: no header → 403, e2b header → 200, cube header → 200.


CubeProxy/lua/sandbox_backend.lua
```diff
@@ -94,6 +125,15 @@ function _M.resolve_backend(ins_id, container_port)
     local host_ip = cache:get(cache_backend_ip_key)
     local host_port = cache:get(cache_backend_port_key)
     if host_ip and host_port then
+        -- Cache-hit path must still enforce the per-sandbox traffic token,
+        -- otherwise a single warm entry would let unauthenticated callers
+        -- bypass the gate for the whole cache TTL. The shared
+        -- "<ins_id>:<field>" cache lines below (set on the first miss) carry
+        -- exactly the same data as Redis.
+        enforce_traffic_token(
+            cache:get(ins_id .. ":AllowPublicTraffic"),
```
fslongjin: Blocking: the auth metadata is cached under separate shared-dict keys with the same initial TTL as the backend, but this hot path refreshes only `backend_ip`/`backend_port` below. Under steady traffic, those backend keys stay alive while `AllowPublicTraffic`/`TrafficAccessToken` expire; then `allow_public` is `nil`, `enforce_traffic_token()` returns as if the sandbox were public, and unauthenticated requests are forwarded. Please refresh the auth metadata with the backend keys, or treat missing auth metadata on a backend cache hit as a cache miss/fail-closed path and add coverage for this case.

sdk/python/cubesandbox/sandbox.py
```diff
@@ -97,6 +97,25 @@ def template_id(self) -> str:
...
+        with HTTP 403 when the sandbox was created with public access
+        restricted.
+
+        ``None`` for sandboxes with the default (publicly reachable)
+        configuration, including instances obtained via :meth:`connect` /
+        :meth:`resume` (the token is only delivered on the original create
+        response). Persist it client-side at create time if you need it
+        later.
+        """
+        token = self._data.get("trafficAccessToken")
```
fslongjin, line 116: This exposes the token but the SDK's own data-plane paths do not use it. `run_code()` sends only `Content-Type`, commands/files add envd/user headers, and `build_client()` has no default traffic-token header, so a sandbox created with `network={"allow_public_traffic": False}` will likely make `run_code`, `commands.run`, and file operations fail with 403 unless callers bypass the SDK. Please attach `e2b-traffic-access-token` (or the cube alias) automatically to all CubeProxy/envd requests when `trafficAccessToken` is present.

docs/guide/restrict-public-access.md
```diff
@@ -0,0 +1,145 @@
...
+but knowledge of it is the only thing standing between the workload and
+the internet. For sensitive use cases (long-running agents handling
+private data, dashboards exposed to partner networks, demos that
+shouldn't be replayed) Cube Sandbox can require every caller to
+authenticate with a per-sandbox token before any request reaches the
+services inside.
+
+This page is the operator-facing counterpart of E2B's
+[Restricting public access][e2b-doc] feature; both share the
+`network.allowPublicTraffic` parameter and the `e2b-traffic-access-token`
```
fslongjin, line 14: These docs introduce the REST wire parameter, but the checked-in `openapi.yml` does not include `allowPublicTraffic` in the `SandboxNetworkConfig` request schema. Generated clients and REST users that rely on the OpenAPI contract cannot discover or set this feature, even though the create path handles it. Please update/regenerate the OpenAPI schema for the new request field (and keep the response token documented).

chenhengqi: Good point. Will do in a follow-up PR.

CubeMaster/pkg/localcache/redis_cache.go
```diff
@@ -365,6 +366,18 @@ func (l *local) getByPassProsyFromRedis(ctx context.Context, key string) (*types
...
+	} else {
+		nodeIdIp.AllowPublicTraffic = true
+	}
+	if v, ok := mapvalues["TrafficAccessToken"]; ok {
+		nodeIdIp.TrafficAccessToken = v
+		delete(mapvalues, "TrafficAccessToken")
+	}
 
 	if len(mapvalues) == 0 {
 		log.G(ctx).Warnf("key: %s,has no ContainerToHostPorts", key)
```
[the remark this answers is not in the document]
chenhengqi: Wrong.

fslongjin (approved):

mthalanesamkelo212-cmyk (approved):

mthalanesamkelo212-cmyk (approved):

PR #651, opened 2026-06-25
feat(cube-bench): add host-mount flag for create requests

zyl1121:
## Motivation

`examples/cube-bench` currently benchmarks the default raw sandbox create path. That makes it hard to measure create-time overhead for `host-mount`, which is the concrete create-time variation we want to compare.

CubeAPI already accepts `host-mount` through `metadata["host-mount"]` as a JSON string. `cube-bench` should be able to send that same create-time payload so host-mount sandbox setups can be benchmarked directly.

## What Changed

This PR adds a `--host-mount '<json-array>'` flag to `cube-bench` and includes the provided host-mount payload in sandbox create requests.

Implementation highlights:

* `examples/cube-bench/main.go` adds the new flag, validates a non-empty host-mount JSON array once during config parsing, preserves the original JSON for config display and exported reports, and caches the compacted `host-mount` string plus request body for request-time reuse
* `examples/cube-bench/runner.go` sends the cached request body and headers on sandbox create requests, so benchmark throughput is not distorted by re-building host-mount metadata on every request
* `examples/cube-bench/main_test.go` covers valid host-mount arrays, invalid JSON, non-array input, empty input, and empty-array rejection
* `examples/cube-bench/README.md` documents the new flag and a host-mount example

This keeps the benchmark helper focused on the current benchmark need without changing the backend metadata contract.

## Validation

* `go test ./...` passed in `examples/cube-bench`
* focused tests cover:
  * valid host-mount arrays being compacted once before request-time use
  * invalid host-mount JSON being rejected
  * non-array host-mount input being rejected
  * empty host-mount arrays being rejected
  * final create requests keeping `metadata["host-mount"]` as a string value
* real-machine validation passed with the `cube-bench` binary using a `--host-mount '[...]'` payload

### DIFF
--- a/examples/cube-bench/README.md
+++ b/examples/cube-bench/README.md
@@ -54,6 +54,7 @@ All env vars can be overridden by the corresponding flag.
 | `-w`, `--warmup` | `0` | Warmup rounds before measurement |
 | `-m`, `--mode` | `create-delete` | `create-delete` or `create-only` |
 | `-o`, `--output` | *(none)* | Export JSON report to file |
+| `--host-mount` | *(none)* | Host mount list as a JSON array |
 | `--api-url` | *(env)* | CubeAPI base URL |
 | `--api-key` | *(env)* | API key |
 | `--theme` | `auto` | Color theme: `dark`, `light`, or `auto` |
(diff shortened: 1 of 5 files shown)


ls-ggg:
Thanks for the PR. Since we won't have many options similar to host-mount for benchmarks, could you just add host-mount as a standalone parameter?

zyl1121:
@ls-ggg Makes sense. I narrowed this PR to a dedicated `--host-mount` flag and updated the title/body/document accordingly.

ls-ggg:
LGTM

PR #652, opened 2026-06-25
fix(cubeapi): forward with_cube_ca when creating templates

xiaojunxiang2023:
## Summary

CubeMaster already supports the `with_cube_ca` flag on template creation, but CubeAPI did not accept or forward it when handling `POST /templates`. As a result, API clients could not control whether the CubeEgress root CA is baked into the template rootfs.

This change:
- Adds `with_cube_ca` to `CreateTemplateRequest`
- Adds the same field to `CreateTemplateFromImageReq` sent to CubeMaster
- Forwards the value in `TemplateService::create_template`

When omitted or `null`, CubeMaster keeps its existing default behavior (`true`).

## Test plan

- [x] `cd CubeAPI && cargo test` (70 passed)
- [x] `cd CubeMaster && go test ./pkg/templatecenter/... -run 'WithCubeCA|CubeEgress'`

### DIFF
--- a/CubeAPI/src/cubemaster/mod.rs
+++ b/CubeAPI/src/cubemaster/mod.rs
@@ -1832,6 +1832,9 @@ pub struct CreateTemplateFromImageReq {
         skip_serializing_if = "Option::is_none"
     )]
     pub cube_network_config: Option<CreateTemplateCubeNetworkConfig>,
+    /// Whether CubeMaster bakes the CubeEgress root CA into the template rootfs.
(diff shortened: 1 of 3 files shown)


chenhengqi (approved):
LGTM. I never knew that we can create template through CubeAPI. Probably we can implement template APIs for SDK.

chenhengqi:
cc @fslongjin @ls-ggg @tinklone @wbzdssm

PR #657, opened 2026-06-26
fix(install): fail-fast when compute node missing control plane address

fslongjin:
## Summary

This PR fixes a long-standing issue where `install-compute.sh` silently succeeded even when the user forgot to configure the control plane address, only failing much later when systemd tried to start the services.

### 1. Fail-fast preflight for compute nodes

Added `check_compute_control_plane_preflight()` — a preflight called during `install.sh` that validates the control plane address configuration (via `ONE_CLICK_CONTROL_PLANE_IP` or `ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR`) **before** any package installation or extraction. If neither variable is set for a compute role, the installer dies immediately with a clear, actionable error message.

### 2. Decouple port derivation from CUBEMASTER_ADDR

The compute-node IP-branch port was previously derived from `CUBEMASTER_ADDR` via `${CUBEMASTER_ADDR##*:}`. `CUBEMASTER_ADDR` is the control node local listen address; borrowing its port for the remote connection was accidental coupling. The port is now a fixed constant (`8089` — the cubemaster protocol port) in all 4 copies of the resolution logic.

Note: `ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR` (which accepts an explicit `host:port`) is completely unaffected.

### 3. Guard against conflicting dual-variable config

When both `ONE_CLICK_CONTROL_PLANE_IP` and `ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR` are set but resolve to different addresses, the preflight now dies with a clear error. Previously the addr would silently take priority, hiding the IP mismatch from the user.

## Files changed

| File | Change |
|------|--------|
| `install.sh` | Wire new preflight into the fail-fast block |
| `lib/common.sh` | New `check_compute_control_plane_preflight()` function + conflict guard |
| `scripts/one-click/common.sh` | `port="${CUBEMASTER_ADDR##*:}"` → `cubemaster_port=8089` |
| `scripts/systemd/common.sh` | Same constant fix |
| `scripts/cube-diag/check-procs.sh` | Same constant fix (standalone function, kept independent) |
| `tests/test_install_mode.sh` | 10 new test cases covering all branches |

## Backward compatibility

- `ONE_CLICK_CONTROL_PLANE_IP=x` → still resolves to `x:8089` (value unchanged)
- `ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR=host:port` → still passed through verbatim
- Control node `CUBEMASTER_ADDR` health check usage is untouched
- Compatible with the Terraform deployer (PR #629) which uses `ONE_CLICK_CONTROL_PLANE_IP`

### DIFF
--- a/deploy/one-click/install.sh
+++ b/deploy/one-click/install.sh
@@ -876,6 +876,7 @@ check_pvm_consistency_preflight
 check_cubelet_fs_preflight
 check_cgroup_cpu_preflight
 check_glibc_preflight
+check_compute_control_plane_preflight
 
 CUBE_SANDBOX_NODE_IP="$(detect_node_ip)"
 export CUBE_SANDBOX_NODE_IP
--- a/deploy/one-click/lib/common.sh
+++ b/deploy/one-click/lib/common.sh
@@ -1756,3 +1756,71 @@ EOF
 
   log "glibc version ${glibc_ver} OK (>= ${min_major}.${min_minor})"
 }
+
+# check_compute_control_plane_preflight: fail fast when a compute node is
+# missing the mandatory control plane address. This mirrors the resolution
+# logic in resolve_control_plane_cubemaster_addr() (both scripts/one-click/
+# and scripts/systemd/) and must run before package extraction or dependency
+# installation so the user gets a friendly, actionable error before any
+# destructive change.
+check_compute_control_plane_preflight() {
+  local role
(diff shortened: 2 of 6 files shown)


chenhengqi (approved):
LGTM.

PR #659, opened 2026-06-26
cubelet: allow read-only host_dir volumes in CommitSandbox

eavae:
## Problem

`CommitSandbox` refuses to snapshot any sandbox that has a mounted `host_dir` volume:

```
host_dir volume <name> is not supported by CommitSandbox
```

The intent is sound for a **writable** host_dir — its writes land on the host, so a self-contained snapshot can't reproduce them. But the check rejects **all** host_dir volumes, including **read-only** ones. A read-only host_dir cannot have received guest writes, so the snapshot loses no state by omitting it; the caller simply re-supplies the same mount on restore (create-from-snapshot re-passes the host-mount metadata).

This blocks a common, useful pattern: mounting a shared, host-supplied **read-only** directory (e.g. a tools dir shared across all sandboxes) while still snapshotting the user's writable rootfs. Today such a sandbox can never be committed.

## Change

`validateCommitVolumeSources` now records which volumes are mounted writable and rejects a `host_dir` volume only when it is mounted writable. Read-only host_dir volumes pass commit. Writable host_dir (and `sandbox_path` Directory/SharedBindMount) volumes remain rejected as before.

## Testing

```
CGO_ENABLED=0 go test ./services/cubebox/ -run TestValidateCommitSandboxTarget
```

All pass, including:
- new `TestValidateCommitSandboxTargetAllowsReadonlyHostDirVolume` (read-only host_dir → commit allowed)
- existing `TestValidateCommitSandboxTargetRejectsHostDirVolume` (writable host_dir → still rejected)

### DIFF
--- a/Cubelet/services/cubebox/service_test.go
+++ b/Cubelet/services/cubebox/service_test.go
@@ -301,6 +301,31 @@ func TestValidateCommitSandboxTargetRejectsHostDirVolume(t *testing.T) {
 	assert.Contains(t, err.Error(), "host_dir")
 }
 
+func TestValidateCommitSandboxTargetAllowsReadonlyHostDirVolume(t *testing.T) {
+	cb := newRunningCommitSandboxForTest([]*cubeboxv1.Volume{{
+		Name: "host",
+		VolumeSource: &cubeboxv1.VolumeSource{
+			HostDirVolumes: &cubeboxv1.HostDirVolumeSources{
+				VolumeSources: []*cubeboxv1.HostDirSource{{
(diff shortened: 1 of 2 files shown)


kinwin-ustc:
Have you tested that after your modifications, after taking a snapshot of the sandbox where the read-only directory is mounted, the sandbox recovered from this snapshot still mounts the read-only directory?

eavae:
Sorry, I tested, it's not working for the following snapshot and remount. There are not a clear path to achieve this, so I will close this PR.

Issue #573, opened 2026-06-16
[CubeMaster] Scheduler binds to first sorted node when scoring is unconfigured


dushulin:
## Summary

  When `scheduler.score` is omitted from `cubemaster.yaml`, the scheduler degenerates to "always pick the first node by NodeID lexicographic order", causing severely unbalanced sandbox placement across compute nodes.

  ## Environment

  - CubeSandbox version / commit: 386a657 (`Improve network performance #469`, branch `master-dev-v0.3.1`)
  - Host OS and kernel version: Linux 6.8.0-54-generic x86_64
  - KVM info (`modinfo kvm`): /lib/modules/6.8.0-54-generic/kernel/arch/x86/kvm/kvm.ko.zst
  - Deployment mode: cluster (10 compute nodes)
  - Relevant component: CubeMaster

  ## Steps to Reproduce

  1. Deploy a multi-node cluster using the shipped `configs/single-node/cubemaster.yaml` (or any config that sets `scheduler.filter` but leaves `scheduler.score` unset).
  2. Concurrently create a large number of sandboxes (e.g. 200 per batch via the SDK) backed by a single template.
  3. Inspect the per-node sandbox distribution via `GET /internal/node?host_id=<ip>` or by counting `containerd-shim-cube-rs` processes on each compute node.

  ## Expected Behavior

  Sandboxes spread roughly evenly across all healthy compute nodes that pass the filters, weighted by their real-time resource availability.

  ## Actual Behavior

  The scheduler always picks `result[0]` of the prefilter output:

  - `prefilter` returns `localcache.GetHealthyNodesByInstanceType` ordered by `Node.Index` (`AllSortByIndex`, ascending).
  - For nodes registered via `nodemeta.toSchedulerNode`, `Node.Index` is never set (defaults to `0`), so the effective order falls back to the upstream `ListNodes` sort, which is `sort.Slice(out, NodeID < )`.
  - `runScoreFilter` returns immediately when `scheduler.score` is empty (see `pkg/selector/score/init.go` and `pkg/scheduler/schedule.go`), leaving the slice unchanged.
  - `LeastRandomSelect(priority_select_num)` with the default `priority_select_num=1` takes `result[0:1]`. The `randomSelect` over a single-element slice is deterministic.

[...]

PR #672, opened 2026-06-29
docs(cubemaster): add multi-node scheduler scoring guidance

YangYuS8:
## Summary

- Document that multi-node CubeMaster deployments should configure `scheduler.score`
- Explain why an unconfigured scheduler score may lead to deterministic placement on the first sorted node
- Add a baseline scoring example based on the workaround discussed in #573

Refs #573

## Checks

- `git diff --check`

Assisted-by: codex:gpt-5.5

### DIFF
--- a/docs/guide/multi-node-deploy.md
+++ b/docs/guide/multi-node-deploy.md
@@ -117,6 +117,38 @@ curl http://127.0.0.1:8089/internal/meta/nodes
 
 The response should include the compute node's IP and a healthy status.
 
+## Configure CubeMaster Scheduler Scoring
+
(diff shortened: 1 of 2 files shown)


YangYuS8:
I’ve iterated on the scheduler scoring guidance based on the review comments and tried to keep the PR focused on the documentation/config workaround requested in #573.

Some of the remaining bot suggestions are starting to go deeper into scheduler tuning policy, such as the exact `resource_weights`, whether to include `req_cpu` / `req_mem`, and larger-cluster formulas for `priority_select_num`. Since those choices may depend on maintainer preference and production assumptions, I’d prefer not to keep expanding this docs PR without human maintainer guidance.

Could a maintainer take a look and let me know whether the current baseline guidance is acceptable, or which specific config you would prefer this document to recommend?

PR #674, opened 2026-06-29
feat(cubemaster): add PostgreSQL support for dao layer

### LINKED ISSUE #517: [Feature Request] PostgreSQL support for CubeMaster dao layer
## Problem / Motivation

CubeMaster currently supports only MySQL as its metadata store. The `dao` package (`CubeMaster/pkg/base/dao`) already ships a driver-agnostic facade — the `Driver` interface, the `Config.Driver` field, and the `dialectSpecs` map all explicitly anticipate additional engines (the code comments even name `"postgres"` as a future driver). However, no PostgreSQL driver exists yet, so operators who standardize on PostgreSQL in their infrastructure cannot deploy CubeMaster without also running a MySQL instance.

Use cases:
- Organizations with a PostgreSQL-only database policy / managed PG offering (e.g. cloud RDS for PostgreSQL) who do not want to introduce MySQL.
- Teams that prefer PostgreSQL operational tooling and want a single database engine across their stack.

## Proposed Solution

Add a PostgreSQL driver alongside the existing MySQL one, with zero changes to business code (which already uses `*gorm.DB` exclusively). Scope:

1. **Driver implementation** — `pkg/base/dao/driver/postgres/`:
   - Implement the `dao.Driver` interface (`Name`, `Open`, `SessionLocker`).
   - Use `gorm.io/driver/postgres` + `pgx` for the connection.
   - Implement `SessionLocker` via `pg_advisory_lock` / `pg_advisory_unlock` (mapping the named MySQL `GET_LOCK` key to an advisory-lock key).

2. **Migrations** — `pkg/base/dao/migrate/migrations/postgres/`:
   - Translate the three existing MySQL migrations to PostgreSQL dialect (BIGSERIAL, boolean, timestamptz, `ALTER TABLE ... IF NOT EXISTS`, `UPDATE ... FROM` instead of `UPDATE ... JOIN`, etc.).
   - Register `"postgres"` in `dialectSpecs` and embed the new SQL files.

3. **Wiring & tests** — blank-import the driver in `cmd/cubemaster`, and add integration tests (testcontainers / docker-compose) so the same schema and dao behavior is validated against PostgreSQL.

All changes are additive — the existing MySQL driver and migrations are not modified. `Config.Driver: "postgres"` selects the new engine; empty/`"mysql"` keeps current behavior.

## Alternatives Considered

- **External MySQL-compatibility proxy in front of PostgreSQL** — fragile, adds an operational component, and does not cover DDL/locking semantics.
- **Keep MySQL-only and document it as a hard requirement** — blocks PG-standardized deployments; leaves the already-built driver abstraction unused.

## Additional Context

- The `dao.Driver` abstraction, `Config.Driver` field, and `dialectSpecs` map already exist and are documented as the intended extension points.
- Happy to split this into sequential PRs (driver skeleton -> migrations + tests) or land it as one, depending on maintainer preference.
- Before writing the migration SQL, I would like to confirm: (a) is PostgreSQL support a direction the maintainers want to accept, and (b) any preference on the PG driver (`pgx` vs `lib/pq`) and on CI running both engines?

cookiebody:
## Summary

Implement the PostgreSQL driver for CubeMaster's data-access layer, fulfilling the multi-database design pre-wired into the `dao.Driver` interface.

Closes #517

## Changes

- **New driver**: `pkg/base/dao/driver/postgres/` — pgx-based `dao.Driver` implementation with `pg_advisory_lock` session locker
- **13 PG migrations**: `migrations/postgres/` — 1:1 translation of all MySQL migrations producing identical logical schema
- **Fingerprint refactor**: extract `fingerprintStore` interface with per-dialect implementations (MySQL: `ON DUPLICATE KEY UPDATE`, PG: `ON CONFLICT DO UPDATE`)
- **Dialect registration**: register `"postgres"` in `dialectSpecs` map + embed PG migrations
- **Blank import**: register postgres driver in `cmd/cubemaster/app/main.go`
- **Dependencies**: add `gorm.io/driver/postgres` + `github.com/jackc/pgx/v5`

## How to use

Set `driver: postgres` in conf.yaml:

```yaml
ossdb_config:
  driver: postgres
  addr: "127.0.0.1:5432"
  user: cube
  pwd: cube_pass
  db_name: cube_mvp
```

## Testing

- All existing MySQL integration tests pass without regression
- New PostgreSQL integration tests (`dockertest` + `postgres:16-alpine`):
  - `TestPostgres_Run_Fresh` — full migration from empty DB to HEAD ✅
  - `TestPostgres_Run_Idempotent` — re-run is a no-op ✅

## Design decisions

- **Pure additive**: zero changes to existing MySQL driver/migrations
- **pgx over lib/pq**: pgx is actively maintained, faster, and the GORM postgres driver uses it natively
- **Advisory locks**: `pg_try_advisory_lock` with retry loop mirrors MySQL `GET_LOCK(name, timeout)` semantics
- **Fingerprint interface**: follows the comment in `fingerprint.go` that explicitly calls for extracting an interface before adding a second dialect


chenhengqi (commented):
Why do we have to duplicate all 13 *.sql files?

kinwin-ustc:
Will this PR affect the MySQL part? If not, I think it's not a big deal. It would be best to have some documentation explaining how to use PostgreSQL as the database for CubeMaster, how to configure CubeMaster, etc. Of course, this can be included as part of a future PR.

cookiebody:
Thanks for the review!

**@chenhengqi** — The 13 SQL files under `migrations/postgres/` are **not copies** of the MySQL migrations. They are PostgreSQL-native rewrites because the two dialects have incompatible DDL syntax:

| MySQL | PostgreSQL equivalent |
|-------|---------------------|
| `AUTO_INCREMENT` | `bigserial` / `GENERATED ALWAYS AS IDENTITY` |
| `tinyint(1)` | `boolean` |
| `mediumtext` | `text` |
| `ENGINE=InnoDB` | (no equivalent, removed) |
| `GET_LOCK()` / `RELEASE_LOCK()` | `pg_advisory_lock()` / `pg_advisory_unlock()` |
| Stored procedures (`DELIMITER //`) | PL/pgSQL `DO $$...$$` blocks |
| `ON DUPLICATE KEY UPDATE` | `ON CONFLICT ... DO UPDATE` |
| `INSERT IGNORE` | `ON CONFLICT DO NOTHING` |
| `UPDATE ... JOIN` | `UPDATE ... FROM ... WHERE` |
| backtick-quoted identifiers | double-quoted or unquoted |

goose runs migrations per-dialect from embedded FS (`//go:embed migrations/postgres/*.sql`), which is the same pattern already used for MySQL (`//go:embed migrations/mysql/*.sql`). Sharing a single set of SQL files across both dialects isn't feasible without a templating layer that would add complexity.

**@kinwin-ustc** — Correct, this PR is purely additive and does **not** touch any existing MySQL code paths. All existing MySQL tests continue to pass unchanged. I'll follow up with a documentation PR explaining PG configuration (config fields, sslmode, example YAML) once this lands.

---

Regarding the automated review findings — I've addressed all Critical and High items in the latest force-push:

1. ✅ **sslmode configurable** via `Extra["sslmode"]` (defaults to `disable` for dev; production users set `require`/`verify-full`)
2. ✅ **Timeouts mapped** — `WriteTimeoutSeconds` → `statement_timeout`, `ReadTimeoutSeconds` → `idle_in_transaction_session_timeout` (via PG connection `options` parameter)
3. ✅ **Exponential backoff** in session locker (200ms → 400ms → … capped at 2s)
4. ✅ **Test locker** now context-aware with select/ctx.Done()
5. ✅ **preflightFingerprints** delegates to `preflightFingerprintsWithStore` (no duplication)
6. ✅ **Open() error path** covered by `TestOpenInvalidDSN`
7. ✅ **Fingerprint drift test** added for PostgreSQL

xiaojunxiang2023:
@cookiebody @kinwin-ustc @chenhengqi Hi, PostgreSQL support would be really useful for my use case. Could we please continue moving this PR forward?

kinwin-ustc:
Yes, we will merge this PR in version 0.6.0 @xiaojunxiang2023

fslongjin:
I'll handle with this PR today.

PR #681, opened 2026-06-30
CubeVS: Allow sandbox traffic to host service

chenhengqi:
Previously, it's not possible to access host service from inside sandboxes even the IP is in allow_out. Let's allow it for specific use cases. Note that by default TCP traffic to local port 80 and 443 are intercept by CubeEgress. if you want to bypass that, you will need to add an iptables rule like:
```
iptables -t mangle -I TRANSPROXY 1 -d $HOST_IP/32 -p tcp --dport 80 -j RETURN
iptables -t mangle -I TRANSPROXY 1 -d $HOST_IP/32 -p tcp --dport 443 -j RETURN
```


zhouxianping (approved):

zyl1121:
I also discovered this issue and implemented the same fix (redirect to `cubegw0_ifindex` when `daddr == nodenic_ip`). Confirmed working on a live server.

A couple of suggestions based on my implementation:

The cached `daddr` variable (line 963) might be a better fit than `l3->daddr`, as the bot also pointed out for consistency.

Regarding the L7 proxy bypass concern: for `nodenic_ip` traffic, bypassing L7 should be the correct behavior. CubeEgress transparent proxy is for outbound internet traffic, not host-local services. Routing through L7 proxy would break the request since the proxy expects external destinations.

I placed my check after `check_net_policy()` (before `pull_headers`/`should_do_nat`). Functionally equivalent since both redirect before `do_tcp_nat()`, but saves one unnecessary `pull_headers` call for the nodenic_ip case. Either placement works fine.

My patch:

```c
if (daddr == nodenic_ip)
    return bpf_redirect(cubegw0_ifindex, BPF_F_INGRESS);
```

Test results with `allow_out=["10.0.16.9/32"]`:

- Before: timeout, dmesg `martian source 10.0.16.9 from 10.0.16.9, on dev eth0`
- After: `code=200`, cube-dev tcpdump shows `192.168.1.x → 10.0.16.9:18099 [S]`
- Default deny still blocks unconfigured sandbox, public network unaffected, cross-node traffic unaffected

PR #688, opened 2026-07-01
Compatible with macOS Bash 3.0

wbzdssm:
compatible with macOS Bash 3.0  for terraform


kinwin-ustc:
There are too many formatting changes in this PR, please only submit the actual changes

PR #691, opened 2026-07-01
fix(cubelet): fail fast on hung hostdir mounts

zyl1121:
## Motivation

An unhealthy host mount, such as a geesefs path with revoked remote access, can block the Cubelet hostdir bind step indefinitely. When that happens, sandbox creation hangs until an upstream API timeout even though the real failure is already at the hostdir mount boundary.

Cubelet should fail fast when a hostdir bind cannot complete, instead of leaving the create path blocked by a hung mount operation.

## What Changed

This PR makes Cubelet hostdir mounts fail fast by moving the bind operation out of the create goroutine and into a bounded helper subprocess.

Implementation highlights:

* `Cubelet/storage/hostdir.go` replaces the in-process `unix.Mount(...)` hostdir bind with `mount --rbind` executed through a helper subprocess
* each hostdir mount step is bounded to 3s
* read-only hostdir mounts keep the follow-up `remount,bind,ro` step under the same bounded helper path
* hostdir backend metadata is recorded before the bind attempt so cleanup still knows the sandbox-scoped hostdir paths after a failed create
* the top-level error category stays `CreateStorageFailed`, while the returned message now makes the hostdir timeout explicit
* `Cubelet/storage/hostdir_test.go` adds focused coverage for command shape, timeout/error handling, and real subprocess timeout cleanup

The timeout is per mount step rather than a flat per-request cap:

* read-write hostdir: up to about 3s per volume
* read-only hostdir: up to about 6s per volume, because bind and remount are bounded separately

## Testing

Focused unit tests:

```bash
go test ./storage -run HostDir -count=1
```

Covered cases:

* read-write hostdir helper command shape
* read-only hostdir bind + remount helper command shape
* timeout handling
* generic mount-error handling
* real subprocess timeout behavior, including prompt return and child process-group cleanup

Real-cluster healthy-path validation:

* with a healthy host mount, sandbox creation still succeeded in about `0.108s`

Real-cluster failure-path validation:

* reproduced an unhealthy geesefs-backed host mount where `ls` on the host path blocked indefinitely
* before this fix, sandbox creation hung until the upstream API timeout
* after this fix, CubeAPI returned `500 Internal Server Error` in about `3010ms`
* Cubelet returned `CreateStorageFailed` with a hostdir-specific timeout message, for example:

```text
prepareHostDirVolume: bind mount /mnt/aime-oss/test -> /data/cubelet/hostdir/<sandbox-id>/rw/hostdir-0 timed out after 3s: context deadline exceeded
```

* CubeMaster propagated the same `CreateStorageFailed` message in `CreateSandbox_rsp fail`
* post-failure checks showed no surviving sandbox in `cubecli cubebox list`
* post-failure checks showed no leftover sandbox hostdir root under `/data/cubelet/hostdir/<sandbox-id>`

Behavioral clarification:

* failure logs may still show a `sandboxID`, `sandboxIP`, and `port_mappings`
* this is expected because `createid` runs before storage, and storage/network run before the final `cubebox` step
* in the reproduced failure, the actual cubebox VM was not created

## Scope

This PR only changes Cubelet hostdir mount failure handling.

### DIFF
--- a/Cubelet/storage/hostdir.go
+++ b/Cubelet/storage/hostdir.go
@@ -5,11 +5,15 @@
 package storage
 
 import (
+	"bytes"
 	"context"
 	"errors"
 	"fmt"
 	"os"
+	"os/exec"
 	"path/filepath"
+	"syscall"
+	"time"
 
 	cubebox "github.com/tencentcloud/CubeSandbox/Cubelet/api/services/cubebox/v1"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/pkg/log"
@@ -20,6 +24,12 @@ import (
 
 var hostDirBasePath = "/data/cubelet/hostdir"
 
+const hostDirMountTimeout = 3 * time.Second
+
+const hostDirMountBinary = "/usr/bin/mount"
+
+var runHostDirCommand = defaultRunHostDirCommand
+
 type HostDirBackendInfo struct {
 	VolumeName string `json:"volume_name"`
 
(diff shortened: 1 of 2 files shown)


Cubelet/storage/hostdir.go
```diff
@@ -30,6 +40,72 @@ type HostDirBackendInfo struct {
 	ReadOnly bool `json:"read_only"`
 }
 
+func defaultRunHostDirCommand(ctx context.Context, name string, args ...string) error {
+	cmd := exec.Command(name, args...)
```
kinwin-ustc, line 44: Some command helpers in shell.go are more robust ; you might consider using it.

zyl1121: Thanks. I looked at the helpers in `storage/shell.go` / `utils.Exec*`.

I kept a dedicated helper here because this path needs one extra guarantee for the stuck-mount case: on timeout, it explicitly kills the whole process group with `Kill(-pid, SIGKILL)`, rather than only relying on the default `exec.CommandContext` cancellation behavior.

The local runner also keeps the timeout / cancel / remount-failure paths directly injectable in unit tests. So I think a small dedicated helper is safer for this hostdir mount path.

kinwin-ustc: Make sense, but in the case of backend storage unresponsiveness, the process is likely to be in the D state, and it will not be killed by SIGKILL. This PR avoids this behavior from blocking the entire cublet, which is a very good design optimization

kinwin-ustc:
LGTM， The only concern is whether the hard-coded mount binary path covers all OS scenarios

PR #714, opened 2026-07-02
fix(cubeproxy): configure resolver for Redis hostnames

zyl1121:
## Motivation

When CubeProxy connects to Redis by hostname, OpenResty Lua can fail with:

```text
no resolver defined to resolve "<redis-hostname>"
```

NGINX/OpenResty requires an explicit `resolver` directive for Lua cosocket hostname resolution. Reading `/etc/resolv.conf` is not enough by itself, because nginx expects concrete resolver addresses in `nginx.conf`.

CubeProxy has two independent Redis connections: the request path reads `$redis_ip` from `global.conf`, while the heartbeat and registration path reads `CUBE_PROXY_REGISTRY_REDIS_HOST`. Following the review feedback, this PR keeps the fix focused on the shared resolver path so both connections use the same container-side nginx resolver configuration.

## What Changed

This PR makes CubeProxy prepare only the resolver include at container startup:

* adds a dedicated `conf/includes/resolver.inc` include hook at the nginx `http` level
* updates `CubeProxy/start.sh` to:
  * reuse the existing Helm `CUBE_PROXY_RESOLVER_ADDRS` override when configured, or discover resolver addresses from the container's `/etc/resolv.conf`
  * validate and normalize IPv4 and IPv6 resolver addresses, including optional ports
  * always render available resolver addresses instead of trying to predict which Lua path will need DNS
  * read the business Redis target from the mounted `global.conf`, matching the source nginx actually uses
  * check `CUBE_PROXY_REGISTRY_REDIS_HOST` when proxy registration is enabled
  * accept trailing comments on the `$redis_ip` declaration and fail clearly when the declaration cannot be classified
  * honor the existing resolver `valid`, `timeout`, and IPv6 settings used by the Helm chart
  * fail fast only when an enabled Redis target is a hostname and no valid resolver is available
  * keep IP-only deployments working when no resolver is available
* adds focused resolver regression coverage in `CubeProxy/tests/test_start.sh` and exposes it through `make -C CubeProxy test`

The final change is limited to the CubeProxy component. It does not rework the broader one-click or Helm rendering flows.

## Validation

* Ran the focused CubeProxy shell tests with `make -C CubeProxy test`.
* Verified the CubeProxy startup helper across the relevant resolver branches:
  * available resolver addresses are always written, including for IP-only Redis targets
  * the business and registry Redis paths both fail fast when a hostname requires DNS but no resolver is available
  * `global.conf` entries with trailing comments are parsed correctly
  * `CUBE_PROXY_REDIS_IP` does not override the `global.conf` value nginx actually consumes
  * IPv4 and IPv6 resolver addresses, optional ports, and resolver tuning values are validated
* Built the CubeProxy image successfully in a Linux environment.
* Verified the container-side startup path in a temporary container: `start.sh` rendered `conf/includes/resolver.inc` with `[2001:db8::53]:5353` and `172.18.0.10:5353`, and the target OpenResty `nginx -t` succeeded.
* Verified `bash -n CubeProxy/start.sh CubeProxy/tests/test_start.sh` and `git diff --check origin/master...HEAD`.

## Scope

This PR fixes Redis hostname resolution through CubeProxy's nginx/OpenResty resolver configuration.

It intentionally does not change:

* CubeProxy's Lua Redis access logic
* non-resolver CubeProxy routing behavior
* the wider one-click nginx rendering flow
* the existing Helm chart rendering flow

### DIFF
--- a/CubeProxy/Makefile
+++ b/CubeProxy/Makefile
@@ -42,6 +42,10 @@ build:
 		-t $(IMAGE_LOCAL):$(IMAGE_TAG)-$(ARCH) \
 		.
 
+.PHONY: test
+test:
+	bash tests/test_start.sh
+
 # Push the architecture-specific image to all registries.
 # Run `make push` on an amd64 host, then again on an arm64 host, then
 # `make manifest` on either host to combine them.
--- a/.github/workflows/unit-test-check.yml
+++ b/.github/workflows/unit-test-check.yml
@@ -7,6 +7,7 @@ on:
       - 'CubeAPI/**'
       - 'CubeOps/**'
       - 'CubeDB/**'
+      - 'CubeProxy/**'
       - 'CubeShim/**'
       - 'CubeNet/**'
       - 'network-agent/**'
@@ -19,6 +20,7 @@ on:
       - 'CubeAPI/**'
       - 'CubeOps/**'
       - 'CubeDB/**'
+      - 'CubeProxy/**'
       - 'CubeShim/**'
       - 'CubeNet/**'
       - 'network-agent/**'
@@ -36,6 +38,7 @@ on:
         options:
           - all
           - cubeapi
+          - cubeproxy
           - cubeshim
           - network-agent
 
@@ -84,6 +87,7 @@ jobs:
           add_all_components() {
             add_component "CubeAPI"
             add_component "CubeOps"
+            add_component "CubeProxy"
             add_component "CubeShim"
             add_component "network-agent"
           }
@@ -94,6 +98,7 @@ jobs:
               all) add_all_components ;;
               cubeapi) add_component "CubeAPI" ;;
               cubeops) add_component "CubeOps" ;;
+              cubeproxy) add_component "CubeProxy" ;;
               cubeshim) add_component "CubeShim" ;;
               network-agent) add_component "network-agent" ;;
               *)
@@ -130,6 +135,9 @@ jobs:
                 CubeOps/*|CubeDB/*)
                   add_component "CubeOps"
                   ;;
+                CubeProxy/*)
+                  add_component "CubeProxy"
+                  ;;
                 CubeShim/*)
                   add_component "CubeShim"
                   ;;
@@ -154,6 +162,7 @@ jobs:
                   display_name: (
                     if . == "CubeAPI" then "CubeAPI"
                     elif . == "CubeOps" then "CubeOps"
+                    elif . == "CubeProxy" then "CubeProxy"
                     elif . == "CubeShim" then "CubeShim"
                     elif . == "network-agent" then "Network Agent"
                     else error("unsupported component: " + .)
@@ -162,11 +171,13 @@ jobs:
                   make_target: (
                     if . == "CubeAPI" then "cube-api-test"
                     elif . == "CubeOps" then "cubeops-test"
+                    elif . == "CubeProxy" then "cube-proxy-test"
                     elif . == "CubeShim" then "shim-test"
                     elif . == "network-agent" then "network-agent-test"
                     else error("unsupported component: " + .)
                     end
-                  )
+                  ),
+                  requires_builder: (. != "CubeProxy")
                 })
             '
           )"
@@ -193,6 +204,7 @@ jobs:
 
       - name: Detect builder changes
         id: builder_changes
+        if: matrix.requires_builder
         shell: bash
         run: |
           set -euo pipefail
@@ -225,7 +237,7 @@ jobs:
 
       - name: Log in to GitHub Container Registry
         id: ghcr_login
-        if: steps.builder_changes.outputs.changed != 'true'
+        if: matrix.requires_builder && steps.builder_changes.outputs.changed != 'true'
         continue-on-error: true
         uses: docker/login-action@v3
         with:
@@ -234,12 +246,13 @@ jobs:
           password: ${{ secrets.GITHUB_TOKEN }}
 
       - name: Warn about GHCR login failure
-        if: steps.ghcr_login.outcome == 'failure'
+        if: matrix.requires_builder && steps.ghcr_login.outcome == 'failure'
         shell: bash
         run: echo "::warning::GHCR login failed - will attempt pull as anonymous or build locally"
 
       - name: Resolve builder image
         id: builder_image
+        if: matrix.requires_builder
         shell: bash
         run: |
           set -euo pipefail
@@ -274,11 +287,15 @@ jobs:
         shell: bash
         env:
           MAKE_TARGET: ${{ matrix.make_target }}
+          REQUIRES_BUILDER: ${{ matrix.requires_builder }}
+          BUILDER_IMAGE: ${{ steps.builder_image.outputs.image }}
         run: |
           set -euo pipefail
 
           echo "Component: ${{ matrix.component }}"
-          echo "Builder image: ${{ steps.builder_image.outputs.image }}"
-
-          make "${MAKE_TARGET}" \
-            BUILDER_IMAGE="${{ steps.builder_image.outputs.image }}"
+          if [[ "${REQUIRES_BUILDER}" == "true" ]]; then
+            echo "Builder image: ${BUILDER_IMAGE}"
+            make "${MAKE_TARGET}" BUILDER_IMAGE="${BUILDER_IMAGE}"
+          else
+            make "${MAKE_TARGET}"
+          fi
--- a/CubeProxy/conf/includes/resolver.inc
+++ b/CubeProxy/conf/includes/resolver.inc
@@ -1,0 +1,2 @@
+# Always regenerated at container startup by CubeProxy/start.sh.
+# Intentionally empty in the source tree.
--- a/CubeProxy/nginx.conf
+++ b/CubeProxy/nginx.conf
@@ -29,6 +29,8 @@ events {
 }
 
 http {
+    # Keep resolver configuration at http scope so every Lua cosocket path inherits it.
+    include /usr/local/openresty/nginx/conf/includes/resolver.inc;
     include mime.types;
     lua_package_path "/usr/local/openresty/nginx/lua/?.lua;;";
     default_type application/octet-stream;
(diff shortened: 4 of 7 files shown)


fslongjin:
Since CubeProxy is a container using the host network, I was wondering whether configuring the resolver inside this container to point to its own `/etc/resolv.conf` would solve the problem.

zyl1121:
> Since CubeProxy is a container using the host network, I was wondering whether configuring the resolver inside this container to point to its own `/etc/resolv.conf` would solve the problem.

I do not think CubeProxy can rely on `/etc/resolv.conf` directly here.

For OpenResty cosocket hostname resolution, nginx still needs an explicit `resolver` directive with concrete nameserver addresses. It cannot point `resolver` at `/etc/resolv.conf` as a file path.

References:
- nginx `resolver` directive syntax expects resolver addresses, not a resolv.conf path:
  https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver
- `lua-resty-redis` also documents that hostname connections need an nginx `resolver` directive:
  https://github.com/openresty/lua-resty-redis

So I think there are two viable ways to provide the resolver addresses:
1. the current approach: discover nameservers during one-click rendering and write them into nginx.conf
2. an alternative container-local approach: read the container's `/etc/resolv.conf` at startup, extract the nameserver IPs, and then render those IPs into nginx.conf

I kept the current one-click render-time approach because it is narrower and fails earlier/clearer when Redis is configured by hostname but no usable resolver can be found.

fslongjin:
> > Since CubeProxy is a container using the host network, I was wondering whether configuring the resolver inside this container to point to its own `/etc/resolv.conf` would solve the problem.
> 
> I do not think CubeProxy can rely on `/etc/resolv.conf` directly here.
> 
> For OpenResty cosocket hostname resolution, nginx still needs an explicit `resolver` directive with concrete nameserver addresses. It cannot point `resolver` at `/etc/resolv.conf` as a file path.
> 
> References:
> 
> * nginx `resolver` directive syntax expects resolver addresses, not a resolv.conf path:
>   https://nginx.org/en/docs/http/ngx_http_core_module.html#resolver
> * `lua-resty-redis` also documents that hostname connections need an nginx `resolver` directive:
>   https://github.com/openresty/lua-resty-redis
> 
> So I think there are two viable ways to provide the resolver addresses:
> 
> 1. the current approach: discover nameservers during one-click rendering and write them into nginx.conf
> 2. an alternative container-local approach: read the container's `/etc/resolv.conf` at startup, extract the nameserver IPs, and then render those IPs into nginx.conf
> 
> I kept the current one-click render-time approach because it is narrower and fails earlier/clearer when Redis is configured by hostname but no usable resolver can be found.

I consider it is better for the cube proxy to use the container's resolv.conf. because in production  deployment, cube proxy usually runs on k8s, which do not use the one-click deploy.sh

zyl1121:
> I consider it is better for the cube proxy to use the container's resolv.conf. because in production deployment, cube proxy usually runs on k8s, which do not use the one-click deploy.sh

That makes sense to me. I'll rework this PR toward the container-based approach.

zyl1121:
@fslongjin I've reworked the PR to use the container-side approach.

CubeProxy now ships `nginx.conf` as a template and renders the final config in `start.sh` before nginx starts. The resolver is derived from the container's own `/etc/resolv.conf`, with `CUBE_PROXY_NGINX_RESOLVER` available as an explicit override.

This removes the one-click-only host rendering path, so the same CubeProxy startup behavior applies to one-click, k8s-style, and manual Docker deployments. The fail-fast behavior is still kept when Redis is configured by hostname but no usable resolver can be found.

I also validated the container-side rendering path and updated the one-click tests for the new flow.

chenhengqi (changes requested):
Let's focus on the resolver-related change and keep the diff minimal. Please drop other changes, they can be in a separate PR.

zyl1121:
@chenhengqi I re-scoped the PR to focus only on the resolver-related change.

The updated version keeps the container-side `/etc/resolv.conf` based resolver discovery, but only writes `conf/includes/resolver.inc` at CubeProxy startup. It no longer renders the full nginx config in the container or reworks the broader one-click flow.

I also dropped the unrelated helper cleanup, docs, and shell hardening changes. Those can be handled separately if still needed.

chenhengqi (commented):
I am OK with the change. @fslongjin Please take a look.

fslongjin (changes requested):
Hey, I took a close look at this PR. Up front:

**You've got the right problem, and the fix is mostly on the right track — but this version is still brittle, and there are a few places where it can fail silently.** I wouldn't merge it as-is; better to tighten up the items below first.

## Main issues

### 1. You only looked at the business Redis path, and missed the registry Redis that's on by default

CubeProxy actually has two independent Redis connections:

1. **Business path**: `sandbox_backend.lua`, which uses `$redis_ip` from `global.conf`
2. **Heartbeat / registration**: `proxy_registry.lua`, which uses `CUBE_PROXY_REGISTRY_REDIS_HOST`, and in one-click it's **enabled by default**

`start.sh` currently ignores the second one entirely. I verified locally: when the main Redis is an IP and the registry host is a hostname, the startup script happily writes:

```text
# no nginx resolver required
```

The container comes up fine, but the heartbeat timer keeps hitting the original `no resolver defined` in the error log every few seconds. Worse, the failure is pretty quiet — CLM just never discovers this proxy.

Those two settings can already be configured separately in docker-compose, so this isn't a theoretical edge case.

### 2. When `global.conf` parsing fails, it pretends "no resolver needed"

The regex that reads `$redis_ip` is too strict — the whole line has to match exactly through the end. If someone writes:

```nginx
set $redis_ip "redis.example.com"; # primary redis
```

parsing fails, the script silently falls back to the default `127.0.0.1`, writes a no-op resolver include, and then **exits 0**.

Meanwhile nginx is still using the hostname, so the original error comes back. Startup looks perfectly healthy, which makes this harder to debug than before. That silent degradation feels worse than just failing fast at startup.

### 3. The config source used to decide "do we need a resolver?" is not the same one nginx actually uses to talk to Redis

`start.sh` prefers the `CUBE_PROXY_REDIS_IP` env var, but the business Lua path actually reads `$redis_ip` from `global.conf`. Even weirder: `CUBE_PROXY_REDIS_IP` isn't even in the `env` whitelist in `nginx.conf`.

I tried it: env set to an IP, `global.conf` set to a hostname → the script decides no resolver is needed → it still blows up at runtime.

In the one-click container setup this env usually isn't passed in, so the main path "looks fine" by accident. That's coincidence, not a design guarantee. Anyone who later changes one place and forgets the other will hit this.

---

## Design-wise, I'd go simpler

Right now the logic is: "only write a resolver if the main redis looks like a hostname; otherwise write nothing."

That minimal strategy is exactly what makes the issues above easy to hit. Honestly, configuring a `resolver` you don't immediately use is basically harmless. A more robust approach:

- If you can discover a usable IPv4 nameserver from `/etc/resolv.conf`, **always write** an http-level `resolver`
- Only fail fast when you clearly know a target is a hostname but can't find any nameserver

That way you don't have to enumerate every cosocket path in CubeProxy, and future hostname-based connections are less likely to get missed again. The K8s path on master already leans toward "always write a resolver."

At minimum, fold `CUBE_PROXY_REGISTRY_REDIS_HOST` into the "do we need DNS?" check. Better still: stop guessing, as above.

---

## Other things worth calling out

- **IPv6-only DNS environments**: startup is rejected outright. That doesn't really fix anything — it just turns a runtime error into "won't start." If short-term support isn't planned, at least call it out clearly in the PR / docs.
- **This branch is fairly behind master**: the Helm chart on master already has another resolver approach (write into `global.conf` + `getent` pre-resolve for the registry host). Same goal, different mechanism. When merging, watch out for stacking two half-overlapping stories on top of each other.
- **Tests**: happy path coverage is good. But the failure cases above — registry on a separate hostname, trailing comments, env vs `global.conf` mismatch — aren't covered. Those are exactly the footguns.

Security-wise I didn't see a big issue: what gets written into the nginx config is validated IPv4 nameserver addresses, and the Redis hostname itself isn't interpolated into the config, so injection risk looks low.

zyl1121:
@fslongjin Thanks for the detailed review. I reworked the resolver path based on your suggestions.

I was initially trying to keep the change as minimal as possible, so I added logic to decide when a resolver was needed based on the business Redis target, that was indeed overthinking it. Now the startup only fails when an enabled Redis target is a hostname and no valid resolver is available.

I also updated the target detection to cover both Redis connections. The request path now checks `$redis_ip` from `global.conf`, while the registry heartbeat path checks `CUBE_PROXY_REGISTRY_REDIS_HOST` when registration is enabled. `CUBE_PROXY_REDIS_IP` is no longer used for this decision since it is not the value consumed by the business Lua path.

The other cases you pointed out are covered as well, including trailing comments in `global.conf`, IPv4 and IPv6 resolver addresses with optional ports, and the resolver settings already passed by the Helm entrypoint. The regression tests have been moved into the CubeProxy component and expanded to cover these cases.

I also rebased the branch onto the latest `master` and verified it with the CubeProxy tests, a Linux image build, and `nginx -t` using both IPv4 and IPv6 resolver addresses.

CubeProxy/Makefile
```diff
@@ -42,6 +42,10 @@ build:
 		-t $(IMAGE_LOCAL):$(IMAGE_TAG)-$(ARCH) \
 		.
 
+.PHONY: test
```
chenhengqi, line 45: It would be better to include this test in CI.

zyl1121: > It would be better to include this test in CI.

Added. CubeProxy changes now trigger the `Test CubeProxy` job in the unit-test workflow. The test runs directly without the builder image, and the CI check has passed.

chenhengqi:
@zyl1121 There is merge conflict, please rebase.

zyl1121:
> @zyl1121 There is merge conflict, please rebase.

Rebased onto the latest master and resolved the conflict.

chenhengqi (approved):

fslongjin (approved):

PR #715, opened 2026-07-02
build: complete aarch64 support across image and bundle builds

lkml-likexu:
Several build and deploy paths still assumed an amd64/x86_64 host, which prevented CubeSandbox from being built and run natively on aarch64. These were the last residual hardcodings left after the earlier aarch64 series: the cube-api container image build, the cube-proxy sidecar build (both the in-tree Makefile flow and the one-click release bundle), and the containerd runtime platform declaration.

Derive the target architecture from the build context so each path builds natively on both amd64 and arm64, and declare both platforms for the runtime so the same config works on either host. The amd64 behavior is preserved unchanged, so existing x86_64 quickstart and install-package deployments are unaffected.


CubeProxy/Makefile
fslongjin, line 1: cube proxy don''t need to cross compile, so the `GOARCH` might be useless and we can remove it.

PR #730, opened 2026-07-03
fix(nodemeta): re-sync node health to localcache on periodic reload

dushulin:
## Summary

In a multi-replica CubeMaster deployment the periodic DB reload never refreshes
localcache's scheduler node cache, so a replica that learns a node only from the
DB keeps an empty healthy-node set for it. templatecenter's on-demand template
locality fallback then cannot match a ready replica to that node, and sandbox
creation intermittently fails with `template has no ready replica` (error 130400)
depending on which replica serves the request.

## Component

CubeMaster (`pkg/nodemeta`)

## Environment

- CubeSandbox version / commit: `master`
- Deployment mode: **cluster**, multi-replica CubeMaster (>= 2 replicas)
- KVM / kernel: not relevant — control-plane scheduler-cache consistency bug.

## Steps to Reproduce

1. Deploy CubeMaster with >= 2 replicas behind a LoadBalancer/VIP (each cubelet
   reports to a single replica).
2. Register/import a template so its ready replicas are learned by whichever
   CubeMaster replica the cubelet reports to.
3. Repeatedly create sandboxes from that template through the API (round-robins
   across CubeMaster replicas).

## Expected Behavior

Sandbox creation succeeds regardless of which replica serves the request.

## Actual Behavior

Only the replica that received the node's direct register/heartbeat has the node
in its localcache healthy-node set. On any other replica,
`EnsureTemplateLocalityReady` (templatecenter's DB fallback) cannot match a
DB-Ready template replica against localcache's healthy nodes, and creation fails
with `template has no ready replica` (error 130400). In a 2-replica setup,
creation succeeded roughly 50/50 depending on which replica the request hit.

## Root Cause

The periodic node-metadata reload (`loopReload` / `applyReloadResult`, added in
#542) rebuilds only the `nodemeta` in-memory node map. It does **not** refresh
`localcache`'s scheduler node cache, which is otherwise populated solely from the
register/heartbeat paths. A replica that learns a node only via the DB reload
therefore keeps an empty healthy-node set for it.

Template locality is not owned by `nodemeta` — it lives in the `imageCache`,
which `templatecenter` owns authoritatively (`warmReadyTemplateLocality` at
startup + `EnsureTemplateLocalityReady` on demand). That on-demand fallback
matches a DB-Ready template replica against localcache's **healthy nodes**; with
the node missing from that set it cannot match, surfacing as 130400.

## Fix

`applyReloadResult` now re-syncs every reloaded node's **node health** into
`localcache` via `syncLocalcacheNodeHealth` (`UpsertNode` only) after releasing
`s.mu`. This makes the node present/healthy on every replica, so
templatecenter's DB fallback can match a ready template replica and self-heal
locality on demand.

It deliberately does **not** call `SyncNodeTemplates`: writing the `imageCache`
from the reload path would race `templatecenter`, the authoritative owner, and
a stale full reconcile could erase freshly-registered replicas. Leaving template
locality to templatecenter keeps a single writer and a clear precedence.

The merge is extracted into `mergeReloadResult` with `defer s.mu.Unlock()` so a
panic in the merge loop cannot leak the lock. The sync reuses the existing
`s.ready` gate (set at the end of `Init`, after the initial reload and before
`localcache.Init`), so the first reload safely skips the sync while
`localcache.Init` loads all nodes from the DB itself.

## Testing

- `TestMergeReloadResultReturnsAllTouchedNodes`: reload returns a decoupled clone
  for every touched node (updated-existing + new-from-DB), carrying derived
  cordon state.
- `TestApplyReloadResultSyncsNodeHealthWhenReady`: with `s.ready = true`, the
  node-health sync is invoked for every touched node.
- `TestApplyReloadResultSkipsSyncBeforeReady`: before `s.ready`, the sync is
  skipped (Init-ordering safety).
- `go test -race ./CubeMaster/pkg/nodemeta/...` passes.
- Verified on a 2-replica staging deployment: a 20x sandbox-create loop that
  previously failed ~50% (130400) now succeeds 20/20.

## Related

- Follow-up to #542, which fixed the same cross-replica gap for node
  heartbeat/health but not for the scheduler node cache used by template-locality
  matching.

### DIFF
--- a/CubeMaster/pkg/nodemeta/service.go
+++ b/CubeMaster/pkg/nodemeta/service.go
@@ -941,18 +941,52 @@ func (s *service) reload() error {
 	return nil
 }
 
-// applyReloadResult merges a DB snapshot (next) into the live in-memory map.
-// Registration fields take the DB value (same eventual-consistency trade-off as
-// other labels). Status/heartbeat keep in-memory when fresher. After unlocking,
-// nodes whose cordon state changed are synced to localcache.
+// applyReloadResult merges a DB snapshot (next) into the live in-memory map and
+// then re-syncs node health into localcache for every node the reload touched.
+//
+// Registration fields and versions always take the DB value; status/heartbeat
+// fields keep the in-memory value when it is fresher than the DB snapshot.
+//
+// The re-sync pushes ONLY the scheduler node cache (health, capacity, cordon
+// state); see syncLocalcacheNodeHealth. Template locality is deliberately left
+// to templatecenter, the authoritative owner of the imageCache (startup
+// warmReadyTemplateLocality + on-demand EnsureTemplateLocalityReady DB fallback).
+//
+// Why node health must be re-synced here: a replica that only learned a node via
+// DB reload (it registered/heartbeated on another replica) otherwise kept an
+// empty healthy-node set for it. EnsureTemplateLocalityReady matches a DB-Ready
+// template replica against localcache's healthy nodes, so with the node absent
+// it could not match and sandbox creation failed with "template has no ready
+// replica" (130400). Pushing node health here lets that DB fallback match and
+// self-heal template locality on demand.
 func (s *service) applyReloadResult(next map[string]*NodeSnapshot) {
-	toSync := make([]*NodeSnapshot, 0)
+	syncSnaps := s.mergeReloadResult(next)
 
+	// s.ready is set true at the end of Init, after the initial reload and
+	// before localcache.Init. The very first reload therefore skips the sync
+	// (localcache caches do not exist yet); localcache.Init then loads all nodes
+	// from the DB itself. Periodic loopReload runs long after both packages are
+	// initialised.
+	if !s.ready {
+		return
+	}
+	// Sync outside s.mu (localcache has its own locks).
+	for _, snap := range syncSnaps {
+		syncNodeHealthFn(snap)
+	}
+}
+
+// mergeReloadResult merges the DB reload snapshot into the in-memory node map
+// under s.mu and returns a clone of every touched node for the caller to sync
+// into localcache outside the lock. The critical section uses
+// defer s.mu.Unlock() so a panic in the merge loop cannot leak the lock and
+// silently deadlock subsequent register/heartbeat handlers.
+func (s *service) mergeReloadResult(next map[string]*NodeSnapshot) []*NodeSnapshot {
 	s.mu.Lock()
+	defer s.mu.Unlock()
+	syncSnaps := make([]*NodeSnapshot, 0, len(next))
 	for nodeID, newSnap := range next {
 		if existing, ok := s.nodes[nodeID]; ok {
-			prevDisabled := snapshotSchedulingDisabled(existing)
-
 			existing.Labels = cloneStringMap(newSnap.Labels)
 			existing.labelsJSONCorrupt = newSnap.labelsJSONCorrupt
 			existing.Capacity = newSnap.Capacity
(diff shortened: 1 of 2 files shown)


fslongjin:
Thanks for your PR! I'll review and test this next Monday~

fslongjin (changes requested):
@dushulin Thanks for putting this together — the root-cause analysis is spot on, and the staging validation (20/20 creates) is really helpful context.

**Verdict: the problem is correctly identified, but I don’t think this PR is ready to merge yet.**

You nailed the root cause: in a multi-replica deployment, only the CubeMaster replica that receives a cubelet’s direct reports knows which templates are ready on which node. Other replicas can reload node metadata from the DB into `nodemeta`, but that path never updates `localcache.imageCache`, so sandbox creation intermittently fails with `template has no ready replica`.

The concern with the current fix is that it adds another `syncLocalcache` call on the reload path without first clarifying **who owns `localcache` and which writer wins**. A 20/20 staging run shows the happy path works; it doesn’t prove correctness under concurrency or over longer runs.

---

## Core issue: treats the symptom, doesn’t block stale writes

`applyReloadResult` merges under `s.mu`, **clones** snapshots, then calls `syncLocalcache` **after** releasing the lock.

There’s no freshness check in between. A typical race:

> reload clones template set A → unlocks → a fresher heartbeat writes set B and syncs to localcache → reload runs `syncLocalcache(A)` from the stale clone and overwrites B.

`SyncNodeTemplates` does a full reconcile and **removes** templates not in the current set — a stale reload isn’t just “filling gaps”; it can erase state that a fresher heartbeat just wrote. Error `130400` can come back.

`syncLocalcache` also calls `UpsertNode`, so scheduler node cache fields like `Healthy` and `MetaDataUpdateAt` can be rolled back by an old snapshot too. This PR focuses on template locality but doesn’t protect the full sync path.

---

## Deeper issue: `localcache` already has multiple writers

`imageCache` isn’t only written by nodemeta:

| Source | API | Meaning |
|--------|-----|---------|
| nodemeta heartbeat / reload | `SyncNodeTemplates` | Node-reported local template list |
| templatecenter | `RegisterTemplateReplica` | Ready replicas from the template replica table |
| templatecenter | `InvalidateImageState` | Explicit invalidation — must not be scheduled |

These aren’t the same semantics. This PR runs periodic per-node full reconcile on reload but **doesn’t define** precedence vs templatecenter register/invalidate — it may drop a freshly registered replica, or re-add one after invalidate from a stale snapshot.

Until source-of-truth priority is clear, adding more sync entry points makes `localcache` harder to reason about, not more reliable.

---

## Other concerns

**`Ready()` is too weak.** It only checks `l.cache != nil`, but `syncLocalcache` also touches `imageCache`, `templateNodeCache`, and `sortedNodesByClusters`. Init assigns fields one by one; we shouldn’t rely on the default 30s sync interval to hope the startup window is “big enough.”

**`SyncNodeTemplates` isn’t atomic.** Read previous → diff → update forward/reverse indexes — no single lock across the reconcile. Interleaved with `RegisterTemplateReplica` / `InvalidateImageState`, outcome depends on timing. New tests only show “template visible after reload,” not stale replay, index consistency, or interleaving with templatecenter.

---

## Suggested path forward

Rather than stacking more `syncLocalcache` calls in this PR, I’d suggest two steps:

1. **Define the rules first**: who overrides whom for `imageCache`, who may only add vs delete; how reload/heartbeat writes avoid stale snapshots overwriting fresh state (both `UpsertNode` and template sync).
2. **Then implement and test**: explicit `Ready()` semantics; race regressions (stale reload after fresh heartbeat, nodemeta vs templatecenter interleaving).

---

## Before merge, I’d like to see

- [ ] Stale reload snapshots cannot overwrite fresher heartbeat state (templates **and** scheduler node cache)
- [ ] Clear write precedence between `nodemeta` and `templatecenter` for `imageCache`
- [ ] `localcache.Ready()` means fully initialized, not “one field is non-nil”
- [ ] Regression tests for the races above, not only the happy path

Again, thanks for the investigation and the repro — the direction is right. I’d recommend tightening the `localcache` consistency model first, then coming back with a revised implementation.

dushulin:
Thank you for your reply. I will make some more modifications according to this suggestion. cc @fslongjin

kinwin-ustc:
@dushulin Can you rebase your code based on the latest master?

fslongjin (changes requested):
Hello, can you add some unitests to cover your changes? I think that this PR's description is outdated, can you modify it?

fslongjin (approved):

PR #741, opened 2026-07-03
fix(template): probe envd readiness before persisting envd capability

zyl1121:
## Motivation

In #566, create-time env injection started relying on `cube.master.components.envd.version` as the envd capability signal, using the mechanism introduced in #650. During review, we agreed that the main envd readiness guarantee should move to template construction time instead of rediscovering the same readiness race on every sandbox creation.

Before this change, the template envd annotation could be written after only collecting `envd --version`. That proved the binary existed, but not that the envd service was reachable and ready on the sandbox runtime endpoint.

As a result, downstream sandbox create flows could treat a template as envd-capable even when envd was absent, stopped, or still starting.

## Summary

* Cubelet now confirms envd readiness via `http://<sandboxIP>:49983/health` before reporting an envd version during template snapshot flows. This uses a dedicated bounded HTTP probe instead of the generic `telnet.Telnet` helper so the readiness path can honor parent context cancellation and reject HTTP redirects explicitly.
* The same readiness check applies to both `AppSnapshot` and `CommitSandbox`, keeping image-based template creation and runtime commit paths consistent.
* CubeMaster persists `cube.master.components.envd.version` only when all READY replicas report the same valid envd version; otherwise, it suppresses the annotation instead of publishing an optimistic capability signal.
* If redo later re-runs the template on the same `templateID` and the stricter convergence check fails, CubeMaster now clears any previously persisted envd annotation instead of leaving stale capability data behind.

## Validation

* `go test -vet=off ./pkg/templatecenter -run 'Envd|Template' -count=1`
* `go test -vet=off ./services/cubebox -run 'Envd|Probe' -count=1`

Manual validation on Linux:

* success path: without the temporary REJECT rule, template info showed:

  ```text
  "cube.master.components.envd.version": "0.5.11"
  ```

  After enabling INFO logs, Cubelet also logged:

  ```text
  collect envd version: readiness confirmed, version=0.5.11
  ```

* failure path: with a temporary REJECT rule blocking Cubelet -> `sandboxIP:49983`, template build still succeeded, but the envd annotation was omitted. Cubelet logged:

  ```text
  collect envd version: envd readiness probe failed on <sandboxIP>:49983 ...
  ```

* redo path: running `cubemastercli cubebox template redo` against the same `templateID` also updated the annotation as expected. A template that previously carried `cube.master.components.envd.version` lost it after a redo under the REJECT rule, confirming that redo re-evaluates and reconciles the envd capability signal instead of leaving the previous value sticky.

## Scope

This change is limited to the template/snapshot paths that produce and persist the envd capability annotation. It does not change sandbox create-time env var propagation semantics from #566.

## Follow-up

A later follow-up can make envd capability suppression visible to users, so a missing annotation can distinguish between "envd is absent" and "envd was detected but not ready during template-time probing". That would let users understand why the template is not marked envd-capable and decide whether to run `redo` for transient startup jitter.

### DIFF
--- a/CubeMaster/pkg/templatecenter/store.go
+++ b/CubeMaster/pkg/templatecenter/store.go
@@ -482,6 +482,7 @@ func createTemplateReplicasOnNodes(ctx context.Context, templateID string, req *
 			replicas = append(replicas, replica)
 			envdVersions = append(envdVersions, nodeEnvdVersion{
 				NodeID:  target.ID(),
+				Ready:   replica.Status == ReplicaStatusReady,
 				Version: envdVersion,
 			})
 			lock.Unlock()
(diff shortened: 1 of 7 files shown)


kinwin-ustc:
Perhaps we just need to use the `probe-path` parameter of `cubemastercli`? For example, `--probe 49983 --probe-path /health`

zyl1121:
> Perhaps we just need to use the `probe-path` parameter of `cubemastercli`? For example, `--probe 49983 --probe-path /health`

I considered that too, but I think it would change the CLI contract too much.

Today, `--probe` / `--probe-path` represents the user-facing readiness probe for the workload being built into the template. If we reuse it for envd (`49983 /health`), envd readiness becomes a hard template-build gate, and users also lose the ability to specify their own service's readiness probe.

This PR intentionally keeps those two concerns separate: the existing probe continues to represent workload readiness, while the envd probe is an internal capability check whose only effect is whether CubeMaster publishes the envd capability annotation.

If we want to express envd readiness through probe configuration in the future, I think we should first add support for multiple probes, or introduce a dedicated internal probe slot, rather than overloading the existing user-facing one.

fslongjin:
I believe the changes in this PR are unnecessary. Whether envd is ready or not is not a prerequisite for the sandbox to be able to run. As far as I know, some users simply use CubeSandbox as a sandbox for running general workloads, and envd is not necessarily required to be installed inside.

For cases where envd is not running correctly, I am more inclined to view this as an implementation error on the part of the application developer, rather than something Cube should guarantee. If a developer needs envd to be ready when creating a template, the application developer should ensure themselves that when the probe path returns success, envd is already ready (this should be implemented by logic inside the container, rather than pushing this complexity onto Cube). Otherwise, it may cause issues when Cube runs workloads that do not contain envd.

zyl1121:
> I believe the changes in this PR are unnecessary. Whether envd is ready or not is not a prerequisite for the sandbox to be able to run. As far as I know, some users simply use CubeSandbox as a sandbox for running general workloads, and envd is not necessarily required to be installed inside.
> 
> For cases where envd is not running correctly, I am more inclined to view this as an implementation error on the part of the application developer, rather than something Cube should guarantee. If a developer needs envd to be ready when creating a template, the application developer should ensure themselves that when the probe path returns success, envd is already ready (this should be implemented by logic inside the container, rather than pushing this complexity onto Cube). Otherwise, it may cause issues when Cube runs workloads that do not contain envd.

@fslongjin I think there may be a misunderstanding about the intent of this PR.

This PR does not make envd readiness a prerequisite for template creation or sandbox execution. Templates still build successfully when envd is absent or not ready. The only behavior change is whether we persist `cube.master.components.envd.version`.

Previously, that annotation was written after `envd --version`, which only proved that the binary existed in the image. It did not prove that envd was started by the workload or reachable at build time. So the goal is to make this annotation a truthful envd capability signal for later envd-specific handling.

That is also why I avoided reusing the normal user-facing probe path: workload readiness and envd capability detection are separate semantics. For workloads without envd, template creation remains unchanged and the envd annotation is simply omitted.

kinwin-ustc:
This introduces an implicit mandatory assumption: for envd templates, the silent mandatory creation of templates only occurs after envd is ready (serviceable). Whether this should be an underlying mandatory behavior or left to the user's discretion is something I think needs discussion.

fslongjin:
See: https://github.com/e2b-dev/infra/blob/main/packages/orchestrator/pkg/template/build/core/envd/envd.go#L10

Cube's method for detecting the envd version is the same as e2b's.

zyl1121:
@fslongjin @kinwin-ustc Thanks for the clarification.

I understand now that `cube.master.components.envd.version` is expected to remain template-time envd version metadata, rather than a Cube-managed readiness or capability guarantee.

This PR was trying to tighten the semantics of that annotation, but that does not align with the current direction. So I’ll close it.

A better follow up may be documentation that makes this contract explicit: the annotation records the detected envd version, while actual envd readiness remains the responsibility of the image / entrypoint and the user-configured probe path.

Thanks again for the review and discussion.

PR #743, opened 2026-07-03
feat: add Python SDK set_timeout() with NEVER_TIMEOUT (-1) support

emailcannotbeblank:
## Summary

This PR adds `set_timeout()` to the CubeSandbox Python SDK, making it
API-compatible with the E2B SDK.  It also extends the endpoint to
accept `NEVER_TIMEOUT (-1)` as a valid sentinel, completing the
three-value timeout semantics introduced in #810.

### Python SDK
- Add `Sandbox.set_timeout(timeout)` → `POST /sandboxes/{id}/timeout`
- Export `NEVER_TIMEOUT = -1` sentinel for never-timeout requests
- Document the new method in `sdk/python/README.md`

### CubeAPI (Rust)
- Replace `#[validate(range(min = 0))]` with a custom validator that
  accepts `>= 0` or `== -1`.  Reject all other negative values.

### CubeMaster (Go)
- Relax `SetTimeout` validation from `timeout < 0` to `timeout < -1`
  so that `-1` (NeverTimeout) is forwarded to the lifecycle provider.

### Tests
- Rust: `set_timeout_request_accepts_never_timeout`
- Go: `TestSetTimeoutValidationAllowsZero`,
  `TestSetTimeoutValidationRejectsNegative`,
  `TestSetTimeoutValidationAllowsNeverTimeout`
- Python: happy path, explicit zero, NEVER_TIMEOUT, 404

## Timeout semantics (after #810)

| `timeout` value | Meaning |
|-----------------|---------|
| `> 0` | Normal TTL in seconds |
| `0` | Immediate expiry |
| `-1` (`NEVER_TIMEOUT`) | Never expire |
| `<-1` | Rejected (invalid) |

## Verification

```bash
# Rust
cargo check --manifest-path CubeAPI/Cargo.toml
cargo test -p cubeapi -- models::mod::tests

# Go
go test ./pkg/service/sandbox -run 'TestSetTimeoutValidation' -count=1

# Python
uv run --project sdk/python --extra dev pytest sdk/python/tests/test_sandbox.py -q
```

## Dependencies

- [x] #810 (merged) — three-value timeout semantics + server-side default

### DIFF
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -42,9 +42,9 @@ func SetTimeout(ctx context.Context, req *types.SetTimeoutRequest) (rsp *types.S
 		rsp.Ret.RetMsg = "should provide sandboxID"
 		return
 	}
-	if req.Timeout < 0 {
+	if req.Timeout < -1 {
 		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
-		rsp.Ret.RetMsg = "timeout must be non-negative (seconds)"
+		rsp.Ret.RetMsg = "timeout must be >= -1 (use -1 for never timeout)"
 		return
 	}
 
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -42,9 +42,9 @@ func SetTimeout(ctx context.Context, req *types.SetTimeoutRequest) (rsp *types.S
 		rsp.Ret.RetMsg = "should provide sandboxID"
 		return
 	}
-	if req.Timeout <= 0 {
+	if req.Timeout < 0 {
 		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
-		rsp.Ret.RetMsg = "timeout must be positive (seconds)"
+		rsp.Ret.RetMsg = "timeout must be non-negative (seconds)"
 		return
 	}
 
--- a/sdk/python/cubesandbox/sandbox.py
+++ b/sdk/python/cubesandbox/sandbox.py
@@ -459,7 +459,8 @@ class Sandbox:
 
         Args:
             timeout: New idle timeout in seconds. ``0`` requests immediate
-                timeout; positive values set a normal TTL.
+                timeout; positive values set a normal TTL; ``NEVER_TIMEOUT``
+                (-1) disables idle timeout entirely.
 
         Raises:
             SandboxNotFoundError: If the sandbox does not exist (HTTP 404).
--- a/sdk/python/cubesandbox/sandbox.py
+++ b/sdk/python/cubesandbox/sandbox.py
@@ -454,6 +454,23 @@ class Sandbox:
         )
         _check_response(resp)
 
+    def set_timeout(self, timeout: int) -> None:
+        """POST /sandboxes/:sandboxID/timeout - Set sandbox idle timeout.
+
+        Args:
+            timeout: New idle timeout in seconds. ``0`` requests immediate
+                timeout; positive values set a normal TTL.
+
+        Raises:
+            SandboxNotFoundError: If the sandbox does not exist (HTTP 404).
+            ApiError: If the timeout is invalid or on unexpected backend error.
+        """
+        resp = self._session.post(
+            f"{self._config.api_url}/sandboxes/{self.sandbox_id}/timeout",
+            json={"timeout": timeout},
+        )
+        _check_response(resp)
+
     def kill(self) -> None:
         """DELETE /sandboxes/:sandboxID - Destroy a sandbox.
 
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -479,8 +479,25 @@ fn default_log_limit() -> i32 {
 /// Request body for POST /sandboxes/{id}/timeout
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct SetTimeoutRequest {
-    #[validate(range(min = 0))]
+    #[validate(custom(function = "validate_timeout_value"))]
     pub timeout: i32,
+}
+
+/// Validates the timeout value for SetTimeoutRequest.
+///
+/// Accepts:
+/// - `>= 0`: normal TTL in seconds (0 = immediate expire)
+/// - `-1`:  never-timeout sentinel (see `NEVER_TIMEOUT` in SDK)
+///
+/// Rejects all other negative values.
+fn validate_timeout_value(timeout: i32) -> Result<(), validator::ValidationError> {
+    if timeout >= 0 || timeout == -1 {
+        Ok(())
+    } else {
+        Err(validator::ValidationError::new(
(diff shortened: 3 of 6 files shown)


CubeMaster/pkg/lifecycle/init.go
```diff
@@ -136,6 +136,13 @@ func onAfterCreate(ctx context.Context, sandboxID, hostID, hostIP string, req *s
 	return nil
 }
 
+func timeoutFields(nowMs int64, timeoutSeconds int) (storedTimeoutSeconds int, endAtMs int64) {
```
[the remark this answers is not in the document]
chengjoey: @emailcannotbeblank Please add the test here, or ensure that this logic is normal; otherwise, the user may be killed just after creating the sandbox.
https://github.com/TencentCloud/CubeSandbox/blob/23786819d399678388d71706f91375781b23c6ca/CubeMaster/pkg/service/sandbox/util.go#L130-L132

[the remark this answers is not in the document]
emailcannotbeblank: Thanks for pointing this out.

The original `timeout=0` behavior is still preserved for sandbox creation: `0` means omitted/default timeout, and the create path normalizes `req.Timeout <= 0` to `CommonTimeoutInsec` in `ConstructCubeletReq()` before the post-create lifecycle hook runs.

Because `0` already has this default-value meaning, this PR uses a separate internal sentinel, `ImmediateExpireTimeoutSeconds = -1`, to represent the new `set_timeout(0)` semantic. CubeMaster writes this sentinel only when handling an explicit timeout update request with `timeout=0`; CubeProxy sidecar then checks this sentinel and treats it as immediate expiry. A plain `TimeoutSeconds=0` remains legacy/missing metadata and still falls back to the default idle timeout.

This keeps the change small and backward-compatible: existing create/default timeout behavior is unchanged, and components that validate `timeout > 0` will naturally reject the internal `-1` sentinel unless they explicitly opt into the new immediate-expiry semantics.

I added tests to lock down this invariant:

- `TestConstructCubeletReqNormalizesCreateTimeoutZeroToDefault`
  Verifies that create-time `Timeout: 0` is normalized to `CommonTimeoutInsec`.

- `TestOnAfterCreate_EffectiveCreateTimeoutDoesNotStoreImmediateExpire`
  Verifies that the post-create lifecycle metadata does not persist `ImmediateExpireTimeoutSeconds = -1` for create/default timeouts.

- `TestStoreTimeoutProvider_RefreshTimeoutZeroStoresImmediateExpire`
  Verifies that only the explicit timeout update path stores `ImmediateExpireTimeoutSeconds = -1` for `set_timeout(0)`.

So a sandbox created with omitted/default timeout will not be killed immediately after creation; only an explicit `set_timeout(0)` request can mark it as immediately expired.

chenhengqi (commented):
We are consider adding support for sandboxes to run indefinitely.

emailcannotbeblank:
> We are consider adding support for sandboxes to run indefinitely.

set timeout to MAX value of int？Or using a magic number like -2 for indefinite？

chenhengqi:
cc @fslongjin

fslongjin:
https://github.com/TencentCloud/CubeSandbox/pull/810

emailcannotbeblank:
@fslongjin @chenhengqi 
PR #810 changes the lifecycle timeout mechanism and adds support for a "never timeout" setting.

From what I can see, creating a sandbox already supports "never timeout" quite well. CubeAPI keeps the difference between "timeout not provided" and "timeout explicitly provided", and CubeMaster later normalizes negative timeout values to the `NeverTimeout` value (`-1`).

What I am not clear about is whether the author also wants "never timeout" to be configurable in places other than sandbox creation.

For example, when updating a sandbox, the CubeAPI sender side already adds a `timeout` field to `SandboxUpdateRequest`:

https://github.com/fslongjin/CubeSandbox/blob/feat/sandbox-timeout-refactor/CubeAPI/src/cubemaster/mod.rs#L1271

```rust
#[derive(Debug, Serialize)]
pub struct SandboxUpdateRequest {
    #[serde(rename = "requestID")]
    pub request_id: String,
    #[serde(rename = "sandbox_id")]
    pub sandbox_id: String,
    #[serde(rename = "instance_type")]
    pub instance_type: String,
    /// "pause" | "resume"
    #[serde(rename = "action")]
    pub action: String,
    /// TTL in seconds (for resume; 0 = keep original). Optional for pause.
    pub timeout: Option<i32>,
}
```

But on the CubeMaster receiver side, the corresponding `UpdateRequest` does not have a `timeout` field:

https://github.com/fslongjin/CubeSandbox/blob/feat/sandbox-timeout-refactor/CubeMaster/pkg/service/sandbox/types/types.go#L684
```go
type UpdateRequest struct {
	RequestID    string `json:"requestID" p:"requestID"  v:"required"`
	SandboxID    string `json:"sandbox_id" p:"sandbox_id"  v:"required"`
	InstanceType string `json:"instance_type" p:"instance_type"  v:"required"`
	Action       string `json:"action" p:"action"  v:"required"`
}
```


So if CubeAPI sends `timeout` during `resume` / `connect`, CubeMaster will lose this field when decoding the request.

Also, at the CubeAPI entry point for setting timeout, the request model requires `timeout >= 0`:

https://github.com/fslongjin/CubeSandbox/blob/feat/sandbox-timeout-refactor/CubeAPI/src/models/mod.rs#L480

```rust
pub struct SetTimeoutRequest {
    #[validate(range(min = 0))]
    pub timeout: i32,
}
```

This means `set_timeout` cannot use `-1` to set a sandbox to "never timeout".

Since the "never timeout" behavior appears to be a CubeSandbox-specific extension beyond e2b's `set_timeout` API, I would like to better understand the intended scope of this feature:

- Is "never timeout" intended to be configured only at sandbox creation time? Should the same behavior also apply when resuming or connecting to an existing sandbox?
- For `set_timeout`, should `-1` be accepted as the "never timeout" sentinel, or is it preferable to keep this API aligned with e2b by accepting only non-negative values?

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -42,9 +42,9 @@ func SetTimeout(ctx context.Context, req *types.SetTimeoutRequest) (rsp *types.S
 		rsp.Ret.RetMsg = "should provide sandboxID"
 		return
 	}
-	if req.Timeout <= 0 {
+	if req.Timeout < 0 {
```
chenhengqi: cc @fslongjin

sdk/python/cubesandbox/sandbox.py
```diff
@@ -454,6 +454,23 @@ def resume(self, timeout: int | None = None) -> None:
         )
         _check_response(resp)
 
+    def set_timeout(self, timeout: int) -> None:
+        """POST /sandboxes/:sandboxID/timeout - Set sandbox idle timeout.
+
+        Args:
+            timeout: New idle timeout in seconds. ``0`` requests immediate
```
chenhengqi, line 461: We have a `NEVER_TIMEOUT` constant.

emailcannotbeblank: Yes, just to align with e2b's language for now, we can set the timeout to 0. The next commit will modify cubeapi to support NEVER_TIMEOUT as an input.

fslongjin (approved):

PR #747, opened 2026-07-04
Allow platform-suffixed versions in version matrix checks

HuChundong:
## Summary
- keep reported component versions unchanged in the Versions matrix
- treat platform-suffixed versions such as `v0.5.0-arm64` as declared when the release manifest declares `v0.5.0`
- apply the same comparison in the WebUI so component badges, filters, and KPI counts match the backend matrix result
- add regression coverage that preserves the displayed reported version while marking it declared

## Root cause
Architecture-specific one-click bundles can report installed component versions with platform suffixes, for example `v0.5.0-arm64`. The release manifest declares the release line as `v0.5.0`, so exact string comparisons marked those platform-specific reports as undeclared even though they belong to the declared release.

## Behavior
The matrix still displays the actual reported value, such as `v0.5.0-arm64`. Only the declared/undeclared comparison normalizes known platform suffixes (`amd64`, `arm64`, `x86_64`, `aarch64`). Pre-release or custom suffixes such as `-rc1` and `-dev` remain undeclared unless explicitly declared.

## Validation
- `go test ./pkg/nodemeta` from `CubeMaster`
- `npm run build` from `web`
- locally deployed CubeMaster and WebUI on the CentOS VM; verified `cube-egress` reports `v0.5.0-arm64`, remains displayed as `v0.5.0-arm64`, and is counted as declared / single-version in the dashboard

Note: this PR intentionally does not change one-click packaging or CubeEgress image defaults.

### DIFF
--- a/CubeMaster/pkg/nodemeta/versionmatrix_test.go
+++ b/CubeMaster/pkg/nodemeta/versionmatrix_test.go
@@ -5,6 +5,7 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/db/models"
 )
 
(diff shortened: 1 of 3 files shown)


CubeMaster/pkg/nodemeta/versionmatrix_test.go
```diff
@@ -87,22 +87,47 @@ func TestVersionIsDeclaredForKernelVariants(t *testing.T) {
...
+func TestVersionIsDeclaredIgnoresPlatformSuffix(t *testing.T) {
+	primary := map[string]string{"cube-egress": "v0.5.0"}
+	sets := map[string]map[string]struct{}{
+		"cube-egress": {
+			"v0.5.0": {},
+		},
+	}
+
+	for _, actual := range []string{"v0.5.0", "v0.5.0-arm64", "v0.5.0-amd64", "v0.5.0-aarch64", "v0.5.0-x86_64"} {
+		if !versionIsDeclared("cube-egress", actual, primary, sets) {
```
chenhengqi: Use `testify/assert`.

fslongjin (approved):
LGTM

PR #607, opened 2026-06-22
feat(templates): E2B-style display names with CubeMaster name cache

HuChundong:
## Summary

- Add optional human-readable template `name` (CubeMaster `display_name`) with **cluster-wide** case-insensitive uniqueness (`20260623051112_template_display_name_index.sql`).
- **Web**: list/show display names; PATCH rename on template detail; optional `name` on create.
- **API**: `POST /sandboxes` and template paths accept **display name or `tpl-*` id**; CubeAPI resolves name→id once at entry, then uses existing tpl-id paths.
- **CubeMaster**: `GET /cube/template/lookup`, `POST /cube/template/display-name`; sync name reservation at submit (`ReserveTemplateDisplayNameForCreate` → 409); release name after failed builds.
- **Name cache (CubeMaster)**: `display_name` → `template_id` cached in `templatecenter` (300s TTL, 4096 entries, singleflight); invalidated on rename/delete/create. CubeAPI is a thin client with no local cache.

## Architecture

| Layer | Approach |
|-------|----------|
| CubeAPI | Thin resolver; delegates all name lookups to CubeMaster |
| CubeMaster | Owns `display_name`, lookup cache, and invalidation |
| Uniqueness | Global per cluster (no tenant/team namespace today) |

Compared to [E2B infra](https://github.com/e2b-dev/infra): E2B uses team-scoped aliases reserved in `RegisterBuild`; we use global `display_name_key` + sync submit check. We **release** names on build failure (E2B keeps aliases until template delete).

## Test plan

- [x] `go test ./pkg/templatecenter/...`
- [x] `go test ./pkg/service/httpservice/cube/...`
- [x] `cargo test -p cube-api`
- [x] `go test ./pkg/base/dao/migrate/ -run TestMigrationFilenames`
- [ ] Apply migration `20260623051112_template_display_name_index` on staging (run audit queries in migration header first)
- [ ] Create template with `name` → create sandbox by name when READY
- [ ] Duplicate name at submit → 409; failed build → name reusable

## Migration note

This migration may clear legacy `tpl-`/`snap-` prefixed names and dedupe display names before adding the unique index. Run preflight audit queries on staging before production.

Assisted-by: Cursor:Composer

PR #749, opened 2026-07-04
feat(cubemaster): add template alias for stable naming across rebuilds

### LINKED ISSUE #584: Feature Request: stable template alias / logical name (decouple from auto-generated tpl-* snapshot IDs)
## Problem

Since v0.4.0 (#474, #476), `cubemastercli tpl create-from-image` no longer accepts `--template-id`, and the server always generates a new `tpl-<random>` ID for each template build.

This creates operational friction in real deployments:

1. **Logical template vs snapshot ID are conflated**
   - Example: we maintain one business template such as `opencode-data` (OpenCode + data stack image).
   - When the Docker image is updated and we rebuild the template, we get a **new** `tpl-xxx` ID, even though it is the same logical template from an operator/application perspective.

2. **Every image refresh breaks client configuration**
   - Applications, SDK env vars (`CUBE_TEMPLATE_ID`), CI/CD, and documentation must all be updated after each rebuild.
   - There is no stable identifier to reference across upgrades.

3. **No alias / display name in API or CLI**
   - `tpl list` and `GET /templates` expose `templateID`, `imageInfo`, `status`, etc., but no human-readable alias.
   - Operators must map friendly names externally (e.g. local `.env` files), which duplicates what the control plane should ideally provide.

4. **In-place update is not possible anyway**
   - Rootfs/spec immutability already forces creating a new template for image/spec changes (understandable).
   - Combined with auto-generated IDs, every upgrade is both a **new snapshot** and a **new opaque identifier**.

## Real-world workflow pain (self-hosted / offline)

Our current upgrade path on v0.4.0:

```bash
# Build/register updated image
cubemastercli tpl create-from-image --image cube-sandbox-opencode:latest ...
# => template_id: tpl-53337f004ee84d11966b7362  (new every time)

# Update all clients
export CUBE_TEMPLATE_ID=tpl-53337f004ee84d11966b7362

# Delete old template
cubemastercli tpl delete --template-id tpl-<old>
```

We cannot keep using names like `tpl-opencode-data` that were stable in v0.3.1.

`tpl redo` only redistributes/fixes replicas for an existing template ID; it does not rebuild from a new OCI image.

## Proposal

Please consider one or more of the following:

### Option A: Template alias (recommended)
- Allow setting a stable **alias** at create time, e.g. `--alias opencode-data` or `alias` field in create JSON.
- Alias rules: unique cluster-wide, `tpl-` prefix optional, validated charset.
- Sandbox creation accepts either `templateID` **or** `templateAlias` (alias resolves to the current default/active revision).

### Option B: Template revisions under one logical ID
- Keep a stable logical ID (operator-chosen or auto `tpl-opencode-data`).
- Each `create-from-image` creates a new **revision**; mark one revision as `active/default`.
- Old sandboxes keep running on their revision; new sandboxes use the active revision.

### Option C: Restore controlled custom IDs (narrower)
- Re-introduce optional `--template-id` for self-hosted deployments only.
- Still enforce immutability: changing rootfs/image requires new ID or explicit replace workflow.

## Expected benefit

- Stable application config across image/template rebuilds
- Clearer ops model: alias/name for humans, snapshot ID for internals
- Better alignment with E2B-style template aliases used by agent platforms

## Environment

- CubeSandbox: v0.4.0 (upgraded from v0.3.1)
- Deployment: self-hosted single-node, bare metal + KVM
- Use case: OpenCode sandbox templates rebuilt from custom Docker images

## Related

- v0.4.0 release: template ID auto-generation (#474), docs removed `--template-id` (#476)
- Template replica compatibility (#510) helps detect version skew, but does not solve stable naming for clients

Happy to help test a design or contribute docs/examples if this direction makes sense.

sparkzky:
## Summary

Resolves #584.

Since v0.4.0 (#474, #476), `cubemastercli tpl create-from-image` no longer accepts `--template-id`, and the server always generates a new `tpl-<random>` ID for each template build. This creates operational friction: every image refresh breaks client configuration (SDK env vars, CI/CD, documentation) because there is no stable identifier to reference across upgrades.

This PR implements **Option A (Template Alias)** from the issue proposal:

- Allow setting a stable **alias** at create time via `--alias` flag (CLI) or `name` field (E2B-compatible CubeAPI)
- Sandbox creation accepts either `templateID` **or** `templateAlias` (alias resolves to the current template)
- Alias moves to the last template that reaches READY — sequential rebuilds always hand off the alias; concurrent builds are serialized by a DB unique constraint (last commit wins)
- Alias validation: `^[a-z0-9][a-z0-9-]{0,63}$`, `tpl-`/`snap-` prefixes rejected

## Architecture

```
┌──────────────┐     POST /templates {name}      ┌──────────┐    CreateTemplateFromImageReq {alias}
│  SDK / CLI   │ ──────────────────────────────▶ │ CubeAPI  │ ────────────────────────────────────▶ ┌────────────┐
│              │   GET  /templates/:alias         │ (Rust)   │   GET /cube/template?template_id=:alias│ CubeMaster │
│              │   DEL  /templates/:alias         │          │   DEL /cube/template  (template_id)    │ (Go)       │
└──────────────┘                                  └──────────┘                                       └─────┬──────┘
                                                                                            resolveTemplateIdentifier
                                                                                                      │
                                                                            ┌─────────────────────────┐
                                                                            │ display_name = alias    │
                                                                            │ STORED generated column │
                                                                            │ alias_key (UNIQUE)      │
                                                                            └─────────────────────────┘
```

### CubeMaster (Go)

| File | Change |
|---|---|
| `types.go` | Add `Alias` field to `CreateTemplateFromImageReq` |
| `store.go` | `GetTemplateByAlias()` (scoped `kind=template`), `ResolveTemplateIdentifier()`, consolidated `createDefinition`/`ensureTemplateDefinition` into thin `//go:noinline` delegators |
| `request_validation.go` | `validateTemplateAlias()` regex validation |
| `snapshot_ops.go` | Alias reassignment guarded to templates only; transactional release+claim |
| `cubeboxutil.go` | Alias resolution in sandbox creation path |
| `template.go` (CLI) | `--alias` flag, `ALIAS` column in `tpl list` |
| `template.go` (server) | Alias resolution in GET **and DELETE** handlers; `DisplayName` in list response |
| `image_job_runner.go` / `redo.go` | Thread `req.Alias` as `DisplayName` |

### CubeAPI (Rust) — E2B compatibility

| File | Change |
|---|---|
| `services/templates.rs` | `name` → alias forwarding (`alias_from_name`), `get_template_by_alias()`, `aliases`/`names` arrays in responses |
| `handlers/templates.rs` | `GET /templates/aliases/:alias` dedicated endpoint; `utoipa` annotations |
| `models/mod.rs` | `CreateTemplateRequest.name`, `TemplateAliasLookupResponse`, `public`/`aliases`/`names` fields |
| `cubemaster/mod.rs` | `is_params_error()` → HTTP 400 mapping |

### Migration (MySQL + PostgreSQL)

| File | Change |
|---|---|
| `mysql/20260704120000_template_alias_unique.sql` | STORED generated column `alias_key` + UNIQUE INDEX, scoped to `kind='template'` |
| `postgres/20260704120000_template_alias_unique.sql` | PostgreSQL counterpart (identical generated column expression) |

The STORED generated column approach replaces a MySQL functional CASE index, which would break the cross-dialect schema-alignment test (`COLUMN_NAME=NULL` in `INFORMATION_SCHEMA`). Both dialects produce identical index signatures.

### SDK (Python)

| File | Change |
|---|---|
| `_template.py` | `Template.build(name=...)` now forwards `name` to the request body (was silently dropped) |

## E2E Tests

6 tests in `sdk/python/tests/test_template_e2e.py`, verified against a live one-click deployment:

| Test | Covers |
|---|---|
| `test_template_list_and_get_existing_template` | List + GET (pre-existing) |
| `test_template_create_from_image_and_cleanup` | Create + cleanup (pre-existing) |
| `test_template_alias_create_get_and_delete` | Full alias lifecycle: create → get by alias → list → delete by alias → post-delete 404 |
| `test_template_alias_validation_rejects_invalid_formats` | Uppercase, `tpl-`/`snap-` prefix, >64 chars → 400 |
| `test_template_alias_dedicated_lookup_endpoint` | `GET /templates/aliases/:alias` E2B compat endpoint |
| `test_template_alias_rebuild_reassignment` | Same alias on second template → alias moves to the newly READY template |

```
6 passed in 24.32s
```

## Design Decisions

- **Reuse `display_name` column**: No new data column needed; a STORED generated column (`alias_key`) derived from it enforces uniqueness
- **Template-only alias scope**: Both write path (`createDefinitionTx`) and read path (`GetTemplateByAlias`) filter `kind=template`; snapshots never participate in alias ownership
- **Transactional reassignment**: Release + claim wrapped in `db.Transaction()` (TOCTOU-safe)
- **Cross-dialect parity**: Identical generated column expression in MySQL and PostgreSQL; passes `TestMigrationVersionParityAcrossDialects`

Assisted-by: OhMyPi:zhipu-coding-plan/glm-5.2


sparkzky:
## Rebase + blind-spot pass

Rebased onto latest `master` (resolved conflict in `models/mod.rs` test imports with `SetTimeoutRequest` from master). One new fix commit on top:

**1. PostgreSQL migration gap (blocks #674)**
`#674` landed PostgreSQL support with cross-dialect parity tests. This PR only had a MySQL migration — `TestMigrationVersionParityAcrossDialects` failed. Added `postgres/20260704120000_template_alias_unique.sql` and rewrote both MySQL + PG to use a **STORED generated column** (`alias_key`) + plain UNIQUE INDEX, replacing the MySQL functional CASE index (which reports `COLUMN_NAME=NULL` in `INFORMATION_SCHEMA` and breaks the schema-alignment test).

**2. Code simplification**
- Consolidated `createDefinition`/`ensureTemplateDefinition` duplicate bodies into thin `//go:noinline` delegators to their `WithOptions` variants (gomonkey patches still reliable).
- Removed dead alias-resolution block from `ResolveTemplate` — zero callers; production sandbox-create path is `dealCubeboxCreateReqWithTemplate`.
- Updated `store_test.go` patch to target `createDefinitionWithOptions`.

**3. E2B compatibility verified**
Confirmed `get_template_by_alias` works correctly — CubeMaster `GET /cube/template` already calls `resolveTemplateIdentifierFn` (template.go:220), so alias lookups resolve through the existing path.

All Go + Rust tests pass.

kinwin-ustc:
Template aliases are a necessary feature, and we need to support this capability in version 0.6.0. The implementation of this PR is much simpler and clearer than PR #607.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -886,6 +899,50 @@ func GetDefinition(ctx context.Context, templateID string) (*models.TemplateDefi
...
+	if !isReady() {
+		return nil, ErrTemplateStoreNotInitialized
+	}
+	alias = strings.TrimSpace(alias)
+	if alias == "" {
+		return nil, ErrTemplateNotFound
+	}
+	def := &models.TemplateDefinition{}
+	err := store.db.WithContext(ctx).Table(constants.TemplateDefinitionTableName).
+		Where("display_name = ? AND status <> ?", alias, StatusDeleting).
```
fslongjin: Small inconsistency here — worth a quick fix.

On the write path you already keep snapshots out of alias ownership: only templates can claim/steal an alias, and a snapshot's `display_name` is just a free-form label even if it collides. That part is clear.

But `GetTemplateByAlias` on the read path looks incomplete: it filters by `display_name` only, with no `kind = template`. Combined with `First()` and no ordering, you get whichever row the DB happens to return first.

So if a snapshot happens to share the same `display_name` as a template alias, resolving by alias (sandbox create / template lookup) can hit the snapshot instead of the template that actually owns the alias. Writes are guarded; reads aren't.

Suggest adding `kind = template` to the query, plus a unit test: when a template and a snapshot share the same name, resolution only returns the template.

CubeMaster/pkg/templatecenter/snapshot_ops.go
```diff
@@ -1092,6 +1092,29 @@ func createDefinitionTx(ctx context.Context, tx *gorm.DB, templateID string, sto
...
+	// scoped to template-kind rows, so the two kinds never contend on
+	// display_name.
+	if kind == TemplateKindTemplate {
+		if alias := strings.TrimSpace(opts.DisplayName); alias != "" {
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where("kind = ? AND display_name = ? AND template_id <> ? AND status <> ?", TemplateKindTemplate, alias, templateID, StatusDeleting).
+				Update("display_name", "").Error; err != nil {
+				return fmt.Errorf("release stale alias %q fail: %w", alias, err)
+			}
+		}
```
fslongjin: **Blocking:** this clears the old template's alias as soon as the new definition is inserted (still `PENDING`). If replica create / finalize fails later, the job is only marked failed — the old alias is not restored, and the new definition keeps the name.

User impact is bad: a failed build can leave the “stable” alias pointing at an unusable template.

Please switch ownership only after the candidate is READY (atomically), keep the old mapping on failure, and preferably don't treat `PENDING`/`FAILED` definitions as valid alias owners. Also add a real-DB regression test: old READY template + new build fails after definition create.

CubeDB/migrate/migrations/mysql/20260704120000_template_alias_unique.sql
```diff
@@ -0,0 +1,72 @@
...
+JOIN (
+  SELECT `display_name`, MAX(`id`) AS keep_id
+  FROM `t_cube_template_definition`
+  WHERE `display_name` <> '' AND `kind` = 'template'
+  GROUP BY `display_name`
+  HAVING COUNT(*) > 1
+) AS d
+  ON t.`display_name` = d.`display_name`
+ AND t.`id` <> d.`keep_id`
+SET t.`display_name` = '';
```
fslongjin, line 41: **Blocking:** the comment says snapshots are intentionally excluded from dedup, but the outer UPDATE only matches `display_name` + `id != keep_id` — it never re-checks `t.kind = 'template'`.

Example: two templates named `my-app` (triggers dedup) plus a snapshot also named `my-app` (allowed today). After upgrade, that snapshot's `display_name` gets cleared too.

The PostgreSQL migration has the same bug. Please add `t.kind = 'template'` on both outer UPDATEs, and add a real migration data test (not just schema alignment).

CubeDB/migrate/migrations/postgres/20260704120000_template_alias_unique.sql
```diff
@@ -0,0 +1,69 @@
...
+   SET display_name = ''
+  FROM (
+    SELECT display_name, MAX(id) AS keep_id
+      FROM t_cube_template_definition
+     WHERE display_name <> '' AND kind = 'template'
+     GROUP BY display_name
+    HAVING COUNT(*) > 1
+  ) AS d
+ WHERE t.display_name = d.display_name
+   AND t.id <> d.keep_id;
```
fslongjin, line 38: Same issue as the MySQL migration: the subquery filters `kind = 'template'`, but the outer UPDATE does not, so a snapshot sharing a duplicated template `display_name` can be wiped. Please add `AND t.kind = 'template'` here too.

CubeMaster/pkg/templatecenter/snapshot_ops.go
```diff
@@ -1092,6 +1092,29 @@ func createDefinitionTx(ctx context.Context, tx *gorm.DB, templateID string, sto
 	if kind == TemplateKindSnapshot && model.StorageBackend == "" {
 		model.StorageBackend = StorageBackendCow
 	}
+	// Alias reassignment: when a new template claims a display_name (alias)
+	// that another non-deleting template already holds, release it from the
+	// old template first. This gives "stable naming across rebuilds" — the
+	// newest template with a given alias wins — without requiring callers to
```
fslongjin: **Blocking:** the comment says “newest template with a given alias wins”, but the implementation doesn't actually guarantee that.

If two different `templateID`s claim the same unused alias concurrently, both UPDATEs can touch 0 rows, then both INSERT; the unique index just makes one succeed and the other hit a uniqueness error. The winner is whoever the DB schedules — not “who claimed / became READY later”. Per-templateID local locks also don't serialize alias ownership across CubeMaster instances.

PostgreSQL unique conflicts also don't look mapped to `ErrDuplicateTemplate` today (the check mostly recognizes MySQL `1062` / `Duplicate entry`).

Please move claim to after READY, add cross-instance serialization for the alias (ownership row + `FOR UPDATE`, advisory lock, or retryable claim), define what “newest” means, and add MySQL + PostgreSQL concurrency tests.

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
```diff
@@ -437,6 +437,19 @@ func dealCubeboxCreateReqWithTemplate(ctx context.Context, reqInOut *types.Creat
...
+	// ID (no tpl-/snap- prefix), treat it as a human-readable alias and
+	// resolve it to the actual template ID. This lets sandboxes reference
+	// templates by a stable name that survives rebuilds.
+	if hasTemplateID && templateID != "" {
+		if resolved, err := templatecenter.ResolveTemplateIdentifier(ctx, templateID); err != nil {
+			return fmt.Errorf("failed to resolve template identifier %q: %w", templateID, err)
+		} else if resolved != "" && resolved != templateID {
+			templateID = resolved
+			reqInOut.Annotations[constants.CubeAnnotationAppSnapshotTemplateID] = templateID
+		}
```
fslongjin: **Blocking:** alias resolution sits before the version split. Non-v2 requests are supposed to keep going to `dealCubeboxReqTemplateByLocalConfig`.

If a non-v2 request carries a legacy / custom identifier without a `tpl-`/`snap-` prefix, we now treat it as an alias, hit the DB, and fail when it's missing — so we never reach the old local-config path. That's a compatibility regression.

Please resolve aliases only on the template-center / v2 branch, leave non-v2 behavior alone, and add a regression test for non-v2 + non-prefixed annotation.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -886,6 +899,57 @@ func GetDefinition(ctx context.Context, templateID string) (*models.TemplateDefi
...
+	if !isReady() {
+		return nil, ErrTemplateStoreNotInitialized
+	}
+	alias = strings.TrimSpace(alias)
+	if alias == "" {
+		return nil, ErrTemplateNotFound
+	}
+	def := &models.TemplateDefinition{}
+	err := store.db.WithContext(ctx).Table(constants.TemplateDefinitionTableName).
+		Where("kind = ? AND display_name = ? AND status <> ?", TemplateKindTemplate, alias, StatusDeleting).
```
fslongjin: The migration indexes `alias_key`, but this query (and the reassignment UPDATE) still filter on `kind + display_name`, which usually won't use that generated-column unique index.

Sandbox create-by-alias is a hot path — large tables will scan. Please query/update by `alias_key = ?` and confirm with `EXPLAIN` on MySQL and PostgreSQL. After that, I wouldn't add a local alias cache; cross-instance moves make caches easy to stale.

CubeAPI/src/services/templates.rs
```diff
@@ -289,6 +303,52 @@ fn string_or(value: String, fallback: &str) -> String {
...
+        return None;
+    }
+
+    let without_tag = name.split_once(':').map_or(name, |(name, _)| name).trim();
+    let alias = without_tag.rsplit('/').next().unwrap_or(without_tag).trim();
+    if alias.is_empty() {
+        None
+    } else {
+        Some(alias.to_string())
+    }
```
fslongjin, line 366: `split_once(':')` strips at the first colon. For `registry:5000/team/app:v2` that becomes `registry`, then we silently claim that alias — and can overwrite an existing `registry` owner.

`team-a/app` and `team-b/app` also collapse to the same global `app`. If that's intentional, please document it; if not, validate strictly or keep the namespace.

Please pin down the E2B `name` grammar (registry/port allowed? how tags are stripped? namespace kept?) and add boundary tests. Don't silently truncate into the wrong alias.

CubeAPI/src/handlers/templates.rs
```diff
@@ -157,6 +218,20 @@ pub struct DeleteTemplateQuery {
     pub sync: Option<bool>,
 }
 
+#[utoipa::path(
+    delete,
+    path = "/templates/{templateID}",
+    params(
+        ("templateID" = String, Path, description = "Template identifier or alias"),
```
fslongjin, line 225: This path now documents delete-by-alias, but `reverse_sync_agenthub_template` still gets the raw path string (which may be the alias). CubeMaster deletes via the resolved template ID; AgentHub lookup matches `template_id` / `source_snapshot_id` exactly, so alias deletes can leave AgentHub registrations behind.

Please resolve to a canonical ID before reverse-sync (or return it from delete), and add an AgentHub integration test.

CubeMaster/pkg/templatecenter/request_validation.go
```diff
@@ -106,6 +110,29 @@ func normalizeTemplateImageRequest(req *types.CreateTemplateFromImageReq) (*type
...
+// format: lowercase alphanumeric and hyphens only, must start with an
+// alphanumeric character, max 64 characters, and must NOT collide with the
+// template ID prefix convention (tpl-/snap-). An empty alias is valid (no
+// alias requested).
+func validateTemplateAlias(alias string) error {
+	alias = strings.TrimSpace(alias)
+	if alias == "" {
+		return nil
+	}
+	if hasValidTemplateIDPrefix(alias) {
```
fslongjin, line 130: The docs/comments say we reject anything starting with `tpl-` / `snap-`, but `hasValidTemplateIDPrefix` only returns true when there are characters after the prefix — so the bare values `tpl-` and `snap-` slip through.

A plain `HasPrefix` check is enough; please add those two boundary cases to the tests.

CubeAPI/src/services/templates.rs
```diff
@@ -93,15 +86,34 @@ impl TemplateService {
...
+        let alias = alias.trim();
+        if alias.is_empty() {
+            return Err(AppError::BadRequest("alias is required".to_string()));
+        }
+
+        let detail = self.get_template(alias).await?;
+        Ok(TemplateAliasLookupResponse {
+            template_id: detail.template_id,
+            public: TEMPLATE_PUBLIC,
         })
```
fslongjin, line 123: OpenAPI says invalid aliases return 400, but this only 400s on empty input. Other invalid formats fall through to generic get and usually 404; `tpl-*` / `snap-*` are even treated as normal IDs.

Either reuse the same alias validation as CubeMaster at the entrance, or update OpenAPI / the endpoint description to match real behavior.

CubeMaster/pkg/templatecenter/image_job_runner.go
```diff
@@ -163,7 +163,9 @@ func runTemplateImageJob(ctx context.Context, jobID string, req *types.CreateTem
 		})
 		return
 	}
-	if _, err := ensureTemplateDefinition(ctx, req.TemplateID, storedReq, generatedReq.InstanceType, constants.GetAppSnapshotVersion(generatedReq.Annotations)); err != nil {
+	if _, err := ensureTemplateDefinitionWithOptions(ctx, req.TemplateID, storedReq, generatedReq.InstanceType, constants.GetAppSnapshotVersion(generatedReq.Annotations), definitionCreateOptions{
+		DisplayName: strings.TrimSpace(req.Alias),
+	}); err != nil {
```
fslongjin: Related to the alias lifecycle blocker: this calls `ensureTemplateDefinitionWithOptions` (and thus alias claim) before replica create / finalize. If anything below fails, the alias has already moved. Prefer claiming only after the template is actually READY.

fslongjin (changes requested):
The overall direction looks right: stable aliases, template-vs-snapshot isolation on both read/write paths, release+claim in a DB transaction, and MySQL/PostgreSQL migrations. CI is green too.

I still don't think this is ready to merge. A few core behaviors don't match the PR description:

1. The alias is taken from the old template before the new one is READY; if the build fails later, it is never restored, so the stable name can point at an unusable template.
2. The migration dedup outer UPDATE is missing `kind='template'`, so a same-named snapshot `display_name` can be cleared.
3. Concurrent claims on the same alias only guarantee uniqueness, not “newest wins”.
4. Alias resolution runs before the version split in sandbox create, which can break non-v2 legacy paths.

Also: alias lookups don't use the new `alias_key` index; `name` with a registry port can be truncated incorrectly; delete-by-alias can leave AgentHub registrations behind. Tests mostly cover validation/forwarding happy paths, but failure rollback, concurrency, migration data semantics, and sandbox-by-alias are still thin.

Requesting changes. Please fix the blockers above and add matching regression tests before we revisit merge.

CubeMaster/pkg/templatecenter/request_validation.go
```diff
@@ -124,7 +124,7 @@ func validateTemplateAlias(alias string) error {
 	if alias == "" {
 		return nil
 	}
-	if hasValidTemplateIDPrefix(alias) {
+	if strings.HasPrefix(alias, "tpl-") || strings.HasPrefix(alias, "snap-") {
```
fslongjin: why not use `hasValidTemplateIDPrefix`

sparkzky: Good catch — I initially switched to `strings.HasPrefix` because `hasValidTemplateIDPrefix` requires `len > len(prefix)`, so the bare values `tpl-` / `snap-`  (with no characters after the prefix) would slip through. But that's not a realistic input scenario, so I've reverted to `hasValidTemplateIDPrefix` in aeeadf4.

fslongjin (approved):

lkml-likexu:
@sparkzky Thanks for your contribution. Could you please add some test cases to the `tests/e2e/sdk_compat/` directory to verify that future changes won't cause regressions for this feature ?

PR #751, opened 2026-07-04
fix(cubelet): avoid FATAL crash when ControllerPlugin does not implement CubeMetaController

### LINKED ISSUE #750: [Bug] Cubelet FATAL: controller run-template-manager is not a valid CubeMetaController
## Bug Description

Cubelet crashes approximately every 5 minutes:

```
LogLevel: FATAL
LogContent: controller run-template-manager is not a valid CubeMetaController
```

This happens on every startup, regardless of database/template/state content. Tested with zero templates, zero sandboxes, and CubeMaster stopped — all failed to mitigate.

## Root Cause

The `NewCubelet` function receives `rtManager runtemplate.RunTemplateManager` as a parameter. The `runtemplate` controller is registered at init time via `GetInMemorySandboxControllers()`, but its implementation fails the `CubeMetaController` interface check.

Source: `Cubelet/pkg/controller/runtemplate/template_manager.go`

Binary strings also reveal: `obj is not *templatetypes.LocalRunTemplate` and `failed to persist recovered local template`, suggesting a runtime type assertion failure.

## Steps to Reproduce

1. Install Cube Sandbox v0.5.0 via one-click bare-metal installer
2. Start cubelet service
3. `systemctl status cube-sandbox-cubelet.service` shows `status=2/INVALIDARGUMENT` every ~5 minutes
4. Check `/data/log/Cubelet/Cubelet-req.log` for the FATAL entry

## Environment

- Cube Sandbox: v0.5.0 (30b4e25ab), also reproduced on v0.4.0 (4004a6ec)
- OS: Fedora 44, Linux 7.0.13, x86_64
- Filesystem: btrfs
- Installation: bare-metal one-click

## Impact

- Cubelet auto-restarts every ~5 minutes (systemd Restart=always)
- ~2-second service interruption per restart
- `cubecli exec` fails during crash windows
- Sandbox TAP networking requires reconnect after crash

## Attempted Mitigations (all failed)

- Deleting all template replicas/definitions from MySQL
- Deleting all cubelet state files and snapshots
- Stopping CubeMaster entirely
- Setting ReconcileInterval to 86400s
- Commenting out controller TOML config section
- Downgrading cubelet binary to v0.4.0
- seccomp filtering exit_group (caused SIGSYS crash)

## Log Snippet

```
{"LogLevel":"FATAL","LogContent":"controller run-template-manager is not a valid CubeMetaController"}
```

kinwin-ustc:
## Summary

- Introduce a dedicated `ResourceManagerPlugin` type (`io.cubelet.resource-manager.v1`) and move `run-template-manager` from `ControllerPlugin` to it
- Downgrade `Fatalf` to `Warnf` in the controller init loop so unexpected interface mismatches degrade gracefully instead of crashing

## Root Cause

The `run-template-manager` was registered as `ControllerPlugin` type but only implements `RunTemplateManager` — it has no `Run(stopCh <-chan struct{}) error` method required by `CubeMetaController`. The controller initialization loop in `cubelet_plugin.go` iterates all `ControllerPlugin` plugins and called `Fatalf` when the type assertion failed, causing Cubelet to crash on every startup (~5 minute restart cycle).

This bug has existed since the initial open-source release (commit 37cb5c6, 2026-04-16).

## Fix

1. Add `ResourceManagerPlugin` plugin type to cleanly separate resource managers (service providers) from controllers (that implement `Run`)
2. Register `run-template-manager` under `ResourceManagerPlugin` instead of `ControllerPlugin`
3. Update `cubelet_plugin.go` and `appsnapshot_plugin.go` to depend on and retrieve from the new type
4. Downgrade `Fatalf` to `Warnf` as a defense-in-depth measure

## Test Plan

- [ ] Verify Cubelet starts without FATAL crash on bare-metal deployment
- [ ] Verify `run-template-manager` is correctly initialized and functional (template listing/ensuring works)
- [ ] Verify `appsnapshot` plugin can still retrieve the template manager
- [ ] Confirm no regression in sandbox creation with snapshot templates

Fixes #750

Signed-off-by: Feng Jin <ronyjin@tencent.com>
Assisted-by: Cursor:claude-opus-4-8-thinking-high

### DIFF
--- a/Cubelet/pkg/constants/const.go
+++ b/Cubelet/pkg/constants/const.go
@@ -90,6 +90,7 @@ const (
 	ControllerConfigPlugin  plugin.Type = "io.cubelet.controller.config.v1"
 	ControllerCubeletPlugin plugin.Type = "io.cubelet.controller.cubelet.v1"
 	ControllerPlugin        plugin.Type = "io.cubelet.controller.v1"
+	ResourceManagerPlugin   plugin.Type = "io.cubelet.resource-manager.v1"
 
 	PluginCubelet                       string   = "cubelet"
 	PluginRunTemplateManager            PluginId = "run-template-manager"
--- a/Cubelet/plugins/controller/cubelet_plugin.go
+++ b/Cubelet/plugins/controller/cubelet_plugin.go
@@ -51,6 +51,7 @@ func registerCubelet() {
 			plugins.CRIServicePlugin,
 			constants.ControllerPlugin,
 			constants.ControllerConfigPlugin,
+			constants.ResourceManagerPlugin,
 		},
 		InitFn: func(ic *plugin.InitContext) (interface{}, error) {
 			obj, err := ic.GetByID(constants.ControllerConfigPlugin, constants.PluginCubelet)
@@ -91,11 +92,11 @@ func registerCubelet() {
 				if ok {
 					controllerMap[name] = c
 				} else {
-					log.G(ic.Context).Fatalf("controller %s is not a valid CubeMetaController", name)
+					log.G(ic.Context).Warnf("controller %s is not a valid CubeMetaController", name)
 				}
 			}
 
-			obj, err = ic.GetByID(constants.ControllerPlugin, constants.PluginRunTemplateManager.ID())
+			obj, err = ic.GetByID(constants.ResourceManagerPlugin, constants.PluginRunTemplateManager.ID())
 			if err != nil {
 				return nil, fmt.Errorf("failed to get run template manager: %w", err)
 			}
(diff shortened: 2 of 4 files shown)


zhanglib1996:
## Verified the updated PR

Rebuilt cubelet with the 4-file commit (87af80c) and tested:

- **Zero crashes** for 2+ minutes (was every ~5 minutes before)
- **Log**: `WARN controller run-template-manager is not a valid CubeMetaController, skipping`
- **Sandbox creation**: ✅ works with real cubecow library
- **Node health**: ✅ 10.42.0.1 RUNNING
- **E2B SDK**: ✅ run_code, DNS, HTTP, HTTPS all pass

The fix is solid. Ready to merge from our testing.

Environment: Fedora 44 bare-metal, go1.24.4, self-compiled cubelet with cubecow v0.1.0

PR #762, opened 2026-07-05
fix(CubeMaster): stably sort sandbox list to stop reordering on refresh

### LINKED ISSUE #742: Sandboxes list lacks stable sorting and reorders every 2 seconds
## Description

The Sandboxes list page currently does not provide any sorting controls, and the list order changes roughly every 2 seconds as the page refreshes/updates.

This makes it difficult to track or locate a specific sandbox, especially when there are many active sandboxes.

## Page

`hostname:12088/sandboxes`

## Actual Behavior

- The sandboxes list refreshes approximately every 2 seconds.
- After each refresh, the order of items appears to change.
- There is no visible sorting option or stable default ordering.

## Expected Behavior

- The list should have a stable default sort order, such as `created_at desc`, `updated_at desc`, or status + creation time.
- Auto-refresh should preserve the current sort order.
- The page should ideally provide sorting controls for common fields such as creation time, update time, status, name, or sandbox ID.

## Impact

- Users lose their position in the list after each refresh.
- It is hard to monitor a specific sandbox.
- The page becomes difficult to use when many sandboxes are present.

## Suggested Fix

Add stable sorting either on the backend API response or in the frontend rendering layer. The selected sort field and direction should remain active across automatic refreshes.

Image

yingjun8:
Fixes #742.
  
### Problem
The Sandboxes list (`:12088/sandboxes`) reorders on every ~auto-refresh. `ListSandbox` fans out to per-node goroutines and appends results to `rsp.Data` in goroutine-completion order, which varies between calls, so the WebUI/SDK receive a reshuffled list each poll. No layer (CubeMaster, CubeAPI, frontend) imposed a stable order.

### Root cause
The request path is:
WebUI → CubeAPI `GET /v2/sandboxes` → CubeMaster `ListSandbox` (gRPC) → per-node Cubelet.
In `CubeMaster/pkg/service/sandbox/sandbox_list.go`, `ListSandbox` fans out to one goroutine per node and aggregates the results through a  channel:

```go
for res := range resChan {
    rsp.Data = append(rsp.Data, res)
}   
```
Results are appended in goroutine-completion order, which is non-deterministic and varies between calls and across nodes. No layer (CubeMaster, CubeAPI, or the frontend) applied any sort, so each refresh returned the same set of sandboxes in a different order.
  
### Fix
Add a stable sort after aggregation in ListSandbox, before returning:

- primary key: CreateAt descending (newest first)
- tie-breaker: SandboxID ascending (deterministic)

This is consistent with the existing cubecli list command, which already sorts by CreatedAt descending. The fix is contained to the CubeMaster aggregation layer, so all consumers (WebUI and SDK) get a stable order without frontend changes.
```
sort.Slice(rsp.Data, func(i, j int) bool {
  if rsp.Data[i].CreateAt != rsp.Data[j].CreateAt {
       return rsp.Data[i].CreateAt > rsp.Data[j].CreateAt
  }
  return rsp.Data[i].SandboxID < rsp.Data[j].SandboxID
})
```

### DIFF
--- a/CubeMaster/pkg/service/sandbox/sandbox_list.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_list.go
@@ -7,6 +7,7 @@ package sandbox
 import (
 	"context"
 	"runtime/debug"
+	"sort"
 	"strconv"
 	"strings"
 	"sync"
@@ -100,6 +101,17 @@ func ListSandbox(ctx context.Context, req *types.ListCubeSandboxReq) (rsp *types
 	wg.Wait()
 	close(resChan)
 	<-done
+
+	// Results arrive over resChan in goroutine-completion order, which varies
+	// between calls and across nodes, so rsp.Data would otherwise be reshuffled
+	// on every list request. Sort by creation time descending (newest first),
+	// falling back to SandboxID for a deterministic, stable order.
+	sort.Slice(rsp.Data, func(i, j int) bool {
+		if rsp.Data[i].CreateAt != rsp.Data[j].CreateAt {
+			return rsp.Data[i].CreateAt > rsp.Data[j].CreateAt
+		}
+		return rsp.Data[i].SandboxID < rsp.Data[j].SandboxID
+	})
 	return
 }
 


chenhengqi (approved):
LGTM. Thanks.

fslongjin (approved):

PR #763, opened 2026-07-05
fix(egress): prevent DNS TTL expiry from killing long-lived TCP connections

kinwin-ustc:
## Summary

DNS-learned `allow_out_v2` entries expire after DNS TTL, causing the eBPF datapath to RST established TCP connections — even though the L7 policy still allows the domain.

### Root Cause

1. `dns_learn_response_ip()` sets `expires_at_ns = now + DNS_TTL` for each learned IP
2. When TTL expires, `check_net_policy()` rejects subsequent packets
3. For TCP, this triggers `tcp_reply_reset()` — the sandbox sees an abrupt connection termination

### Fix

**Activity-based TTL refresh (rate-limited)** — DNS-learned entries retain their original TTL. `check_net_policy()` now accepts a `refresh_ttl` flag: when a non-SYN TCP packet matches a dynamic entry whose remaining TTL has dropped below half of `DNS_LEARNED_REFRESH_NS` (5 min), the expiration is extended by 10 min. The refresh reuses the existing LPM lookup pointer — zero extra map traversals.

| Property | Behavior |
|----------|----------|
| Active connections | Kept alive — non-SYN data packets trigger rate-limited TTL refresh |
| Idle/stale entries | Naturally expire and are cleaned up by `reapDNSLearnedPolicies` |
| Static CIDR rules | Untouched (`expires_at_ns == 0`) |
| SYN packets | No refresh — new connections validated against current DNS state |
| Extra LPM lookups per packet | **0** (reuses existing lookup pointer) |
| Write frequency | At most **once per ~5 min** per destination IP |
| Max idle timeout | **600s** (eBPF + Nginx aligned) |

CubeEgress `proxy_read_timeout` / `proxy_send_timeout` increased from 60s to 600s, aligned with the eBPF refresh window, to support long-lived HTTP streams while limiting resource holding on upstream hangs.

### Changed Files

| File | Change |
|------|--------|
| `CubeNet/src/cubevs.h` | Add `DNS_LEARNED_REFRESH_NS` constant (600s) |
| `CubeNet/src/mvmtap.bpf.c` | Add rate-limited TTL refresh inside `check_net_policy()` for non-SYN TCP on dynamic entries |
| `CubeEgress/nginx.conf` | Increase proxy timeouts from 60s to 600s |

## Test Plan

- [ ] Verify non-SYN TCP packets refresh entry expiration only when remaining TTL < 5 min
- [ ] Verify SYN packets do **not** refresh the entry
- [ ] Verify long-lived TCP connection (e.g. gRPC stream) survives multiple DNS TTL cycles
- [ ] Verify idle DNS-learned entries expire and get reaped after 10 min without traffic
- [ ] Verify static AllowOut CIDR entries (`expires_at_ns == 0`) are never modified by refresh
- [ ] Verify CubeEgress proxy timeout supports 10-min idle HTTP streams without reset
- [ ] Verify no measurable performance regression under high-pps TCP workload

### DIFF
--- a/CubeEgress/nginx.conf
+++ b/CubeEgress/nginx.conf
@@ -123,8 +123,8 @@ http {
 
             proxy_pass http://$server_addr:$server_port;
             proxy_connect_timeout 10s;
-            proxy_send_timeout 60s;
-            proxy_read_timeout 60s;
+            proxy_send_timeout 600s;
+            proxy_read_timeout 600s;
 
             header_filter_by_lua_block {
                 require("debug_dump").dump_response_headers()
@@ -192,8 +192,8 @@ http {
             proxy_ssl_name $cube_original_sni;
             proxy_pass https://$server_addr:$server_port;
             proxy_connect_timeout 10s;
-            proxy_send_timeout 60s;
-            proxy_read_timeout 60s;
+            proxy_send_timeout 600s;
+            proxy_read_timeout 600s;
 
             header_filter_by_lua_block {
                 require("debug_dump").dump_response_headers()
--- a/CubeNet/src/cubevs.h
+++ b/CubeNet/src/cubevs.h
@@ -38,6 +38,11 @@
 #define NET_POLICY_FLAG_L7_REQUIRED	1
 #define NSEC_PER_SEC			1000000000ULL
 #define DNS_QUERY_TRACK_TTL_NS		(10ULL * NSEC_PER_SEC)
+/* Duration to extend DNS-learned allow_out_v2 entries when non-SYN TCP
+ * traffic hits them.  Keeps long-lived connections alive across DNS TTL
+ * rotation while still letting the reaper reclaim truly idle entries.
+ */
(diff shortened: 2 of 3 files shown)


CubeNet/src/dns_response.h
```diff
@@ -98,29 +98,33 @@ static __always_inline bool dns_response_learning_enabled(__u32 ifindex)
...
+ * connections are not killed when DNS TTL elapses. Entries are cleaned up on
+ * policy change (replaceNetPolicy flushes the inner map) or sandbox teardown.
+ */
 static __always_inline void dns_learn_response_ip(__u32 ifindex, __u32 ip, __u32 ttl,
 						  __u8 flags)
 {
 	struct lpm_key key = { .prefixlen = 32, .ip = ip };
 	struct net_policy_value_v2 value = {
-		.expires_at_ns = bpf_ktime_get_ns() + ((__u64)ttl * NSEC_PER_SEC),
+		.expires_at_ns = 0,
```
chenhengqi: This will block DNS reaper to do its job and the allow_out_v2 map will get full.

Issue #644, opened 2026-06-25
【2026犀牛鸟issue】CubeSandbox 与 Claude Code / CodeBuddy / OpenCode 集成


fslongjin:
【产品名称/功能模块】
CubeSandbox 与 Claude Code / CodeBuddy / OpenCode 集成

【问题描述】
CubeSandbox 定位为面向 AI Agent 的安全沙箱服务，并与 e2b 协议兼容。Claude Code、CodeBuddy、OpenCode 是当前主流的终端型 AI 编码 Agent——它们在隔离环境中运行代码、编辑文件、执行命令，天然适合以 CubeSandbox 沙箱作为执行后端或运行环境：既可以把编码 Agent 本身跑在沙箱里获得隔离、可复现的运行环境，也可以让 Agent 编排系统按需拉起 CubeSandbox 沙箱执行其生成的代码。

目前 CubeSandbox 已有面向 OpenAI Agents、OpenClaw 等的集成示例，但尚未提供 Claude Code / CodeBuddy / OpenCode 这三类终端型编码 Agent 的集成文档与可运行示例，社区缺少可直接参照的"上手即用"材料。本 issue 为开放性文档与示例贡献任务，允许多人并行参与、各自认领一个子方向，以增量方式合入，不要求单次提交完整 PR。

【期望内容/功能改进】（每人可认领其中一个或多个子方向，渐进式贡献）
1）为 Claude Code、CodeBuddy、OpenCode 中至少一个 Agent 编写端到端集成指南，覆盖：在 CubeSandbox 中运行该 Agent 所需的模板构建、运行环境与依赖、API Key 等敏感配置的注入方式、网络与出口策略（Agent 需访问 LLM API）、以及会话保持与状态持久化（借助快照保存 Agent 运行状态、跨会话恢复）；
2）提供可运行的示例工程（置于 examples/ 下），包含模板定义与构建脚本、最小可复现的调用流程，以及能在本地部署环境跑通的 README；
3）给出典型使用场景与最佳实践：如把编码 Agent 跑在沙箱内做隔离开发、用沙箱执行 Agent 生成的代码并回收结果、利用快照实现 Agent 长时任务的断点续跑等；
4）补充常见问题与排错（鉴权失败、出口被拦截、镜像拉取、超时等），并与平台既有出口管控、鉴权、模板机制对齐；
5）文档纳入 docs/guide/integrations 体系，遵循既有集成文档模板与风格。

【参与说明与验收】
[...]

PR #765, opened 2026-07-05
docs: add Claude Code integration guide and examples

shsaihdsaiudh:
## Problem

Claude Code runs on the host, so **every `Bash` command it issues executes on that host**. Existing sandbox integrations (MCP- or SDK-based) only help when the agent *chooses* a sandbox tool — a plain `Bash` call still lands on the host, with no isolation.

Refs #644.

## Solution

Add one runnable example centered on a **`PreToolUse` hook** that transparently redirects **every `Bash` command** Claude Code runs into an isolated **CubeSandbox** MicroVM. The model issues normal `Bash` calls; the hook rewrites them before they execute. No prompt, tool, or workflow changes are required.

```
Claude Code (host)
    ├── Read / Write / Edit ─────────────► host project files
    │
    └── Bash ──► PreToolUse hook ──► cubesandbox_exec ──► CubeAPI ──► MicroVM
                 (cubesandbox_rewrite.py)                 (:3000)     └─ reusable per session
```

Only `Bash` is redirected. `Read`, `Write`, and `Edit` keep operating on host files, so Claude Code edits your project locally while its shell commands run in a throwaway kernel / filesystem / network.

This PR also folds in the hook backend from #783 (now closed as superseded); the earlier "run Claude Code inside the sandbox and connect from outside" surfaces were dropped as too common per maintainer feedback.

## Changes

- `examples/claude-code-integration/hooks/`
  - `cubesandbox_rewrite.py` — the `PreToolUse` hook that rewrites each `Bash` call into an executor invocation via `updatedInput`.
  - `cubesandbox_exec.py` — the executor: reuses one MicroVM per Claude Code `session_id`, persists working directory and exported environment across calls, serializes concurrent calls in a session with a per-session file lock.
  - `install.sh` — idempotent install / uninstall that merges a `Bash` matcher into `~/.claude/settings.json` without clobbering other settings.
- Bilingual guides (`docs/guide/integrations/claude-code.md`, `docs/zh/...`) and index entries, plus example `README.md` / `README_zh.md` / `TROUBLESHOOTING.md`, all centered on the hook.
- Tests for the rewrite, executor, and install lifecycle.

## Security boundaries

- **Examples/docs only** — no change to CubeSandbox runtime or API behavior.
- **Fail-closed** — if the hook cannot rewrite a call safely, it exits non-zero and blocks the command rather than letting it run on the host.
- **Injection-safe** — the original command is passed as a single `shlex`-quoted argument, so shell metacharacters and newlines can never break out onto the host.
- **Unconditional wrapping** — every `Bash` call is rewritten; an already-wrapped executor invocation fed back through the hook simply fails inside the sandbox (the host hook path does not exist there), never on the host.
- **Scope** — the hook covers `Bash` only; `Read`/`Write`/`Edit` still run on the host.
- **Read-only host mount** — the optional project mount is read-only; if it cannot be created, Bash stays isolated in a sandbox without the shared file view.
- **No credential leakage** — the installer copies only whitelisted `CUBE_*` values into the hook config; LLM provider API keys are excluded, and dangerous self-executing variables (`BASH_ENV`, `ENV`, `LD_PRELOAD`, `PROMPT_COMMAND`) are scrubbed from the persisted environment.

## Verification

Verified on a real single-node cluster (CubeAPI reachable, live template):

```text
pytest examples/claude-code-integration/tests -q
71 passed

# Same `uname -a; hostname` on the host vs. through the hook shows a different
# kernel and hostname — commands run in a MicroVM, not on the host.
# `cd` / `export` persist across separate hook calls within a session.
# `echo start; touch /tmp/PWNED_FROM_HOOK; echo end` through the hook creates the
# file inside the sandbox and NOT on the host — the `;` never executes on the host.
```

`#783` is closed as superseded by this PR.


fslongjin:
Refs #644

fslongjin:
Hello, thank you for your contribution! Since we have a large number of participants in this event, to speed up the PR review process, could you please attach some screenshots and logs (both are needed) showing the verification process and results for this PR? Thank you for your participation!

shsaihdsaiudh:
## Review fixes and verification

I addressed the Claude Code integration review findings in commit `cf9c9a8`.

Highlights:

- Session reuse state is now UID-scoped and written with `0600` permissions.
- Provider credentials are read when used rather than frozen at import time; `CC_MODEL` is the single documented override.
- The secure-network example now reuses the shared command and setup helpers, including consistent non-zero exit handling.
- The duplicate Node version probe was removed, the standalone helper dependency is declared, and both READMEs now list the complete example layout.

### Verification

```text
uv run --with pytest --with-requirements examples/claude-code-integration/requirements.txt pytest examples/claude-code-integration/tests -q
62 passed in 0.12s

python3 -m py_compile examples/claude-code-integration/{_common.py,network_policy.py,run_claude_code.py,resume_claude_code.py,env_utils.py,sandbox_exec.py,mcp_server.py}
```

Full verification log: https://gist.github.com/shsaihdsaiudh/00b0c60c994dd7a5e9bbccb491289cee

A terminal screenshot has been prepared and will be added once GitHub binary attachment upload is available.

shsaihdsaiudh:
image

shsaihdsaiudh:
Issue linkage: this PR implements the Claude Code integration scope tracked in #644. The PR description closes #644.

fslongjin (changes requested):
Thanks for your contribution! I think the content of PR #783  can be merged into this. The pre-exec hook feature in PR # 783 is excellent, in my opinion, we can merge the two PRs into one.

shsaihdsaiudh:
Implemented the requested consolidation in commit `13674c9`.

- Moved only the unique PreToolUse hook capability from #783 into `examples/claude-code-integration/hooks/`.
- Reused #765's existing MCP server, standalone executor, requirements, and bilingual docs instead of adding a second copy.
- Made hook failures block Bash execution, serialized per-session command/reset state, kept host mounts read-only, and prevented provider keys from being copied into the installed hook config.
- Documented that Cubelet `hostPath` values require co-location or an already shared/synchronized identical path.

Verification: `110 passed`; repeated concurrency test passed 25/25; Python/Shell/Ruff checks and VitePress build passed. The GitHub `review` check is green. #783 is now superseded by this PR.

shsaihdsaiudh:
## Requested consolidation and follow-up fixes completed

The unique `PreToolUse` hook from #783 remains consolidated into this PR, and the branch is now rebased onto current `master`.

Additional fixes in the latest commit:

- Corrected MCP stdio from LSP-style `Content-Length` framing to newline-delimited JSON-RPC.
- Preserved exit code, stdout, and stderr and marked failed tools with `isError`.
- Added deterministic cached-sandbox cleanup on reset, EOF, and process exit.
- Removed the unsupported OpenAI-compatible execution claim and documented Claude Code's Anthropic-protocol requirement.
- Updated tests for protocol framing, failed command results, provider validation, and cleanup.

Verification: `112 passed`; full Ruff, Python compile, shell syntax, and VitePress builds passed.

Autonomously-by: Codex:GPT-5

shsaihdsaiudh:
Latest review follow-up is complete in `6f5490d`: state/session files now use descriptor-based `O_NOFOLLOW` handling, hook timeouts are capped, reconnect transport errors fall back safely, and Ctrl+C propagates after cleanup. The duplicate findings were covered by the same fixes; the `except Exception`/`KeyboardInterrupt` finding was a Python inheritance false positive and is explained inline. All 8 threads are resolved, the full example suite passes (117 tests), and the current automated review check is green. @fslongjin The earlier request to merge the hook work from #783 is also satisfied; please re-review when convenient.

fslongjin:
Hello, thank you for your contribution.

I think the feature in your PR that leverages the Claude Code hook to enable "executing code inside the sandbox" is very interesting and a major highlight.

I believe this PR would be better simplified into an example of using the hook. Because directly running Claude Code inside the sandbox and then connecting to it from outside is actually a common example that doesn't stand out much; on the contrary, the hook functionality you implemented is really intriguing.

shsaihdsaiudh:
Reworked this PR per your review, and folded in the hook backend from the now-closed #783.

**Changes**
- The example is now centered on the transparent **PreToolUse hook** (`hooks/cubesandbox_rewrite.py` + `cubesandbox_exec.py`): every Claude Code `Bash` call is rewritten into a per-session MicroVM exec, while `Read`/`Write`/`Edit` stay on the host.
- Removed the "run Claude Code inside the sandbox and connect from outside" surfaces you flagged as too common — `run_claude_code.py`, `resume_claude_code.py`, `network_policy.py`, `Dockerfile` — plus `mcp_server.py`, `sandbox_exec.py` and their helpers/tests.
- Merged #783's hardening intent into the hook: an idempotency guard so a command the hook already wrapped isn't wrapped again (executor-in-executor). The existing always-wrap + `shlex.join` + fail-closed design already keeps wrapping injection-safe.
- Docs (EN/ZH README, docs-site guide, troubleshooting) rewritten around the hook; deps trimmed to `cubesandbox` + `python-dotenv`. Net **+471 / −2651**.

**Verification on a real single-node cluster** (you asked for real-cluster logs)
- `pytest tests` → **57 passed** (35 rewrite + 19 executor + 3 install).
- Same `uname -a; hostname`: on the host → kernel `...cubesandbox.pvm.host...`, hostname `VM-0-15-opencloudos`; through the hook → kernel `...cubesandbox.pvm.guest...`, hostname `tpl-c703`. It runs in a MicroVM.
- The session state file records `sandbox_id=237c9fafc72e4ec09d9b3e8c87ff535a`, which matches the live sandbox reported by cube-api `/sandboxes`.
- `cd` and `export` persist across separate hook calls in a session; a file written in one call is present in the next (per-session sandbox reuse).
- `echo start; touch /tmp/PWNED_FROM_HOOK; echo end` routed through the hook creates the file **inside the sandbox** and **not on the host** (`ls` on host → No such file) — the `;` never executes on the host.

Full logs + checksums: https://gist.github.com/shsaihdsaiudh/63d038cd3995c49ad51549491ae8fe2e

shsaihdsaiudh:
## End-to-end verification on a real single-node cluster

@fslongjin As requested, here are verification logs for the `PreToolUse` hook, run against a live CubeSandbox deployment (CubeAPI on `:3000`, template `tpl-c703…` `READY`). The commit under test is the current PR head.

### 1. Unit tests — `71 passed`

```text
$ python -m pytest examples/claude-code-integration/tests -q
.......................................................................  [100%]
71 passed in 2.96s
```

### 2. Bash runs in a MicroVM, not on the host

Host vs. the same command routed through the hook — different kernel (`pvm.host` → `pvm.guest`) and hostname:

```text
# HOST
Linux VM-0-15-opencloudos 6.6.69-...cubesandbox.pvm.host-gb85200d80fa2 ... x86_64 GNU/Linux
hostname: VM-0-15-opencloudos

# THROUGH THE HOOK (cubesandbox_rewrite.py -> cubesandbox_exec.py)
Linux tpl-c703 6.6.69-...cubesandbox.pvm.guest-gb85200d80fa2 ... x86_64 GNU/Linux
hostname=tpl-c703
```

### 3. Hook rewrite + auto-approve (dangerous command stays a single quoted arg)

```json
// echo '{"tool_name":"Bash","cwd":"/root","session_id":"demo",
//        "tool_input":{"command":"rm -rf /; echo pwned"}}' | cubesandbox_rewrite.py
{
  "hookSpecificOutput": {
    "hookEventName": "PreToolUse",
    "permissionDecision": "allow",
    "updatedInput": {
      "command": ".../cubesandbox_exec.py --session=demo --mount=/root -- 'rm -rf /; echo pwned'"
    }
  }
}
```

### 4. Session state file → live sandbox

```text
state file: ~/.cache/cubesandbox-hook/29430d….json   (perms: 600)
{"sandbox_id": "1eafa7449e94422284686d2c37c6cf88", "mount": "...", "state_token": "…"}
cube-api /sandboxes live ids: ['1eafa7449e94422284686d2c37c6cf88']   MATCH: True
```

### 5. `cd` / `export` persist across separate hook calls in one session

```text
>> call A:  cd /tmp && export MYVAR=persisted_across_calls   -> set-done
>> call B:  echo cwd=$(pwd); echo MYVAR=$MYVAR
cwd=/tmp
MYVAR=persisted_across_calls          # same reused sandbox, no re-create
```

### 6. Injection-safe — the `;` never executes on the host

```text
>> through hook:  echo start; touch /tmp/PWNED_FROM_HOOK_MARKER_15705; echo end; ls -la <marker>
start
end
-rw------- 1 root root 0 Jul 29 03:01 /tmp/PWNED_FROM_HOOK_MARKER_15705   # created INSIDE the sandbox
>> HOST check:  No such file  ->  the ";" never ran on the host  (GOOD)
```

Read-only host-mount rejection was also exercised: when the mount path is outside `allowed_host_mount_prefixes`, the executor emits a warning and **falls back to an isolated sandbox without the mount** (fail-safe), rather than failing the command.

`--reset` destroys the per-session sandbox; `/sandboxes` returns empty afterwards.

---

*Note on environment:* persistence/reconnect was verified with the `cubesandbox` SDK version matching this deployment. The hook's reconnect path is fail-safe — if a cached sandbox cannot be reconnected it transparently creates a fresh one — so please ensure the installed SDK and the target CubeSandbox deployment are on compatible versions for session reuse to take effect.

shsaihdsaiudh:
@fslongjin Requesting a re-review when you have time — your `CHANGES_REQUESTED` from 07-12 was against the old three-mode version, and the PR has since been restructured around exactly what you asked for.

**What changed since you last looked**

You said the run-inside-then-connect approach was "a common example that does not stand out," while "the hook functionality you implemented is really intriguing." So in `c12b117` I deleted the run-inside mode entirely and recentered the example on the `PreToolUse` hook, folding in the hook from #783 (which I closed as superseded). Net effect of that commit was **+471 / −2651** — the PR got substantially smaller, not larger.

The shape now: `cubesandbox_rewrite.py` intercepts the `PreToolUse` Bash payload and rewrites the command to run inside a MicroVM; `cubesandbox_exec.py` reuses one sandbox per `session_id` and persists cwd/exported env between calls. Nothing in `CubeAPI` or any other component is touched — it is purely `examples/` plus the bilingual guide.

**Review threads**

All eight bot threads are addressed. Seven are resolved; I left the `install.sh` heredoc one open because I chose to explain the tradeoff rather than refactor the install path — your call if you want it done before merge.

The last two commits fix the substantive ones: a lock-file leak in the state directory (`reset()` now reaps them, gated on a non-blocking flock so a lock another run is waiting on is never unlinked out from under it), plus exception-type reporting and explicit precedence in the symlink guards.

**Verification**

73 tests pass on Linux, including the two new lock-lifecycle tests. Earlier I posted real single-node cluster results in the gist linked above — host vs guest kernel differing, confirming commands genuinely execute inside the MicroVM rather than on the host, plus the shell-injection cases.

One thing worth your judgement: the guide is filed as a `docs:` PR but roughly 1900 lines are hook code and tests. If you would rather see the docs and the example land separately I am happy to split it — though the guide and the example are fairly coupled, so I left them together.

fslongjin (approved):

PR #705, opened 2026-07-02
lifecycle: Extract auto-pause coordinator into cube-lifecycle-manager

chenhengqi:
Split cube-proxy-sidecar out of the CubeProxy image into a standalone control-plane service (CLM), add Redis-backed service discovery so CubeProxy can run in multiple replicas, and wire it into one-click under cube-sandbox-control.target. Wire protocol, Redis schema, and SDK behavior are unchanged.

PR #767, opened 2026-07-05
cubeproxy: use atomic Lua CAS for lifecycle transitions

novahe:
## Background
Currently, acquiring the lifecycle state lock relies on `SETNX`, which presents three significant issues in highly concurrent and complex state transitions:
1. **Get-and-Set Race Conditions**: Evaluating state before an update using separate operations is inherently racy. Between a `GET` to read the state and a `SET`/`SETNX` to write the new state, another sidecar instance could claim the key, leading to inconsistent state machine transitions.
2. **Wait Latency on State Override**: Existing states (e.g., `running`, `paused`) share the same state key. Under pure `SETNX`, transitioning from `running` to `pausing` is blocked by the current `running` state's TTL. In some scenarios, this causes an idle pause latency of up to one `StateLockTTL` period.
3. **Imprecise Rollback on Failure**: When a CubeMaster RPC (e.g., Pause or Kill) fails, the system struggles to precisely rollback the state to its exact original form because it cannot perceive the original state while acquiring the lock.

## What Changed
This commit atomically refactors CubeProxy's lifecycle state transitions:

1. **Lua CAS Atomic Operation**
   * Deprecates `SETNX` and introduces the `TryTransitionState` Lua script in Redis to perform a Check-And-Swap (CAS).
   * Supports conditional state updates (e.g., allowing a transition to `pausing` only if the state is `missing` or `running`), ensuring atomic transitions and eliminating both the Get-and-Set race condition and override wait times.
2. **Precise Failure Rollback Mechanism**
   * Upon successfully acquiring the lock, the CAS operation now returns the pre-transition original state (`origState`).
   * In the error handling logic within `resumer.go` and `sweeper.go`, if a Pause/Kill/Resume request fails, `origState` is utilized to accurately restore both Redis and Proxy states.
3. **Steady-state TTL Optimization**
   * For long-term stable states like `running`, `paused`, and `killed`, the short anti-deadlock TTL is replaced with a TTL of `0` (persistent). This reduces unnecessary Redis writes and redundant CubeMaster calls.
4. **Resumer Fail-fast Strategy**
   * Introduces a fail-fast mechanism in the Resume flow: if the sandbox is already detected as `killing` or `killed`, the operation is immediately aborted, avoiding futile polling.

PTAL @chenhengqi

### DIFF
--- a/CubeProxy/sidecar/internal/config/config.go
+++ b/CubeProxy/sidecar/internal/config/config.go
@@ -40,8 +40,8 @@ type Config struct {
 	DefaultIdleTimeout time.Duration
 
 	// Loop intervals.
-	StreamReadBlock  time.Duration // XREADGROUP BLOCK arg
-	LastActivePoll   time.Duration // GET /admin/last_active cadence
+	StreamReadBlock   time.Duration // XREADGROUP BLOCK arg
+	LastActivePoll    time.Duration // GET /admin/last_active cadence
 	IdleSweepInterval time.Duration // sweeper cadence
 	// BootstrapWarmup: after sidecar restart, wait this long before pausing
 	// any sandbox that was loaded via HGETALL bootstrap. Lets the
@@ -49,7 +49,7 @@ type Config struct {
 	// that arrive AFTER startup are not affected by this delay.
 	BootstrapWarmup time.Duration
 
-	// Pause/resume locks (SETNX TTL). Long enough to outlive a slow
+	// Pause/resume locks (CAS TTL). Long enough to outlive a slow
 	// CubeMaster RPC, short enough that a crashed sidecar releases the lock.
 	StateLockTTL time.Duration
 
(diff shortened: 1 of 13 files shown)


chenhengqi:
We will land #705 soon, please defer changes to sidecar.

novahe:
> We will land #705 soon, please defer changes to sidecar.

Since my changes modify the lifecycle state management logic (which is being moved in #705), I wanted to check if this PR is still needed?

novahe:
> We will land #705 soon, please defer changes to sidecar.

Could you please review these changes and confirm if they are still necessary in the context of the new modifications?

chenhengqi:
> > We will land #705 soon, please defer changes to sidecar.
> 
> Could you please review these changes and confirm if they are still necessary in the context of the new modifications?

We will split the sidecar to a lifecycle manager and it will run on control node only for simplicity.
I think the change here is not necessary.

PR #768, opened 2026-07-06
fix(examples): align host-mount example paths with default allowed prefix

zyl1121:
## Motivation

After #756, the default host-mount validation only allows host paths under `/data/shared/`.

The dedicated runnable host-mount example still used the old `/tmp/...` paths, so it could fail on a fresh default deployment even though users were following the example as written.

## What Changed

* Update `examples/host-mount/create_with_mount.py` to use `/data/shared/rw` and `/data/shared/ro`.
* Update `examples/host-mount/README.md` so the runnable example, prerequisite commands, and diagram all use the same default prefix.
* Clarify troubleshooting guidance for disallowed hostPath prefixes, read-only sandbox mount paths, and host-directory permission mismatches.

## Scope

This PR only updates the dedicated host-mount example.

## Validation

* Verified the worktree only changes:

  * `examples/host-mount/create_with_mount.py`
  * `examples/host-mount/README.md`
* Verified the updated host-mount example no longer uses `/tmp/rw` or `/tmp/ro`.
* No code tests were run because this PR only updates example and documentation content.

### DIFF
--- a/examples/host-mount/README.md
+++ b/examples/host-mount/README.md
@@ -13,8 +13,8 @@ the standard E2B API: they are requested through the `metadata` field of
 ```
 Cubelet host node
 ┌──────────────────────────────────────────────────────┐
-│  /tmp/rw  ───────────────────► /mnt/rw  (read-write) │
-│  /tmp/ro  ───────────────────► /mnt/ro  (read-only)  │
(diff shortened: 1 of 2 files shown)


ls-ggg:
LGTM

PR #780, opened 2026-07-06
CubeVS: Skip network policy check for established connections

chenhengqi:
This PR fixes connections broken by DNS TTL expiry.

# CubeEgress sessions

DNS response TTL is recorded in allow_out_v2 map, and the userspace DNS reaper periodically removes stale entries. CubeVS relies on these records to enforce egress network policies.

If a DNS record expires (removed by the reaper) while a connection is still open, subsequent packets will be rejected by the policy check, even though the TCP session already exists. This causes active connections to break when the DNS TTL elapses.

Fix this by checking whether a TCP socket already exists for the connection (bpf_skc_lookup_tcp) before enforcing the egress network policy. If a socket is found, the traffic is from an already-established session, so the policy check is skipped and the packet is redirected directly to CubeEgress.

# NAT sessions

Move the egress network policy gate off the sandbox egress fast path and into the shared create_nat_session() helper so TCP, UDP and ICMP all funnel through a single enforcement point at session creation time.

Previously, from_cube() called check_net_policy() on every packet before NAT. That call site is now removed; the check runs once per new NAT session and its result is cached in the session record's lifetime. Packets that hit an existing session are not re-evaluated so that an established connection survived even when DNS record expired.

To preserve the TCP "RST on policy deny" behavior, create_nat_session() receives the skb and communicates the deny reason back to callers via skb->cb[0]. do_tcp_nat() translates NAT_CB_DENIED_BY_POLICY into TCP_NAT_RESET, which the caller already dispatches to tcp_reply_reset(). UDP and ICMP drops stay silent, matching their previous behavior.

should_redirect_to_l7_proxy() no longer takes a plumbed policy_value. It performs its own allow_out_v2 lookup — cheap since the branch only fires for TCP dports 80/443.

### DIFF
--- a/CubeNet/src/cubevs.h
+++ b/CubeNet/src/cubevs.h
@@ -70,6 +70,9 @@
 #define ICMP_CSUM_OFF(LEN)		(sizeof(struct ethhdr) + LEN + offsetof(struct icmphdr, checksum))
 #define ICMP_ECHO_ID_OFF(LEN)		(sizeof(struct ethhdr) + LEN + offsetof(struct icmphdr, un.echo.id))
 
+/* Current network namespace */
+#define BPF_F_CURRENT_NETNS		(-1L)
+
 /* IP and MAC address inside MVMs */
 const volatile __u32 mvm_inner_ip       = 0x0644fea9;	/* 169.254.68.6, network byte order */
 const volatile __u32 mvm_macaddr_p1     = 0xfc6f9020;	/* 20:90:6f:fc:fc:fc */
--- a/CubeNet/src/cubevs.h
+++ b/CubeNet/src/cubevs.h
@@ -242,6 +242,26 @@ struct snat_ip {
 	__u16 reserved;
 };
 
+/* skb->cb[0] is reserved as a per-invocation NAT status word used by
+ * create_nat_session() to communicate the failure reason back to callers
+ * in from_cube(). skb->cb[] is 5 * u32 scratch that survives across
+ * bpf-to-bpf calls within a single program invocation, so this works even
+ * when the session helpers are compiled as subprogs.
+ */
+#define NAT_CB_STATUS_INDEX		0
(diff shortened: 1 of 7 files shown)


kinwin-ustc:
The performance impact of querying the hash table for each package is significant, and in the future, if we need to support update_net_rules to completely block connections to certain domains, how should we do it?

chenhengqi:
> The performance impact of querying the hash table for each package is significant, and in the future, if we need to 

I also hate that we have to do it. Currently this is limit to TCP 80/443 to mitigate the performance degrade.

> support update_net_rules to completely block connections to certain domains, how should we do it?

As far as I can tell:

* We will need to record IP <=> domain mapping in BPF side
* Mark sessions as blocked or delete sessions on update_network() in userspace

ZhengkaiWang (approved):
Thanks! I tested this against our CubeSandbox staging deployment with the DeepSeek streaming workload that reproduced our >60s CubeEgress issue.

Tested binary: `network-agent pr-780 (8bc7936cb9c0a6911e6acb3add52a6eebde9887f)`.

Baseline on our current staging `network-agent v0.4.0 (4004a6ec34a9d045a9789a1fd438d6518eedb3d3)`:
- CubeEgress header-injection path to `api.deepseek.com` ended early at about 50.052s with EOF before `[DONE]` (exit 4).
- Control case bypassing CubeEgress injection, with the API key passed directly as command env, survived 75.053s (exit 0).

With this PR:
- Same CubeEgress injection path; the process inside the sandbox had no API key (`has_guest_key: false`).
- The stream reached 75.617s / 2732 chunks / 889,679 bytes and crossed the >60s threshold successfully (exit 0).
- I also tried a 120s stress run, but that hit the sandbox command control-plane deadline, so I am not counting that as evidence either way.

I restored the staging host back to v0.4.0 after testing. This fixes my reported bug. Approved.

PR #801, opened 2026-07-07
fix(network-agent): retain taps until cleanup and pool prep complete

FakeLearne:
Keep TAP devices out of the reusable pool when sandbox cleanup fails. Port mappings and CubeVS policy metadata must be cleaned successfully before a TAP can be recycled, so stale allow/deny state cannot leak into a later sandbox that reuses the same ifindex.

Move the default private/link-local deny ranges out of the sandbox creation hot path. TAPs now receive those invariant deny_out entries when they are initialized or asynchronously prepared for the pool, while replace/recover paths still replay them after flushing policy maps.

### DIFF
--- a/CubeNet/cubevs/netpolicy.go
+++ b/CubeNet/cubevs/netpolicy.go
@@ -209,10 +209,18 @@ func cleanupNetPolicy(ifindex uint32) error {
 	return flushInnerMap(denyOut, ifindex)
 }
 
-// PrepareTAPDevicePolicy installs policy entries that are invariant for a TAP
-// while it sits in the free pool. Per-sandbox policy application can then skip
-// rewriting these default private/link-local deny ranges on every create.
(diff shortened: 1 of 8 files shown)


CubeNet/cubevs/netpolicy.go
```diff
@@ -36,10 +36,11 @@ type denyOutPolicyEntry struct {
...
 
 type netPolicyPlan struct {
-	allowOutEntries []allowOutPolicyEntry
-	dnsAllowRules   []dnsAllowRule
-	denyOutEntries  []denyOutPolicyEntry
-	dnsPolicyFlags  uint8
+	allowOutEntries       []allowOutPolicyEntry
+	dnsAllowRules         []dnsAllowRule
+	denyOutEntries        []denyOutPolicyEntry
+	includeDefaultDenyOut bool
```
chenhengqi: No need for this bool flag.

PR #802, opened 2026-07-07
fix(sdk/go): align allowOut domain guard with CubeAPI

zyl1121:
## Motivation

The Go SDK currently treats `network.allowPublicTraffic = false` as satisfying the domain `allowOut` guard during sandbox creation.

That does not match the CubeAPI contract. Server-side validation only treats `network.allowInternetAccess = false` or an explicit `network.denyOut: ["0.0.0.0/0"]` as deny-all for domain-based `allowOut` validation.

This was observed on a real deployment when using the Go SDK with `AllowPublicTraffic=false`, `AllowOut=["api.example.com"]`, and default `AllowInternetAccess`. The SDK accepted the request and sent it to CubeAPI, which then rejected it with HTTP 400:

```text
When specifying allowed domains in allow_out, you must disable public outbound traffic or include '0.0.0.0/0' in deny_out to block all other traffic.
```

This PR aligns the SDK's local validation with CubeAPI, so this invalid parameter combination is rejected before the request is sent.

## What Changed

* Stop treating `AllowPublicTraffic = false` as deny-all in Go SDK sandbox-create validation.
* Add focused coverage for the case where domain `allowOut` is used while only `AllowPublicTraffic` is disabled.

## Scope

This PR only updates Go SDK sandbox-create validation and its tests.

It does not change template-build validation or broader network-policy behavior.

## Validation

In `sdk/go`:

```bash
go test -vet=off . -run 'TestCreateSerializesPolicyAndPublicTraffic|TestCreateRejectsAllowOutDomainWithoutDenyAll|TestCreateRejectsAllowOutDomainWhenOnlyAllowPublicTrafficDisabled|TestCreateSendsPythonCompatiblePayload|TestBuildTemplateForwardsCreateFromImageOptions|TestBuildTemplateRequiresImage|TestGetTemplateParsesNetworkFields' -count=1
```

### DIFF
--- a/sdk/go/aligned_test.go
+++ b/sdk/go/aligned_test.go
@@ -30,7 +30,7 @@ func TestCreateSerializesPolicyAndPublicTraffic(t *testing.T) {
 	_, err := client.Create(context.Background(), CreateOptions{
 		Network: NetworkOptions{
 			AllowPublicTraffic: &allowPublic,
-			AllowOut:           []string{"api.github.com"},
+			AllowOut:           []string{"172.67.0.0/16"},
 			Rules: []Rule{{
 				Name:   "gh",
 				Match:  Match{Host: "api.github.com", Scheme: "https"},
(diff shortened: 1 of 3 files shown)


sdk/go/aligned_test.go
```diff
@@ -30,7 +30,7 @@ func TestCreateSerializesPolicyAndPublicTraffic(t *testing.T) {
 	_, err := client.Create(context.Background(), CreateOptions{
 		Network: NetworkOptions{
 			AllowPublicTraffic: &allowPublic,
-			AllowOut:           []string{"api.github.com"},
+			AllowOut:           []string{"172.67.0.0/16"},
```
chenhengqi, line 33: What's the purpose of this change here?

chenhengqi (approved):
The change is correct, we've fix it in Python side already.

sdk/go/aligned_test.go
zyl1121: > What's the purpose of this change here?

This test is meant to cover serialization, not domain `allowOut` validation. After this fix, `api.github.com` with only `AllowPublicTraffic=false` is rejected locally, so I switched it to a CIDR target to keep this test on the valid serialization path. The domain rejection behavior is covered separately by `TestCreateRejectsAllowOutDomainWhenOnlyAllowPublicTrafficDisabled`.

chenhengqi: Why it is rejected?

zyl1121: > Why it is rejected?

Because `api.github.com` is a domain-based `allowOut` target. For domain `allowOut`, CubeAPI requires a deny-all baseline so the allowlist is enforceable. The contract currently treats only `allowInternetAccess=false` or `denyOut` containing `0.0.0.0/0` as that deny-all baseline; `allowPublicTraffic=false` alone is not considered sufficient.

So the old fixture in `TestCreateSerializesPolicyAndPublicTraffic` became invalid after aligning the SDK validation with CubeAPI. Since that test is only meant to verify serialization, I changed the target to a CIDR to avoid triggering the domain guard there. The rejection case is covered separately by another test I added.

PR #810, opened 2026-07-08
feat: sandbox timeout refactor — server-side default + three-value timeout semantics

fslongjin:
## Summary

Implement three-value sandbox timeout semantics (`>0`=TTL, `0`=immediate expiry, `<0`=never timeout), moving the default-timeout decision from SDK/CubeAPI to CubeMaster so "not set" is preserved end-to-end.

## Motivation

Currently every layer in the call chain applies its own default timeout (SDK: 300s → CubeAPI: 15s → CubeMaster: 30s → CLM: 5min), making it impossible to express "never timeout" or to let the server decide the default. This refactor makes the server the single authority for timeout defaults.

### SDK (Go/Python)
- `CreateOptions.Timeout` changed to pointer/optional — `nil`/`None` means "not set" (do not send the field)
- `Create`/`Connect`/`Resume` no longer auto-fill a default timeout
- Expose `NeverTimeout`/`NEVER_TIMEOUT` sentinel (`-1`) for explicit never-timeout requests

### CubeAPI (Rust)
- `NewSandbox`/`ConnectSandbox`/`ResumedSandbox` timeout fields: `i32` → `Option<i32>`
- Remove `serde default_timeout()` fallback; pure passthrough

### CubeMaster (Go)
- `CreateCubeSandboxReq.Timeout`: `int` → `*int` to distinguish "not set" from explicit `0`
- Three-value resolution: not-set→server default, explicit negative→normalize to `-1` (never), explicit `≥0`→passthrough
- Add `default_timeout_insec` (server-side default) and `create_timeout_insec` (RPC deadline) config options
- Decouple create RPC context deadline from sandbox idle TTL
- Lifecycle meta `TimeoutSeconds`: `int` → `*int` for old-data compatibility

### cube-lifecycle-manager
- Sweeper: `<0` → skip recycling; `==0` → immediate recycling; `nil` → fallback to `DefaultIdleTimeout`
- Schema `TimeoutSeconds` updated to `*int` (no `omitempty`)

### Docs
- Update lifecycle docs (en+zh) with the new timeout behavior matrix

## Deployment Order (Important)

1. Upgrade CLM first (sweeper `<0` skip support)
2. Then CubeMaster (sentinel `-1` + config + deadline decouple)
3. Then SDK/CubeAPI ("not set" passthrough)
Rollback in reverse order.

### DIFF
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -171,9 +171,9 @@ pub struct NewSandbox {
     #[serde(rename = "templateID")]
     pub template_id: String,
 
-    #[validate(range(min = 0))]
-    #[serde(default = "default_timeout")]
-    pub timeout: i32,
+    /// Optional idle TTL in seconds. See docs/guide/lifecycle.md.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub timeout: Option<i32>,
 
     /// Sandbox lifecycle configuration. Maps to e2b's `lifecycle` object so
     /// callers that already speak e2b can pass through unchanged. Absent
@@ -213,10 +213,6 @@ pub struct NewSandbox {
 
     #[serde(rename = "volumeMounts", skip_serializing_if = "Option::is_none")]
     pub volume_mounts: Option<Vec<SandboxVolumeMount>>,
-}
-
-fn default_timeout() -> i32 {
-    15
 }
 
 // ─── Sandbox — create / connect response ──────────────────────────────────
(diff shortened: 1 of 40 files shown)


CubeAPI/src/models/mod.rs
```diff
@@ -215,10 +215,6 @@ pub struct NewSandbox {
     pub volume_mounts: Option<Vec<SandboxVolumeMount>>,
 }
 
-fn default_timeout() -> i32 {
-    15
```
chenhengqi, line 219: 🥇

chenhengqi (approved):
Looks good overall.

PR #815, opened 2026-07-08
feat(sdk): add PTY (interactive terminal) support to Go SDK

chaojixinren:
## Summary

Adds PTY (pseudo-terminal) support to the Go SDK (`sdk/go`), closing the
cross-SDK parity gap called out in #811. Go users can now run interactive
programs that need a real TTY — shells/REPLs, full-screen tools (`vim`,
`top`), agent-driven terminals — mirroring the Python (`sandbox.pty`,
#579) and Node (`sandbox.pty`, #792) surface.

## What's added

`sandbox.Pty()` namespace over envd's Connect-JSON RPC:

- **Streaming RPCs** — `Create` (`/process.Process/Start`, runs `/bin/bash
  -i -l`) and `Connect` (`/process.Process/Connect`, reattach by PID).
- **Selector RPCs** — `Kill` (`SendSignal` + `SIGNAL_SIGKILL`),
  `SendStdin` (`SendInput`), `Resize` (`Update`).
- **`PtyHandle`** — streams raw output via `Output()` (channel) or
  `Wait(onData)`; exposes `PID()` / `ExitCode()` / `ErrorMessage()`;
  `Disconnect()` to detach without killing; and per-handle
  `Kill`/`SendStdin`/`Resize`.

Behavior matches Python/Node:

- Seeds `TERM=xterm-256color`, `LANG`/`LC_ALL=C.UTF-8` (caller-overridable).
- `Kill` returns `false` (not an error) on Connect `not_found` / HTTP 404.
- `timeout` is sent to envd as `Connect-Timeout-Ms` **and** enforced
  client-side as an idle abort that resets on every frame (default 60s).
- `Wait` surfaces envd's end-event error (e.g. `signal: killed`) and parses
  exit codes from `exitCode`/`exit_code`/`status`.

## Testing

- **Unit** (`sdk/go/pty_test.go`, `httptest`-mocked envd): create/connect,
  kill (success + 404 + Connect `not_found` body + error), sendStdin,
  resize, wait error propagation, handle→PID delegation, disconnect, and
  idle timeout. `gofmt`/`go vet` clean; `go test -race ./...` green.
- **Live E2E** against a real CubeSandbox: create → `echo` roundtrip →
  resize verified via `stty size` readback (`40 120`) → disconnect +
  reattach-by-PID → kill (`Wait` reports `signal: killed`) → second kill
  reports not-found. All pass.

## Scope

[...]

PR #821, opened 2026-07-08
fix(sdk): route Go SDK envd RPCs to the envd port (49983)

### LINKED ISSUE #816: [Bug Report] Go SDK sends envd data-plane RPCs to the Jupyter port (49999) instead of envd (49983)
## Problem

Every envd data-plane call in the Go SDK (`sdk/go`) is sent to the sandbox's **Jupyter** port (`49999`) instead of the **envd** port (`49983`). Against a real deployment this makes the following fail with a non-2xx error (observed as `HTTP 404 Not Found` or `502 Bad Gateway`, depending on what the Jupyter port fronts in a given template):

- `Commands.Run` (`/process.Process/Start`)
- `Files.Read` / `Write` / `WriteFiles` (`/files`)
- `Filesystem.List` / `Stat` / `Exists` / `MakeDir` / `Rename` / `Remove` / `WatchDir` (`/filesystem.Filesystem/*`)
- `Pty` (`/process.Process/*`, added in #815)

Only `RunCode` (`/execute`) works, because `/execute` genuinely lives on the Jupyter port.

## Root Cause

The SDK only defines the Jupyter port:

```go
// sdk/go/sandbox.go
const JupyterPort = 49999
```

and `newEnvdRequest` — the shared builder for **all** envd RPCs (process, files, filesystem) — hardcodes it as the host port:

```go
// sdk/go/envd.go
func (s *Sandbox) newEnvdRequest(ctx context.Context, method, path string, query url.Values, body io.Reader) (*http.Request, error) {
	target := url.URL{
		Scheme: s.client.config.ProxyScheme,
		Host:   s.GetHost(JupyterPort), // <-- should be the envd port (49983)
		Path:   path,
		...
	}
}
```

CubeProxy routes by the port prefix in the virtual host (`<port>-<sandboxID>.<domain>`), so these requests are forwarded to the Jupyter kernel gateway, which does not serve `/process.Process/*`, `/files`, or `/filesystem.Filesystem/*` → non-2xx (404 or 502).

## Evidence (live run)

Against a real CubeSandbox (`api=http://127.0.0.1:13000`, `proxy=127.0.0.1:11080`, `domain=cube.app`), with the proxy DNS-bypassed via `ProxyNodeIP`.

**1) Existing `Commands.Run` integration test — 404 on port 49999 (template `tpl-58f2b99af8684ca48973a61b`):**

```
=== RUN   TestIntegrationSandboxExecutionCommandsFilesAndErrors
    integration_test.go:57: CubeSandbox integration target api=http://127.0.0.1:13000 template=tpl-58f2b99af8684ca48973a61b proxy=127.0.0.1:11080 domain=cube.app
    integration_test.go:150: Commands.Run returned error: Not Found (HTTP 404)
--- FAIL: TestIntegrationSandboxExecutionCommandsFilesAndErrors (2.62s)
```

(Note: `RunCode` at line 90 — `/execute` on 49999 — succeeded; the failure is the first `/process.Process/*` call.)

**2) PTY E2E — same 404 on port 49999:**

```
=== RUN   TestIntegrationPtyLifecycle
    pty_integration_test.go:60: CubeSandbox integration target api=http://127.0.0.1:13000 template=tpl-58f2b99af8684ca48973a61b proxy=127.0.0.1:11080 domain=cube.app
    pty_integration_test.go:80: pty.Create returned error: Not Found (HTTP 404)
--- FAIL: TestIntegrationPtyLifecycle (1.40s)
```

**3) After pointing `newEnvdRequest` at the envd port `49983` — full PTY lifecycle passes (create → echo → resize/`stty size` → reattach → kill):**

```
=== RUN   TestIntegrationPtyLifecycle
    pty_integration_test.go:60: CubeSandbox integration target api=http://127.0.0.1:13000 template=tpl-58f2b99af8684ca48973a61b proxy=127.0.0.1:11080 domain=cube.app
    pty_integration_test.go:82: PTY started, pid=27
    pty_integration_test.go:131: reattached.Wait -> code=-1 err=PTY exited with error: signal: killed exitKnown=true endErr="signal: killed"
    pty_integration_test.go:141: second Kill correctly reported not-found (false)
--- PASS: TestIntegrationPtyLifecycle (1.74s)
PASS
ok  	github.com/tencentcloud/CubeSandbox/sdk/go	1.742s
```

Same `api`/`proxy`/`template`; the **only** change between runs #2 and #3 was `GetHost(JupyterPort)` → `GetHost(49983)` in `newEnvdRequest`.

**4) Same-sandbox A/B (raw HTTP, only the host port differs), template `tpl-d635607121a94ef5b9413339`:**

```
GET /files @49999 -> HTTP 502 (openresty "Bad Gateway")
GET /files @49983 -> HTTP 200, body = /etc/hostname contents
```

Identical request; flipping the virtual-host port prefix from `49999` to `49983` is the only change, and it flips failure → success. (This template returns `502` where the one in runs #1–#3 returned `404` — the failure code depends on the Jupyter upstream, but the wrong-port routing is the same bug.)

## Cross-SDK parity

Python and Node both keep the two ports separate:

- Python: `ENVD_PORT = 49983` (`_commands.py`) for process/filesystem; `JUPYTER_PORT = 49999` (`sandbox.py`) for `/execute`.
- Node: `ENVD_PORT = 49983` (`commands.ts`) for process/pty; `JUPYTER_PORT = 49999` (`sandbox.ts`) for `/execute`.

The architecture docs also state envd is health-probed on `:49983`.

## Impact

All data-plane envd functionality of the Go SDK (commands, files, filesystem, and PTY) is unusable against real deployments. Existing unit tests do not catch this: they route through a single `httptest` server via `ProxyNodeIP`, so the virtual host's port prefix never selects a backend and the wrong port is never exercised.

## Proposed Solution

1. Add `const EnvdPort = 49983`.
2. In `newEnvdRequest`, use `s.GetHost(EnvdPort)` (process/files/filesystem).
3. Leave `RunCode`'s `/execute` on `JupyterPort` (49999).
4. Add an integration-tagged E2E (or extend the existing one) that actually drives `Commands.Run`/`Files`/`Pty` against a live sandbox so the port is covered.

This is a one-line routing fix; every `newEnvdRequest`-based API (including the PTY added in #815) recovers automatically.

chaojixinren:
## Summary

Fixes #816. Every envd data-plane RPC in the Go SDK was routed to the sandbox's **Jupyter** port (`49999`) instead of the **envd** port (`49983`), so `Commands.Run`, `Files.*`, `Filesystem.*`, and the `Pty` from #815 all failed against real deployments (non-2xx — `404` or `502` depending on the template's Jupyter upstream). Only `RunCode` (`/execute`) worked, since `/execute` genuinely lives on the Jupyter port.

## Root cause

`newEnvdRequest` — the shared request builder for **all** envd RPCs — hardcoded `s.GetHost(JupyterPort)`. CubeProxy routes by the `<port>-<sandboxID>.<domain>` host prefix, so these requests reached the Jupyter kernel gateway, which does not serve `/process.Process/*`, `/files`, or `/filesystem.Filesystem/*`.

## Changes

- Add a dedicated `EnvdPort = 49983` constant (`sandbox.go`).
- Use `s.GetHost(EnvdPort)` in `newEnvdRequest` (`envd.go`); `RunCode`'s `/execute` stays on `JupyterPort`.
- Update the affected unit-test host assertions (`sdk_test.go`) and the README proxy example.

This matches the Python (`ENVD_PORT`) and Node SDKs, which keep the two ports separate. It's effectively a one-line routing fix — every `newEnvdRequest`-based API (including the PTY in #815) recovers automatically.

## Testing

- `gofmt` / `go vet ./...` clean; `go test -race ./...` green.
- **Live E2E** against a real CubeSandbox (`api=127.0.0.1:13000`, `proxy=127.0.0.1:11080`, `domain=cube.app`): `/files` read/write and `/filesystem.Filesystem/*` (mkdir/list/stat) now succeed on `49983`. Same-sandbox A/B confirms the routing: identical `GET /files` returns non-2xx on `49999` but `200` on `49983` — the port prefix is the only difference.

## Scope

Routing fix only. A separate, independent framing bug on the streaming `/process.Process/Start` request (its body isn't Connect-enveloped in `startProcess`, unlike Python and the Go PTY) is tracked separately and will be its own PR; it does not affect `/files` or `/filesystem.Filesystem/*`.

### DIFF
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -198,10 +198,14 @@ sb, err := client.Create(ctx, cubesandbox.CreateOptions{
 When `CUBE_PROXY_NODE_IP` is set, data-plane requests connect directly to that IP and port while preserving the virtual sandbox host:
 
 ```text
-URL:  <CUBE_PROXY_SCHEME>://49999-<sandboxID>.<CUBE_SANDBOX_DOMAIN>/<envd-endpoint>
+URL:  <CUBE_PROXY_SCHEME>://49983-<sandboxID>.<CUBE_SANDBOX_DOMAIN>/<envd-endpoint>
 TCP:  <CUBE_PROXY_NODE_IP>:<CUBE_PROXY_PORT_HTTP>
-Host: 49999-<sandboxID>.<CUBE_SANDBOX_DOMAIN>
+Host: 49983-<sandboxID>.<CUBE_SANDBOX_DOMAIN>
 ```
+
+The host prefix is the sandbox's internal service port: `49983` (envd) for
+commands/filesystem/files/PTY, and `49999` (Jupyter) for the code interpreter
+(`RunCode` / `/execute`).
 
 You can also set it directly:
 
--- a/sdk/go/envd.go
+++ b/sdk/go/envd.go
@@ -225,7 +225,7 @@ func multipartFileBody(path string, data []byte) (io.Reader, string, error) {
 func (s *Sandbox) newEnvdRequest(ctx context.Context, method, path string, query url.Values, body io.Reader) (*http.Request, error) {
 	target := url.URL{
 		Scheme:   s.client.config.ProxyScheme,
-		Host:     s.GetHost(JupyterPort),
+		Host:     s.GetHost(EnvdPort),
 		Path:     path,
 		RawQuery: query.Encode(),
 	}
--- a/sdk/go/sandbox.go
+++ b/sdk/go/sandbox.go
@@ -13,7 +13,15 @@ import (
 	"time"
 )
 
-const JupyterPort = 49999
+const (
+	// JupyterPort is the code-interpreter (Jupyter) port, used only by the
+	// /execute endpoint (RunCode).
+	JupyterPort = 49999
+	// EnvdPort is the envd daemon port. All envd data-plane RPCs — commands,
+	// filesystem, files, and PTY — route here, matching the Python/Node SDKs
+	// (ENVD_PORT = 49983). Sending these to JupyterPort yields 404s.
+	EnvdPort = 49983
+)
 
 func (s *Sandbox) GetHost(port int) string {
 	domain := s.Domain
--- a/sdk/go/sdk_test.go
+++ b/sdk/go/sdk_test.go
@@ -655,7 +655,7 @@ func TestCommandsRunUsesEnvdProcessStart(t *testing.T) {
 		t.Fatalf("Run: %v", err)
 	}
 
-	if gotHost != "49999-sb-proc.cube.test" {
+	if gotHost != "49983-sb-proc.cube.test" {
 		t.Fatalf("Host=%q", gotHost)
 	}
 	if gotHeaders.Get("Content-Type") != connectContentType || gotHeaders.Get("Connect-Protocol-Version") != connectProtocolVersion {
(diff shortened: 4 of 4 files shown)


ls-ggg:
LGTM  @tinklone @wbzdssm cc

chenhengqi (approved):

PR #828, opened 2026-07-08
fix(sdk): Connect-envelope process.Process/Start and parse exit-0 end…

### LINKED ISSUE #825: [Bug Report] Go SDK sends /process.Process/Start request body without a Connect envelope
## Summary

The Go SDK sends the streaming `/process.Process/Start` request body as raw JSON without a Connect envelope, so envd rejects every `Commands.Run` (and `Pty`) call with a protocol framing error.

## Environment

- CubeSandbox version / commit: `master` @ `79434dc` (`sdk/go`)
- Host OS and kernel version: Linux `5.15.0-124-generic`
- KVM info (`modinfo kvm`): N/A — client-side SDK bug (reproduces against any live deployment)
- Deployment mode: single-node dev (`cube-devenv`)
- Relevant component: Go SDK (`sdk/go`) — envd data-plane client

## Steps to Reproduce

> Note: this is currently masked by #816 (all envd RPCs 404/502 on the wrong port). To reach envd you must either apply the #816 port fix (`49983`) first, or target the envd port directly.

1. Create a real sandbox and ensure envd RPCs reach the **envd** port (`49983`).
2. Call `sb.Commands().Run(ctx, "echo hi", ...)` — i.e. a `POST /process.Process/Start` on the envd port.
3. envd returns a Connect end-stream error instead of a process stream.

Isolating the single variable (Connect framing) with two raw HTTP requests to the **same** endpoint on the **same** port (`49983`), one with a raw-JSON body and one wrapped in a 5-byte Connect envelope:

## Expected Behavior

`/process.Process/Start` is a server-streaming Connect RPC, so the request body must be a 5-byte length-prefixed enveloped message (as Python and the Go PTY already do). It should return a stream of process events (`start` → `data` → `end`).

## Actual Behavior

`startProcess` posts the payload as **raw JSON** while advertising `Content-Type: application/connect+json`, and omits `Connect-Content-Encoding: identity`. envd reads the first 5 bytes of the JSON (`{"pr…`) as a frame header, decodes a bogus length, and aborts.

Live A/B against a real CubeSandbox (`api=127.0.0.1:13000`, `proxy=127.0.0.1:11080`, `domain=cube.app`), identical endpoint/port, only the body framing differs:

```
[FRAME] raw JSON      -> {"error":{"code":"invalid_argument","message":"protocol error: promised 577794671 bytes in enveloped message, got 50 bytes"}}

[FRAME] Connect frame -> {"event":{"start":{"pid":11}}}
                         {"event":{"data":{"stdout":"aGkK"}}}
                         {"event":{"end":{"exited":true,"status":"exit status 0"}}}
```

Raw JSON → envd's protocol framing error. The 5-byte-enveloped request runs cleanly: `pid=11`, `stdout="aGkK"` (base64 of `hi\n`), `exit status 0`. The envelope is the only difference — confirming the framing as the sole root cause.

## Root Cause

```go
// sdk/go/envd.go — startProcess
req, err := s.newEnvdRequest(ctx, http.MethodPost, "/process.Process/Start",
    nil, bytes.NewReader(raw)) // <-- raw JSON, not a Connect envelope
req.Header.Set("Content-Type", connectContentType) // application/connect+json
req.Header.Set("Connect-Protocol-Version", connectProtocolVersion)
// missing: Connect-Content-Encoding: identity
```

The response side is already parsed as Connect envelopes (`parseProcessStartStream` → `readConnectEnvelope`), so only the request side is inconsistent.

## Cross-SDK parity

- **Python** (`_commands.py`): `content=_encode_connect_envelope(json.dumps(payload)…)` + `Connect-Content-Encoding: identity`.
- **Go PTY** (`pty.go`, #815): already wraps via `encodeConnectEnvelope(raw)` + `identity` — correct.
- Only Go's `startProcess` (commands) is missing it.

## Additional Context

- Distinct from #816 (that's the wrong-port routing bug). Both sit on the same code path and were surfaced by the same live run; `/files` and `/filesystem.Filesystem/*` (plain octet-stream / unary JSON) are unaffected by this framing bug.
- Not caught by unit tests: the `httptest` mocks don't validate request framing.
- **Proposed fix:** in `startProcess`, wrap the body with `encodeConnectEnvelope(raw)` and set `Connect-Content-Encoding: identity`, matching Python and the Go PTY path.

chaojixinren:
## Summary

Two fixes on the Go SDK's `Commands.Run` (`/process.Process/Start`) path, both
required to run a process end-to-end against a real deployment:

1. **Request framing (#825)** — `startProcess` sent the streaming request body as
   raw JSON while advertising `Content-Type: application/connect+json`. A
   server-streaming Connect RPC requires a 5-byte length-prefixed enveloped
   message, so envd read the leading JSON bytes (`{"pr…`) as a frame header,
   decoded a bogus length, and aborted. Now wrapped via a new
   `encodeConnectEnvelope` helper + `Connect-Content-Encoding: identity`,
   matching the Python SDK and the Go PTY path. `watchDir` reuses the same
   helper (previously inlined identical framing).

2. **Exit-0 end-event parsing** — the end event is proto3 JSON, which omits a
   zero-valued `exitCode` entirely, so a successful (exit 0) process arrives as
   only `{"exited":true,"status":"exit status 0"}` and `parseProcessStartStream`
   failed with `process EndEvent missing exit code`. Now recovers the code from
   the `status` string and falls back to the `exited` flag, so exit-0 commands
   (the common case) no longer spuriously fail.

## Changes

- `sdk/go/envd.go`
  - Add `encodeConnectEnvelope`; use it in `startProcess` + set
    `Connect-Content-Encoding: identity`; refactor `watchDir` to reuse it.
  - `processEndEvent.exitCode()`: fall back to parsing `status` (`exit status N`)
    then the `exited` flag when `exitCode`/`exit_code` are absent.
- `sdk/go/sdk_test.go`
  - `TestCommandsRunUsesEnvdProcessStart`: assert the request body is a valid
    Connect envelope and `Connect-Content-Encoding: identity`.
  - Add `TestProcessEndEventExitCode` (table-driven) and
    `TestCommandsRunExitZeroStatusOnly` (exit-0 stream reproduction).

## Root cause (framing) — live A/B, same endpoint/port, only body framing differs

```
[FRAME] raw JSON      -> {"error":{"code":"invalid_argument","message":"protocol error: promised 577794671 bytes in enveloped message, got 50 bytes"}}

[FRAME] Connect frame -> {"event":{"start":{"pid":11}}}
                         {"event":{"data":{"stdout":"aGkK"}}}
                         {"event":{"end":{"exited":true,"status":"exit status 0"}}}
```

Raw JSON → envd's protocol framing error. The 5-byte-enveloped request runs
cleanly (`pid=11`, `stdout=aGkK` = base64 of `hi\n`, `exit status 0`). The
envelope is the only variable.

## Testing

- `go vet ./...` clean; `go test ./...` green (unit tests, incl. the new
  envelope + exit-0 coverage).
- **Live E2E** against a real CubeSandbox (`api=127.0.0.1:13000`,
  `proxy=127.0.0.1:11080`, `domain=cube.app`), with the #816 envd-port fix
  applied so requests reach envd (`49983`):

Framing fix — `/process.Process/Start` streams cleanly:

```
=== RUN   TestIntegrationProcessConnectEnvelope
OK  /process.Process/Start streamed cleanly -> stdout="envd-ok\n" stderr="e\n" exit=3
--- PASS: TestIntegrationProcessConnectEnvelope
```

Exit-0 parsing fix — exit 0 and exit 3 both return correctly:

```
=== RUN   TestIntegrationCommandsExitCodes
OK  exit 0 -> stdout="line-1\nline-2\nline-3\n" exit=0
OK  exit 3 -> stdout="bye\n" exit=3
--- PASS: TestIntegrationCommandsExitCodes (0.76s)
PASS
```

## Notes

- Depends on #816 (envd-port routing) to be exercisable end-to-end; the two are
  distinct bugs on the same code path. Unit tests do not catch either
  (`httptest` mocks never validate request framing and always include a nonzero
  `exitCode`).
- Response parsing was already Connect-aware (`parseProcessStartStream` →
  `readConnectEnvelope`); only the request side and the exit-0 case were wrong.

Closes #825

### DIFF
--- a/sdk/go/envd.go
+++ b/sdk/go/envd.go
@@ -102,12 +102,13 @@ func (s *Sandbox) startProcess(ctx context.Context, payload processStartRequest,
 		return nil, err
 	}
 
-	req, err := s.newEnvdRequest(ctx, http.MethodPost, "/process.Process/Start", nil, bytes.NewReader(raw))
+	req, err := s.newEnvdRequest(ctx, http.MethodPost, "/process.Process/Start", nil, encodeConnectEnvelope(raw))
 	if err != nil {
 		return nil, err
 	}
 	req.Header.Set("Content-Type", connectContentType)
 	req.Header.Set("Connect-Protocol-Version", connectProtocolVersion)
+	req.Header.Set("Connect-Content-Encoding", "identity")
 	req.Header.Set("Authorization", basicAuthUser(opts.User))
 	setConnectTimeout(req, opts.Timeout)
 
@@ -347,6 +348,18 @@ func readConnectEnvelope(r io.Reader) (byte, []byte, error) {
 	return header[0], payload, nil
 }
 
+// encodeConnectEnvelope wraps payload in a Connect streaming message: a 5-byte
+// header (1 flag byte + big-endian uint32 length) followed by the payload. Both
+// request and response messages on a streaming Connect RPC use this framing.
+func encodeConnectEnvelope(payload []byte) io.Reader {
+	buf := bytes.NewBuffer(make([]byte, 0, 5+len(payload)))
+	var header [5]byte
+	binary.BigEndian.PutUint32(header[1:], uint32(len(payload)))
+	buf.Write(header[:])
+	buf.Write(payload)
+	return buf
+}
+
 func parseConnectEndStream(raw []byte) error {
 	if len(raw) == 0 {
 		return nil
@@ -539,15 +552,8 @@ func (s *Sandbox) watchDir(ctx context.Context, path string) (*Watcher, error) {
 		return nil, err
 	}
 
-	var envelope bytes.Buffer
-	var header [5]byte
-	header[0] = 0
-	binary.BigEndian.PutUint32(header[1:], uint32(len(payload)))
-	envelope.Write(header[:])
-	envelope.Write(payload)
-
 	streamCtx, cancel := context.WithCancel(ctx)
-	req, err := s.newEnvdRequest(streamCtx, http.MethodPost, "/filesystem.Filesystem/WatchDir", nil, &envelope)
+	req, err := s.newEnvdRequest(streamCtx, http.MethodPost, "/filesystem.Filesystem/WatchDir", nil, encodeConnectEnvelope(payload))
 	if err != nil {
 		cancel()
 		return nil, err
@@ -643,6 +649,19 @@ func (e *processEndEvent) exitCode() (int, bool) {
 	}
 	if e.ExitCodeSnake != nil {
 		return *e.ExitCodeSnake, true
+	}
+	// envd serializes the end event as proto3 JSON, which omits a zero-valued
+	// exitCode field entirely. A successful (exit 0) process therefore arrives
+	// with no exitCode key at all — only status="exit status 0" and
+	// exited=true. Recover the code from the status string, then fall back to
+	// the exited flag so exit-0 commands don't spuriously fail.
+	if s := strings.TrimSpace(e.Status); strings.HasPrefix(s, "exit status ") {
+		if code, err := strconv.Atoi(strings.TrimSpace(strings.TrimPrefix(s, "exit status "))); err == nil {
+			return code, true
+		}
+	}
+	if e.Exited {
+		return 0, true
 	}
 	return 0, false
 }
--- a/sdk/go/sdk_test.go
+++ b/sdk/go/sdk_test.go
@@ -10,6 +10,7 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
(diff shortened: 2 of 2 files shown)


wbzdssm:
Overall: fixes are correct and well-tested. A few notes:

1.  watchDir missing  Connect-Content-Encoding: identity
 startProcess sets this header after wrapping with  encodeConnectEnvelope, but  watchDir doesn't, even though it now uses the same helper. Both are server-streaming Connect RPCs — the request framing should be consistent.

2. Signal-killed exit code
 "signal: killed" falls back to exit code  0 via the  exited flag, but the conventional mapping is  128 + signal_number. The Python SDK already handles this. Worth checking what envd actually puts in  status for signal-terminated processes.

3. Nit:  encodeConnectEnvelope returns  *bytes.Buffer instead of  io.Reader
No bug, but returning the concrete type leaks an implementation detail.

Issue #228, opened 2026-05-13
【Feature Request】Support pause / resume for sandboxes with host-mount


kinwin-ustc:
### Discussed in https://github.com/TencentCloud/CubeSandbox/discussions/209

<div type='discussions-op-text'>

<sup>Originally posted by **ggrp-china** May 12, 2026</sup>
## Background
Cube Sandbox currently provides two capabilities:
- `host-mount`: mount a host directory into the sandbox; (see #106 for the observed behavior).
- `pause` / `resume`: pause a sandbox to release resources, and resume it later.
However, these two capabilities are not supported together 
## Request
We'd like sandboxes with `host-mount` to also support `pause` / `resume` properly. The current workaround is to destroy and recreate the sandbox, but that loses all in-sandbox processes and runtime state, which defeats the purpose of pause / resume.</div>

ls-ggg:
CubeSandbox already supports pause/resume for host-mount sandboxes in the latest version

PR #834, opened 2026-07-09
feat(shim): support pause/resume for host-mount sandboxes (#228)

sparkzky:
Host-mount sandboxes (native virtiofsd) could not survive pause/resume: the virtio-fs device-state migration is not enabled in this branch, so the pause snapshot's fs device state cannot be trusted. Route A rebuilds the FsConfig vec from the in-memory config (conf.virtiofs is preserved across pause since the shim process is not restarted) and passes it via RestoreConfig.fs to VmResumeFromSnapshot. The VMM then spawns fresh virtiofsd backends at the shareDirs retained on the Cubelet node, reusing the same vm_restore path already exercised by cold-restore (restore_vm) and the integration test test_snapshot_restore_from_config - zero new VMM code.

- Extract build_fs_configs() shared by restore_vm (cold-restore) and build_pause_resume_restore_config (pause/resume), removing duplication.
- resume_vm_with_config None branch now constructs RestoreConfig{fs:Some} and calls resume_vm_cube_with_config.
- When no virtio-fs is configured, fs:None matches the old resume_vm_cube behavior byte-for-byte (no regression).
- 3 unit tests cover host-mount / no-virtiofs(no-regression) / combined virtiofs+block-fs.

Refs #228
Assisted-by: OhMyPi:zai/glm-5.2

### DIFF
--- a/CubeShim/shim/src/sandbox/sb.rs
+++ b/CubeShim/shim/src/sandbox/sb.rs
@@ -862,12 +862,7 @@ impl SandBox {
         );
         infof!(self.log, "snapshot dir:{}", snapshot.clone());
         let restore_memory_vol_url = self.conf.snapshot_memory_vol_url.clone();
-        let mut fss = vec![];
-        if let Some(fs) = self.conf.fs.as_ref() {
-            let f = Utils::restore_fs_configs(fs);
-            fss.push(f);
-        }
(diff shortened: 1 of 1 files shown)


ls-ggg:
Thanks for the contribution! CubeSandbox already supports pause/resume for host-mount sandboxes in the latest version

PR #679, opened 2026-06-30
feat(cubeproxy): add plaintext gRPC ingress on port 9090

xiaojunxiang2023:
## Summary

Add a plaintext HTTP/2 gRPC ingress listener on CubeProxy (default port `9090`) so clients can reach sandbox gRPC services without wildcard DNS or TLS on CubeProxy.

Clients dial the CubeProxy IP and identify the target sandbox via the gRPC `:authority` pseudo-header, using the same `<container_port>-<sandbox_id>` format as host-based HTTP routing.

## Changes

| Area | What |
|------|------|
| `CubeProxy/nginx.conf` | Add `grpc_backend` upstream and HTTP/2 server block on port `9090` |
| `CubeProxy/Dockerfile` | Expose port `9090` |
| One-click deploy | `CUBE_PROXY_GRPC_PORT` templating, startup/postcheck, `env.example` |
| TKE | CLB/service exposure and security group for port `9090` |
| Docs | gRPC access mode in `https-and-domain.md`; port/config in `network-hardening.md`, `self-build-deploy.md`, `service-management.md` |
| Tests | Extend `test_runtime_file_safety.sh` for gRPC postcheck port |

## Motivation

CubeProxy already supports sandbox ingress via:

- **Host mode**: `<container_port>-<sandbox_id>.<domain>` on ports 80/443
- **Path mode**: `/sandbox/<sandbox_id>/<container_port>/...`

gRPC clients often cannot rely on wildcard DNS or custom TLS. A dedicated plaintext HTTP/2 listener lets them connect directly to the CubeProxy IP.

## Usage

```
dial: <cube-proxy-ip>:9090
:authority: <container_port>-<sandbox_id>
```

Example:

```
dial: 10.0.0.5:9090
:authority: 49983-abc123
```

See `docs/guide/https-and-domain.md` for details.

## Configuration

| Variable | Default | Description |
|----------|---------|-------------|
| `CUBE_PROXY_GRPC_PORT` | `9090` | Plaintext gRPC (HTTP/2) listen port in one-click deploy |

PR #837, opened 2026-07-09
feat(cubeproxy): add plaintext gRPC ingress on port 9090

xiaojunxiang2023:
## Summary

Add a plaintext HTTP/2 gRPC ingress listener on CubeProxy (default port `9090`) so clients can reach sandbox gRPC services without wildcard DNS or TLS on CubeProxy.

Clients dial the CubeProxy IP and identify the target sandbox via the gRPC `:authority` pseudo-header, using the same `<container_port>-<sandbox_id>` format as host-based HTTP routing.

This supersedes the closed #679, rebased on current `master` after #705 (cube-lifecycle-manager). Review feedback from #679 is addressed:

- Port `9090` opened on the **CLB** security group (not jumpserver)
- `grpc_set_header` for client IP forwarding
- `specify-protocol` includes `9090` in both public and internal CLB modes

Note on streaming / buffering: OpenResty `1.21.4` (nginx `1.21.4`) has no `grpc_buffering` directive (`nginx -t` reports `unknown directive`). `ngx_http_grpc_module` forwards responses to the client as they are received; there is nothing to toggle here.

## Changes

| Area | What |
|------|------|
| `CubeProxy/nginx.conf` | Add `grpc_backend` upstream and HTTP/2 server block on port `9090` |
| `CubeProxy/Dockerfile` | Expose port `9090` |
| One-click deploy | `CUBE_PROXY_GRPC_PORT` templating, startup/postcheck, `env.example` |
| TKE | CLB/service exposure and security group for port `9090` |
| Docs | gRPC access mode in `https-and-domain.md`; port/config in `network-hardening.md`, `self-build-deploy.md`, `service-management.md` |
| Tests | Extend `test_runtime_file_safety.sh` for gRPC postcheck port |
| Examples | `examples/grpc-ingress/` for native gRPC client usage |

## Motivation

CubeProxy already supports sandbox ingress via:

- **Host mode**: `<container_port>-<sandbox_id>.<domain>` on ports 80/443
- **Path mode**: `/sandbox/<sandbox_id>/<container_port>/...`

gRPC clients often cannot rely on wildcard DNS or custom TLS. A dedicated plaintext HTTP/2 listener lets them connect directly to the CubeProxy IP.

## Usage

```
dial: <cube-proxy-ip>:9090
:authority: <container_port>-<sandbox_id>
```

Example:

```
dial: 10.0.0.5:9090
:authority: 50051-abc123
```

(`:authority` must target a **native gRPC** service port in the sandbox. envd on `49983` speaks Connect/HTTP, not native gRPC.)

See `docs/guide/https-and-domain.md` and `examples/grpc-ingress/` for details.

## Configuration

| Variable | Default | Description |
|----------|---------|-------------|
| `CUBE_PROXY_GRPC_PORT` | `9090` | Plaintext gRPC (HTTP/2) listen port in one-click deploy |

## Test plan

- [x] `bash deploy/one-click/tests/test_runtime_file_safety.sh` (local macOS) — **PASS**
- [x] Same test suite on ubuntu22 test host — **PASS**
- [x] Real-cluster unary RPC via CubeProxy `:9090` (native gRPC echo in sandbox) — happy path OK; unknown sandbox → `UNAVAILABLE`; bad `:authority` → `INVALID_ARGUMENT`

Manual verification (operator):

1. Deploy with one-click; confirm `ss -lnt` shows `9090` on cube-proxy node
2. Run a native gRPC client against `:9090` with `:authority=<port>-<sandbox_id>` (see `examples/grpc-ingress/`)
3. (TKE) Confirm CLB exposes port `9090` and security group allows inbound TCP/9090


xiaojunxiang2023:
@chenhengqi  hi~, #705 is merged — I've rebased the gRPC ingress work from #679 onto master and opened a new PR. Same feature, CLM conflicts fixed, plus the review items from #679. Would appreciate a look when you can.

CubeProxy/nginx.conf
```diff
@@ -361,6 +370,53 @@ http {
...
 
+    server {
+        listen 9090 http2 reuseport;
+        server_name _;
+        set $cube_proxy_host_ip "";
+        set $ins_id "";
+        set $container_port "";
+        set $backend_ip "";
+        set $backend_port "";
+        set $host_proxy_port 9090;
```
chenhengqi: I think we don't need `$host_proxy_port`.

CubeProxy/nginx.conf
```diff
@@ -361,6 +370,53 @@ http {
...
+            proxy_pass http://$cube_sidecar_addr/internal/resume?$args;
+        }
+
+        location / {
+            include /usr/local/openresty/nginx/conf/global/global.conf;
+
+            grpc_connect_timeout 3s;
+            grpc_read_timeout 7206s;
+            grpc_send_timeout 7206s;
+            grpc_buffering off;
```
chenhengqi: Why disable buffering?

CubeProxy/nginx.conf
```diff
@@ -97,6 +97,15 @@ http {
         keepalive_timeout 80;
     }
 
+    # Plaintext gRPC ingress. Clients dial this IP:port and set :authority to
+    # "<container_port>-<sandbox_id>" (no DNS required).
+    upstream grpc_backend {
+        server 0.0.0.1:1235;
```
chenhengqi, line 105: It seems like the `server` instruction can be omited.

CubeProxy/nginx.conf
xiaojunxiang2023: Thanks! Removed `$host_proxy_port` from the gRPC server block — it isn't consumed on that path. Force-pushed in b819f55.

CubeProxy/nginx.conf
xiaojunxiang2023: Good catch — removed `grpc_buffering off`.

That directive isn't available in our OpenResty 1.21.4 (nginx 1.21.4) build; the gRPC module already keeps buffering disabled by default (`upstream.buffering = 0`), so the line was both unnecessary and would fail config load.

Note: the `process.Process/Start` locations use HTTP `proxy_pass` with `proxy_buffering off`, not the gRPC path — different module.

Force-pushed in bbcccba.

CubeProxy/nginx.conf
xiaojunxiang2023: The `server 0.0.0.1:1235` entry is a dummy placeholder required by the upstream block syntax. The real backend peer is chosen dynamically in `balancer_by_lua` via `balancer.set_current_peer()` — same pattern as the HTTP `backend` upstream (`0.0.0.1:1234`). Removing it would make the upstream block invalid.

xiaojunxiang2023:
@chenhengqi Addressed review feedback — removed unused `$host_proxy_port` from the gRPC server block (b819f55). Kept `grpc_buffering off` for server-streaming RPCs and the dummy upstream `server` entry for `balancer_by_lua`. Happy to adjust if you see a cleaner approach.

xiaojunxiang2023:
> ## Review: PR #837 — Plaintext gRPC Ingress on Port 9090
> Overall this is a well-structured PR. The core implementation (nginx config, Terraform plumbing, shell-script readiness probes, test coverage, and documentation) is thorough and consistent. I have a few findings, including **one correctness bug** that should be addressed before merging.
> 
> ### 1. (Critical) Missing grpc_buffering off in nginx config
> **File:** CubeProxy/nginx.conf lines 399-416 The PR description says "grpc_buffering off for streaming RPCs" was addressed from the #679 review, but this directive does not appear in the config. nginx grpc_buffering defaults to on, buffering the entire response body before sending. Unary RPCs work fine but server-streaming/bidirectional streaming RPCs break - the client receives zero data until the stream completes. **Fix:** Add grpc_buffering off inside the location / block before grpc_pass.
> 
> ### 2. (High) Docs say 403, code returns 404 for token violations
> **Files:** docs/guide/restrict-public-access.md and docs/zh/guide/restrict-public-access.md The error-model table says traffic token violations return 403, but sandbox_backend.lua calls utils:respond_not_found() which returns 404 (deliberately, to avoid leaking sandbox existence). **Fix:** Update both docs to say 404 with a footnote explaining the deliberate choice.
> 
> ### 3. (Medium) Public plaintext gRPC with no TLS alternative
> **File:** deploy/one-click/terraform/tencentcloud/main.tf lines 389-395 When enable_public_network=true, TCP/9090 is open to 0.0.0.0/0 with no encryption. Unlike port 80 (which has 443/HTTPS), port 9090 has no TLS alternative - all gRPC frames including traffic access tokens are in cleartext. **Consideration:** Add a TLS-enabled gRPC port or document that this port is for trusted networks only.
> 
> ### 4. (Medium) Missing security warning in example README
> **File:** examples/grpc-ingress/README.md Describes plaintext gRPC but has no warning about using grpc.insecure_channel() over public networks. **Suggestion:** Add a note that port 9090 uses plaintext gRPC and should only be used over trusted networks.
> 
> ### 5. (Low) Auto-pause gating on gRPC port not documented
> **Files:** docs/guide/https-and-domain.md and docs/zh/guide/https-and-domain.md The gRPC block runs rewrite_phase.lua which invokes state.gate() - auto-pause/resume lifecycle applies but is not mentioned. **Suggestion:** Document that lifecycle gating applies transparently.
> 
> ### 6. (Low) Missing test: custom gRPC port with default HTTP port
> **File:** deploy/one-click/tests/test_runtime_file_safety.sh Test matrix covers default+default, custom HTTP+default gRPC, and custom HTTP+custom gRPC but not default HTTP+custom gRPC.
> 
> Summary: Findings 1 and 2 are the most actionable - one is a correctness bug for streaming RPCs, the other is a documentation inaccuracy that would mislead operators debugging access-control issues.

Thanks for the auto-review — addressed the actionable items in 6069af4.

1. `grpc_buffering off` — not added back. That directive is not available in our OpenResty 1.21.4 (nginx 1.21.4) build and would fail config load. The gRPC module already hardcodes `upstream.buffering = 0`; buffering is not on by default for `grpc_pass`.

2. 403 vs 404 — fixed both `restrict-public-access` docs (EN/ZH) to match `sandbox_backend.lua`: token failures return **404** by design (same as unknown sandbox, to avoid existence leaks).

3. Public plaintext 9090 — by design for clients that cannot use TLS/SNI wildcard DNS. Clarified in the gRPC example READMEs that this port is plaintext and should only be used over trusted networks (or behind a TLS terminator).

4. Example security warning — added.

5–6 (auto-pause docs / extra test matrix) — leaving for a follow-up unless you'd like them in this PR.

deploy/one-click/terraform/tencentcloud/main.tf
```diff
@@ -386,6 +386,14 @@ resource "tencentcloud_security_group_rule_set" "clb" {
     description = "Allow CLB HTTPS (cube-proxy)"
   }
 
+  ingress {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Thanks — this is intentional and mirrors how ports 80/443 are gated by `enable_public_network`.

Port 9090 exists for clients that cannot use TLS / wildcard SNI DNS. Operators who enable public network already accept exposing CubeProxy ingress; we document the plaintext risk in the gRPC example READMEs (trusted networks / TLS terminator in front).

A separate `enable_public_grpc` knob would be a follow-up if operators need asymmetric exposure; keeping the same boolean for this PR to avoid extra deploy surface.

examples/grpc-ingress/env_utils.py
```diff
@@ -0,0 +1,22 @@
...
+    ]
+
+    seen_paths = set()
+    for path in candidate_paths:
+        resolved_path = path.resolve()
+        if resolved_path in seen_paths:
+            continue
+        seen_paths.add(resolved_path)
+
+        if path.is_file():
```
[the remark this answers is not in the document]
xiaojunxiang2023: Good catch on the broken-symlink edge case. This helper is only used by the minimal example to load `.env` from the script dir or cwd; for the common paths we care about, `path` and `path.resolve()` agree.

Happy to tighten it in a follow-up if we expand the example; leaving as-is for now to keep the demo tiny.

docs/guide/network-hardening.md
```diff
@@ -182,6 +182,7 @@ sudo ufw allow from 10.0.0.0/24 to any port 12088 proto tcp  # WebUI
 sudo ufw allow 22/tcp     # SSH
 sudo ufw allow 80/tcp     # cube-proxy (public if needed)
 sudo ufw allow 443/tcp    # cube-proxy TLS
+sudo ufw allow 9090/tcp   # cube-proxy plaintext gRPC
 sudo ufw enable
 ```
```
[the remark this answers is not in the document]
xiaojunxiang2023: Thanks — updated the Option B ufw comment for port 9090 to call out that it is plaintext and should be restricted to trusted networks when opened publicly (EN/ZH). Force-pushed in 78db4a0.

CubeProxy/nginx.conf
```diff
@@ -361,6 +370,51 @@ http {
...
     }
 
+    server {
+        listen 9090 http2 reuseport;
+        server_name _;
+        set $cube_proxy_host_ip "";
+        set $ins_id "";
+        set $container_port "";
+        set $backend_ip "";
+        set $backend_port "";
```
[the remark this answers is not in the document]
xiaojunxiang2023: Intentionally omitted — @chenhengqi asked to drop `$host_proxy_port` from the gRPC server block, and nothing in the Lua phases / access log currently reads it on that path.

Re-adding it only for “future consistency” would undo that review feedback. If we later start consuming `$host_proxy_port`, we can set it in the gRPC block then.

CubeProxy/nginx.conf
```diff
@@ -361,6 +370,51 @@ http {
...
+            proxy_read_timeout 30s;
+            proxy_pass http://$cube_sidecar_addr/internal/resume?$args;
+        }
+
+        location / {
+            include /usr/local/openresty/nginx/conf/global/global.conf;
+
+            grpc_connect_timeout 3s;
+            grpc_read_timeout 7206s;
+            grpc_send_timeout 7206s;
```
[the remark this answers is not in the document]
xiaojunxiang2023: Noted. HTTP has `proxy_next_upstream` at the http level; those directives don't apply to `grpc_pass`.

For this path the Lua balancer already picks the peer via `balancer.set_current_peer()`. Transient connect failures are rare after that, and gRPC clients typically retry themselves. Skipping `grpc_next_upstream*` for now to keep the gRPC location minimal; we can add them later if we see real need in production.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
```diff
@@ -1083,7 +1088,7 @@ resource "kubernetes_service" "cube_proxy" {
     # Public mode: a public CLB billed by traffic (internet-charge-type).
```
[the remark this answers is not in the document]
xiaojunxiang2023: Good catch — fixed.

`local.cubeproxy_nginx_conf` now replaces `__CUBE_PROXY_GRPC_PORT__` → `9090`, and `create.sh`'s template generator also emits/checks that token (aligned with the release-bundle sed). Force-pushed in 8adaa7a.

CubeProxy/nginx.conf
```diff
@@ -361,6 +370,51 @@ http {
         }
     }
 
+    server {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Reproduced on our base image `openresty/1.21.4.1`: Reproduced on our base image `openresty/1.21.4.1`:

`strings` on the nginx binary shows `grpc_buffer_size` but not `grpc_buffering`. nginx 1.13.10 introduced the gRPC *module* (`grpc_pass`), not a `grpc_buffering` directive — that name isn't in the OSS `ngx_http_grpc_module` command table we checked (1.21.4 / 1.25.1 / master). The module hardcodes `upstream.buffering = 0`.

So we cannot add the directive without breaking CubeProxy startup. Happy to walk through any specific streaming failure if you see one in practice.

CubeProxy/nginx.conf
```diff
@@ -361,6 +370,51 @@ http {
...
+        location / {
+            include /usr/local/openresty/nginx/conf/global/global.conf;
+
+            grpc_connect_timeout 3s;
+            grpc_read_timeout 7206s;
+            grpc_send_timeout 7206s;
+            grpc_set_header X-Real-IP $remote_addr;
+            grpc_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+
+            rewrite_by_lua_file lua/rewrite_phase.lua;
```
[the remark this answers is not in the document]
xiaojunxiang2023: Agreed this is a real UX gap for native gRPC clients: routing failures currently reuse the HTTP JSON helpers (`respond_not_found` / etc.), which aren't valid gRPC status frames.

I'd treat proper `grpc-status` mapping as a follow-up rather than expand this ingress PR — same Lua phases are shared with HTTP today. Open to doing it next if you want that tracked.

tinklone:
LGTM

xiaojunxiang2023:
@fslongjin @tinklone hi~, If there are no further comments, could you help merge this PR? I've had to resolve merge conflicts multiple times because of other merged PRs.

CubeProxy/nginx.conf
```diff
@@ -368,6 +377,55 @@ http {
...
     }
 
+    server {
+        listen 9090 http2 reuseport;
+        server_name _;
+        set $cube_proxy_host_ip "";
+        set $ins_id "";
+        set $container_port "";
+        set $backend_ip "";
+        set $backend_port "";
```
kinwin-ustc, line 389: If access_time is not set, an error will occur in the subsequent log_phase.lua file

CubeProxy/nginx.conf
```diff
@@ -368,6 +377,55 @@ http {
...
+        # Required by rewrite_phase.lua → request_host.apply (maskRequestHost).
+        set $cube_upstream_host $http_host;
+        # Auto-pause / auto-resume wiring; mirrors the 8081 server. See that block.
+        set $cube_sidecar_addr "127.0.0.1:8083";
```
kinwin-ustc, line 400: The auto pause/resume service has been removed from cube-proxy, should this also be modified here?

kinwin-ustc (commented):
Should gRPC error responses follow the gRPC specification itself? Have you tested this PR in a real-world cluster deployment?

xiaojunxiang2023:
@kinwin-ustc I did test the initial version with a real deployment. However, after some changes were made based on community discussions, I haven't validated it through an actual deployment again. I'll perform another real deployment and run a regression test afterward.

CubeProxy/nginx.conf
xiaojunxiang2023: Thanks — fixed. The 9090 server now declares `set $access_time "";` (same contract as 8080/8081) so `log_phase.lua` can assign it. Also declared on the admin server to silence the same uninitialized-variable warnings there.

Force-pushed in 282b0e1.

CubeProxy/nginx.conf
xiaojunxiang2023: Good question. The in-process cube-proxy-sidecar was replaced by `cube-lifecycle-manager`, but `$cube_sidecar_addr` / `location = /_sidecar_resume` are still the dataplane resume hook — chart/entrypoint wires them to the LCM Service address (same as the 8080/8081 servers). So this block stays on purpose; it is not leftover sidecar code.

I also added the missing `include .../global.conf` on the 9090 `_sidecar_resume` location so operator overrides of `$cube_sidecar_addr` apply, matching 8081.

Force-pushed in 282b0e1.

CubeProxy/nginx.conf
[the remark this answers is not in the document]
xiaojunxiang2023: Follow-up done in 282b0e1: routing failures on the plaintext gRPC ingress now emit `grpc-status` / `grpc-message` instead of JSON.

xiaojunxiang2023:
@kinwin-ustc Follow-up on the real-cluster validation (after the `grpc-status` / `grpc-message` fix in 282b0e1).

### Cluster setup
- Single-node K3s, CubeProxy image built from this PR branch with plaintext gRPC ingress on `:9090`
- Template built from a gateway-client base image with a native gRPC echo service baked in on port `50051` (envd still on `49983`)

### Results (unary RPC through CubeProxy `:9090`)

| Case | `:authority` | Result |
|------|----------------|--------|
| Happy path | `50051-<sandbox_id>` | **OK** — `/cube.e2e.Echo/Ping` echoed `b'hello-pr837'`; access log upstream `10.42.0.30:20007` HTTP 200 |
| Unknown sandbox | `50051-does-not-exist` | `UNAVAILABLE` / `unavailable` (Redis miss) |
| Bad authority | `badhost` | `INVALID_ARGUMENT` / `bad request` |

Also verified envd health over HTTP `:80` (`Host: 49983-<sandbox_id>` → HTTP 204) on the same sandbox.

So routing failures now return gRPC status trailers as discussed, and a real unary RPC through `:9090` succeeds end-to-end against a native gRPC backend in the sandbox.

Note: dialing envd (`49983`) over `:9090` is **not** a valid happy-path check — envd speaks Connect/HTTP, not native gRPC. The example in this PR only asserts channel `READY` to the proxy; the table above is the actual ingress E2E.

CubeProxy/lua/sandbox_state.lua
```diff
@@ -62,7 +62,7 @@ function _M.gate(ins_id)
             string.format("request %s sandbox %s is %s; returning 410",
                 ngx.var.http_x_cube_request_id or "-", ins_id, state))
         ngx.var.cube_retcode = "310410"
```
kinwin-ustc: dead code？

xiaojunxiang2023: Not dead code — still live.

`cube-lifecycle-manager`'s sweeper pushes `killing` / `killed` into CubeProxy's `cube_sandbox_state` dict (via `ProxyPush.SetState`) so in-flight requests fail fast with 410 instead of hanging on a sandbox being torn down. See `cube-lifecycle-manager/internal/sweeper/sweeper.go` around the kill path.

This PR only changes `ngx.exit(410)` → `utils:respond_gone()` so the same gate returns `grpc-status` / `grpc-message` on the plaintext gRPC ingress, instead of a bare HTTP 410 that breaks native gRPC clients.

kinwin-ustc: cube_retcode has been deprecated, perhaps we can delete it by the way

xiaojunxiang2023: Good catch — leftover from before #653 removed `$cube_retcode` / `X-Cube-Retcode`. Dropped the assignment in 490580e.

CubeProxy/lua/utils.lua
```diff
@@ -42,18 +60,42 @@ function _M.is_null(self, str)
     return str == nil or str == ""
 end
 
+-- True on the dedicated plaintext gRPC server, or when the client already
+-- advertised application/grpc (covers Content-Type-based detection).
+function _M.is_grpc_request(self)
+    if ngx.var.cube_ingress_protocol == "grpc" then
+        return true
+    end
+    local ct = ngx.var.http_content_type or ""
```
kinwin-ustc: If a request with Content-type: application/grpc is sent to port 80/443, it may enter the grpc processing logic, but many variables that grpc depends on are not defined in 80/443, leading to internal errors. Perhaps we can simplify this here

xiaojunxiang2023: Agreed — Content-Type sniffing on :80/:443 was unsafe (`@grpc_lua_error` / `$cube_grpc_*` only exist on the 9090 server).

Simplified `is_grpc_request()` to only check `$cube_ingress_protocol == "grpc"` (set exclusively on the plaintext gRPC listener). Force-pushed in 2f0c309.

kinwin-ustc (approved):

fslongjin (approved):

PR #861, opened 2026-07-09
fix(actions): guard claude-response on explicit /cubebot trigger

### LINKED ISSUE #844: [Bug Report] AI Assistant workflow fails on ordinary PR comments/reviews without CubeBot trigger
## Summary

The `AI Assistant / claude-response` workflow can produce false failing checks on PRs when ordinary comments or reviews are submitted without explicitly requesting CubeBot.

## Environment

* CubeSandbox version / commit: current `master`
* Relevant component: GitHub Actions workflow / AI Assistant / CubeBot

## Steps to Reproduce

1. Open a PR in this repository.
2. Submit a normal PR review or PR comment without requesting cubebot.
3. Observe that the `AI Assistant / claude-response` workflow can still be triggered.
4. If the actor does not have write permission to the repository, the workflow fails during the Claude action permission check.

## Expected Behavior

Ordinary PR comments or reviews that do not explicitly request CubeBot should not produce a failing workflow check.

If CubeBot is not requested, the workflow job should be skipped before running the Claude action.

For example:

* PR comment/review without CubeBot trigger: skipped
* Maintainer explicitly invokes CubeBot: workflow runs
* User without write access: skipped before entering the action, instead of producing a red failed check

## Actual Behavior

A normal PR review/comment can trigger the workflow even when CubeBot was not requested. The Claude action then detects that no trigger phrase was matched, checks the actor permission, and fails if the actor only has read access.

This can create a misleading red check on the PR, even though the PR itself did not fail any code/test validation and did not request CubeBot.

Example log:

```text
No trigger was met for /cubebot
Auto-detected mode: agent for event: pull_request_review
Using provided GITHUB_TOKEN for authentication
Checking permissions for actor: zyl1121
Permission level retrieved: read
Warning: Actor has insufficient permissions: read
Error: Action failed with error: Actor does not have write permissions to the repository
Error: Process completed with exit code 1.
```

## Additional Context

The Claude action's internal write-access permission check appears to be correct, users without write access should not be allowed to run the agent.

The issue is that the GitHub Actions workflow itself is triggered too broadly, so ordinary PR comments/reviews can start the workflow job before the Claude action rejects the event. This results in false red checks on PRs.

A possible fix is to add a job-level if guard so the workflow only runs when the comment/review body explicitly contains the intended CubeBot trigger.

A possible fix is to add a job-level if guard so the workflow only runs when the comment/review body explicitly contains the intended CubeBot trigger. The guard could match the current `/cubebot` trigger, or use `@cubebot` if that is the preferred invocation style. It could also include an actor association check, such as `OWNER`, `MEMBER`, or `COLLABORATOR`, so unauthorized users are skipped before the Claude action starts.

Could you confirm the preferred CubeBot trigger phrase and whether an actor association check should be added at the workflow level?

zyl1121:
## Motivation

The `AI Assistant / claude-response` workflow currently starts on ordinary PR comments and reviews even when CubeBot was not explicitly requested.

In those cases, `anthropics/claude-code-action` reports that no trigger phrase was found and can still fail during actor permission checks, creating false red checks on PRs.

## What Changed

* Add a job-level guard so `claude-response` only runs when the event body contains `/cubebot`.
* Limit the guarded execution path to trusted GitHub `author_association` values.
* Keep the guard aligned with the action's existing `/cubebot` `trigger_phrase`, so the workflow does not start for comments the action would later reject as non-triggers.
* Preserve the existing `/cubebot` trigger contract instead of broadening trigger handling.

## Scope

This PR only updates `.github/workflows/claude-response.yml`.

It does not change the existing `/cubebot` trigger phrase, and it does not add broader aliases such as `@cubebot`.

## Related

Fixes #844.

### DIFF
--- a/.github/workflows/claude-response.yml
+++ b/.github/workflows/claude-response.yml
@@ -11,6 +11,24 @@ on:
 
 jobs:
   claude-response:
+    if: |
+      (
+        github.event_name == 'issue_comment' &&
+        contains(github.event.comment.body || '', '/cubebot') &&
+        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
+      ) || (
+        github.event_name == 'pull_request_review_comment' &&
+        contains(github.event.comment.body || '', '/cubebot') &&
+        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.comment.author_association)
+      ) || (
+        github.event_name == 'pull_request_review' &&
+        contains(github.event.review.body || '', '/cubebot') &&
+        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.review.author_association)
+      ) || (
+        github.event_name == 'issues' &&
+        contains(github.event.issue.body || '', '/cubebot') &&
+        contains(fromJSON('["OWNER","MEMBER","COLLABORATOR"]'), github.event.issue.author_association)
+      )
     runs-on: ubuntu-latest
     permissions:
       contents: read


kinwin-ustc:
@fslongjin take a look at this review workflow change

fslongjin (approved):

PR #862, opened 2026-07-09
fix(cubemaster): sync timeout metadata after resume

emailcannotbeblank:
## Summary

`UpdateRequest` had no `timeout` field, so the timeout CubeAPI forwards with a
resume request was silently dropped — a resumed sandbox always kept its
pre-pause timeout. This PR carries it through, with regression tests and
lifecycle docs.

## Changes

- `types.UpdateRequest`: new optional `Timeout *int`; `nil` preserves the stored timeout.
- `Update()` (resume only):
  - rejects `timeout < -1` with `MasterParamsError` — same policy as `SetTimeout`
  - after a successful resume, publishes an explicit timeout to lifecycle metadata via `refreshTimeoutMeta`
  - best-effort by design: metadata failures are logged at error level and never alter the resume response
- CubeMaster unit tests for validation, `nil`/`0`/`-1`/`N`, success-only publish, and provider failure
- `sdk_compat` e2e: `resume(timeout=-1)`, `resume(timeout=-2)` → 400, `resume(timeout=0)` preserves TTL
- English and Chinese lifecycle docs: per-API timeout tables

## Behavior

| resume `timeout` | before | after |
|---|---|---|
| omitted / `nil` | dropped | preserve stored timeout |
| `0` | dropped | preserve stored timeout (immediate expiry is `set_timeout(0)`-only) |
| `-1` | dropped | never-timeout |
| `N > 0` | dropped | N-second window from now |
| `< -1` | dropped | `MasterParamsError` |

Publishing runs only after the resume RPC succeeds.

## Out of scope

This PR does **not** implement:

- refresh never-timeout / omitted-duration cluster default
- CubeAPI inbound `Validate` on `ResumedSandbox` / `ConnectSandbox`
- E2B `endAt` far-future sentinel

## Test plan

- [x] `go test -short -gcflags=all=-l ./pkg/service/sandbox/`
- [ ] `make cubemaster-test` / CI Unit Test Check
- [ ] `tests/e2e/sdk_compat/cases/lifecycle/test_resume_timeout.py` against a live cluster (`CUBE_TEMPLATE_ID` required)
- [ ] Format Check

Assisted-by: Cursor:Grok-4.6

### DIFF
--- a/CubeAPI/src/cubemaster/mod.rs
+++ b/CubeAPI/src/cubemaster/mod.rs
@@ -1397,7 +1397,8 @@ pub struct SandboxUpdateRequest {
     /// "pause" | "resume"
     #[serde(rename = "action")]
     pub action: String,
-    /// TTL in seconds (for resume; 0 = keep original). Optional for pause.
+    /// TTL in seconds for resume. None keeps the existing timeout; 0 requests
+    /// immediate timeout; -1 disables timeout.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub timeout: Option<i32>,
 }
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -540,6 +540,9 @@ pub async fn refresh_sandbox(
     Path(sandbox_id): Path<String>,
     Json(body): Json<RefreshRequest>,
 ) -> AppResult<impl IntoResponse> {
+    body.validate()
+        .map_err(|e| AppError::BadRequest(e.to_string()))?;
+
     let duration = body.duration.unwrap_or(0);
     state
         .logger
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -4,22 +4,68 @@
 
 use axum::{
     extract::{Path, Query, State},
-    http::StatusCode,
+    http::{header::USER_AGENT, HeaderMap, StatusCode},
     response::IntoResponse,
     Json,
 };
+use chrono::{DateTime, Utc};
 use validator::Validate;
 
 use crate::{
     error::{AppError, AppResult},
     logging::{LogEvent, LogLevel},
     models::{
-        ApiError, ConnectSandbox, ListSandboxesQuery, ListSandboxesV2Query, NewSandbox,
-        RefreshRequest, ResumedSandbox, Sandbox, SandboxDetail, SandboxLogsQuery,
+        ApiError, ConnectSandbox, ListSandboxesQuery, ListSandboxesV2Query, ListedSandbox,
+        NewSandbox, RefreshRequest, ResumedSandbox, Sandbox, SandboxDetail, SandboxLogsQuery,
         SandboxLogsV2Query, SandboxLogsV2Response, SetTimeoutRequest,
     },
     state::AppState,
 };
+
+// CubeSandbox represents a never-timeout sandbox as `end_at=None`, which the
+// public Cube response serializes by omitting `endAt`. The E2B Python SDK's
+// generated SandboxDetail/ListedSandbox models require `endAt` to be a valid
+// datetime string, so E2B SDK callers get a far-future sentinel at the handler
+// boundary while Cube-native callers keep the no-deadline shape.
+const E2B_NEVER_TIMEOUT_END_AT_RFC3339: &str = "9999-12-31T23:59:59Z";
+
+fn is_e2b_sdk_request(headers: &HeaderMap) -> bool {
+    // E2B SDKs identify API requests through User-Agent. Keep this compatibility
+    // branch narrow so curl, Cube SDKs, and other clients preserve Cube's
+    // native omitted-endAt semantics for never-timeout sandboxes.
+    headers
+        .get(USER_AGENT)
+        .and_then(|value| value.to_str().ok())
+        .map(|value| {
+            let user_agent = value.to_ascii_lowercase();
+            user_agent.contains("e2b-python-sdk/") || user_agent.contains("e2b-code-interpreter/")
+        })
+        .unwrap_or(false)
+}
+
+fn e2b_never_timeout_end_at() -> DateTime<Utc> {
+    DateTime::parse_from_rfc3339(E2B_NEVER_TIMEOUT_END_AT_RFC3339)
+        .expect("E2B never-timeout sentinel must be valid RFC3339")
+        .with_timezone(&Utc)
+}
+
+fn apply_e2b_sandbox_detail_compat(headers: &HeaderMap, detail: &mut SandboxDetail) {
+    if is_e2b_sdk_request(headers) && detail.end_at.is_none() {
+        detail.end_at = Some(e2b_never_timeout_end_at());
+    }
+}
+
+fn apply_e2b_listed_sandbox_compat(headers: &HeaderMap, sandboxes: &mut [ListedSandbox]) {
+    if !is_e2b_sdk_request(headers) {
+        return;
+    }
+
+    for sandbox in sandboxes {
+        if sandbox.end_at.is_none() {
+            sandbox.end_at = Some(e2b_never_timeout_end_at());
+        }
+    }
+}
 
 // ─── GET /sandboxes ───────────────────────────────────────────────────────────
 
@@ -34,6 +80,7 @@ use crate::{
 )]
 pub async fn list_sandboxes(
     State(state): State<AppState>,
+    headers: HeaderMap,
     Query(params): Query<ListSandboxesQuery>,
 ) -> AppResult<impl IntoResponse> {
     state
@@ -51,7 +98,8 @@ pub async fn list_sandboxes(
         .list(params.metadata.as_deref(), None, 200)
         .await
     {
-        Ok(list) => {
+        Ok(mut list) => {
+            apply_e2b_listed_sandbox_compat(&headers, &mut list);
             state
                 .logger
                 .log(
@@ -91,6 +139,7 @@ pub async fn list_sandboxes(
 )]
 pub async fn list_sandboxes_v2(
     State(state): State<AppState>,
+    headers: HeaderMap,
     Query(params): Query<ListSandboxesV2Query>,
 ) -> AppResult<impl IntoResponse> {
     state
@@ -103,7 +152,7 @@ pub async fn list_sandboxes_v2(
         )
         .await;
 
-    let list = state
+    let mut list = state
         .services
         .sandboxes
         .list(
@@ -112,6 +161,7 @@ pub async fn list_sandboxes_v2(
             params.limit,
         )
         .await?;
+    apply_e2b_listed_sandbox_compat(&headers, &mut list);
 
     state
         .logger
@@ -140,6 +190,7 @@ pub async fn list_sandboxes_v2(
 )]
 pub async fn get_sandbox(
     State(state): State<AppState>,
+    headers: HeaderMap,
     Path(sandbox_id): Path<String>,
 ) -> AppResult<impl IntoResponse> {
     state
@@ -151,7 +202,8 @@ pub async fn get_sandbox(
         )
         .await;
 
-    let detail = state.services.sandboxes.get_sandbox(&sandbox_id).await?;
+    let mut detail = state.services.sandboxes.get_sandbox(&sandbox_id).await?;
+    apply_e2b_sandbox_detail_compat(&headers, &mut detail);
     state
         .logger
         .log(
@@ -574,4 +626,114 @@ pub async fn refresh_sandbox(
         )
         .await;
     Ok(StatusCode::NO_CONTENT)
+}
+
+#[cfg(test)]
+mod tests {
+    use super::{
+        apply_e2b_listed_sandbox_compat, apply_e2b_sandbox_detail_compat,
+        E2B_NEVER_TIMEOUT_END_AT_RFC3339, USER_AGENT,
+    };
+    use crate::models::{ListedSandbox, SandboxDetail, SandboxState};
+    use axum::http::{HeaderMap, HeaderValue};
+    use chrono::TimeZone;
+
+    fn sandbox_detail_without_end_at() -> SandboxDetail {
+        SandboxDetail {
+            template_id: "tpl-test".to_string(),
+            alias: None,
+            sandbox_id: "sbx-test".to_string(),
+            client_id: "node-test".to_string(),
+            started_at: chrono::Utc
+                .timestamp_opt(1_800_000_000, 0)
+                .single()
+                .expect("valid timestamp"),
+            end_at: None,
+            envd_version: "0.5.11".to_string(),
+            envd_access_token: None,
+            domain: Some("cube.app".to_string()),
+            cpu_count: 2,
+            memory_mb: 2000,
+            disk_size_mb: Some(0),
+            metadata: None,
+            state: SandboxState::Running,
+            volume_mounts: None,
+        }
+    }
+
+    fn listed_sandbox_without_end_at() -> ListedSandbox {
+        ListedSandbox {
+            template_id: "tpl-test".to_string(),
+            alias: None,
+            sandbox_id: "sbx-test".to_string(),
+            client_id: "node-test".to_string(),
+            started_at: chrono::Utc
+                .timestamp_opt(1_800_000_000, 0)
+                .single()
+                .expect("valid timestamp"),
+            end_at: None,
+            cpu_count: 2,
+            memory_mb: 2000,
+            disk_size_mb: Some(0),
+            metadata: None,
+            state: SandboxState::Running,
+            envd_version: "0.5.11".to_string(),
+            volume_mounts: None,
+        }
+    }
+
+    fn e2b_headers() -> HeaderMap {
+        let mut headers = HeaderMap::new();
+        headers.insert(
+            USER_AGENT,
+            HeaderValue::from_static("e2b-python-sdk/2.34.0"),
+        );
+        headers
+    }
+
+    #[test]
+    fn e2b_sandbox_detail_compat_adds_end_at_for_never_timeout() {
+        let mut detail = sandbox_detail_without_end_at();
+
+        apply_e2b_sandbox_detail_compat(&e2b_headers(), &mut detail);
+
+        let json = serde_json::to_value(detail).expect("detail should serialize");
+        assert_eq!(json["endAt"], E2B_NEVER_TIMEOUT_END_AT_RFC3339);
+    }
+
+    #[test]
+    fn non_e2b_sandbox_detail_keeps_end_at_omitted_for_never_timeout() {
+        let mut detail = sandbox_detail_without_end_at();
+
+        apply_e2b_sandbox_detail_compat(&HeaderMap::new(), &mut detail);
+
+        let json = serde_json::to_value(detail).expect("detail should serialize");
+        assert!(!json
+            .as_object()
+            .expect("detail should serialize as object")
+            .contains_key("endAt"));
+    }
+
+    #[test]
+    fn e2b_listed_sandbox_compat_adds_end_at_for_never_timeout() {
+        let mut sandboxes = vec![listed_sandbox_without_end_at()];
+
+        apply_e2b_listed_sandbox_compat(&e2b_headers(), &mut sandboxes);
+
+        let json = serde_json::to_value(&sandboxes[0]).expect("listed sandbox should serialize");
+        assert_eq!(json["endAt"], E2B_NEVER_TIMEOUT_END_AT_RFC3339);
+    }
+
+    #[test]
+    fn non_e2b_listed_sandbox_keeps_end_at_omitted_for_never_timeout() {
+        let mut sandboxes = vec![listed_sandbox_without_end_at()];
+
+        apply_e2b_listed_sandbox_compat(&HeaderMap::new(), &mut sandboxes);
+
+        let json = serde_json::to_value(&sandboxes[0]).expect("listed sandbox should serialize");
+        assert!(!json
+            .as_object()
+            .expect("listed sandbox should serialize as object")
+            .contains_key("endAt"));
+    }
 }
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -357,6 +357,8 @@ pub async fn connect_sandbox(
     Path(sandbox_id): Path<String>,
     Json(body): Json<ConnectSandbox>,
 ) -> AppResult<impl IntoResponse> {
+    body.validate()
+        .map_err(|e| AppError::BadRequest(e.to_string()))?;
     state
         .logger
         .log(
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -310,6 +310,8 @@ pub async fn resume_sandbox(
     Path(sandbox_id): Path<String>,
     Json(body): Json<ResumedSandbox>,
 ) -> AppResult<impl IntoResponse> {
+    body.validate()
+        .map_err(|e| AppError::BadRequest(e.to_string()))?;
     state
         .logger
         .log(
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -329,11 +329,12 @@ pub struct SandboxDetail {
 // ─── Sandbox — pause/resume/connect/snapshot ──────────────────────────────
 
 /// Request body for POST /sandboxes/{id}/resume (deprecated).
-#[derive(Debug, Deserialize, ToSchema)]
+#[derive(Debug, Deserialize, Validate, ToSchema)]
 #[allow(dead_code)]
 pub struct ResumedSandbox {
     /// Idle timeout in seconds; None when the client did not send one.
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[validate(custom(function = "validate_timeout_value"))]
     pub timeout: Option<i32>,
     #[serde(rename = "autoPause", default)]
     pub auto_pause: bool,
@@ -513,6 +514,8 @@ fn validate_timeout_value(timeout: i32) -> Result<(), validator::ValidationError
 /// Request body for POST /sandboxes/{id}/refreshes
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct RefreshRequest {
+    /// Refresh duration in seconds. If omitted, CubeAPI treats it as `0`,
+    /// which means immediate timeout. Use `-1` for never-timeout.
     #[validate(range(min = -1, max = 3600))]
     pub duration: Option<i32>,
 }
@@ -546,8 +549,8 @@ fn default_page_limit() -> i32 {
 #[cfg(test)]
 mod tests {
     use super::{
-        CreateTemplateRequest, NewSandbox, RefreshRequest, SandboxNetworkConfig, SetTimeoutRequest,
-        TemplateAliasLookupResponse,
+        CreateTemplateRequest, NewSandbox, RefreshRequest, ResumedSandbox, SandboxNetworkConfig,
+        SetTimeoutRequest, TemplateAliasLookupResponse,
     };
     use validator::Validate;
 
@@ -601,6 +604,31 @@ mod tests {
                 "duration={duration} should be rejected"
             );
         }
+    }
+
+    #[test]
+    fn resumed_sandbox_accepts_omitted_never_zero_and_positive_timeout() {
+        for timeout in [None, Some(-1), Some(0), Some(60)] {
+            let req = ResumedSandbox {
+                timeout,
+                auto_pause: false,
+            };
+            req.validate()
+                .unwrap_or_else(|e| panic!("resume timeout={timeout:?} should be valid: {e}"));
+        }
+    }
+
+    #[test]
+    fn resumed_sandbox_rejects_invalid_negative_timeout() {
+        let req = ResumedSandbox {
+            timeout: Some(-2),
+            auto_pause: false,
+        };
+
+        assert!(
+            req.validate().is_err(),
+            "resume timeout=-2 should be rejected"
+        );
     }
 
     #[test]
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -345,6 +345,7 @@ pub struct ResumedSandbox {
 pub struct ConnectSandbox {
     /// Idle timeout in seconds; None when the client did not send one.
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[validate(custom(function = "validate_timeout_value"))]
     pub timeout: Option<i32>,
 }
 
@@ -549,8 +550,8 @@ fn default_page_limit() -> i32 {
 #[cfg(test)]
 mod tests {
     use super::{
-        CreateTemplateRequest, NewSandbox, RefreshRequest, ResumedSandbox, SandboxNetworkConfig,
-        SetTimeoutRequest, TemplateAliasLookupResponse,
+        ConnectSandbox, CreateTemplateRequest, NewSandbox, RefreshRequest, ResumedSandbox,
+        SandboxNetworkConfig, SetTimeoutRequest, TemplateAliasLookupResponse,
     };
     use validator::Validate;
 
@@ -628,6 +629,21 @@ mod tests {
         assert!(
             req.validate().is_err(),
             "resume timeout=-2 should be rejected"
+        );
+    }
+
+    #[test]
+    fn connect_sandbox_validates_timeout_like_resume() {
+        for timeout in [None, Some(-1), Some(0), Some(60)] {
+            let req = ConnectSandbox { timeout };
+            req.validate()
+                .unwrap_or_else(|e| panic!("connect timeout={timeout:?} should be valid: {e}"));
+        }
+
+        let req = ConnectSandbox { timeout: Some(-2) };
+        assert!(
+            req.validate().is_err(),
+            "connect timeout=-2 should be rejected"
         );
     }
 
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -513,7 +513,7 @@ fn validate_timeout_value(timeout: i32) -> Result<(), validator::ValidationError
 /// Request body for POST /sandboxes/{id}/refreshes
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct RefreshRequest {
-    #[validate(range(min = 0, max = 3600))]
+    #[validate(range(min = -1, max = 3600))]
     pub duration: Option<i32>,
 }
 
@@ -546,7 +546,7 @@ fn default_page_limit() -> i32 {
 #[cfg(test)]
 mod tests {
     use super::{
-        CreateTemplateRequest, NewSandbox, SandboxNetworkConfig, SetTimeoutRequest,
+        CreateTemplateRequest, NewSandbox, RefreshRequest, SandboxNetworkConfig, SetTimeoutRequest,
         TemplateAliasLookupResponse,
     };
     use validator::Validate;
@@ -576,6 +576,30 @@ mod tests {
             let req = SetTimeoutRequest { timeout };
             req.validate()
                 .unwrap_or_else(|e| panic!("timeout={timeout} should be valid: {e}"));
+        }
+    }
+
+    #[test]
+    fn refresh_request_accepts_never_zero_and_positive_duration() {
+        for duration in [-1, 0, 60, 3600] {
+            let req = RefreshRequest {
+                duration: Some(duration),
+            };
+            req.validate()
+                .unwrap_or_else(|e| panic!("duration={duration} should be valid: {e}"));
+        }
+    }
+
+    #[test]
+    fn refresh_request_rejects_out_of_range_duration() {
+        for duration in [-2, 3601] {
+            let req = RefreshRequest {
+                duration: Some(duration),
+            };
+            assert!(
+                req.validate().is_err(),
+                "duration={duration} should be rejected"
+            );
         }
     }
 
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -95,9 +95,9 @@ func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.
 		rsp.Ret = ret
 		return
 	}
-	if req.Duration < -1 {
+	if req.Duration < -1 || req.Duration > 3600 {
 		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
-		rsp.Ret.RetMsg = "duration must be >= -1 (use -1 for never timeout)"
+		rsp.Ret.RetMsg = "duration must be in [-1, 3600]"
 		return
 	}
 
@@ -130,6 +130,9 @@ func refreshTimeoutMeta(ctx context.Context, sandboxID string, timeoutSeconds in
 		} else if endAt > 0 {
 			return endAt
 		}
+	}
+	if timeoutSeconds < 0 {
+		return 0
 	}
 	return time.Now().UnixMilli() + int64(timeoutSeconds)*1000
 }
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -65,6 +65,7 @@ func SetTimeout(ctx context.Context, req *types.SetTimeoutRequest) (rsp *types.S
 
 // Refresh implements POST /cube/sandbox/refresh. Semantically identical to
 // SetTimeout: refresh(d) rebases the idle clock and sets TimeoutSeconds = d.
+// Duration accepts -1 for never-timeout, 0 for immediate timeout, or a positive TTL.
 func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.RefreshSandboxRes) {
 	rsp = &types.RefreshSandboxRes{
 		RequestID: req.RequestID,
@@ -90,13 +91,13 @@ func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.
 		rsp.Ret.RetMsg = "should provide sandboxID"
 		return
 	}
-	if req.Duration <= 0 {
-		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
-		rsp.Ret.RetMsg = "duration must be positive (seconds)"
-		return
-	}
 	if ret := normalizeSandboxIDInReq(ctx, &req.SandboxID); ret != nil {
 		rsp.Ret = ret
+		return
+	}
+	if req.Duration < -1 {
+		rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
+		rsp.Ret.RetMsg = "duration must be >= -1 (use -1 for never timeout)"
 		return
 	}
 
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -125,7 +125,7 @@ func refreshTimeoutMeta(ctx context.Context, sandboxID string, timeoutSeconds in
 	if p := getTimeoutProvider(); p != nil {
 		endAt, err := p.RefreshTimeout(ctx, sandboxID, timeoutSeconds)
 		if err != nil {
-			log.G(ctx).Warnf("lifecycle: RefreshTimeout sandbox=%s failed: %v", sandboxID, err)
+			log.G(ctx).Errorf("lifecycle: RefreshTimeout sandbox=%s failed: %v", sandboxID, err)
 		} else if timeoutSeconds < 0 {
 			return 0
 		} else if endAt > 0 {
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
@@ -8,6 +8,8 @@ import (
 	"context"
 	"testing"
 
+	"github.com/stretchr/testify/assert"
+
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/localcache"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
@@ -27,25 +29,26 @@ func TestSetTimeoutValidationAllowsZero(t *testing.T) {
 		Timeout:   0,
 	})
 
-	if rsp.Ret.RetCode != int(errorcode.ErrorCode_Success) {
-		t.Fatalf("timeout=0 should be accepted, got ret=%+v", rsp.Ret)
-	}
-	if rsp.EndAt <= 0 {
-		t.Fatalf("timeout=0 should return an immediate endAt, got %d", rsp.EndAt)
-	}
+	assert.Equal(t, int(errorcode.ErrorCode_Success), rsp.Ret.RetCode)
+	assert.Greater(t, rsp.EndAt, int64(0))
 }
 
 func TestSetTimeoutValidationRejectsNegative(t *testing.T) {
+	const sandboxID = "sb-timeout-negative-validation"
+	localcache.SetSandboxCache(sandboxID, &localcache.SandboxCache{
+		SandboxID: sandboxID,
+		HostIP:    "127.0.0.1",
+	})
+	defer localcache.DeleteSandboxCache(sandboxID)
+
 	// Only -1 (NeverTimeout) is accepted as a valid negative value.
 	rsp := SetTimeout(context.Background(), &types.SetTimeoutRequest{
 		RequestID: "req-negative",
-		SandboxID: "sb-timeout-negative-validation",
+		SandboxID: sandboxID,
 		Timeout:   -2,
 	})
 
-	if rsp.Ret.RetCode != int(errorcode.ErrorCode_MasterParamsError) {
-		t.Fatalf("timeout<-1 should be rejected as params error, got ret=%+v", rsp.Ret)
-	}
+	assert.Equal(t, int(errorcode.ErrorCode_MasterParamsError), rsp.Ret.RetCode)
 }
 
 func TestRefreshValidationRejectsNonPositiveDuration(t *testing.T) {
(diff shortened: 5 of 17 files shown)


emailcannotbeblank:
## Note About E2B Refresh Semantics

In E2B, refresh-sandbox can only extend the sandbox lifetime; it does not shorten it.

The E2B refresh handler calls:

```go
a.orchestrator.KeepAliveFor(ctx, teamID, sandboxID, duration, false)
```

The last argument is `allowShorter=false`. Inside `KeepAliveFor`, E2B skips the update when the newly calculated `endTime` is earlier than the current `sbx.EndTime`:

```go
if !allowShorter && endTime.Before(sbx.EndTime) {
    return sbx, nil
}
```

By contrast, CubeSandbox refresh currently rewrites `CreatedAt`, `TimeoutSeconds`, and `EndAt` directly. This means CubeSandbox refresh may shorten the sandbox lifetime if the new duration is smaller than the currently remaining lifetime.

CubeMaster/pkg/service/sandbox/types/types.go
```diff
@@ -686,6 +686,7 @@ type UpdateRequest struct {
 	SandboxID    string `json:"sandbox_id" p:"sandbox_id"  v:"required"`
 	InstanceType string `json:"instance_type" p:"instance_type"  v:"required"`
 	Action       string `json:"action" p:"action"  v:"required"`
+	Timeout      *int   `json:"timeout,omitempty" p:"timeout"`
```
chenhengqi: What is this `p` struct tag?

CubeAPI/src/models/mod.rs
```diff
@@ -503,7 +503,7 @@ fn validate_timeout_value(timeout: i32) -> Result<(), validator::ValidationError
 /// Request body for POST /sandboxes/{id}/refreshes
 #[derive(Debug, Deserialize, Validate, ToSchema)]
 pub struct RefreshRequest {
-    #[validate(range(min = 0, max = 3600))]
+    #[validate(range(min = -1, max = 3600))]
```
chenhengqi: Is it possible to bump the upper limit?

CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
```diff
@@ -66,3 +66,56 @@ func TestSetTimeoutValidationAllowsNeverTimeout(t *testing.T) {
...
+	})
+	defer localcache.DeleteSandboxCache(sandboxID)
+
+	rsp := Refresh(context.Background(), &types.RefreshSandboxRequest{
+		RequestID: "req-refresh-zero",
+		SandboxID: sandboxID,
+		Duration:  0,
+	})
+
+	if rsp.Ret.RetCode != int(errorcode.ErrorCode_Success) {
```
chenhengqi: Use `assert`.

CubeMaster/pkg/service/sandbox/types/types.go
emailcannotbeblank: This is to receive the timeout field in the SandboxUpdateRequest (https://github.com/TencentCloud/CubeSandbox/blob/master/CubeAPI/src/cubemaster/mod.rs#L1283).

CubeAPI/src/models/mod.rs
emailcannotbeblank: Yes. Currently, it's set this way just to align with e2b's behavior. We can set it to -1 to represent 'never timeout'.

CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
emailcannotbeblank: Yes sir. Update it later.

chenhengqi (commented):
I am fine with the changes. @fslongjin Please double check.

emailcannotbeblank:
Hi maintainers, friendly ping.

This PR has been rebased onto the latest master, with conflicts resolved and commits cleaned up.

The remaining Claude Auto Review failure is caused by the `pull_request_target` workflow rejecting fork PR checkout, rather than this PR’s code. 
@fslongjin

CubeMaster/pkg/service/sandbox/types/types.go
chenhengqi: Please drop the p struct tag, see #1078.

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -467,6 +467,9 @@ pub async fn refresh_sandbox(
     Path(sandbox_id): Path<String>,
     Json(body): Json<RefreshRequest>,
 ) -> AppResult<impl IntoResponse> {
+    body.validate()
```
fslongjin, line 606: 👍 Good catch adding `body.validate()` here — the `RefreshRequest` struct had validation annotations that were never being invoked before this PR.

**However,** the same gap also exists in `resume_sandbox` (line 289). The `ResumedSandbox` struct at `models/mod.rs:324` has neither a `Validate` derive nor validation annotations on its `timeout` field, and the handler never calls `body.validate()` either.

This means an attacker or buggy client can currently send arbitrary `i32` values (e.g., `i32::MIN`, `i32::MAX`) as a resume `timeout`, bypassing the CubeAPI validation layer entirely. CubeMaster does normalize negative values to `-1` as defense-in-depth, but this validation gap should be closed at the API boundary for consistency with the other two timeout endpoints.

**Suggested fix:**

In `models/mod.rs`, update `ResumedSandbox`:
```rust
#[derive(Debug, Deserialize, Validate, ToSchema)]  // add Validate
#[allow(dead_code)]
pub struct ResumedSandbox {
    #[serde(default, skip_serializing_if = "Option::is_none")]
    #[validate(custom(function = "validate_timeout_value"))]  // add this
    pub timeout: Option<i32>,
    #[serde(rename = "autoPause", default)]
    pub auto_pause: bool,
}
```

Then in the handler, add `body.validate().map_err(|e| AppError::BadRequest(e.to_string()))?;` after the destructuring, matching the pattern used here at line 470.

CubeAPI/src/models/mod.rs
fslongjin: 👍 The range change from `min = 0` to `min = -1` is the right fix — refresh should accept the same `-1` never-timeout sentinel that `SetTimeoutRequest` already does.

**Side note:** while we're fixing validation coverage on the Rust side for this PR, the same gap exists in the `ResumedSandbox` struct (line 324). It has no `#[derive(Validate)]` and its `timeout` field has no `#[validate(...)]` annotation, so resume requests bypass CubeAPI-level timeout validation. Would be great to fix it in the same PR for consistency — `validate_timeout_value` (line 495) already exists and can be reused.

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -50,6 +50,10 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
 		rsp.Ret = ret
 		return
 	}
+	if req.Action == "resume" && req.Timeout != nil && *req.Timeout < 0 {
```
fslongjin: **issue:** The validation strategy here is inconsistent with `SetTimeout` and `Refresh`.

`SetTimeout` (`sandbox_timeout.go:49`) and `Refresh` (`sandbox_timeout.go:98`) both **reject** values `< -1` with `ErrorCode_MasterParamsError`:
```go
if req.Timeout < -1 {
    rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
    rsp.Ret.RetMsg = "timeout must be >= -1 (use -1 for never timeout)"
    return
}
```

But this resume path **silently coerces** any negative value to `-1`:
```go
if req.Action == "resume" && req.Timeout != nil && *req.Timeout < 0 {
    timeout := types.NeverTimeout
    req.Timeout = &timeout
}
```

A caller sending `timeout=-2` to `SetTimeout` gets a clear error. The same caller sending `timeout=-2` to `resume` gets a `200` with the value silently changed to never-timeout. This is a behavioral trap.

**Suggested fix:** align with `SetTimeout`/`Refresh`:
```go
if req.Action == "resume" && req.Timeout != nil && *req.Timeout < -1 {
    rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
    rsp.Ret.RetMsg = "timeout must be >= -1 (use -1 for never timeout)"
    return
}
```

Bonus: the current `< 0` condition also captures `-1` (the valid `NeverTimeout` sentinel) and does a no-op `-1 → -1` assignment, which is correct but confusing to read. Changing to `< -1` fixes that readability issue too.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -94,9 +95,9 @@ func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.
 		rsp.Ret = ret
 		return
 	}
-	if req.Duration <= 0 {
+	if req.Duration < -1 {
```
fslongjin: **issue:** This check only validates the lower bound (`Duration < -1`), with no corresponding upper-bound check.

CubeAPI's `RefreshRequest` enforces `#[validate(range(min = -1, max = 3600))]` (`models/mod.rs:506`), capping the duration at 3600 seconds. But here in CubeMaster there's no equivalent — if someone calls CubeMaster directly (or CubeAPI has a bug), they could send `duration=999999` and it would be accepted silently.

**Suggested fix:**
```go
if req.Duration < -1 || req.Duration > 3600 {
    rsp.Ret.RetCode = int(errorcode.ErrorCode_MasterParamsError)
    rsp.Ret.RetMsg = "duration must be in [-1, 3600]"
    return
}
```

(Note: `SetTimeout` doesn't need an upper bound since arbitrarily large positive timeouts are valid there, but refresh has a specific window-extension semantic where a ceiling is appropriate.)

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -467,6 +467,9 @@ pub async fn refresh_sandbox(
...
+        .map_err(|e| AppError::BadRequest(e.to_string()))?;
+
     let duration = body.duration.unwrap_or(0);
```
fslongjin: **question (non-blocking):** The default for an omitted `duration` is `0`, which in this PR's semantics means "immediate timeout":
```rust
let duration = body.duration.unwrap_or(0);
```

A caller who forgets to set `duration` will have their sandbox immediately scheduled for expiry rather than getting a refreshed lease. I understand this likely matches E2B's behavior, but could we either:

1. Add an explicit doc comment on `RefreshRequest` warning that omitting `duration` equals `duration=0` (immediate timeout), or
2. Consider whether `None` should mean "just rebase the idle clock without changing the timeout" instead?

Not blocking — just raising it since it's a sharp edge for API consumers.

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -94,10 +99,18 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
...
+		publishUpdateTimeout(ctx, req)
 		// Only on genuine success — IsAlreadyInState / NotFound are handled
 		// upstream by CLM's own reconciliation and would send misleading
 		// state signals through the lifecycle channel.
 		runAfterUpdateSandboxSuccessHook(ctx, req.SandboxID, req.InstanceType, req.Action, req.RequestID)
 	}
 	return
 }
+
+func publishUpdateTimeout(ctx context.Context, req *types.UpdateRequest) {
```
fslongjin, line 109: **nit:** This function is one of the two core additions in this PR but has no doc comment. A brief explanation would help future maintainers:

```go
// publishUpdateTimeout syncs the requested timeout to the lifecycle
// metadata channel after a successful sandbox resume. It is a
// best-effort operation: failures are logged but do not affect the
// resume response.
//
// Only acts on "resume" action with a non-nil Timeout. Pause
// requests and resumes without an explicit Timeout are silently
// skipped (preserving the existing timeout without rebasing the
// idle clock).
func publishUpdateTimeout(ctx context.Context, req *types.UpdateRequest) {
```

This makes it immediately clear that: (1) it's resume-only, (2) failures are non-fatal, and (3) `nil` Timeout means "don't touch the timeout."

CubeMaster/pkg/service/sandbox/types/types.go
```diff
@@ -701,6 +701,7 @@ type UpdateRequest struct {
...
```
fslongjin: **nit:** The new `Timeout` field would benefit from a comment explaining its semantics, similar to how the `SetTimeoutRequest` type has detailed field documentation:

```go
// Timeout is the new idle TTL in seconds for "resume" actions.
// Accepted values: -1 (never timeout), 0 (immediate timeout),
// or a positive number of seconds. nil preserves the existing
// timeout without rebasing the idle clock.
Timeout *int `json:"timeout,omitempty" p:"timeout"`
```

The `nil` vs `0` vs `-1` distinction is subtle enough that future readers will appreciate the hint.

fslongjin (commented):
Thanks for this PR! The core fix is solid — adding the missing `timeout` field to `UpdateRequest` and fixing the refresh validation gap are both real bugs that needed fixing.

I did a thorough review across several dimensions (security, correctness, architecture, test coverage, docs). The functional logic is correct, but I found a few things worth addressing before merging:

### Blocking items

1. **Missing input validation on `resume_sandbox`** — `ResumedSandbox` has no `Validate` derive and the handler never calls `body.validate()`, unlike `set_sandbox_timeout` and `refresh_sandbox` which both validate. This is a defense-in-depth gap (see inline comments).

2. **Inconsistent timeout validation strategy** — `SetTimeout` and `Refresh` reject `timeout < -1` with a clear error, but `Update/resume` silently coerces any negative value to `-1`. A caller sending `timeout=-2` gets an explicit error from `SetTimeout` but a 200 from `resume`. This is a behavioral trap for SDK authors (see inline comment on `sandbox_update.go:53`).

3. **Missing upper-bound validation in CubeMaster's `Refresh()`** — CubeAPI enforces `max=3600` on refresh duration, but the Go side only checks `Duration < -1` with no upper bound (see inline comment on `sandbox_timeout.go:98`).

### Non-blocking suggestions

- The default `duration=0` (immediate timeout) when no duration is provided in refresh is a destructive default — worth documenting explicitly or reconsidering.
- New function `publishUpdateTimeout` and field `UpdateRequest.Timeout` are missing doc comments.
- The `refreshTimeoutMeta` fallback for `-1` timeout computes `time.Now().UnixMilli() + (-1)*1000` which returns a ~1-second-past timestamp as `EndAt`. The lifecycle sweeper isn't affected (it reads from Redis directly), but API consumers receiving a past timestamp for a never-timeout sandbox might be confused. Consider adding a guard:
  ```go
  if timeoutSeconds < 0 {
      return 0 // never-timeout: sentinel for "no expiry deadline"
  }
  return time.Now().UnixMilli() + int64(timeoutSeconds)*1000
  ```

All three blocking items are small-scope fixes (~15 lines total across Rust and Go). Once addressed, this PR looks good to merge.

fslongjin (changes requested):
Hello, can you add some e2e tests under `tests/e2e/sdk_compat` ?

chenhengqi:
@emailcannotbeblank Please also add e2e tests. See https://github.com/TencentCloud/CubeSandbox/tree/master/tests/e2e/sdk_compat

CubeAPI/src/handlers/sandboxes.rs
emailcannotbeblank: Thanks, fixed as suggested above.

CubeAPI/src/models/mod.rs
emailcannotbeblank: Thanks, fixed. `ResumedSandbox` now derives `Validate`, `timeout` reuses `validate_timeout_value`, and `resume_sandbox` calls `body.validate()` before forwarding the request.

CubeMaster/pkg/service/sandbox/sandbox_update.go
emailcannotbeblank: Fixed. `Update/resume` now rejects explicit timeout values `< -1` with `ErrorCode_MasterParamsError`, matching `SetTimeout` and `Refresh`.

In the previous version, I noticed that sandbox creation normalized timeout values `< -1` to `-1`, so the earlier change intentionally followed that normalization behavior for consistency with create. After your feedback, I aligned resume with the stricter validation strategy used by `SetTimeout` and `Refresh` instead.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
emailcannotbeblank: Fixed. CubeMaster `Refresh()` now validates `duration` in `[-1, 3600]`, matching CubeAPI's `RefreshRequest` constraint.

CubeAPI/src/handlers/sandboxes.rs
emailcannotbeblank: Documented. I kept the existing behavior for compatibility, but added an explicit `RefreshRequest.duration` comment noting that omitted duration is treated as `0`, which means immediate timeout, and `-1` means never-timeout.

CubeMaster/pkg/service/sandbox/sandbox_update.go
emailcannotbeblank: Thanks, fixed as suggested above.

CubeMaster/pkg/service/sandbox/types/types.go
emailcannotbeblank: Thanks, fixed as suggested above.

emailcannotbeblank:
> Hello, can you add some e2e tests under `tests/e2e/sdk_compat` ?
>
> @emailcannotbeblank Please also add e2e tests. See https://github.com/TencentCloud/CubeSandbox/tree/master/tests/e2e/sdk_compat


Yes, sir. Added `sdk_compat` lifecycle E2E coverage for resume timeout semantics as requested:

- valid never-timeout resume (`timeout=-1`);
- invalid negative resume (`timeout=-2`) returning 400.

emailcannotbeblank:
**Changes in this commit**

This commit mainly fixes inconsistent timeout semantics across the refresh / resume / update paths, and adds corresponding tests.

Main changes:

1. Added input validation for CubeAPI resume requests. `ResumedSandbox` now derives `Validate`, and `ResumedSandbox.timeout` reuses the existing `validate_timeout_value`. The `resume_sandbox` handler now calls `body.validate()` before forwarding the request to the service layer, so `timeout < -1` is rejected at the CubeAPI layer.

2. CubeMaster update/resume no longer silently normalizes invalid negative values. Previously, `timeout < 0` was normalized to `-1`. Now only `-1` means never-timeout, and explicit values `< -1`, such as `-2`, return a parameter error.

3. Added upper-bound validation for CubeMaster refresh. CubeAPI already limits `duration` to `[-1, 3600]`, but CubeMaster previously only checked the lower bound. CubeMaster now also checks `duration < -1 || duration > 3600`, matching CubeAPI.

4. The never-timeout `EndAt` fallback is now explicitly `0`. `refreshTimeoutMeta` falls back to local computation when there is no provider or the provider fails; for `timeoutSeconds < 0`, it now returns `0`, meaning there is no expiry deadline.

5. Added documentation comments. `RefreshRequest.duration` now documents that an omitted value is treated as `0`, meaning immediate timeout; `publishUpdateTimeout` documents that it only handles resume requests with an explicit timeout; `UpdateRequest.Timeout` documents the semantics of `nil`, `-1`, `0`, and positive values.

6. Added tests. Rust model tests cover valid and invalid `ResumedSandbox.timeout` values; Go tests cover refresh/update timeout behavior; E2E tests cover successful `resume(timeout=-1)` and `resume(timeout=-2)` returning 400.

**Previous design rationale**

In the previous version, resume was designed to align with create: negative timeout values were normalized to `-1` instead of rejecting values `< -1`. This has now been changed to reject `< -1`; only `-1` means never-timeout.

The previous version also did not specially handle never-timeout `EndAt`. On one hand, CubeSandbox itself did not use this returned value; on the other hand, it was unclear whether this value should be `-1` or `0`. It is now standardized to return `EndAt = 0` for never-timeout.

**Remaining open question**

According to E2B semantics, `refresh` is only used to extend a sandbox's lifetime. In the current implementation, however, `refresh(duration)` effectively resets the timeout deadline, so it can either extend or shorten the sandbox lifetime. It is still unclear whether CubeSandbox should strictly align with E2B's “extend only, never shorten” behavior; this should be confirmed separately.

tests/e2e/sdk_compat/cases/lifecycle/test_pause_resume.py
```diff
@@ -18,6 +18,13 @@
 ]
 
 
+@pytest.fixture()
+def cubesandbox_resume_sandbox(sdk_backend, sdk_sandbox):
+    if sdk_backend != "cubesandbox":
+        pytest.skip("explicit resume timeout validation is CubeSandbox SDK specific")
```
chenhengqi: Why this is CubeSandbox specific?

chenhengqi: @tinklone Please take a look at the test.

emailcannotbeblank: Fixed. Updated the E2E coverage to avoid making this CubeSandbox-only.

The resume timeout cases now go through the shared `sdk_sandbox.resume(timeout=...)` adapter method, and both backends are collected:

- `test_resume_accepts_never_timeout[e2b]`
- `test_resume_accepts_never_timeout[cubesandbox]`
- `test_resume_rejects_invalid_negative_timeout[e2b]`
- `test_resume_rejects_invalid_negative_timeout[cubesandbox]`

I also added `resume(timeout)` support to the E2B adapter via its existing `resume_or_connect(timeout)` path, and made the CubeSandbox adapter forward timeout through explicit resume before reconnecting.

emailcannotbeblank:
> Hello, can you add some e2e tests under `tests/e2e/sdk_compat` ?
>
> Why this is CubeSandbox specific?

During E2B e2e testing, we found that the E2B GET /sandboxes/{sandboxID} response requires endAt to be present and formatted as a valid datetime string.

In the current implementation, CubeMaster aligns with the create behavior for never-timeout sandboxes and stores endAt as 0. On the CubeAPI side, this value is deserialized into None, which causes the E2B GET sandbox response to have an incompatible data shape for the E2B SDK.

To fix this, I updated the E2B-facing response behavior in CubeAPI. When handling an E2B SDK request, CubeAPI now checks the sandbox endAt value; if the sandbox is configured to never expire, it returns a far-future datetime sentinel instead of omitting endAt.

I also added documentation describing this compatibility behavior and clarifying that E2B SDK requests with timeout = -1 are supported.

Friendly ping to @fslongjin @ls-ggg

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -467,6 +523,9 @@ pub async fn refresh_sandbox(
...
```
fslongjin: Hmm, this default looks dangerous once paired with the CubeMaster change in this PR.

`unwrap_or(0)` used to be mostly harmless because CubeMaster rejected `Duration <= 0`. Now that `0` is accepted, a refresh with no `duration` (or an empty body) will actually go through and immediately expire the sandbox. That’s a pretty nasty footgun for a keep-alive API.

E2B doesn’t do this either — if duration is omitted they fall back to the default TTL, they don’t treat it as “kill now”.

Please change this so an omitted `duration` falls back to the **cluster default timeout**, instead of filling in `0`. That defaulting belongs in CubeMaster: our cluster default is never-timeout, and only if the user has explicitly configured a timeout should we use that configured value. CubeAPI shouldn’t turn “not provided” into “expire immediately” via `unwrap_or(0)`.

Also please update the `RefreshRequest` comment that currently documents “omitted = 0 = immediate timeout” as intentional.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -94,9 +95,9 @@ func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.
 		rsp.Ret = ret
 		return
 	}
-	if req.Duration <= 0 {
+	if req.Duration < -1 || req.Duration > 3600 {
```
fslongjin: I’m not comfortable with `duration=0` becoming a valid refresh path here.

Previously `<= 0` was rejected; now `0` succeeds, so refresh acts like `set_timeout(0)` and can immediately expire the sandbox. Immediate timeout belongs on `set_timeout`, not refresh. `-1` (never-timeout) is fine.

Please reject `0` on refresh, and update `TestRefreshValidationAllowsZero`.

CubeAPI/src/cubemaster/mod.rs
```diff
@@ -1380,7 +1380,8 @@ pub struct SandboxUpdateRequest {
     /// "pause" | "resume"
     #[serde(rename = "action")]
     pub action: String,
-    /// TTL in seconds (for resume; 0 = keep original). Optional for pause.
+    /// TTL in seconds for resume. None keeps the existing timeout; 0 requests
+    /// immediate timeout; -1 disables timeout.
```
fslongjin: This comment rewrite quietly changes the resume contract — I’d push back on it.

Previously both the docs and the behavior treated `0` as keep-original (`UpdateRequest` had no `timeout` field, so `0` was dropped). Now that it’s wired through, `resume`/`connect` with `timeout: 0` immediately expires the sandbox. That’s a silent breaking change.

`-1` and omitted are fine. Please keep `0` as preserve-original on resume, and leave immediate timeout to `set_timeout(0)`.

fslongjin (changes requested):
Thanks for the work on this — the core bugfixes (missing `UpdateRequest.timeout`, calling `validate()`, and never-timeout `endAt=0` fallback) look solid.

Before we can merge, though, there are a few timeout-semantics regressions that need to be addressed. I’ve left inline comments on the three main spots:

1. Omitted refresh `duration` must not become immediate timeout — please fall back to the CubeMaster cluster default instead.
2. Refresh should not accept `duration=0` as “kill now”; that belongs on `set_timeout`.
3. `resume(timeout=0)` should keep the old preserve-original meaning, not silently flip to immediate timeout.

Happy to re-review once those are sorted.

emailcannotbeblank:
> Thanks for the work on this — the core bugfixes (missing `UpdateRequest.timeout`, calling `validate()`, and never-timeout `endAt=0` fallback) look solid.
> 
> Before we can merge, though, there are a few timeout-semantics regressions that need to be addressed. I’ve left inline comments on the three main spots:
> 
> 1. Omitted refresh `duration` must not become immediate timeout — please fall back to the CubeMaster cluster default instead.
> 2. Refresh should not accept `duration=0` as “kill now”; that belongs on `set_timeout`.
> 3. `resume(timeout=0)` should keep the old preserve-original meaning, not silently flip to immediate timeout.
> 
> Happy to re-review once those are sorted.

Fixed, thanks for the reviews.

I updated the timeout semantics according to the three review comments:

- omitted `refresh.duration` is no longer converted to `0`; CubeMaster now
  applies the cluster default when `duration` is omitted;
- `refresh(duration=0)` is rejected, because immediate timeout should go through
  `set_timeout(0)`;
- `resume/connect timeout=0` now preserves the existing timeout, instead of
  immediately expiring the sandbox.

I also added E2E coverage for the related lifecycle timeout behavior:

| Area | Backend / entry | Input | Expected result |
|---|---|---:|---|
| refresh | CubeAPI REST | omitted `duration` / `{}` | success; sandbox remains usable |
| refresh | CubeAPI REST | `duration=-1` | success; never-timeout |
| refresh | CubeAPI REST | `duration=60` | success; positive TTL |
| refresh | CubeAPI REST | `duration=0` | 400 bad request |
| refresh | CubeAPI REST | `duration=-2` | 400 bad request |
| refresh | CubeAPI REST | `duration=3601` | 400 bad request |
| set_timeout | CubeSandbox SDK + E2B SDK | `timeout=-1` | success; never-timeout; sandbox remains usable |
| set_timeout | CubeSandbox SDK + E2B SDK | `timeout=60` | success; positive TTL; sandbox remains usable |
| set_timeout | CubeSandbox SDK + E2B SDK | `timeout=0` | success; immediate timeout is accepted |
| set_timeout | CubeSandbox SDK + E2B SDK | `timeout=-2` | rejected; sandbox remains usable |
| resume/connect | CubeSandbox SDK + E2B SDK | `timeout=-1` | success; never-timeout; sandbox resumes and remains usable |
| resume/connect | CubeSandbox SDK + E2B SDK | `timeout=0` | success; preserves existing timeout; sandbox resumes and remains usable |
| resume/connect | CubeSandbox SDK + E2B SDK | `timeout=60` | success; positive TTL; sandbox resumes and remains usable |
| resume/connect | CubeSandbox SDK + E2B SDK | `timeout=-2` | rejected with bad request |

Local verification passed against the updated CubeAPI and CubeMaster binaries:

```text
28 passed in 20.74s
```

fslongjin:
Hi, could you please review and address the bot’s feedback? Thanks.

emailcannotbeblank:
> Hi, could you please review and address the bot’s feedback? Thanks.

Ok, probably push an update tomorrow.

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -4,27 +4,74 @@
...
+}
+
+fn apply_e2b_sandbox_detail_compat(headers: &HeaderMap, detail: &mut SandboxDetail) {
+    if is_e2b_sdk_request(headers) && detail.end_at.is_none() {
+        detail.end_at = Some(e2b_never_timeout_end_at());
+    }
+}
+
+fn apply_e2b_listed_sandbox_compat(headers: &HeaderMap, sandboxes: &mut [ListedSandbox]) {
+    if !is_e2b_sdk_request(headers) {
```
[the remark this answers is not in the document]
emailcannotbeblank: Fixed in 41732bf6, and taken one step further than suggested: the sentinel is now a compile-time `const` built via `DateTime::from_timestamp(253_402_300_799, 0)` — zero runtime parsing, zero conversion, and no lazy-init overhead. This thread can be resolved.

CubeMaster/pkg/service/sandbox/types/types.go
```diff
@@ -815,15 +819,14 @@ type SetTimeoutRes struct {
 	Ret       *Ret   `json:"ret,omitempty"`
 }
 
-// RefreshSandboxRequest extends the sandbox idle window by `duration`
-// seconds. Semantically `refresh(d)` is identical to `set_timeout(d)` in
-// this implementation: both rebase CreatedAt to "now" and set the new
-// TimeoutSeconds. Mirrors e2b's refresh-then-set-timeout convergence.
+// RefreshSandboxRequest refreshes the sandbox idle timeout. nil Duration uses
```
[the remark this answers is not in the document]
emailcannotbeblank: Thanks for flagging. CubeMaster and CubeAPI ship from the same repo and upgrade as a pair; even during a brief rolling window, an old CubeMaster fails cleanly with a 400 — easy to spot, nothing silent. The Duration pointer change is required to distinguish omitted from explicit 0. Skipping dedicated deploy docs for now; easy to add later if needed.

CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
```diff
@@ -62,7 +65,65 @@ func TestSetTimeoutValidationAllowsNeverTimeout(t *testing.T) {
...
+	})
+	defer localcache.DeleteSandboxCache(sandboxID)
+
+	rsp := Refresh(context.Background(), &types.RefreshSandboxRequest{
+		RequestID: "req-refresh-never",
+		SandboxID: sandboxID,
+		Duration:  types.NeverTimeout,
+	})
+
+	assert.Equal(t, int(errorcode.ErrorCode_Success), rsp.Ret.RetCode)
```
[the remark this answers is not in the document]
emailcannotbeblank: Added: the end-to-end `TestRefreshValidationAllowsNeverTimeout` now asserts `rsp.EndAt == 0` (sandbox_timeout_test.go:103)

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -46,6 +46,10 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
 		rsp.Ret.RetMsg = "action should be pause or resume"
 		return
 	}
+	if req.Action == "resume" && req.Timeout != nil && *req.Timeout < 0 {
+		timeout := types.NeverTimeout
+		req.Timeout = &timeout
+	}
```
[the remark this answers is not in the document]
emailcannotbeblank: Fixed per option 1: `Update` now explicitly rejects values `< -1` (consistent with SetTimeout/Refresh) and no longer mutates `req.Timeout` — a test verifies that -2 returns the error with the struct left untouched (sandbox_update_test.go:29-33).

CubeMaster/pkg/service/sandbox/sandbox_update_test.go
```diff
@@ -0,0 +1,160 @@
...
+	"context"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/config"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/localcache"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
```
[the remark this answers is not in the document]
emailcannotbeblank: Done as suggested: the shared double now lives in `timeout_provider_test.go` (renamed `recordingTimeoutProvider`, timeout_provider_test.go:9-28) and is used by the update/timeout/info test files.

CubeMaster/pkg/service/sandbox/sandbox_update_test.go
```diff
@@ -0,0 +1,160 @@
...
+	provider := &updateTimeoutProvider{}
+	SetTimeoutProvider(provider)
+	defer SetTimeoutProvider(nil)
+
+	timeout := 600
+	rsp := Update(context.Background(), &types.UpdateRequest{
+		RequestID:    "req-update-timeout-metadata",
+		SandboxID:    sandboxID,
+		InstanceType: "cubebox",
+		Action:       "resume",
```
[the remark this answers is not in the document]
emailcannotbeblank: Added: `TestUpdateResumeNeverTimeoutPublishesMetadata` (sandbox_update_test.go:54-72) covers the resume + timeout=-1 path and asserts the provider receives NeverTimeout.

tests/e2e/sdk_compat/cases/lifecycle/test_pause_resume.py
```diff
@@ -65,6 +65,75 @@ def test_pause_and_connect_resume_preserves_env_vars(sdk_sandbox, sdk_e2e_config
...
+)
+def test_resume_accepts_zero_and_positive_timeout(
+    sdk_sandbox,
+    sdk_e2e_config,
+    timeout,
+    stdout,
+):
+    sdk_sandbox.pause(timeout=sdk_e2e_config.default_timeout)
+
+    sdk_sandbox.resume(timeout=timeout)
```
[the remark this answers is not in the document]
emailcannotbeblank: Intentionally kept broad: this compatibility suite runs against multiple SDKs that wrap API errors in different exception types (noted in the noqa B017 comment), so narrowing to `requests.HTTPError` isn't feasible across them. The false-positive surface is already contained — `_assert_invalid_timeout_error` enforces HTTP 400 plus message content, so unrelated exceptions like network timeouts fail the 400 assertion.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -90,9 +93,15 @@ func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.
 		rsp.Ret.RetMsg = "should provide sandboxID"
 		return
 	}
-	if req.Duration <= 0 {
+	var duration int
+	if req.Duration == nil {
```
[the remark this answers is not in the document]
emailcannotbeblank: This behavior is intentional (omitted duration falls back to the cluster default, consistent with create semantics and the direction the maintainer requested), and the requested docs warning is in place: both the English and Chinese lifecycle docs carry a dedicated warning box plus a behavior-table entry stating that an omitted duration becomes never-timeout under the default config, and to pass an explicit positive duration when a finite TTL is needed.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -90,9 +93,15 @@ func Refresh(ctx context.Context, req *types.RefreshSandboxRequest) (rsp *types.
...
+		duration, _ = resolveTimeoutSeconds(nil, config.GetConfig().CubeletConf.DefaultTimeoutInsec)
```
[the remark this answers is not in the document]
emailcannotbeblank: Both edges are addressed in the final revision: refresh no longer has a 3600 cap (validation now only requires -1 or a positive value), so a >3600 cluster default no longer 400s; the omitted-duration → cluster-default behavior is intentional and covered by a dedicated warning and behavior-table entry in both the English and Chinese docs.

tests/e2e/sdk_compat/adapters/cubesandbox_adapter.py
```diff
@@ -122,8 +122,12 @@ def pause(self, *, timeout: int = 60) -> None:
         self._sandbox.pause(timeout=timeout)
 
     def resume_or_connect(self, *, timeout: int = 60) -> "CubeSandboxAdapter":
+        self._sandbox.resume(timeout=timeout)
         return type(self).connect(self.sandbox_id, self._e2e_config or SdkE2EConfig.from_env())
```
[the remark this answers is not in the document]
emailcannotbeblank: Server-side behavior is verified and documented: in the final design, `connect` without a timeout preserves the current timeout and merely rebases the idle window (and leaves the timeout untouched for running sandboxes) — see the "Timeout input behavior by API" table in lifecycle.md. So the `connect()` after `resume(timeout=...)` cannot overwrite the value resume just set; the two calls each own their concern (resume owns timeout semantics, connect owns session establishment).

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -4,23 +4,69 @@
...
+        return;
+    }
```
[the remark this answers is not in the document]
emailcannotbeblank: Background: for never-timeout (timeout=-1) sandboxes, Cube-native responses omit endAt, while E2B SDK models require endAt to be a valid datetime — this PR therefore adds a compat branch that returns the 9999-12-31 sentinel only on E2B-facing paths. 

Broader compatibility, however, isn't straightforward: 32-bit time_t caps at 2038 (Y2038), so no far-future sentinel can carry the never-timeout meaning there. We're leaving this as-is for now and can revisit if such clients turn out to exist.

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -4,23 +4,76 @@
...
+#[cfg(test)]
+const E2B_NEVER_TIMEOUT_END_AT_RFC3339: &str = "9999-12-31T23:59:59Z";
+// 9999-12-31T23:59:59Z as Unix seconds. Build it at compile time to avoid
+// per-request parsing or runtime synchronization.
+const E2B_NEVER_TIMEOUT_END_AT: DateTime<Utc> =
+    DateTime::from_timestamp(253_402_300_799, 0).expect("E2B sentinel must be valid");
+const NEVER_TIMEOUT_SECONDS: i32 = -1;
+const E2B_USER_AGENT_MARKERS: [&str; 3] =
+    ["e2b-python-sdk/", "e2b-js-sdk/", "e2b-code-interpreter/"];
+
```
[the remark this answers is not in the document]
emailcannotbeblank: Background: for never-timeout (timeout=-1) sandboxes, Cube-native responses omit endAt, while E2B SDK models require endAt to be a valid datetime — this PR therefore adds a compat branch that returns the 9999-12-31 sentinel only on E2B-facing paths. 

The design question we think more worth discussing here is: given that E2B SDKs must receive an endAt, should Cube change its behavior uniformly (returning endAt to Cube-native callers as well), or should CubeAPI differentiate by SDK and respond differently? This PR takes the latter route — distinguishing via User-Agent so Cube-native semantics stay untouched — and the tradeoffs (spoofability, proxy rewriting) are documented in lifecycle.md. So we're not making further changes here for now.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
```diff
@@ -130,5 +139,9 @@ func refreshTimeoutMeta(ctx context.Context, sandboxID string, timeoutSeconds in
 			return endAt
 		}
 	}
+	// All callers reject values below NeverTimeout before reaching this helper.
+	if timeoutSeconds == types.NeverTimeout {
```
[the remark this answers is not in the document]
emailcannotbeblank: Fixed: `refreshTimeoutMeta` now explicitly excludes NeverTimeout before adopting the provider's endAt and returns 0 for NeverTimeout at the end — the endAt=0 contract no longer depends on providers special-casing -1. The provider is still invoked because it also stores the new timeout and publishes the lifecycle event (documented in the function comment), which doesn't conflict with making the contract provider-independent.

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -94,10 +100,27 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
...
+// publishUpdateTimeout syncs the requested timeout to the lifecycle
+// metadata channel after a successful sandbox resume. It is a
+// best-effort operation: failures are logged but do not affect the
+// resume response.
+//
+// Only acts on "resume" action with a non-nil Timeout. Pause
+// requests and resumes without an explicit Timeout are silently
+// skipped (preserving the existing timeout without rebasing the
+// idle clock).
+func publishUpdateTimeout(ctx context.Context, req *types.UpdateRequest) {
```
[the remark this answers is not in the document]
emailcannotbeblank: Intentionally keeping Warn: metadata publishing is best-effort by design (the function comment notes failures never affect the resume response) — a failure only stalemates the metadata channel, not correctness, so Error-level would create unnecessary alert noise.

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -96,6 +101,28 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
 	return
 }
 
+// publishUpdateTimeout starts a new lifecycle timeout window after a successful
+// sandbox resume. A non-zero Timeout replaces the stored timeout; nil or 0
+// preserves the stored timeout while moving its CreatedAt and EndAt forward
+// from the resume time. Metadata updates are best effort and never change the
+// resume response.
+func publishUpdateTimeout(ctx context.Context, req *types.UpdateRequest) {
```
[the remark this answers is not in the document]
emailcannotbeblank: Fixed as suggested: the store's Lua script returns an explicit not-found marker for sandboxes without metadata, and `SetTimeoutWindow`/`RebaseTimeoutWindow` surface it as a "lifecycle metadata for sandbox %s was not found" error (lifecycle/store.go:26-28, 227-228), so this path now logs clearly instead of silently returning `(0, nil)`. The resume response stays Success by the best-effort design (see the function comment).

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -96,6 +102,28 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
...
+		if p := getTimeoutProvider(); p != nil {
+			if _, err := p.RebaseTimeoutWindow(ctx, req.SandboxID); err != nil {
+				log.G(ctx).Warnf("lifecycle: RebaseTimeoutWindow sandbox=%s failed: %v", req.SandboxID, err)
+			}
+		}
+		return
+	}
+	// refreshTimeoutMeta updates lifecycle metadata through the timeout provider.
+	// Resume does not return endAt, so the computed value is intentionally ignored.
+	refreshTimeoutMeta(ctx, req.SandboxID, *req.Timeout)
```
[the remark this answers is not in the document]
emailcannotbeblank: Fixed in exactly this direction: `storeTimeoutProvider.RefreshTimeout` now routes through `SetTimeoutWindow`'s single-EVAL script, atomically replacing the timeout, starting a new window, and publishing the event (lifecycle/init.go:61-68, lifecycle/store.go:189-216) — the explicit-timeout path shares the same atomic script as the nil/0 path, so the protection is no longer asymmetric.

tests/e2e/sdk_compat/cases/lifecycle/test_pause_resume.py
```diff
@@ -65,6 +65,46 @@ def test_pause_and_connect_resume_preserves_env_vars(sdk_sandbox, sdk_e2e_config
...
+def _assert_invalid_timeout_error(exc: Exception) -> None:
+    status_code = getattr(exc, "status_code", None)
+    response = getattr(exc, "response", None)
+    if status_code is None and response is not None:
+        status_code = getattr(response, "status_code", None)
+    if status_code is not None:
+        assert status_code == 400
+
+    message = str(exc).lower()
+    assert "timeout" in message
```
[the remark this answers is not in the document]
emailcannotbeblank: The helper already asserts both HTTP 400 and "timeout" in the message, so unrelated exceptions like network timeouts fail the 400 check; tightening further to exact fragments (e.g. "must be"/">= -1") is possible but couples the test to error wording and becomes brittle if the message changes — an optional tradeoff.

fslongjin:
please rebase your code onto the latest `origin/master`, thanks.

emailcannotbeblank:
> please rebase your code onto the latest `origin/master`, thanks.

Thanks. I’ve merged the latest origin/master, resolved the conflicts, and pushed the updated branch.

CubeMaster/pkg/service/sandbox/sandbox_timeout.go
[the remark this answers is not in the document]
emailcannotbeblank: Thanks for calling this out. This behavior is intentional and follows the earlier maintainer review: an omitted `refresh.duration` must use CubeMaster's cluster default, while only an explicit `0` is rejected. Therefore, with the shipped `default_timeout_insec: -1`, the result is deliberately never-timeout. This is called out in a dedicated warning and in the timeout behavior table in both lifecycle guides, and the default-resolution path is covered by unit and E2E tests.

CubeAPI/src/handlers/sandboxes.rs
```diff
@@ -4,23 +4,76 @@
...
+    // E2B SDKs identify API requests through User-Agent. Keep this compatibility
+    // branch narrow so curl, Cube SDKs, and other clients preserve Cube's
+    // native omitted-endAt semantics for never-timeout sandboxes.
+    headers
+        .get(USER_AGENT)
+        .and_then(|value| value.to_str().ok())
+        .map(|value| {
+            let user_agent = value.to_ascii_lowercase();
+            E2B_USER_AGENT_MARKERS
+                .iter()
```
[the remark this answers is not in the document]
emailcannotbeblank: Thanks. This User-Agent check selects a response compatibility shape; it is not an authentication or authorization boundary, so spoofing it does not grant access or mutate lifecycle state. The substring match is deliberate because E2B User-Agents may be compound (covered by the `e2b-js-sdk/... e2b-cli/...` unit case), while the trailing `/` in every marker narrows accidental matches. The sentinel still requires both a confirmed `timeout_seconds == -1` and a missing `end_at`, so finite or unresolved lifecycle metadata is never rewritten as never-timeout.

CubeAPI and CubeMaster ship from the same repository and are upgraded as a pair. During a brief version-skew window, an older CubeMaster omits `timeout_seconds`, so the handler retains the native omitted-`endAt` shape instead of inferring never-timeout. The recognized User-Agent markers and fallback behavior are explicitly documented in both lifecycle guides.

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -91,6 +96,9 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
 			*rsp = *pauseSandbox(ctx, req, hostIP)
 		case "resume":
 			*rsp = *resumeFromPauseSnapshot(ctx, req, hostIP)
+			if rsp.Ret.RetCode == int(errorcode.ErrorCode_Success) {
+				publishUpdateTimeout(ctx, req)
```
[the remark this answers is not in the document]
emailcannotbeblank: Outdated

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -100,6 +108,22 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
...
 }
 
+// publishUpdateTimeout persists a resume-supplied timeout into lifecycle
+// metadata once the sandbox is running again. A positive Timeout replaces the
+// stored timeout and opens a new window from now; -1 (NeverTimeout) marks the
+// sandbox as never expiring. Nil or 0 keeps the stored timeout unchanged.
+// Best effort by design: metadata failures are logged and never alter the
+// already-successful resume response.
+func publishUpdateTimeout(ctx context.Context, req *types.UpdateRequest) {
+	if req == nil || req.Action != "resume" || req.Timeout == nil || *req.Timeout == 0 {
```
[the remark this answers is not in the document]
emailcannotbeblank: Deliberate divergence, locked in by this PR's own contract: both the PR description and the e2e coverage (`test_resume_accepts_zero_and_positive_timeout`) specify `resume(timeout=0)` → success + preserve stored timeout, so rejecting 0 now would break documented, tested behavior. Immediate timeout is intentionally `set_timeout(0)`-only — a caller that wants the resumed sandbox to expire immediately can say so explicitly after resume.

CubeMaster/pkg/service/sandbox/sandbox_update.go
```diff
@@ -100,6 +108,22 @@ func Update(ctx context.Context, req *types.UpdateRequest) (rsp *types.Res) {
...
+		return
+	}
+	// refreshTimeoutMeta updates lifecycle metadata through the timeout
+	// provider. Resume does not return endAt, so the computed value is
+	// intentionally ignored.
+	refreshTimeoutMeta(ctx, req.SandboxID, *req.Timeout)
```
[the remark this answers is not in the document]
emailcannotbeblank: Acknowledged — resume is not exceptional here: `set_timeout`/`refresh` share this exact persistence path (`refreshTimeoutMeta`) and the same failure modes, so this is the established best-effort stance rather than a resume-specific gap. Fair point on the `meta == nil` branch being completely silent — noted for a follow-up log line. The provider is shared by `set_timeout`/`refresh`, so that change is outside this PR's scope and belongs in a separate focused fix. Surfacing persistence failures in the resume response is a wire change, also tracked as a follow-up.

CubeMaster/pkg/service/sandbox/sandbox_update.go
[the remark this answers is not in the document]
emailcannotbeblank: Outdated

fslongjin:
Ready for re-review.

Addressed the merge blockers from the last review:

- CubeMaster regression tests for resume `timeout < -1`, the `nil`/`0`/`-1`/`N` publish matrix, publish-only-on-success, and best-effort provider failure
- `tests/e2e/sdk_compat` coverage: `resume(timeout=-1)`, `resume(timeout=-2)` → 400, and `resume(timeout=0)` preserves the stored TTL
- PR title aligned with the slim resume-only fix (refresh never-timeout is out of scope)
- EN/ZH lifecycle docs now split timeout semantics by API (`resume(0)` ≠ `set_timeout(0)`)

@fslongjin this should close the e2e CHANGES_REQUESTED.

fslongjin (approved):

PR #922, opened 2026-07-11
refactor(cubemaster): migrate HTTP API from gorilla/mux to gin-gonic/gin

sparkzky:
## What this PR does

Migrate CubeMaster's HTTP API from `gorilla/mux` to `gin-gonic/gin`. This is a **router-layer + handler-layer** migration (not a thin adapter): the giant `cube.HttpHandler` / `inner.HttpHandler` / `notify.HttpHandler` switch dispatchers are replaced by gin's radix-tree routing, and the handlers are native `gin.HandlerFunc`s.

### Router layer
- `gorilla/mux` → `gin-gonic/gin` (release mode). Routes registered directly via `RegisterCubeRoutes` / `RegisterInnerRoutes` / `RegisterNotifyRoutes` / `RegisterMetaRoutes`.
- Request middleware (logging / auth / recovery / trace) mounted on a root group so **matched routes** keep it, while **NoRoute / NoMethod stay bare** — gorilla/mux does not run router middleware on unmatched/method-mismatch requests, so neither does this (probes/mistyped URLs don't enter `checkAuth`).

### Handler layer
- All handlers are `func(c *gin.Context)`, reading path params via `c.Param` (build_id / snapshot_id / operation_id / sandbox_id / filename) — the manual path-parsing helpers are gone.
- **Standardized response writes**: success-path responses in cube/inner/notify go through `common.WriteAPI` / `common.WriteListAPI` (HTTP 200 + the JSON envelope serialized via **FastestJsoniter**, not gin's `encoding/json`), so no handler hand-rolls `c.Writer` + Content-Type. The `meta` package keeps its own error writer for its intentional HTTP-400 paths (malformed request body), which intentionally differ from the cube contract (HTTP 200 + ret_code); `common.WriteErr` is a convenience error-envelope helper currently exercised by the common package's contract tests rather than the hot path.

### Backward compatibility (preserved + tested)
- Response envelope `{ret:{ret_code,ret_msg},requestID}` and **HTTP 200** (errors carried in `ret_code`) — unchanged.
- **FastestJsoniter** serialization — unchanged (Content-Type is exactly `application/json`, not gin's default `application/json; charset=utf-8`).
- NoRoute → plain **HTTP 404**; method-mismatch → **HTTP 405** (empty body). gin additionally emits an RFC-7231 `Allow` header listing the allowed methods (mux's 405 did not) — intentional, locked by test.
- `POST /internal/fake_create` kept (legacy `200 + {ret_code:-1, Not Found}`).
- URL paths, auth flow, error codes — unchanged.

Two intentional deltas vs a literal mux mirror:
- `GET /cube/snapshot/storage` is registered as an explicit static route ahead of `/cube/snapshot/:snapshot_id` (safer than the old path-parse-inside-handler approach).
- gin's 405 response includes an RFC-7231 `Allow` header (mux's 405 did not).

### Tests
- Request-level contract tests: bare 404/405 body+status, middleware skipped on 404/405 (no auth), static-priority snapshot/storage, method-agnostic `/internal/{query,ws}`.
- Engine-level contract tests: `WriteAPI`/`WriteListAPI`/`WriteErr` emit HTTP 200 + `application/json` (FastestJsoniter) + the envelope; `fake_create` envelope test.
- meta/notify request-level tests (gomonkey, run with `-gcflags=all=-l`): meta `node_id` path param, labels `node_id`+`key` query, isolation PUT/DELETE, malformed register body → 400; notify `/notify/host` good/bad body, `/notify/health`. Lock the intentional notify-200-on-bad-body and meta-400+MasterInternalError contract details.
- Gin routing micro-benchmark + PVM A/B (see comments): no creation-latency regression vs mux at c=1/c=10.

Verified: `go build ./...`, `go vet`, `go test ./pkg/server/... ./pkg/service/httpservice/...`, `gofmt` all clean.

Signed-off-by: sparkzky <sparkhhhhhhhhhh@outlook.com>
Assisted-by: Orca:zai/glm-5.2


fslongjin:
Need some performance tests.

CubeMaster/docs/gin-refactor-plan.md
fslongjin, line 1: remove this

docs/zh/blog/posts/2026-06-03-cubesandbox-perf-benchmark-pvm.md
fslongjin, line 1: don't change this file

CubeMaster/pkg/service/httpservice/middleware/gin_middleware.go
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) 2024 Tencent Inc.
```
fslongjin: 2026

CubeMaster/go.mod
```diff
@@ -13,6 +13,7 @@ require (
 	github.com/charmbracelet/x/term v0.2.2
 	github.com/containerd/containerd v1.7.32
 	github.com/fsnotify/fsnotify v1.9.0
+	github.com/gin-gonic/gin v1.10.0
```
chenhengqi, line 16: With this, can we remove the `github.com/gorilla/mux` dependency?

sparkzky:
## Performance test: gorilla/mux → gin-gonic/gin (PVM A/B)

To address the "need some performance tests" comment, I ran a controlled A/B on a PVM node: same machine, same template, same Go toolchain — isolating **only** the gin-refactor commits. Each binary was swapped in via `systemctl stop → cp → start`, verified by sha256 + health + a smoke create before each run, and the original restored after.

**Node:** PVM kernel `6.6.69-opencloudos9.cubesandbox.pvm.host`, 16 vCPU, 15 GiB RAM · **Template:** `stable-python` (2 vCPU / 2 GiB) · **Tool:** `cube-bench`, create-delete, 2 warmup rounds.

### CREATE latency (ms)

| concurrency | variant | avg | P50 | P95 | P99 | success |
|:---:|---|---:|---:|---:|---:|---:|
| 1  | master (mux) | 78.7  | 77.2  | 83.7  | 85.0  | 100% |
| 1  | gin          | 77.7  | 76.2  | 86.6  | 88.1  | 100% |
| 10 | master (mux) | 143.6 | 142.9 | 178.0 | 189.9 | 100% |
| 10 | gin          | 145.3 | 143.0 | 182.8 | 200.0 | 100% |

**Δ (gin − master):** avg **−1.0 ms** at c=1, **+1.7 ms** at c=10 (≤ 1.3%); P50 effectively identical.

### Conclusion
- **c=1 / c=10 (100% success on both): no measurable regression.** Differences are within run-to-run noise on a shared node.
- This matches first principles: gin route lookup ≈ **4.8 µs/op** (`BenchmarkGin*`, 31 allocs) vs sandbox creation ≈ 67–150 ms → routing is < 0.01% of end-to-end creation.
- c=20 hits this smaller (15 GiB) node's scheduler memory-quota ceiling (`SelectNodesNoRes`) for **both** builds — a capacity limit, not a framework effect. (The published 32 GiB benchmark sustains c=20 at 100%.)

Net: the HTTP-layer migration from `gorilla/mux` to `gin-gonic/gin` does not regress sandbox creation latency or throughput.

fslongjin:
It looks like the switch to gin has some impact on p99 latency under concurrent load. I also noticed that gin mode doesn't seem to be explicitly set to release — would you mind setting `gin.SetMode(gin.ReleaseMode)` in the code and re-running the benchmark?

sparkzky:
## Update: re-benchmark with gin release mode + interleaved rounds

Thanks for the catch — gin was running in **debug mode** (default), which adds route-printing and debug-writer overhead. Fixed in c2c2d6d (`gin.SetMode(gin.ReleaseMode)`).

I also re-ran with **interleaved rounds** (master, gin, master, gin) to cancel the run-order resource-accumulation bias that inflated the earlier numbers. PVM node, same template/tooling, 2 rounds each, averaged.

### CREATE latency (ms) — interleaved, release mode

**c=1 (n=30/round):**

| round | master avg | gin-rel avg |
|---|---:|---:|
| r1 | 77.9 | 79.2 |
| r2 | 79.2 | 77.3 |
| **mean** | **78.6** | **78.2** |

**c=10 (n=200/round):**

| round | master avg | gin-rel avg | master P99 | gin-rel P99 |
|---|---:|---:|---:|---:|
| r1 | 145.6 | 142.4 | 211.6 | 190.5 |
| r2 | 143.7 | 146.7 | 197.7 | 203.5 |
| **mean** | **144.6** | **144.6** | **204.6** | **197.0** |

### Delta (gin-release − master)
| conc | Δ avg | Δ P99 |
|---|---|---|
| 1  | **−0.4 ms (−0.5%)** | −1.5 ms |
| 10 | **0.0 ms (0.0%)** | −7.6 ms |

### Conclusion
With release mode and run-order bias removed, **gin matches gorilla/mux within noise** — c=10 average is identical (144.6 ms), c=1 within 0.5%. The earlier apparent gap was debug-mode overhead + gin running second in a non-interleaved sequence. No regression.

CubeMaster/pkg/server/server.go
```diff
@@ -48,82 +52,73 @@ func New(ctx context.Context, cfg *config.Config) (*Server, error) {
...
 func NewInternalHttp(ctx context.Context, cfg *config.Config) (*internalHttp, error) {
 	if cfg == nil || cfg.Common == nil {
 		return nil, errors.New("config is nil")
 	}
 
-	router := mux.NewRouter()
+	gin.SetMode(gin.ReleaseMode)
+	engine := gin.New()
+	engine.RedirectTrailingSlash = false
+	engine.HandleMethodNotAllowed = true
```
fslongjin, line 65: **[blocking][api-contract]** Please don’t enable `HandleMethodNotAllowed` unless the response matches the old mux contract.

gorilla/mux (no custom `MethodNotAllowedHandler`) serves **HTTP 405** with an empty body, and **does not** attach middleware on `ErrMethodMismatch` (`MatchErr != nil` → middleware skipped; see mux `Match()` / `methodNotAllowedHandler`).

With `HandleMethodNotAllowed = true` + the `NoMethod` handler below, wrong-method requests become **HTTP 200 + JSON `MasterParamsError`** and still go through `GinRequestMiddleware` (auth). That is an external API contract change / regression.

Please either:
1. set `HandleMethodNotAllowed = false` and keep overall NotFound/405 parity with mux, or
2. keep explicit 405 handling but respond with **HTTP 405**, **no auth**, mux-compatible body.

Add a regression test, e.g. `GET /cube/sandbox` (only POST/DELETE registered).

CubeMaster/pkg/server/server.go
```diff
@@ -48,82 +52,73 @@ func New(ctx context.Context, cfg *config.Config) (*Server, error) {
...
+		if rt != nil {
+			rt.RetCode = -1
+		}
+		common.WriteResponse(c.Writer, http.StatusOK, &types.Res{
+			Ret: &types.Ret{
+				RetCode: -1,
+				RetMsg:  http.StatusText(http.StatusNotFound),
+			},
+		})
+	})
```
fslongjin, line 69: **[blocking][api-contract]** This `NoRoute` handler rewrites the public contract for unmatched paths.

Old mux (no custom `NotFoundHandler`):
- `http.NotFoundHandler()` → **HTTP 404**, body `404 page not found`
- router middleware **does not run** on 404 (mux `TestMiddlewareNotFound`; middleware only when `MatchErr == nil`)

Current:
- **HTTP 200** + JSON `{ret:{ret_code:-1,ret_msg:"Not Found"}}`
- `Engine.Use` combines into `allNoRoute` → **auth/mock/trace run first**

With `AuthConf.Enable=true`, probing an unknown path used to be plain 404; now it can surface as **AuthFailed**. Clients/monitors that key off HTTP status also break.

This PR must not introduce that change. Please restore mux-compatible NotFound:
- HTTP **404** (plain text like mux is fine; switching to JSON 200 would still be a contract change — keep that out of this router PR)
- **do not** run `checkAuth` on unmatched routes

Table-driven Auth on/off × unknown path tests should lock this.

**[should]** After restoring mux-compatible 404/405 (status + no auth), consider extracting the NotFound/MethodNotAllowed write path to one helper (e.g. under `common`) so future status-policy edits don’t drift. Extracting a helper ≠ switching to the business JSON-200 envelope — contract first, DRY second.

CubeMaster/pkg/server/server.go
```diff
@@ -48,82 +52,73 @@ func New(ctx context.Context, cfg *config.Config) (*Server, error) {
...
+		if rt != nil {
+			rt.RetCode = int64(errorcode.ErrorCode_MasterParamsError)
+		}
+		common.WriteResponse(c.Writer, http.StatusOK, &types.Res{
+			Ret: &types.Ret{
+				RetCode: int(errorcode.ErrorCode_MasterParamsError),
+				RetMsg:  http.StatusText(http.StatusMethodNotAllowed),
+			},
+		})
+	})
```
fslongjin, line 75: **[blocking][api-contract]** Same issue as NotFound: folding method-mismatch into the business envelope (`HTTP 200` + `ErrorCode_MasterParamsError`) changes external semantics and conflates “wrong HTTP method” with “bad request params”.

Please keep method-mismatch **HTTP 405** and **outside auth**, matching mux. If the product wants “always HTTP 200 JSON”, that is a separate intentional API redesign — not this router migration.

CubeMaster/pkg/server/server.go
```diff
@@ -48,82 +52,73 @@ func New(ctx context.Context, cfg *config.Config) (*Server, error) {
...
-	metaGroup.HandleFunc(metahttp.RegisterNodeAction(), metahttp.RegisterNodeHandler).Methods(http.MethodPost)
-	metaGroup.HandleFunc(metahttp.NodesAction(), metahttp.ListNodesHandler).Methods(http.MethodGet)
-	metaGroup.HandleFunc(metahttp.VersionMatrixAction(), metahttp.VersionMatrixHandler).Methods(http.MethodGet)
-	metaGroup.HandleFunc(metahttp.NodeAction(), metahttp.GetNodeHandler).Methods(http.MethodGet)
-	metaGroup.HandleFunc(metahttp.NodeStatusAction(), metahttp.UpdateNodeStatusHandler).Methods(http.MethodPost)
-	metaGroup.HandleFunc(metahttp.NodeLabelsAction(), metahttp.UpdateNodeLabelsHandler).Methods(http.MethodPost)
-	metaGroup.HandleFunc(metahttp.NodeLabelsAction(), metahttp.DeleteNodeLabelHandler).Methods(http.MethodDelete)
+func (s *internalHttp) registerRoutes() {
+	r := s.engine
+	r.Use(middleware.GinRequestMiddleware())
```
fslongjin: **[blocking][api-contract]** In gin, `Engine.Use` rebuilds `allNoRoute` / `allNoMethod` via `combineHandlers`, so global middleware **does** run for 404/405 handlers. gorilla/mux only wraps middleware when `MatchErr == nil`.

Whatever NotFound/MethodNotAllowed strategy we pick, **unmatched / method-mismatch requests must not enter `checkAuth`**, matching mux. Options:
- attach auth only on business `Group`s; leave Engine-level 404/405 ungated, or
- short-circuit in middleware before auth when handling no-route / no-method, or
- register 404/405 handlers that are not wrapped by the auth middleware

Otherwise merely enabling auth changes the observable contract for probes and mistyped URLs — a real regression.

CubeMaster/pkg/service/httpservice/cube/gin_adapter.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/gin-gonic/gin"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/common"
+	"github.com/tencentcloud/CubeSandbox/cubelog"
+)
+
+// Adapt wraps a handler returning interface{}, writing the response via
+// common.WriteResponse. Streaming handlers return nil on success (they already
+// wrote directly to the ResponseWriter); on error they return a *types.Res.
+func Adapt(fn func(http.ResponseWriter, *http.Request, *CubeLog.RequestTrace) interface{}) gin.HandlerFunc {
```
fslongjin: **[blocking][native-gin]** Please do not land `Adapt` / `AdaptList` as the merge shape.

This is a half-migration: the route table is gin, but handlers stay `(http.ResponseWriter, *http.Request, *RequestTrace)` and the adapter calls `WriteResponse`. Cost:
- we pay for the gin dependency tree while barely using `c.Param` / a unified middleware model
- `cube` / `notify`+`inner` / `meta` end up with three styles
- a second migration is deferred into mainline

Before merge (this PR or a must-land-first stacked PR):
1. convert cube (and notify/inner) handlers to native `gin.HandlerFunc`
2. delete `gin_adapter.go`
3. keep matched-API response semantics unchanged (FastestJsoniter + existing envelope / HTTP 200 + `ret_code`)

Keeping net/http signatures “to avoid touching unit tests” is not an acceptable merge reason — tests should move with the handlers.

CubeMaster/pkg/service/httpservice/cube/template_commit.go
```diff
@@ -212,3 +211,12 @@ func handleTemplateBuildStatusAction(w http.ResponseWriter, r *http.Request, rt
...
 	}
 }
+
+// TODO: [shortcut] Temporary manual path param extraction because handler
+// signature stays (w,r,rt) to preserve unit test compatibility.
+// Future migration to *gin.Context can use c.Param("build_id").
+func buildIDFromPath(path string) string {
+	prefix := actionURI(TemplateBuildStatusAction) + "/"
+	trimmed := strings.TrimPrefix(path, prefix)
+	return strings.TrimSuffix(trimmed, "/status")
```
fslongjin: **[blocking][native-gin]** The route already declares `:build_id`, but this PR adds `buildIDFromPath` (`TrimPrefix` + `TrimSuffix("/status")`) — a second source of truth and a direct artifact of the Adapt bridge.

Please switch to `c.Param("build_id")` after handlers are native gin, and delete this shortcut. Same class of debt:
- `resourceIDFromPath` (snapshot / operation)
- `sandboxIDFromRollbackPath`
- CA `filepath.Base(r.URL.Path)` → should be `c.Param("filename")`

Path params should be read from gin once only, so prefix/shape changes cannot silently mis-parse.

CubeMaster/pkg/service/httpservice/cube/routes.go
```diff
@@ -0,0 +1,64 @@
...
+
+import "github.com/gin-gonic/gin"
+
+// RegisterCubeRoutes registers all /cube routes onto the given gin.RouterGroup.
+// The method/path registrations mirror the original gorilla/mux wiring in
+// server.go:86-110 exactly — only handlers that were registered there are
+// registered here.
+func RegisterCubeRoutes(g *gin.RouterGroup) {
+	// Sandbox CRUD
+	g.POST(SandboxAction, Adapt(handleSandboxAction))
```
fslongjin: **[blocking][native-gin][docs]** Two asks on this file:

1) **Unify on native gin** (meta’s style). Today:
| Package | Style |
|---|---|
| `cube` | `Adapt(handleXxx)` |
| `notify` / `inner` | hand-written `*gin.Context` wrappers |
| `meta` | native gin + `c.Param` |

Please make cube/notify/inner match meta: `Register*Routes` mounts `gin.HandlerFunc`s, use `c.Param` / `c.Query` / `c.Request`, write via existing `common.WriteResponse` (thin helper OK). Matched API wire semantics must stay unchanged.

2) The “mirror … exactly” comment is inaccurate:
- this file **adds** explicit `GET SnapshotStorageAction` (good/safer than mux param+switch, but not “exact wiring”)
- `POST /internal/fake_create` was on mux and is omitted under `inner.RegisterInnerRoutes`

Please reword to “parity with previous observable behavior” and list known deltas.

CubeMaster/pkg/server/server_test.go
```diff
@@ -6,21 +6,34 @@ package server
...
 )
 
-func TestRegisterHandlersIncludesCADownloadRoute(t *testing.T) {
-	s := &internalHttp{router: mux.NewRouter()}
-	s.registerHandlers()
+// TestCADownloadRouteRegistration verifies that the CA download route is
+// registered for both GET and HEAD. It inspects the gin route table directly
+// rather than executing a request, so no middleware runs and no config/redis
+// initialization is required.
+func TestCADownloadRouteRegistration(t *testing.T) {
```
fslongjin: **[blocking][tests]** Scanning `engine.Routes()` only proves registration, not request-level contract or native-gin param behavior. The old test used `httptest` + `mux.Match`.

Please add request-level table-driven coverage (httptest + minimal config / auth mock) at least for:
1. Auth on/off × unknown path → status/body/whether auth runs (mux parity)
2. Auth on/off × wrong method on a registered path → mux parity
3. `GET /cube/snapshot/storage` hits storage, not `:snapshot_id == "storage"`
4. param routes after native gin (`c.Param`): rollback / build status / CA filename
5. CA/artifact streaming success does not double-`WriteResponse`; errors still use JSON envelope
6. `/internal/ws` and `/internal/query` remain method-agnostic (`Any`)

A route-table smoke test can stay, but it must not be the only guard for this refactor.

CubeMaster/pkg/service/httpservice/inner/routes.go
```diff
@@ -0,0 +1,43 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/httpservice/common"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+	"github.com/tencentcloud/CubeSandbox/cubelog"
+)
+
+func RegisterInnerRoutes(g *gin.RouterGroup) {
+	g.GET(NodeAction, nodeGinHandler)
+	g.Any(StateWs, websocketGinHandler)
+	g.Any(StateQuery, queryGinHandler)
```
fslongjin, line 17: **[should]** mux registered `POST` `FakeCreateAction` (`/fake_create`); this registration is gone. Old `HttpHandler` also had no real case (JSON NotFound after middleware).

Under zero-contract / clear-diff rules, please either:
- keep registering it and preserve the old observable behavior for that path, or
- delete the leftover `FakeCreateAction` constant in the same PR and call out the removal in the PR description

Constant present + route absent is easy to misread later.

CubeMaster/pkg/service/httpservice/middleware/gin_middleware.go
```diff
@@ -0,0 +1,115 @@
...
+			if err := recover(); err != nil {
+				log.G(ctx).Fatalf("HandlerFunc panic:%s", string(debug.Stack()))
+				common.WriteResponse(c.Writer, http.StatusOK, &types.Res{
+					Ret: &types.Ret{
+						RetCode: -1,
+						RetMsg:  http.StatusText(http.StatusInternalServerError),
+					},
+				})
+			c.Abort()
+			}
```
fslongjin, line 68: **[nit]** Pre-existing Fatalf-on-panic behavior is inherited; `WriteResponse` + `c.Abort()` after `Fatalf` are effectively dead. While touching this file, please fix the `c.Abort()` indentation. Not blocking vs the contract / native-gin bars.

fslongjin (changes requested):
Thanks for the mux → gin migration and the PVM A/B data.

Per review, this PR should not merge until **both** bars are met:

### 1) Zero external API contract change vs gorilla/mux
| Case | Old (mux) | Current (this PR) |
|---|---|---|
| Unknown path | HTTP **404** text; **middleware/auth not run** | HTTP **200** JSON `ret_code=-1`; **middleware + auth run** |
| Wrong method | HTTP **405**; middleware not run | HTTP **200** JSON `MasterParamsError`; middleware + auth run |

“All backward compatibility contracts preserved” is not accurate for these paths. Please restore mux-compatible status + auth coverage, with Auth on/off × NotFound × MethodNotAllowed tests.

### 2) Native gin — no half-migration
| Package | Today |
|---|---|
| `cube` | `Adapt`/`AdaptList` over `(w,r,rt)` |
| `notify` / `inner` | hand-written gin wrappers calling old funcs |
| `meta` | native gin + `c.Param` |

Routes already declare `:param`, but handlers still parse `r.URL.Path` (`buildIDFromPath`, etc.). Please finish native gin before merge: one `gin.HandlerFunc` style, `c.Param` only, delete `gin_adapter.go`, keep FastestJsoniter + existing envelope for **matched** APIs unchanged.

Happy to re-review after both are addressed. Inline comments map to specific lines.

---
**Must before merge**
- [ ] 404: HTTP 404 + no auth (mux parity)
- [ ] 405: HTTP 405 + no auth (mux parity); or disable HandleMethodNotAllowed with equivalent parity
- [ ] Delete cube Adapt/AdaptList and gin_adapter.go
- [ ] Unify cube/notify/inner/meta on gin.HandlerFunc
- [ ] Path params via c.Param only; delete buildIDFromPath and siblings
- [ ] Matched business APIs: envelope / status 200+ret_code / FastestJsoniter unchanged
- [ ] Request-level tests: Auth×NotFound×MethodNotAllowed + key param routes + CA streaming

sparkzky:
## Re: review round 2 — all blocking items addressed

Pushed `ed8007e` + `26cde2b`. Each point below maps to a comment above.

### [blocking][api-contract] — server.go (`ed8007e`)
- **NoRoute → bare 404.** Now `http.NotFound(c.Writer, c.Request)` (plain `404 page not found`), matching mux's `http.NotFoundHandler()`.
- **NoMethod → bare 405.** `c.AbortWithStatus(http.StatusMethodNotAllowed)` (empty body), matching mux's `methodNotAllowedHandler()` for `ErrMethodMismatch`.
- **No auth on unmatched/method-mismatch.** Moved `GinRequestMiddleware` off the engine (`r.Use`) onto a root group, so matched routes keep logging/auth/recovery while the engine-level NoRoute/NoMethod stay bare — `checkAuth` no longer runs on probes/mistyped URLs.
- Extracted `newEngine()` so the policy is shared with tests.

### [blocking][native-gin] — cube handlers (`26cde2b`)
- **Deleted the `Adapt`/`AdaptList` bridge.** All 24 cube top-level handlers are now native `func(c *gin.Context)`, matching the `meta` style (handlers read `c.Param`/`c.Query`/`c.Request`, write via `common.WriteResponse`). `gin_adapter.go` removed.
- **Path params via `c.Param`** (removed the manual parsers): `build_id`, `snapshot_id`, `operation_id`, `sandbox_id`, `filename` (CA). Deleted `buildIDFromPath`/`resourceIDFromPath`/`sandboxIDFromRollbackPath`.
- **Streaming handlers** (CA / artifact / rootfs download) use a bare `return` after writing directly; error paths keep the JSON envelope. (`filepath.Base(record.Ext4Path)` in `ServeContent` is on a file path, not the URL — left as-is.)
- **routes.go comment** reworded to "parity with previous observable behavior" and notes the two deltas (explicit `GET /cube/snapshot/storage`; mux `POST /internal/fake_create` not registered here). notify/inner were already native gin — unchanged.

### [blocking][tests]
- **Request-level contract tests** added in `server_test.go` (httptest through the real engine): bare 404 body+status, bare 405 empty body, middleware-skipped on 404/405 (sentinel), static-priority `/cube/snapshot/storage`, method-agnostic `/internal/{query,ws}`.
- **Param/streaming behavior** covered by updated cube unit tests driving handlers via `gin.CreateTestContext` + `c.Params`.

### [should] / [nit]
- Deleted the dead `inner.FakeCreateAction` constant (route was already gone, no consumers).
- Fixed the `c.Abort()` indentation in the gin panic-recovery middleware.

`go build ./...`, `go vet` (migrated packages), `go test ./pkg/server/... ./pkg/service/httpservice/...`, and `gofmt` are all clean. Response semantics (JSON envelope, HTTP 200, FastestJsoniter, error codes) unchanged.

fslongjin (changes requested):
I think this needs some changes. This PR introduced Gin but didn't leverage its native capabilities to reduce code complexity. Our original intention for introducing Gin was to make the code simpler, avoiding manual URL path matching and structured JSON response writing. This PR brings in Gin, yet it still retains a lot of old coding patterns—halfway done, not fully refactored. Additionally, some API semantics have been altered or are missing. I believe further modifications are needed.

fslongjin (commented):
Left two suggestions below — neither is a hard blocker. Happy to +1 once the description/tests story is clarified (in this PR or via a follow-up).

CubeMaster/pkg/service/httpservice/common/gin_response.go
```diff
@@ -0,0 +1,39 @@
...
+	"github.com/gin-gonic/gin"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
+)
+
+// WriteAPI writes a standard API response onto a gin.Context: HTTP 200 with the
+// JSON envelope serialized via FastestJsoniter (NOT gin's encoding/json). It is
+// the single write path for success/error responses so handlers stop hand-rolling
+// c.Writer + Content-Type + status. The request trace (rt.RetCode) is still set
+// by the caller, since some handlers record a sentinel value for metrics.
+func WriteAPI(c *gin.Context, data interface{}) {
```
fslongjin, line 23: nit / suggestion: the PR says every response goes through `WriteAPI` / `WriteListAPI` / `WriteErr`, but `meta` error paths still go `writeErr(c.Writer, …)` → `WriteResponse`, and `WriteErr` doesn’t seem to be used in production at all — only in unit tests.

Also, the comment here calling `WriteAPI` “the single write path” doesn’t quite match reality.

I’m not saying we have to fully converge this in this PR — `meta` still returns HTTP 400 on parse failure, which is already different from most cube APIs (`HTTP 200` + `ret_code`). Forcing everything through `WriteErr` could actually change behavior for the worse.

Would prefer one of these (or a bit of both):

1. Tone down the PR description / comments: drop “every” / “single”; clarify whether `WriteErr` is a reserved helper or something we plan to delete; call out that meta’s 400 is intentional.
2. If we do want to unify this round, fold `meta.writeErr` into a gin-aware helper and spell out which errors stay 400 vs go 200.

Otherwise the next person who touches the response contract is very likely to miss meta.

CubeMaster/pkg/service/httpservice/meta/routes.go
```diff
@@ -0,0 +1,22 @@
...
+	g.POST(registerNodeAction, registerNodeGinHandler)
+	g.GET(nodesAction, listNodesGinHandler)
+	g.GET(versionMatrixAction, versionMatrixGinHandler)
+	g.GET(nodeAction, getNodeGinHandler)
+	g.POST(nodeStatusAction, updateNodeStatusGinHandler)
+	g.POST(nodeLabelsAction, updateNodeLabelsGinHandler)
+	g.DELETE(nodeLabelsAction, deleteNodeLabelGinHandler)
+	g.PUT(nodeIsolationAction, isolateNodeGinHandler)
+	g.DELETE(nodeIsolationAction, unisolateNodeGinHandler)
+}
```
fslongjin, line 22: The cube / server contract tests look solid (404/405, snapshot/storage priority, `fake_create`, etc.). But `meta` and `notify` are still no-test packages — after the gin migration, things like `c.Param` / `c.Query`, body parse, and health `RetCode` are basically only covered by compile + the route table.

Regression cost here isn’t cheap. Would be good to add at least a few request-level cases:

- meta: `node_id` path params, labels query, isolation PUT/DELETE, bad register body → 400
- notify: `/notify/host` good/bad body, `/notify/health`

Doesn’t need to be exhaustive — just pin down “where params come from after the migration.” If it doesn’t fit this PR, a follow-up issue is fine; please leave a pointer.

sparkzky:
## meta/notify request-level tests added (`04ab71a`)

Addressing the nit on `meta/routes.go:22` — meta and notify were no-test packages. Added 9 request-level tests (`meta/meta_test.go`, `notify/notify_test.go`) that pin **where params come from after the gin migration**.

**Approach**: register the real routes (`RegisterMetaRoutes`/`RegisterNotifyRoutes`) on a test gin engine + inject a request trace (`cubelog.WithRequestTrace`, like `inner/fakecreate_test.go`), and `gomonkey`-patch the DB-backed business funcs (`nodemeta.*`, `hostChangeNotify`) so no cluster/DB is needed. Run with `-gcflags=all=-l` (repo convention — gomonkey needs inlining disabled). Each patch captures the args it received, so the tests prove `c.Param`/`c.Query`/body-parse actually flow through, not just that something compiled.

**Coverage**
- meta: `node_id` path param → `nodemeta.GetNode`; labels `node_id`+`key` query → `DeleteNodeLabel`; isolation PUT → `SetNodeSchedulingDisabled(true)`, DELETE → `(false)`; malformed register body → **400**; valid body → `RegisterNode`.
- notify: `/notify/host` good body → 200; bad body → 200+MasterParamsError; `/notify/health` → 200.

All 9 pass under `-gcflags=all=-l`; `go vet` + `gofmt` clean; no non-test files changed.

**Two contract details these tests now lock (so reviewers know they're intentional, not drift):**
1. **notify returns HTTP 200 (not 4xx) on a malformed `/notify/host` body** — `hostChangeGinHandler` writes the MasterParamsError envelope via `WriteAPI`, which uses HTTP 200. This mirrors the old mux behavior and is intentionally different from meta.
2. **meta's malformed `/internal/meta/nodes/register` body yields HTTP 400 + `MasterInternalError` (130593)** — `meta.writeErr` only special-cases `gorm.ErrRecordNotFound` and the two label errors; a generic JSON parse error falls through to `MasterInternalError` while still emitting 400. (If we'd prefer `MasterParamsError` there, that's a separate cleanup — flagged, not changed in this test-only commit.)

fslongjin (approved):

PR #831, opened 2026-07-08
fix(cubeapi): return conflict for paused sandbox delete

ihanzh:
## Summary

Fixes #219.

This PR changes CubeAPI sandbox deletion behavior so deleting a paused sandbox
returns `409 Conflict` instead of surfacing a CubeMaster internal error.

When CubeMaster reports the paused-delete backend error, CubeAPI now checks the
sandbox state. If the sandbox is paused, the API returns a client-actionable
conflict telling the caller to resume the sandbox before deleting it. Other
CubeMaster internal errors keep the existing internal-error behavior.

## Changes

- Map paused sandbox delete failures to `AppError::Conflict`.
- Preserve the existing not-found mapping for missing sandbox deletion.
- Add focused CubeAPI coverage for the paused-delete conflict path.

## Validation

Unit tests — CubeAPI sandbox delete path

```bash
cargo test --manifest-path CubeAPI/Cargo.toml kill_sandbox -- --nocapture
```

```text
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.30s
     Running unittests src/main.rs (target/debug/deps/cube_api-580ae8bdea1a4d0b)

running 2 tests
test services::sandboxes::tests::kill_sandbox_returns_conflict_for_paused_delete_state_error ... ok
test services::sandboxes::tests::kill_sandbox_maps_cubemaster_not_found_to_app_not_found ... ok

test result: ok. 2 passed; 0 failed; 0 ignored; 0 measured; 80 filtered out; finished in 1.55s
```

Broader sandbox service regression tests:

```bash
cargo test --manifest-path CubeAPI/Cargo.toml services::sandboxes::tests -- --nocapture
```

```text
    Finished `test` profile [unoptimized + debuginfo] target(s) in 0.14s
     Running unittests src/main.rs (target/debug/deps/cube_api-580ae8bdea1a4d0b)

running 23 tests
test services::sandboxes::tests::create_sandbox_accepts_valid_env_var_names ... ok
test services::sandboxes::tests::create_sandbox_request_serializes_lifecycle_flags ... ok
test services::sandboxes::tests::create_sandbox_rejects_invalid_env_var_value ... ok
test services::sandboxes::tests::create_sandbox_rejects_dangerous_env_var_names ... ok
test services::sandboxes::tests::create_sandbox_rejects_invalid_env_var_name_format ... ok
test services::sandboxes::tests::create_sandbox_request_timeout_wire_shape ... ok
[...]

PR #940, opened 2026-07-13
fix: delete paused sandboxes transparently

### LINKED ISSUE #219: [Bug Report] DELETE /sandboxes/<id> returns 500 for paused sandboxes
## Description

Attempting to delete a paused sandbox via `DELETE /sandboxes/<id>` returns HTTP 500 with error message: `"sandbox not in normal state"`.

## Steps to Reproduce

1. Create a sandbox
2. Pause it via `POST /sandboxes/<id>/pause`
3. Confirm state is `paused` via detail API
4. Call `DELETE /sandboxes/<id>`

## Expected Behavior

Delete should succeed (HTTP 204) regardless of sandbox state, or return a clear 4xx error with guidance (e.g., "resume first, then delete").

## Actual Behavior

```json
{"code": 500, "message": "sandbox not in normal state"}
```

## Workaround

Resume the sandbox first via `POST /sandboxes/<id>/connect`, then delete:
```bash
curl -X POST http://host:3000/sandboxes/<id>/connect
curl -X DELETE http://host:3000/sandboxes/<id>  # now returns 204
```

## Environment

- CubeSandbox master branch (2026-05-13)
- Tested on cubesandbox002

zyl1121:
## Motivation

Deleting a paused sandbox currently fails because the runtime rejects the normal delete operation while the VM is paused. Users must manually resume the sandbox before deleting it.

This PR makes `DELETE /sandboxes/{sandboxID}` work for a stable paused CubeBox. The sandbox is resumed as part of the delete operation and then removed through the existing destroy workflow. DELETE remains synchronous: success is returned only after cleanup completes.

## Design

CubeMaster remains the external DELETE entry point and continues to coordinate both synchronous and asynchronous destroy requests. The atomic resume-before-destroy transition is implemented in Cubelet's `Destroy` path.

Cubelet owns the authoritative runtime state, containerd task, and the per-sandbox lifecycle lock shared by pause, resume, rollback, and commit. Only paused or pausing Cube runtime deletes enter this bounded gate, so running deletes retain their existing direct path. Keeping the transition in Cubelet keeps state inspection, resume, reconciliation, and normal cleanup together.

This supersedes the API-only paused-delete workaround in #831. That workaround translated a CubeMaster internal error into `409 Conflict` after an extra sandbox-status lookup. This PR removes that fallback: DELETE now performs the bounded wake-up itself and returns typed capacity or retryable lifecycle errors only when the actual wake-up cannot proceed.

## Behavior

- A stable `PAUSED` sandbox is resumed internally, then follows the normal destroy workflow.
- A `RUNNING` sandbox keeps the existing delete path and does not resume.
- If a resume RPC reports an error but runtime reconciliation proves the VM is `RUNNING`, DELETE continues. Otherwise the sandbox remains retryable.
- The internal wake-up is bounded to 5 seconds, with 20 seconds reserved for cleanup and 5 seconds reserved for the Cubelet RPC response. The existing CubeAPI 30-second request timeout is unchanged.
- DELETE does not emit a caller-visible `sandbox.resumed` event, reset idle timeout, or create a background retry.

## Error Semantics

DELETE for running sandbox and existing `404` behavior are unchanged. For the paused-delete path:

- Capacity or missing resource error returns `409 Conflict`.
- `PAUSING` or another lifecycle operation in progress returns `503` with `Retry-After: 2`.
- A failed or unproven internal resume, or insufficient remaining Cubelet RPC budget, returns `503` with `Retry-After: 5`.

`Retry-After: 5` follows CubeProxy's existing auto-resume failure backoff, while `Retry-After: 2` follows its PAUSING gate. These headers tell the client when to retry; DELETE is not continued or retried in the background.

## Scope

This PR does not add a public DELETE parameter or protobuf field, bypass resume admission, change the existing CubeShim delete guard, alter ordinary running DELETE behavior, or add a new metric.

## Validation

Focused Linux builder checks passed:

```bash
cd Cubelet && go test ./pkg/utils -run TestResourceLock -count=1
cd Cubelet && go test ./services/cubebox -run 'TestResumeTaskLocked|TestDeleteAutoResume|TestDestroy' -count=1
cd CubeMaster && go test ./pkg/service/sandbox -run TestSetSyncDestroyFailure -count=1
cd CubeAPI && cargo test --locked
cd CubeAPI && cargo test --locked delete_paused_sandbox_maps_business_errors_from_cubemaster
```

The full CubeAPI suite completed with `88 passed; 0 failed`. The focused HTTP route test verifies CubeMaster `130409` -> HTTP `409`, alongside the `503` and `Retry-After` cases.

Real-server validation built and atomically deployed the final Cubelet, CubeMaster, and CubeAPI binaries. All three services remained active and both health endpoints returned success.

Live E2E validation created a temporary sandbox, paused it, and deleted it directly without an explicit `resume` or `connect` request:

- pause returned `204`, and GET reported `paused`;
- DELETE returned `204 No Content` in `629 ms`;
- subsequent GET returned `404`;
- no matching containerd task or sandbox data path remained.

## References

Fixes #219.

Replaces the API-only paused-delete workaround from #831.


kinwin-ustc:
Good work. I noticed the automatic resume timeout is 5 seconds. If we support cross-node pause resume in the future, the resume time might be longer (due to retrieving snapshots from remote shared storage). Could these timeouts be made more configurable?

zyl1121:
> Good work. I noticed the automatic resume timeout is 5 seconds. If we support cross-node pause resume in the future, the resume time might be longer (due to retrieving snapshots from remote shared storage). Could these timeouts be made more configurable?

Thanks, that makes sense. The current 5-second resume timeout is actually part of the existing 30-second synchronous DELETE budget, which also has to cover the RPC call and cleanup. Exposing it as an independent setting could therefore create inconsistent deadlines or leave too little time for cleanup.

Making this configuration effective would require revisiting the timeout budget across the full DELETE path. That would involve the API timeout, the CubeMaster-to-Cubelet RPC deadline, the Cubelet cleanup budget, and the corresponding configuration and tests.

Since this PR is focused on fixing deletion of paused sandboxes within the existing synchronous DELETE behavior, I think adding an end-to-end configurable timeout model would broaden its scope considerably. I’d prefer to keep the current conservative budget here and track the timeout configuration work as a follow-up for cross-node pause/resume scenario.

Cubelet/services/cubebox/delete_resume.go
```diff
@@ -0,0 +1,196 @@
...
+
+const (
+	// CubeMaster gives its synchronous Cubelet Destroy RPC a 30-second budget.
+	// Keep at most five seconds for the internal wake-up, at least twenty
+	// seconds for normal cleanup, and five seconds for the RPC response to
+	// return to CubeMaster.
+	deleteAutoResumeMaxDuration = 5 * time.Second
+	deleteDestroyReserve        = 20 * time.Second
+	deleteResponseReserve       = 5 * time.Second
+	deleteLifecycleLockMaxWait  = 2 * time.Second
```
ls-ggg, line 30: The 5/20/5-second split ( deleteAutoResumeMaxDuration,  deleteDestroyReserve,  deleteResponseReserve) in  delete_resume.go is hardcoded in absolute seconds, but it implicitly assumes the caller's ctx deadline is ~30s (i.e. CubeMaster's  common_timeout_insec, currently defaulting to 30s). These two configs live in different services and can drift independently — if  common_timeout_insec is ever tuned down (e.g. to 10s) without updating these constants,  latestResumeDeadline will always fail the  !latestResumeDeadline.After(now) check, and every delete of a paused sandbox will short-circuit to the "not enough time" error path, not just occasionally but permanently.
Suggest deriving the resume budget
proportionally from the actual remaining time on  ctx ( deadline, _ := ctx.Deadline(); remaining := time.Until(deadline)) instead of subtracting fixed absolute reserves. e.g. reserve a percentage of the remaining budget for destroy/response rather than a fixed number of seconds, so this stays correct regardless of what  common_timeout_insec /  destroy_dead_line are configured to. At minimum this coupling should be documented so operators don't hit a silent, hard-to-diagnose regression when tuning timeouts.

Cubelet/services/cubebox/service.go
```diff
@@ -568,20 +570,81 @@ func (s *service) Destroy(ctx context.Context, req *cubebox.DestroyCubeSandboxRe
...
-			runtimeType = sb.OciRuntime.Type
+
+	// Preserve the existing destroy path's fresh state read after its deadline
+	// is established. Running deletes must not use the object observed before
+	// annotation handling and timeout setup.
+	destroyCtx := ctx
+	sb, ctx, getErr = s.loadDestroySandboxRuntime(destroyCtx, req.SandboxID)
+	ctx = log.WithLogger(ctx, stepLog)
+
+	needsPausedPreflight := getErr == nil && constants.IsCubeRuntime(ctx) &&
```
ls-ggg: needsPausedPreflight is computed from a status read taken before acquiring  sandboxLifecycleLocks. If the sandbox is observed as  RUNNING at that point (so  needsPausedPreflight == false), the request skips the lock
entirely and falls straight through to  engine.Destroy(...) unprotected. If a concurrent  Update(pause) wins  sandboxLifecycleLocks and pauses the sandbox in the gap between that read and the actual destroy call, this Destroy request ends up doing a plain destroy against what is now a paused sandbox — the exact scenario this PR is meant to guard against — and it does so without holding the lock at all.
This isn't a regression introduced by this PR (the old code read status unlocked too), but since the PR's stated goal is "protect destroy from racing with pause/resume/commit/rollback via the lock," it's worth calling out that the guarantee is currently best-effort rather than fully closed: the decision to take the lock is itself made outside the lock. Would it be worth acquiring  sandboxLifecycleLocks
unconditionally on entry (before reading status) and deciding the resume-vs-plain-destroy branch under the lock, rather than deciding whether to lock based on a stale read? If the current scope is intentionally narrower than that, a code comment noting the residual race would help future maintainers avoid assuming this is fully serialized.

ls-ggg:
Thanks for the PR! Please address the CI failures and the review comments above. Also, could you add more comprehensive e2e test cases (covering concurrency, PAUSING state, and resume timeouts)?

zyl1121:
> Thanks for the PR! Please address the CI failures and the review comments above. Also, could you add more comprehensive e2e test cases (covering concurrency, PAUSING state, and resume timeouts)?

@ls-ggg Thanks for the detailed review. Both inline issues have been addressed in the latest commit:

1. The paused-delete budget is now calculated from the actual Cubelet request deadline. The default 30-second path still allocates `5s` for resume, `20s` for cleanup, and `5s` for the response. For shorter RPC deadlines, DELETE now derives a smaller bounded budget instead of rejecting the request solely because the previous fixed reservations do not fit.
2. Cube runtime DELETE now acquires the per-sandbox lifecycle lock before making its final state decision, then reloads the sandbox while holding that lock. This closes the race where a sandbox could transition from `RUNNING` to `PAUSED` between the initial read and destruction.

Additional focused coverage now includes task-loading failures, reconciliation failures, unexpected shim states, and the fallback for empty CubeMaster error messages. The duplicated resume-task test fake has also been consolidated.

For validation, Cubelet was rebuilt and paused DELETE was verified against a real server. For concurrency validation, I ran five real-server trials with Pause and DELETE issued concurrently. In two trials, DELETE reached Cubelet while Pause still held the lifecycle lock and returned `503 Service Unavailable` with `Retry-After: 2`. Retrying after the advertised delay returned `204 No Content`. In the other three trials, DELETE did not overlap with the lock-holding window and completed successfully. All five sandboxes were ultimately deleted, with no remaining containerd resources or sandbox data.

The real-server validation covers normal paused deletion and concurrent Pause/DELETE behavior. The `PAUSING` state and Resume timeout paths are covered by service-level tests that set the relevant lifecycle state or inject failures into the resume and status operations, then verify the resulting DELETE response and retry behavior.

Stable full-stack E2E coverage for those two fault paths (`PAUSING` state and Resume timeout) cannot currently be added because CubeSandbox does not provide a CI-managed harness spanning `CubeAPI -> CubeMaster -> Cubelet`, or controlled fault-injection support for holding a sandbox in `PAUSING` or forcing a shim Resume timeout. Without that infrastructure, a real-server test would be timing-dependent and flaky. Once those capabilities are available, I would be happy to add the corresponding E2E cases.

The previous failed claude-response CI check was unrelated to this PR’s code or tests. It is the known workflow permission issue described in #844: normal review or comment activity can trigger the workflow, which then fails the actor permission check. All 10 PR checks were passing before this comment, as shown below.

image

ls-ggg:
> Without that infrastructure, a real-server test would be timing-dependent and flaky. Once those capabilities are available, I would be
Good job， And could you test this locally to confirm it works as expected? I’m planning to include this in the v0.6.0 release, which is currently targeted for July 24th

zyl1121:
> Good job, And could you test this locally to confirm it works as expected? I’m planning to include this in the v0.6.0 release, which is currently targeted for July 24th

Thanks. I reran the validation against the latest commit.

The focused unit and service-level tests pass. I also verified the full `CubeAPI -> CubeMaster -> Cubelet` path on a real server. A stable paused sandbox can be deleted successfully with `204 No Content`, and the related containerd resources are cleaned up.

I additionally validated the two retryable failure paths:

* For a `PAUSING` sandbox, I held the shim while Pause was in progress. Cubelet reported the sandbox as `PAUSING`, and `DELETE` returned the expected `503 Service Unavailable` with `Retry-After: 2`. After Pause completed, retrying `DELETE` returned `204`.

* For an internal Resume timeout, I paused the sandbox and temporarily held the shim during `DELETE`. The auto-resume path exhausted its bounded budget, and `DELETE` returned the expected `503 Service Unavailable` with `Retry-After: 5`. After releasing the shim, retrying `DELETE` returned `204`.

I also repeated the concurrent `Pause`/`DELETE` scenario five times. All sandboxes were eventually deleted successfully. One run returned the expected retryable `503` while Pause held the lifecycle lock; the other runs completed with `DELETE` winning the race.

ls-ggg:
LGTM， I ran some initial tests and they all passed.

Issue #244, opened 2026-05-14
[good first issue] docs: Help us build integration guides with mainstream AI agent frameworks


caoqianyun:
## What this issue is about

Cube Sandbox doesn't have an Integrations section yet. We'd love your help building one — so developers around the world can plug Cube into the AI agent frameworks they already use.

> 💡 Cube is compatible with the E2B SDK — in most cases, switching from E2B to Cube is just **a single env var change**, while gaining MicroVM-level isolation. That makes each integration guide pretty light to write.

## What you can contribute

An integration guide for one AI agent framework.

Candidate frameworks (pick one — first come, first served; you're also welcome to propose frameworks not listed here):

| Framework | Type |
| --- | --- |
| LangChain | Power your agent with Cube |
| LlamaIndex | Power your agent with Cube |
| Dify | Power your agent with Cube |
| AutoGen | Power your agent with Cube |
| CrewAI | Power your agent with Cube |
| Google ADK | Power your agent with Cube |
| Claude Code | Run your agent inside Cube |
| OpenClaw | Run your agent inside Cube |
| OpenHands | Run your agent inside Cube |
| Others |  |

Each guide just needs to cover four things:
- **Setup** (dependencies and how you deployed Cube)
- **Integration snippet** (ideally a diff showing the original code vs. the Cube-powered version — highlighting the near-zero refactor)
- **Runnable demo** (the agent doing a real task inside a Cube sandbox)
- **Going further** (how to configure Cube-specific features like network isolation, timeouts, or file mounts)

## Where to submit a PR

Create a new standalone markdown file for your chosen framework — **one file per framework**. The site will automatically render it as a sub-page.

- English: `docs/guide/integrations/<framework>.md`
- Chinese: `docs/zh/integrations/<framework>.md`

> Only one contributor per framework — before you start, please drop a `/claim <framework>` comment on this issue (e.g. `/claim langchain`) so we don't end up with overlapping work.

> ⚠️ Both English and Chinese versions are **required in the same PR** — a CI check enforces this and PRs missing either language cannot be merged. 
> If you're more comfortable writing in just one language, feel free to post a draft in this issue and we'll help with the translation.

Suggested PR title prefix: `docs(integrations):`

## Need help?

Drop a comment on this issue, or reach out to us:
- 🇨🇳 **In China:** add Cube Girl on WeChat (ID: `CubeSandbox`)
- 🌍 **Outside China:** join our [X](https://x.com/CubeSandbox_AI)

[...]

PR #952, opened 2026-07-14
docs(integrations): add LangChain integration guide

peerless-hero:
issue: [244](https://github.com/TencentCloud/CubeSandbox/issues/244)

### DIFF
--- a/docs/guide/integrations/langchain.md
+++ b/docs/guide/integrations/langchain.md
@@ -1,0 +1,514 @@
+---
+title: LangChain Integration Guide
+author: peerless-hero
+date: 2026-07-07
+tags:
+  - integration
+  - langchain
+  - agent
+lang: en-US
+---
+
+# LangChain Integration Guide
+
+[中文](../../zh/guide/integrations/langchain.md)
+
+Run a LangChain agent that calls a Python tool inside a
+[CubeSandbox](https://github.com/TencentCloud/CubeSandbox) MicroVM. Because Cube exposes an
+**E2B-compatible API**, migrating a LangChain app from E2B to Cube usually means changing a few
+environment variables, while giving every line of agent-generated code KVM-level isolation. This guide
+ships a runnable
+[`examples/langchain-integration`](https://github.com/TencentCloud/CubeSandbox/tree/master/examples/langchain-integration)
+sample project.
+
+## Two major LangChain versions (0.x and 1.x)
+
+This guide provides **two variants** that differ only in the LangChain Agent API and the required
+Python version. Both share the same `run_python` tool (built on the official `cubesandbox` SDK) and
+the same sandbox template.
+
+| Variant | LangChain | langchain-openai | Python | Agent API | Example path |
+|---|---|---|---|---|---|
+| **0.x** (legacy) | `0.3.x` | `0.3.x` | 3.9+ | `AgentExecutor` + `create_react_agent` | `examples/langchain-integration/0.x` |
+| **1.x** (modern) | `1.x` | `1.x` | 3.10+ | `langchain.agents.create_agent` + `@tool` | `examples/langchain-integration/1.x` |
+
+If you are on Python 3.9, or already have LangChain 0.3.x code → use **0.x**. If you are on
+Python 3.10+ and starting fresh (or already on LangChain 1.x) → use **1.x**.
+
+### Components and versions
+
+| Component | 0.x | 1.x |
+|---|---|---|
+| LangChain | `langchain==0.3.23` | `langchain>=1.3.14,<2.0` |
+| langchain-openai | `langchain-openai==0.3.12` | `langchain-openai>=1.0,<2.0` |
+| LangGraph (create_agent runtime) | — (not used; `AgentExecutor` + `create_react_agent` have no LangGraph dep) | `langgraph>=0.2` (declared in `1.x/requirements.txt` so the sample is self-contained) |
+| cubesandbox SDK (main driver) | `cubesandbox>=0.6.0` | `cubesandbox>=0.6.0` |
+| CubeSandbox base image | `ghcr.io/tencentcloud/cubesandbox-base:2026.16` | same |
+| CubeSandbox platform | `>= 0.3.0` (core); higher for optional features below | `>= 0.3.0` (core); higher for optional features below |
+
+Per-feature platform minimums (the SDK floor above covers the base workflow; raise
+`cubesandbox`/platform only when you use the feature):
+
+| Feature | Minimum platform | Minimum SDK |
+|---|---|---|
+| Base workflow (`commands.run` / `files.write` over the envd process API on 49983, context-manager teardown) | `>= 0.3.0` | `>= 0.6.0` |
+| CubeEgress secret injection (advanced usage, default-deny egress) | `>= 0.4.0` | `>= 0.3.0` |
+| Volume mounts (`volume_mounts=`, `Volume` API) | `>= 0.6.0` | `>= 0.6.0` |
+
+## Prerequisites
+
+- CubeSandbox is deployed and CubeAPI is reachable at `http://<node>:3000`.
+- `cubemastercli` is on your `$PATH` and connected to the cluster.
+- The build machine has Docker and the registry is pullable by Cube nodes.
+- Python 3.10+ (for the 1.x variant). The 0.x variant also runs on Python 3.9+.
+- Reachability to the Cube cluster node. The `cubesandbox` SDK reaches CubeAPI via `CUBE_API_URL`
+  and the data plane via `CUBE_PROXY_NODE_IP` (so `*.cube.app` resolves without DNS); if the proxy
+  does not listen on port 80, set `CUBE_PROXY_PORT_HTTP`.
+- An OpenAI-compatible LLM endpoint (the sample uses TokenHub; any OpenAI-compatible endpoint works
+  via `OPENAI_BASE_URL` / `OPENAI_API_KEY`).
+
+## Why run the LangChain agent in a sandbox
+
+LangChain agents often expose a code-execution tool (data analysis, file conversion, shell calls).
+Running that tool on the host mixes the agent's blast radius with your dev machine. Inside
+CubeSandbox you get:
+
+| Concern | What CubeSandbox provides |
+|---|---|
+| **Isolation** | One KVM MicroVM per session, with its own guest kernel — agent code cannot touch the host |
+| **Reproducibility** | Every session boots from the same template snapshot |
+| **Fast startup** | Cold start under 60ms, so running many agents in parallel is cheap |
+| **Long tasks** | `sandbox.pause()` snapshots the VM + root filesystem for later resume |
+| **Secret hygiene** | CubeEgress can inject LLM auth headers at the network layer — the real secret is never visible inside the VM |
+| **Egress audit** | Every outbound request is logged to the egress audit log |
+
+## Integration steps
+
+### 1. Build the template image
+
+Layer a Python data-science stack on top of `cubesandbox-base` (envd listens on `:49983`).
+
+```dockerfile
+# examples/langchain-integration/Dockerfile
+ARG CUBE_BASE_IMAGE=ghcr.io/tencentcloud/cubesandbox-base:2026.16
+FROM ${CUBE_BASE_IMAGE}
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        python3 python3-pip ca-certificates git \
+    && rm -rf /var/lib/apt/lists/*
+
+# Pin versions for reproducible builds; upgrade as needed
+RUN python3 -m pip install --no-cache-dir --upgrade pip \
+    && python3 -m pip install --no-cache-dir --break-system-packages \
+        pandas==2.2.3 numpy==1.26.4 matplotlib==3.9.3 scikit-learn==1.6.1
+
+WORKDIR /workspace
+
+# Seed a demo dataset so the sample agent runs without external input
+COPY sales.csv /workspace/sales.csv
+
+EXPOSE 49983
+```
+
+Build and push:
+
+```bash
+docker build --platform linux/amd64 \
+  -t <your-registry>/langchain-cube:latest \
+  examples/langchain-integration
+docker push <your-registry>/langchain-cube:latest
+```
+
+### 2. Register as a Cube template
+
+```bash
+cubemastercli tpl create-from-image \
+  --image <your-registry>/langchain-cube:latest \
+  --writable-layer-size 2G \
+  --expose-port 49983 \
+  --probe       49983 \
+  --probe-path  /health
+```
+
+`create-from-image` blocks and watches the job to completion by default; pass `--detach` if you
+want to submit-and-exit and poll later with `cubemastercli tpl watch --job-id <job_id>`.
+
+Once the job reaches `READY`, note the `template_id` — you must pass it to every
+`Sandbox.create()`. The `2G` writable layer suits medium analysis tasks; raise it to `4G+` if the
+agent installs large packages at runtime.
+
+### 3. Install dependencies and configure env vars
+
+```bash
+cd examples/langchain-integration/1.x      # or 0.x, depending on your LangChain version
+cp ../.env.example .env
+# Fill in CUBE_API_URL, CUBE_API_KEY, CUBE_TEMPLATE_ID, CUBE_PROXY_NODE_IP, and your LLM key
+pip install -r requirements.txt
+```
+
+| Variable | Flows to | Notes |
+|---|---|---|
+| `CUBE_API_URL` | `cubesandbox` SDK | CubeAPI address (`http://<node>:3000`) |
+| `CUBE_API_KEY` | `cubesandbox` SDK | (optional) `X-API-Key` header; only set it for an auth-enabled CubeAPI backend — the SDK sends no auth header when unset |
+| `CUBE_TEMPLATE_ID` | `Sandbox.create(template=...)` | From step 2 |
+| `CUBE_PROXY_NODE_IP` | `cubesandbox` SDK | CubeProxy node IP so `*.cube.app` resolves without DNS |
+| `CUBE_PROXY_PORT_HTTP` | `cubesandbox` SDK | Proxy HTTP port (default `80`; set `8081` if the proxy listens on 8081) |
+| `OPENAI_API_KEY` / `TOKENHUB_API_KEY` | LLM client | OpenAI-compatible key |
+| `OPENAI_BASE_URL` | LLM client | e.g. `https://tokenhub.tencentmaas.com/v1` |
+| `CHAT_MODEL` | LLM client | e.g. `deepseek-v3` |
+| `CUBE_SSL_CERT_FILE` | demo | (optional) self-signed CA cert path for HTTPS CubeAPI. The demo exports it as `SSL_CERT_FILE` (process-global) so the gRPC client trusts it — the bundle should include public root CAs or the LLM endpoint is affected too |
+
+### 4. Wire the code-execution tool to Cube
+
+The agent logic is identical across versions; only the tool wrapper changes from a local REPL to the
+Cube sandbox. The snippet below is the **shared `run_python` tool for both `0.x` and `1.x`** —
+built on the official **`cubesandbox` Python SDK** (`from cubesandbox import Sandbox`), using
+`sandbox.files.write` to upload code and `sandbox.commands.run` to execute, so **no raw HTTP** is
+needed.
+
+```python
+# run_python tool — official cubesandbox SDK (used by the snippets below;
+# the shipped demos inline equivalent logic inside build_agent)
+import itertools
+from cubesandbox import Sandbox
+
+def make_run_python(sandbox: Sandbox):
+    """Return a run_python tool bound to the given created `sandbox`."""
+    script_counter = itertools.count()
+    def run_python(code: str) -> str:
+        """Execute Python inside the Cube Sandbox MicroVM; return stdout, with
+        stderr delimited below it when present.
+
+        The image preinstalls pandas / numpy / matplotlib / scikit-learn. Each call
+        writes the snippet to a unique /workspace/_agent_<n>.py and runs it, so
+        concurrent tool calls don't overwrite each other. Charts can be saved under
+        /workspace (e.g. /workspace/revenue.png).
+        """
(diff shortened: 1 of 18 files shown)


docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,337 @@
...
+### 5. Run the agent
+
+```python
+from langchain_openai import ChatOpenAI
+from langchain_core.messages import HumanMessage
+from langgraph.prebuilt import create_react_agent
+
+llm = ChatOpenAI(
+    model=os.getenv("CHAT_MODEL", "deepseek-v3"),
+    api_key=os.getenv("OPENAI_API_KEY") or os.getenv("TOKENHUB_API_KEY"),
```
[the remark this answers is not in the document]
peerless-hero: Pedagogical excerpt, not a standalone file. Step 4's excerpt is labeled (excerpt — env validation and tool definition) — its purpose is to show the tool wiring pattern, not to reproduce the full 100-line script. The validation guard (sys.exit(1)) is one-time boilerplate that distracts from the core pattern.
Guard is already documented. Step 4's comment # validates ... the LLM API key explicitly tells readers that validation exists upstream.
Full script is the source of truth. Users who need the complete guard can read langchain_agent_demo.py directly.

examples/langchain-integration/langchain_agent_demo.py
```diff
@@ -0,0 +1,147 @@
...
+from dotenv import load_dotenv
+
+load_dotenv()
+
+# ---------------------------------------------------------------------------
+# CUBE_SSL_CERT_FILE → SSL_CERT_FILE (for HTTPS CubeAPI deployments)
+# ---------------------------------------------------------------------------
+_cube_ssl = os.environ.get("CUBE_SSL_CERT_FILE")
+if _cube_ssl and os.path.isfile(_cube_ssl):
+    os.environ["SSL_CERT_FILE"] = _cube_ssl
```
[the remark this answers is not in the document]
peerless-hero: Consistent with every other example in the repo. openai-agents-example, openai-agents-code-interpreter, mini-rl-training all set os.environ["SSL_CERT_FILE"] — this is an established repo convention.
E2B SDK lacks a per-connection CA override API. It only reads SSL_CERT_FILE from the environment; there is no per-session certificate option.
No practical conflict. A custom CA for CubeAPI (self-signed / internal CA) and public CAs for the LLM endpoint coexist without issues. Documenting the process-wide scope in Caveats is sufficient.
Fixing this would require changing the SDK or connection pattern — cost far outweighs benefit.

examples/langchain-integration/.env.example
```diff
@@ -0,0 +1,18 @@
...
+TOKENHUB_API_KEY="<your-tokenhub-api-key>"
+# or set OPENAI_API_KEY + OPENAI_BASE_URL directly:
+# OPENAI_API_KEY="<your-api-key>"
+# OPENAI_BASE_URL="https://tokenhub.tencentmaas.com/v1"
+
+# Model name for the chat model (change per your provider)
+CHAT_MODEL="deepseek-v3"
+
+# cube-sandbox (E2B compatible) connection
+E2B_API_URL="http://<your-cube-sandbox-ip>:3000"
```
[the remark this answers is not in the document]
peerless-hero: Default URL https:// → Cube's default deployment serves plain HTTP. Changing the example to https:// would mislead users into a connection that will fail. The runtime stderr warning we added (line 43–48) is sufficient.

examples/langchain-integration/langchain_agent_demo.py
```diff
@@ -0,0 +1,168 @@
...
+
+_api_key = os.getenv("OPENAI_API_KEY") or os.getenv("TOKENHUB_API_KEY")
+if not _api_key:
+    print(
+        "Missing LLM API key. Set either OPENAI_API_KEY or TOKENHUB_API_KEY in .env",
+        file=sys.stderr,
+    )
+    sys.exit(1)
+
+import e2b.envd.rpc as _e2b_rpc
```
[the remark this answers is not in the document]
peerless-hero: Import ordering / private API → Same e2b.envd.rpc.default_username pattern used by every other example in this repo (pi-agent, code-interpreter, mini-rl-training). Established convention, not worth diverging.

nnWhisperer:
Hello, I'm not a maintainer but am thinking of using the changes, hence I was reviewing. They seem fine, the only thing is, you assumed that the cli will be run on the developer machine but it isn't a valid assumption I believe. For example, I have a baremetal machine which I'm connecting over a VPN, that's running the cubesandbox, where the cli is only installed. The documentation assumes that the machines where cube api is called needs to have the cli, but this is incorrect. I believe most users will use it as a remote machine.
Thanks for your work

examples/langchain-integration/langchain_agent_demo.py
```diff
@@ -0,0 +1,168 @@
...
+
+    Use this to compute with pandas/numpy/matplotlib, read files under
+    /workspace, or install nothing (the stack is preinstalled in the template).
+
+    The template image pre-seeds /workspace/sales.csv so the default demo
+    question works without external inputs.
+    """
+    sandbox = None
+    try:
+        sandbox = Sandbox.create(
```
nnWhisperer: Have you run this code locally?
I doubt that it would work, because [Sandbox.create seems to take env dict as input, doesn't look it up itself](https://github.com/e2b-dev/E2B/blob/b5119539ca7efc12c51e4ef2434e8a4f43956efc/packages/python-sdk/e2b/sandbox_sync/main.py#L166). Hence, if you run a non-trivial example where there is a non-empty API key or a API URL, the code, this will likely to not work.

nnWhisperer (commented):
I reviewed, particularly the flow of the api keys and urls, and it seems that the e2b.sandbox.create does need to take the variables directly.
Secondly, langchain has a langchain_e2b library with the implementation [here](https://github.com/e2b-dev/langchain-e2b/) you might want to check what they are doing, for an improved support of the langchain's tools(langchain/langgraph/deepagents) etc.

examples/langchain-integration/langchain_agent_demo.py
peerless-hero: Sandbox.create reads E2B_API_KEY and E2B_API_URL from env automatically.

examples/langchain-integration/langchain_agent_demo.py
```diff
@@ -0,0 +1,170 @@
...
+
+
+def build_agent():
+    llm = ChatOpenAI(
+        model=os.getenv("CHAT_MODEL") or "deepseek-v3",
+        api_key=_api_key,
+        base_url=os.getenv("OPENAI_BASE_URL") or "https://tokenhub.tencentmaas.com/v1",
+        timeout=60,
+        max_retries=2,
+        temperature=0,
```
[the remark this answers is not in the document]
peerless-hero: This is a copy-paste example, not a library or production script. Its primary goal is to demonstrate the Cube ↔ LangChain integration with minimal noise. Every environment variable adds cognitive overhead for the reader and maintenance burden to .env.example. temperature=0 is the correct default for deterministic code-generation agents; users who want to experiment can change the literal in one place. Exposing tuning knobs as env vars would lead to scope creep (timeout, max_retries, _MAX_OUTPUT_CHARS...) and dilute the example's core narrative. Keep essential connection params as env vars; keep tuning params as hardcoded literals.

examples/langchain-integration/langchain_agent_demo.py
```diff
@@ -0,0 +1,170 @@
...
+# CUBE_SSL_CERT_FILE → SSL_CERT_FILE (for HTTPS CubeAPI deployments)
+# ---------------------------------------------------------------------------
+_cube_ssl = os.environ.get("CUBE_SSL_CERT_FILE")
+if _cube_ssl and os.path.isfile(_cube_ssl):
+    os.environ["SSL_CERT_FILE"] = _cube_ssl
+
+# Warn when running over plain HTTP (sandbox traffic in cleartext)
+if os.environ.get("E2B_API_URL", "").startswith("http://"):
+    print(
+        "Warning: E2B_API_URL uses plain HTTP — sandbox traffic (code, keys, data) "
```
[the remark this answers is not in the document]
peerless-hero: Same principle as the temperature case — this is a demo, not a production harness. Adding CUBE_SANDBOX_TIMEOUT and CUBE_CMD_TIMEOUT would balloon .env.example with two more optional knobs that almost nobody will touch, while making the file harder to skim. The timeout values are already documented in the Caveats section where the user will look when they actually hit the limit. Changing a literal is a one-line edit; the added env-var surface area isn't worth the marginal convenience.

examples/langchain-integration/langchain_agent_demo.py
nnWhisperer: No, it doesn't. I showed you the code.

peerless-hero:
@tinklone Please review this PR.

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,430 @@
...
+
+### File mounts
+
+Mount a host directory into the sandbox so the agent reads/writes shared data directly, without
+round-tripping through `files.write`:
+
+```python
+sandbox = Sandbox.create(
+    template=os.environ["CUBE_TEMPLATE_ID"],
+    mounts=[{"source": "/data/datasets", "target": "/workspace/datasets"}],
```
fslongjin: This `mounts` API doesn't seem to match the current Python SDK — `Sandbox.create` doesn't take `mounts=[{source, target}]`. Host directory mounts go through `metadata={"host-mount": ...}` (see `examples/host-mount`); volumes use `volume_mounts=`. As written, readers who copy-paste this will likely hit a confusing failure. Worth switching to the supported API, and maybe noting that `hostPath` is on the Cubelet node, not the local machine.

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,430 @@
...
+python langchain_agent_demo.py "Train a linear model on the dataset and report the RMSE."
+```
+
+### Expected output
+
+With the default prompt, the **agent's final answer** printed on the host includes the per-month
+revenue numbers (2780.0 / 3375.5 / 3872.0). (The 1.x `create_agent` is not verbose by default; the
+0.x `AgentExecutor(verbose=True)` also prints the full reasoning trace to the console.)
+
+If teardown fails, a `Warning: failed to kill sandbox <id>: <err>` line is written to **stderr**,
```
fslongjin: This says a failed teardown prints `Warning: failed to kill sandbox ...` to stderr, but in the current `cubesandbox` SDK `__exit__` just swallows `CubeSandboxError` from `kill()` — it doesn't emit that warning. The docs don't match the behavior here; I'd update this (and the matching notes in the example READMEs) so people don't look for a log line that never shows up when debugging leaks.

examples/langchain-integration/0.x/langchain_agent_demo.py
```diff
@@ -0,0 +1,90 @@
...
+            out += f"\n[non-zero exit code: {result.exit_code}]"
+        return out
+
+    tools = [Tool(
+        name="run_python",
+        func=run_python,
+        description="Execute Python code in a Cube Sandbox MicroVM with pandas/numpy/matplotlib "
+                    "preinstalled. Returns the combined stdout and stderr. Save charts under /workspace.",
+    )]
+    prompt = PromptTemplate.from_template(
```
fslongjin, line 67: The custom prompt for `create_react_agent` looks like it's missing the usual ReAct format instructions (`Action` / `Action Input` / `Final Answer`). With only the tool blurb and scratchpad, the model output is pretty easy to mis-parse — `handle_parsing_errors=True` can paper over some of it, but it's not very reliable. I'd either add the standard format block or pull in the hub ReAct template. (Same issue in the 0.x snippet in the integration guide.)

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,430 @@
...
+)
+```
+
+### Long-task pause / resume
+
+```python
+sandbox = Sandbox.create(template=os.environ["CUBE_TEMPLATE_ID"], timeout=1800)
+try:
+    turn_1(sandbox)
+    sandbox_id = sandbox.pause() or sandbox.sandbox_id
```
fslongjin: `pause()` returns `None`, not a sandbox id. `or sandbox.sandbox_id` happens to work, but it reads like `pause` itself returns the id. Clearer to call `sandbox.pause()` and then `Sandbox.connect(sandbox.sandbox_id)`.

docs/zh/guide/integrations/langchain.md
```diff
@@ -0,0 +1,413 @@
...
+```
+
+### 文件挂载
+
+把宿主机目录挂载进沙箱，让 Agent 直接读写共享数据，无需经 `files.write` 来回搬运：
+
+```python
+sandbox = Sandbox.create(
+    template=os.environ["CUBE_TEMPLATE_ID"],
+    mounts=[{"source": "/data/datasets", "target": "/workspace/datasets"}],
```
fslongjin: Same issue as in the English guide: this `mounts=[{source, target}]` shape isn't a supported `Sandbox.create` API. Please mirror the fix here.

docs/zh/guide/integrations/langchain.md
```diff
@@ -0,0 +1,413 @@
...
+python langchain_agent_demo.py "在数据集上训练一个线性模型并报告 RMSE。"
+```
+
+### 预期效果
+
+使用默认提示词时，宿主机上打印的 **Agent 最终回答**会包含各月营收数字
+（2780.0 / 3375.5 / 3872.0）。（1.x 的 `create_agent` 默认不显示详细过程；0.x 的
+`AgentExecutor(verbose=True)` 还会把完整推理过程打印到控制台。）
+
+若销毁失败，会向 **stderr** 写入一行 `Warning: failed to kill sandbox <id>: <err>`，而非静默吞掉。
```
fslongjin: Same as the English guide — the SDK `__exit__` doesn't print this kill-failure warning today. Please keep EN/ZH in sync when updating.

docs/zh/guide/integrations/langchain.md
```diff
@@ -0,0 +1,413 @@
...
+)
+```
+
+### 长任务暂停 / 恢复
+
+```python
+sandbox = Sandbox.create(template=os.environ["CUBE_TEMPLATE_ID"], timeout=1800)
+try:
+    turn_1(sandbox)
+    sandbox_id = sandbox.pause() or sandbox.sandbox_id
```
fslongjin: Same as the English guide: `pause()` returns `None`; please align this snippet when you fix the EN version.

fslongjin (commented):
A few doc/example nits I noticed while reviewing — details on the inline comments.

docs/guide/integrations/langchain.md
peerless-hero: Good catch — Sandbox.create() indeed has no mounts parameter. Fixed in 970c97a: host directory mounts now use metadata["host-mount"] with a JSON-encoded {hostPath, mountPath, readOnly} list (matching examples/host-mount), and volume mounts use volume_mounts={path: volume_id} (with a Volume.create() example). The zh-CN guide is synced too.

docs/guide/integrations/langchain.md
peerless-hero: You're right — I verified in sdk/python/cubesandbox/sandbox.py that __exit__ swallows CubeSandboxError silently (except CubeSandboxError: pass), so no warning is printed. Fixed in 0173dbf: the guide and all four READMEs now describe teardown as best-effort with silent error swallowing, and explain how to detect failures via an explicit sandbox.kill().

examples/langchain-integration/0.x/langchain_agent_demo.py
peerless-hero: Agreed, a custom prompt without the format block is fragile even with handle_parsing_errors=True. Fixed in 222639a: added the standard ReAct format instructions (Thought / Action / Action Input / Observation / Final Answer) to both the 0.x demo and the matching snippet in the guide (en + zh).

docs/guide/integrations/langchain.md
peerless-hero: Thanks — pause() returns None; the sandbox ID never changes. Fixed in 836a792: the snippet now calls sandbox.pause() directly and resumes via Sandbox.connect(sandbox.sandbox_id).

docs/zh/guide/integrations/langchain.md
peerless-hero: The zh-CN guide was updated in the same commits (970c97a for mounts, 836a792 for pause/resume) so both languages stay in sync.

examples/langchain-integration/Dockerfile
```diff
@@ -0,0 +1,23 @@
...
+FROM ${CUBE_BASE_IMAGE}
+
+ARG DEBIAN_FRONTEND=noninteractive
+
+# Pin data-science versions for reproducible builds. Bump as needed.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends \
+        python3 python3-pip ca-certificates git \
+    && rm -rf /var/lib/apt/lists/* \
+    && python3 -m pip install --no-cache-dir --break-system-packages --upgrade pip \
```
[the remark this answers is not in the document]
peerless-hero: Good catch — the base image ships pip 22.0.2, which predates --break-system-packages (added in pip 23.0). Fixed in e065d60: pip is upgraded first, then packages are installed with the flag, matching the inline Dockerfile in the guide.

examples/langchain-integration/.env.example
```diff
@@ -0,0 +1,22 @@
...
+# (optional) defaults to http://127.0.0.1:3000
+CUBE_API_URL="http://<your-node-ip>:3000"
+# (optional) only for an auth-enabled CubeAPI backend; SDK sends no auth header when unset
+CUBE_API_KEY="<your-cube-api-key>"
+# (required) sandbox template used by the demo
+CUBE_TEMPLATE_ID="<your-template-id>"
+# (optional) direct IP to reach CubeProxy; unset = use normal DNS
+CUBE_PROXY_NODE_IP="<your-node-ip>"
+# (optional) CubeProxy HTTP port; defaults to 80
+CUBE_PROXY_PORT_HTTP=8081
```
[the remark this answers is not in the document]
peerless-hero: Fixed in 90e89e5 — CUBE_PROXY_PORT_HTTP=8081 contradicted the "defaults to 80" comment, and copying the file verbatim would point the data plane at port 8081 while CubeProxy listens on 80. The line is now commented out so the SDK default applies.

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,471 @@
...
+
+## Two major LangChain versions (0.x and 1.x)
+
+This guide provides **two variants** that differ only in the LangChain Agent API and the required
+Python version. Both share the same `run_python` tool (built on the official `cubesandbox` SDK) and
+the same sandbox template.
+
+| Variant | LangChain | langchain-openai | Python | Agent API | Example path |
+|---|---|---|---|---|---|
+| **0.x** (legacy) | `0.3.x` | `0.3.x` | 3.8+ | `AgentExecutor` + `create_react_agent` | `examples/langchain-integration/0.x` |
```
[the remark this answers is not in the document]
peerless-hero: You're right — the SDK declares requires-python = ">=3.9" and LangChain 0.3.x also requires ≥3.9. Fixed in 45f1d85 across the version table, recommendation line, prerequisites, and all READMEs (0.x/1.x/top-level, en + zh).

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,471 @@
...
+# Before: code runs on the host via langchain_experimental
+from langchain_experimental.tools import PythonREPLTool
+tools = [PythonREPLTool()]
+```
+
+Swap `PythonREPLTool` for the `run_python` Cube tool above to get isolation for free.
+
+## Caveats
+
+- **envd user.** Cube's envd serves `root` only, and the `cubesandbox` SDK's `files.write` /
```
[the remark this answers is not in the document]
peerless-hero: Fixed in 4b12148 — the caveat now states the SDK runs as root by default and notes the base image provisions a non-root uid=1000 user.

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,471 @@
...
+|---|---|---|
+| `CUBE_API_URL` | `cubesandbox` SDK | CubeAPI address (`http://<node>:3000`) |
+| `CUBE_API_KEY` | `cubesandbox` SDK | Sent as `X-API-Key`; set any non-empty string when auth is disabled |
+| `CUBE_TEMPLATE_ID` | `Sandbox.create(template=...)` | From step 2 |
+| `CUBE_PROXY_NODE_IP` | `cubesandbox` SDK | CubeProxy node IP so `*.cube.app` resolves without DNS |
+| `CUBE_PROXY_PORT_HTTP` | `cubesandbox` SDK | Proxy HTTP port (default `80`; set `8081` if the proxy listens on 8081) |
+| `OPENAI_API_KEY` / `TOKENHUB_API_KEY` | LLM client | OpenAI-compatible key |
+| `OPENAI_BASE_URL` | LLM client | e.g. `https://tokenhub.tencentmaas.com/v1` |
+| `CHAT_MODEL` | LLM client | e.g. `deepseek-v3` |
+| `CUBE_SSL_CERT_FILE` | `cubesandbox` SDK | (optional) CA cert path for HTTPS CubeAPI |
```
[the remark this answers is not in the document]
peerless-hero: Fixed in 4b12148 — the table now marks it as a host-env variable and explains the SDK doesn't read it directly; it must be exported as SSL_CERT_FILE / REQUESTS_CA_BUNDLE.

docs/guide/integrations/langchain.md
```diff
@@ -0,0 +1,471 @@
...
+
+```python
+import json
+from cubesandbox import Sandbox
+
+sandbox = Sandbox.create(
+    template=os.environ["CUBE_TEMPLATE_ID"],
+    metadata={
+        "host-mount": json.dumps([
+            {"hostPath": "/data/datasets", "mountPath": "/workspace/datasets",
```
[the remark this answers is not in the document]
peerless-hero: Fixed in 4b12148 — the example now uses /data/shared/datasets (inside the default allowed prefix) and notes how to extend allowed_host_mount_prefixes.

lkml-likexu:
@peerless-hero Thanks for your contribution and you're wonderful. Would you help fix the DCO (Developer Certificate of Origin) in the commits ?

fslongjin (approved):
LGTM

fslongjin:
Hello, thanks for your contribution! Please add a line `signed-off-by` in your commit, and I'll merge this into `master`.

PR #955, opened 2026-07-14
feat(CubeProxy): make admin port configurable and detect port conflicts

### LINKED ISSUE #945: [Feature Request] Make the CubeProxy admin port configurable and detect port conflicts during one-click installation
## Problem / Motivation

CubeProxy uses host networking and its internal admin server is currently hard-coded to `<node-ip>:8082`. If another process is already listening on `0.0.0.0:8082`, CubeProxy repeatedly fails to start with: `nginx: [emerg] bind() to <node-ip>:8082 failed (98: Address in use)`. Meanwhile, `install.sh` waits at: `systemctl enable --now cube-sandbox-control.target` without clearly reporting the underlying CubeProxy failure.

The public HTTP and HTTPS ports are configurable through `CUBE_PROXY_HTTP_PORT` and `CUBE_PROXY_HTTPS_PORT`, but there is no corresponding option for the admin port.

## Proposed Solution

Introduce a configuration variable such as: `CUBE_PROXY_ADMIN_PORT=8082`. The default should remain `8082` for backward compatibility.

Use this variable consistently for:
  - the CubeProxy nginx admin listener
  - `CUBE_PROXY_ID`
  - `CUBE_PROXY_ADMIN_URL`
  - `CUBE_PROXY_RESUME_URL`
  - Docker Compose and Terraform deployment templates
  - health checks, tests, and documentation

The one-click installer could also check whether the configured address and port are available and fail early with a clear error message.

## Additional Context

The systemd post-start check only checks `CUBE_PROXY_HTTP_PORT`; it does not check the admin endpoint or the container's restart state. This allows the service to appear healthy even when the admin listener cannot be created.

Relevant code:
  - `CubeProxy/nginx.conf:379`
  - `deploy/one-click/build-release-bundle.sh:409`
  - `deploy/one-click/scripts/one-click/up-cube-proxy.sh:69-95`
  - `deploy/one-click/scripts/one-click/up-cube-proxy.sh:206-213`
  - `deploy/one-click/scripts/systemd/cube-proxy-postcheck.sh:6-16`

yingjun8:
## Summary

CubeProxy runs with host networking and its `/admin/*` server was hard-coded to port `8082`. When another process on the node already holds that port, nginx aborts at startup with `bind() to <node-ip>:8082 failed (98: Address already in use)` and the container crash-loops — while the one-click installer stalls at `cube-sandbox-control.target` without ever surfacing the real cause (#945).

This PR fixes both halves of the problem:

1. **The port is now configurable** via `CUBE_PROXY_ADMIN_PORT` (default `8082`, fully backward-compatible), threaded through the entire stack — systemd one-click path, nginx templating, and the Terraform/TKE deployment.
2. **Conflicts now fail fast with a clear message** instead of a silent hang: the pre-start conflict check, the post-start readiness loop, and the systemd postcheck all cover the admin port.

### Root cause

Two independent defects:

- **Admin port not configurable.** A prior change parametrized the admin *listen address* (`CUBE_PROXY_ADMIN_LISTEN`) but left the *port* `8082` hard-coded — in the nginx template transforms, in the derived `CUBE_PROXY_ID` / `CUBE_PROXY_ADMIN_URL`, and in the Terraform deployment — unlike HTTP/HTTPS/gRPC, which already have `CUBE_PROXY_HTTP_PORT` / `CUBE_PROXY_HTTPS_PORT` / `CUBE_PROXY_GRPC_PORT`.
- **Silent failure.** The pre-start conflict check and all readiness/postcheck probes only covered HTTP/HTTPS/gRPC. An admin-bind failure kills the whole nginx master, so the container crash-loops and the installer just waits.

### Changes

**1. Admin port configurable — systemd one-click path**
- `up-cube-proxy.sh`: new `CUBE_PROXY_ADMIN_PORT` (default 8082); validates it is a TCP port in 1–65535 and distinct from the HTTP/HTTPS/gRPC ports; derives `CUBE_PROXY_ID` / `CUBE_PROXY_ADMIN_URL` from it; renders the new `__CUBE_PROXY_ADMIN_PORT__` nginx template placeholder.
- `env.example`: documents the new variable.

**2. Fail-fast conflict & readiness detection**
- `up-cube-proxy.sh` pre-start check now covers the admin port, and the error names the specific override variable for whichever port collided (e.g. port 80 occupied → suggests `CUBE_PROXY_HTTP_PORT`, not the admin port).
- The post-start readiness loop also waits on the admin port; on timeout it names each missing port with its override variable and points at `docker logs cube-proxy` for the `bind()` failure.
- `cube-proxy-postcheck.sh` (systemd `ExecStartPost`) probes the admin port as well, so a bind failure marks the unit failed with a clear message.
- `quickcheck.sh`: `check_unit_active` now dies immediately if the unit is already in `failed` state (with `systemctl status` output) instead of polling until the 120s readiness budget expires — relevant because `control.target` only `Wants=` its children, so a failed child never fails `enable --now`.

**3. nginx template placeholder**
- `build-release-bundle.sh`, `terraform/tencentcloud/create.sh`, `tests/test_package_layout.sh`: the sed transforms now emit `__CUBE_PROXY_ADMIN_LISTEN__:__CUBE_PROXY_ADMIN_PORT__` (previously `:8082`), and the placeholder-validation loops require the new token. `create.sh` regenerates a stale template that lacks it.

**4. Terraform / TKE**
- `variables.tf`: new `cube_proxy_admin_port` variable (number, validated 1–65535, default 8082).
- `create.sh` / `destroy.sh`: `TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT` → `TF_VAR_cube_proxy_admin_port`, persisted to `.env` and `resolved.auto.tfvars.json` (destroy keeps parity).
- `tke-addons.tf`: the admin-port `replace()` is folded into the main token chain (`tostring(var.cube_proxy_admin_port)`); `containerPort` and the `CUBE_PROXY_ADMIN_URL` / `CUBE_PROXY_RESUME_URL` env values follow the variable.
- `terraform/tencentcloud/env.example`: documents the new variable.

**5. Docs**
- `docs/guide/lifecycle.md` + `docs/zh/guide/lifecycle.md`: note that the admin port defaults to 8082 and can be overridden with `CUBE_PROXY_ADMIN_PORT` when the port is taken (EN/ZH in sync).

**6. Tests**
- `test_runtime_file_safety.sh`: the postcheck harness now stubs the admin port (the `ss` stub appends across the three probes); new case `test_quickcheck_check_unit_active_dies_fast_on_failed_unit` verifying the failed-unit fail-fast (dies immediately, prints `systemctl status`, never sleeps).
- `test_package_layout.sh`: covers the new template placeholder (and picks up the gRPC placeholder it was missing).

### Scope note

Earlier revisions of this PR also added an installer-level port preflight and a full systemd state machine in `quickcheck.sh`. Per review feedback these were removed as out of scope / redundant with the service-level checks above; general fail-fast behavior for the installer under `Wants=` units will be handled in a follow-up. What remains here is scoped to the admin port: configurability plus service-level conflict/readiness detection.

### Testing

- `deploy/one-click/tests/test_runtime_file_safety.sh` — OK
- `deploy/one-click/tests/test_package_layout.sh` — OK

Closes #945


mon3stera:
When I ran install.sh, I encounter a problem of silent failure: `install.sh` calls: `systemctl enable --now "${target}"` and does not print the failed service's status or recent journal entries. Therefore the installer may still stop around this command without directly showing the actionable `port 8082 is already in use` message.

yingjun8:
@mon3stera Fixed in 9bc711a. install.sh ran a bare systemctl enable --now, so when cube-proxy failed on the 8082 conflict, set -e aborted before quickcheck and systemd swallowed the child's logs. Now it dumps status + journal for failed cube-sandbox-* units and dies with an actionable hint.

mon3stera:
The CubeProxy preflight itself works. However, I found that the new installer diagnostic path is not triggered with the current target dependencies.

`cube-sandbox-control.target` has an empty `Requires=` and pulls `cube-sandbox-cube-proxy.service` in through `Wants=`. Therefore, when cube-proxy fails, `systemctl enable --now "${target}"` can still return 0, so this branch doesn't work:

```
if !systemctl enable --now "${target}"; then
  dump_failed_cube_units "${target}"
fi
```

The installer then proceeds to quickcheck, while the actionable port-conflict message remains only in the service journal.

Other issues:
1. `test_runtime_fail_safety.sh` fails because the existing CubeProxy postcheck test stub only simulates the HTTP listener, while the postcheck now also waits for port 8082.

yingjun8:
> The CubeProxy preflight itself works. However, I found that the new installer diagnostic path is not triggered with the current target dependencies.
> 
> `cube-sandbox-control.target` has an empty `Requires=` and pulls `cube-sandbox-cube-proxy.service` in through `Wants=`. Therefore, when cube-proxy fails, `systemctl enable --now "${target}"` can still return 0, so this branch doesn't work:
> 
> ```
> if !systemctl enable --now "${target}"; then
>   dump_failed_cube_units "${target}"
> fi
> ```
> 
> The installer then proceeds to quickcheck, while the actionable port-conflict message remains only in the service journal.
> 
> Other issues:
> 
> 1. `test_runtime_fail_safety.sh` fails because the existing CubeProxy postcheck test stub only simulates the HTTP listener, while the postcheck now also waits for port 8082.

  Thanks, fixed in 9262837:
 
  1. I moved detection into quickcheck's check_unit_active — it now dumps systemctl status + journalctl -n 40 on timeout, surfacing the actionable port-8082 message.
  2. Test stub now models the admin listener too and asserts on both ports.

  Please test it when you have a chance.

deploy/one-click/scripts/one-click/quickcheck.sh
```diff
@@ -129,7 +129,25 @@ unit_is_active() {
...
+  # `systemctl enable --now <target>` return non-zero. This readiness gate is
+  # the exit-code-independent place that catches it, so on the deadline we dump
+  # the unit's status and journal -- otherwise the actionable root cause (like a
+  # port already in use) stays buried in the service journal and the operator
+  # only sees a generic "unit not active".
+  while :; do
+    if unit_is_active "${unit}"; then
+      return 0
+    fi
+    if (( $(date +%s) >= QUICKCHECK_DEADLINE )); then
```
mon3stera: if the service has already failed when the port conflicts (where the corresponding systemctl service status is `failed`), the check will wait for `QUICKCHECK_DEADLINE`. It is better to definitely determine the status. Actually, I think we can directly check the port in `install.sh` first,  which can provide better UX.  However we still need to validate it in the service, because we can bypass the script with `systemctl restart cube-sandbox-cube-proxy.service`

mon3stera (changes requested):

deploy/one-click/scripts/one-click/up-cube-proxy.sh
```diff
@@ -176,12 +180,15 @@ fi
...
+# cube-proxy uses network_mode: host, so its HTTP/HTTPS and admin ports must be
+# free on the host before we attempt to start the container; otherwise the
+# failure mode is a cryptic "address already in use" from nginx inside the
+# container that crash-loops without surfacing the real reason. The admin port
+# in particular has no default liveness check, so a conflict there would
+# otherwise stall the installer at cube-sandbox-control.target.
+for port in "${CUBE_PROXY_HTTP_PORT}" "${CUBE_PROXY_HTTPS_PORT}" "${CUBE_PROXY_ADMIN_PORT}"; do
   if command_output_contains_fixed_string "LISTEN" ss -lnt "( sport = :${port} )"; then
-    die "port ${port} is already in use; cube-proxy uses host networking and requires it to be free"
+    die "port ${port} is already in use; cube-proxy uses host networking and requires it to be free (override the admin port with CUBE_PROXY_ADMIN_PORT if 8082 is taken)"
```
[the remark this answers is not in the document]
mon3stera: bot is right

deploy/one-click/terraform/tencentcloud/create.sh
```diff
@@ -634,13 +634,13 @@ EOF
 				-e 's|^\(\s*listen \)8080\( ssl reuseport;\)|\1__CUBE_PROXY_HTTPS_PORT__\2|' \
 				-e 's|^\(\s*set \$host_proxy_port \)8081;|\1__CUBE_PROXY_HTTP_PORT__;|' \
 				-e 's|^\(\s*set \$host_proxy_port \)8080;|\1__CUBE_PROXY_HTTPS_PORT__;|' \
-				-e 's|^\(\s*listen \)127\.0\.0\.1:8082;|\1__CUBE_PROXY_ADMIN_LISTEN__:8082;|' \
+				-e 's|^\(\s*listen \)127\.0\.0\.1:8082;|\1__CUBE_PROXY_ADMIN_LISTEN__:__CUBE_PROXY_ADMIN_PORT__;|' \
```
mon3stera, line 653: I think you need to define a variable `TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT` and export it like `TF_VAR_cube_proxy_admin_port="${TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT:-8082}"` following the existing code style. In addition, you may want to refer to the code around line 4161 to ensure it is correctly persisted.

mon3stera (changes requested):

deploy/one-click/terraform/tencentcloud/create.sh
yingjun8: Updated in d1c26cc: terminal-state short-circuit in check_unit_active, cube-proxy port preflight in install.sh (after the upgrade stop), and TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT → TF_VAR_cube_proxy_admin_port with .env/resolved.auto.tfvars.json persistence.

deploy/one-click/scripts/one-click/quickcheck.sh
```diff
@@ -127,9 +127,58 @@ unit_is_active() {
...
+# instead -- e.g. when cube-proxy loses a bind() port race under
+# `systemctl restart cube-sandbox-cube-proxy.service` (bypassing install.sh's
+# preflight), the unit flips straight to `failed` and the actionable root cause
+# is in the journal. `is-active` returns "activating" while a unit is still
+# starting (a legit transient state) and only a non-zero exit + a state string
+# other than "activating" means we are permanently stuck.
+unit_is_terminally_stuck() {
+  local unit="$1"
+  local state
+  systemctl is-active "${unit}" >/dev/null 2>&1 && return 1
```
mon3stera: Race condition: if the status changes from inactive to active between two calls,  it will actually return an incorrect result. I think a better way is acquiring the status only once and then branch based on it.

deploy/one-click/terraform/tencentcloud/create.sh
```diff
@@ -4220,6 +4223,7 @@ TENCENTCLOUD_CUBE_LIFECYCLE_MANAGER_HEARTBEAT_TTL='${TF_VAR_cube_lifecycle_manag
 TENCENTCLOUD_CUBE_LIFECYCLE_MANAGER_DISCOVERY_REFRESH='${TF_VAR_cube_lifecycle_manager_discovery_refresh:-${TENCENTCLOUD_CUBE_LIFECYCLE_MANAGER_DISCOVERY_REFRESH:-3s}}'
 TENCENTCLOUD_CUBE_ADMIN_TOKEN='${TF_VAR_cube_admin_token:-${TENCENTCLOUD_CUBE_ADMIN_TOKEN:-}}'
 TENCENTCLOUD_CUBE_PROXY_HEARTBEAT_INTERVAL_MS='${TF_VAR_cube_proxy_heartbeat_interval_ms:-${TENCENTCLOUD_CUBE_PROXY_HEARTBEAT_INTERVAL_MS:-5000}}'
+TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT='${TF_VAR_cube_proxy_admin_port:-${TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT:-8082}}'
```
mon3stera, line 4246: you may also need to add it to `terraform/tencentcloud/env.example`

deploy/one-click/scripts/one-click/quickcheck.sh
```diff
@@ -127,9 +127,58 @@ unit_is_active() {
...
+  state="$(systemctl is-active "${unit}" 2>/dev/null || true)"
+  [[ "${state}" != "activating" && -n "${state}" ]]
```
mon3stera: The code treats all non-active statuses as failure, which is not accurate. Actually, systemctl services have multiple status categories (such as `LoadState`, `SubState` and so on). 

Recommended logic:

```bash

# This includes oneshot units configured with RemainAfterExit=yes
check_unit_active() {
  local unit="$1"
  local output key value
  local load_state active_state sub_state result job restarts
  local now remaining delay

  while :; do
    load_state=""
    active_state=""
    sub_state=""
    result=""
    job=""
    restarts=""

    output="$(
      systemctl show "${unit}" \
        --no-pager \
        --property=LoadState \
        --property=ActiveState \
        --property=SubState \
        --property=Result \
        --property=Job \
        --property=NRestarts \
        2>/dev/null || true
    )"

    while IFS='=' read -r key value; do
      case "${key}" in
        LoadState) load_state="${value}" ;;
        ActiveState) active_state="${value}" ;;
        SubState) sub_state="${value}" ;;
        Result) result="${value}" ;;
        Job) job="${value}" ;;
        NRestarts) restarts="${value}" ;;
      esac
    done <<< "${output}"

    case "${load_state}" in
      not-found|bad-setting|error|masked)
        _unit_dump_diagnostics "${unit}"
        die "systemd unit cannot be loaded: ${unit} \
(load=${load_state}, active=${active_state:-unknown}, \
sub=${sub_state:-unknown})"
        ;;
    esac

    case "${active_state}" in
      active|reloading|refreshing)
        return 0
        ;;

      failed)
        _unit_dump_diagnostics "${unit}"
        die "systemd unit failed: ${unit} \
(sub=${sub_state:-unknown}, result=${result:-unknown}, \
restarts=${restarts:-0})"
        ;;

      inactive)
        # A pending job may still start or restart the unit.
        if [[ -z "${job}" && -n "${load_state}" ]]; then
          _unit_dump_diagnostics "${unit}"
          die "systemd unit is inactive with no pending job: ${unit} \
(sub=${sub_state:-unknown}, result=${result:-unknown})"
        fi
        ;;

      activating|deactivating|maintenance)
        ;;

      *)
        # Unknown or temporarily unavailable state: let the deadline decide.
        ;;
    esac

    now="$(date +%s)"
    if (( now >= QUICKCHECK_DEADLINE )); then
      _unit_dump_diagnostics "${unit}"
      die "systemd unit did not become active: ${unit} \
(load=${load_state:-unknown}, active=${active_state:-unknown}, \
sub=${sub_state:-unknown}, result=${result:-unknown}, \
restarts=${restarts:-0}; timeout=${QUICKCHECK_READY_TIMEOUT}s)"
    fi

    remaining=$((QUICKCHECK_DEADLINE - now))
    delay="${QUICKCHECK_READY_INTERVAL}"
    if (( delay > remaining )); then
      delay="${remaining}"
    fi

    sleep "${delay}" || true
  done
}

```

yingjun8: Updated in f447f75: check_unit_active now reads all systemd state categories in a single systemctl show call and branches on that snapshot (fixing the race and adopting your status taxonomy), and TENCENTCLOUD_CUBE_PROXY_ADMIN_PORT='8082' is now in terraform/tencentcloud/env.example.

chenhengqi:
@mon3stera There is a requested change from your previous review. I would like to get an explicit approve from you on this PR. Thanks.

deploy/one-click/scripts/one-click/up-cube-proxy.sh
```diff
@@ -35,12 +35,15 @@ CUBE_PROXY_HTTPS_PORT="${CUBE_PROXY_HTTPS_PORT:-443}"
...
-# Address the /admin/* server (port 8082) binds to. Defaults to the node's
+# Address the /admin/* server binds to. Defaults to the node's
 # cluster-reachable IP so cube-lifecycle-manager (running on the control
 # node) can push meta / state updates here. Token auth via
 # $cube_admin_token below prevents outside abuse; operators may further
 # tighten this with iptables / security groups.
 CUBE_PROXY_ADMIN_LISTEN="${CUBE_PROXY_ADMIN_LISTEN:-${CUBE_SANDBOX_NODE_IP}}"
+# Port the /admin/* server binds to. cube-proxy uses host networking, so this
+# port must be free on the host; override when 8082 is already taken.
+CUBE_PROXY_ADMIN_PORT="${CUBE_PROXY_ADMIN_PORT:-8082}"
```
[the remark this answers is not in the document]
mon3stera: bot is right

mon3stera:
Overall, the changes look good. Please address the remaining suggestions left by the bot

mon3stera (approved):

chenhengqi:
@yingjun8  @mon3stera Thanks. Will take this in our next release.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
```diff
@@ -118,8 +112,15 @@ locals {
       "__CUBE_PROXY_SSL_KEY__",
       "cube.app+3-key.pem"
     ),
-    "__CUBE_PROXY_ADMIN_LISTEN__:8082",
-    "0.0.0.0:8082"
+    "__CUBE_PROXY_ADMIN_LISTEN__",
+    "0.0.0.0"
+  )
+  # Pods run with per-pod networking, so the admin server binds 0.0.0.0 and the
```
fslongjin: @tinklone Hi, could you please take a look and see if this change is correct?

fslongjin (commented):
## Scope review

The core change (`CUBE_PROXY_ADMIN_PORT` threaded through templates / IDs / URLs / TF, plus admin in the service-level conflict/readiness checks) looks necessary and focused.

However, roughly ~2/3 of the diff is a broader installer/quickcheck diagnostics rewrite that is only loosely related to making the admin port configurable. Suggest splitting into:
1. **This PR**: admin port config + service/postcheck conflict detection
2. **Follow-up**: installer UX / `check_unit_active` fail-fast rewrite

If kept together, please at least address the P1 dead-path issue on `dump_failed_cube_units`, and call out the installer UX work explicitly in the PR summary.

deploy/one-click/install.sh
```diff
@@ -855,13 +904,57 @@ install_systemd_units() {
...
   local target
   target="$(systemd_target_for_role "${DEPLOY_ROLE}")"
   systemctl disable --now \
     cube-sandbox-control.target \
     cube-sandbox-compute.target >/dev/null 2>&1 || true
-  systemctl enable --now "${target}"
+  # `enable --now` returns non-zero as soon as a required child unit fails to
+  # start. Without this guard `set -e` would abort the installer right here --
+  # before the quickcheck below ever runs -- leaving no diagnostics on screen.
+  if ! systemctl enable --now "${target}"; then
```
fslongjin: **[P1] `dump_failed_cube_units` likely never runs on the #945 failure path**

This diagnostics dump is gated on `if ! systemctl enable --now "${target}"`. But this PR (and the new `quickcheck.sh` comments) already document that `cube-sandbox-control.target` only `Wants=` its members — so a failed child unit (including cube-proxy) typically still leaves `enable --now` returning **0**.

That means the actual #945 path — "8082 taken → cube-proxy crash-loops → installer appears to succeed / stalls silently" — **almost never enters** this branch. The helper covers a Requires=/hard-dependency failure mode, not the Wants= silent-failure mode this PR claims to fix.

Please either:
1. Drop this from the PR and move it to a separate installer-UX change; or
2. Run the scan **after a successful** `enable --now`, looking for non-active `cube-sandbox-*.service` units (so Wants= failures are actually covered).

As written, ~45 lines of new code don't help close #945 but do add maintenance surface.

deploy/one-click/install.sh
```diff
@@ -855,13 +904,57 @@ install_systemd_units() {
...
+}
+
 start_systemd_target() {
   local target
   target="$(systemd_target_for_role "${DEPLOY_ROLE}")"
   systemctl disable --now \
     cube-sandbox-control.target \
     cube-sandbox-compute.target >/dev/null 2>&1 || true
-  systemctl enable --now "${target}"
+  # `enable --now` returns non-zero as soon as a required child unit fails to
```
fslongjin: **[P1] Internal contradiction with the Wants= narrative**

This comment says `enable --now` returns non-zero as soon as a **required** child fails.

But lines 258–260 in this file, and `quickcheck.sh` lines 135–137, emphasize that the control target only uses `Wants=`, so a member failure does **not** make `enable --now` fail.

Please reconcile: if Wants= is the real model, this guard does not cover #945; if some path really fails `enable --now`, cite the concrete dependency graph. As written, future readers will wrongly assume "installer dump covers port conflicts".

deploy/one-click/install.sh
```diff
@@ -252,6 +252,55 @@ check_proxy_cert_preflight() {
...
+# one is the source of truth for the `systemctl restart cube-sandbox-cube-proxy
+# .service` path which bypasses install.sh entirely. There is an inherent
+# TOCTOU window between this preflight and the actual container start, so the
+# service-level check must stay.
+#
+# MUST run after stop_existing_systemd_deployment: on an upgrade the *previous*
+# deployment legitimately holds these ports, so probing before the stop would
+# always false-positive. After the stop our own services have released the
+# ports and a remaining listener means a genuine external conflict.
+check_cube_proxy_port_conflict_preflight() {
```
fslongjin: **[P2] Duplicate of the `up-cube-proxy.sh` conflict check — not required for the feature loop**

`up-cube-proxy.sh` already checks HTTP/HTTPS/admin conflicts and (per the comments here) remains the source of truth for `systemctl restart cube-proxy`. This copies the same logic again and explicitly admits a TOCTOU window.

For closing #945, service-level detection + postcheck probing the admin port is sufficient. An install-time preflight only buys "fail before starting unrelated units" UX — it is not part of making the admin port configurable.

Suggest splitting to a follow-up, or at least sharing one implementation so the two paths cannot drift (e.g. install still lacks the same port-format validation as `up-cube-proxy.sh`).

deploy/one-click/scripts/one-click/quickcheck.sh
```diff
@@ -123,13 +123,133 @@ wait_until() {
...
+# SubState, Result, plus any in-flight Job and NRestarts). Reading only
+# `is-active` conflates them and produced two bugs in the previous version:
+#   1. A TOCTOU race -- it queried `is-active --quiet` then `is-active` again,
+#      so a state flip between the two calls could return the wrong verdict.
+#   2. It treated every non-active status as terminal (e.g. `activating`,
+#      `inactive` with a pending start Job) and would fail-fast on a unit that
+#      was legitimately still coming up.
+# We therefore fetch every category in a single `systemctl show` call and branch
+# exactly once per iteration on the snapshot, so the verdict is consistent.
 check_unit_active() {
```
fslongjin, line 128: **[P2] Related, but scope-creep — consider a separate PR**

This grew from "don't burn the readiness budget on a already-`failed` unit" into a full systemd state machine (Load/Active/Sub/Result/Job/NRestarts + journal dumps + three new tests). That is a general fix for silent failures under `Wants=`, not "add `CUBE_PROXY_ADMIN_PORT`".

Helpful for #945, but the volume (~+120 implementation, ~+110 tests) already dwarfs the core feature. Prefer a dedicated PR (e.g. *fail-fast systemd readiness*). Keep this PR to admin-port wiring + service/postcheck probing.

If it must stay: please say so explicitly in the title/summary so reviewers don't evaluate this as a small config knob.

deploy/one-click/scripts/one-click/up-cube-proxy.sh
```diff
@@ -86,6 +89,36 @@ case "${COMPOSE_DETACH}" in
...
+   || "${CUBE_PROXY_ADMIN_PORT}" == "${CUBE_PROXY_HTTPS_PORT}" ]]; then
+  die "CUBE_PROXY_ADMIN_PORT (${CUBE_PROXY_ADMIN_PORT}) must differ from CUBE_PROXY_HTTP_PORT (${CUBE_PROXY_HTTP_PORT}) and CUBE_PROXY_HTTPS_PORT (${CUBE_PROXY_HTTPS_PORT})"
+fi
+# HTTP and HTTPS must also differ from each other: nginx generates two `listen`
+# directives for them (one plain, one `ssl reuseport`) from the same template,
+# so a duplicated port makes both try to bind the same host port. The `ss`
+# pre-check only catches an *external* listener, not this self-conflict, so an
+# equal pair would otherwise crash-loop the container with a cryptic
+# "address already in use" from nginx.
+if [[ "${CUBE_PROXY_HTTP_PORT}" == "${CUBE_PROXY_HTTPS_PORT}" ]]; then
```
fslongjin: **[P3] HTTP≠HTTPS check is orthogonal to admin-port configurability**

Validating admin ≠ HTTP/HTTPS belongs with this feature. Requiring HTTP ≠ HTTPS is sensible hardening, but it is unrelated to making the admin port configurable (likely review/bot follow-on).

Non-blocking — just further amplifies the "rewriting the whole port subsystem" feel. Fine as a follow-up if we want a tighter PR.

deploy/one-click/install.sh
```diff
@@ -252,6 +252,55 @@ check_proxy_cert_preflight() {
   :
 }
 
+# cube-proxy runs with network_mode: host, so its HTTP/HTTPS/admin ports must be
```
fslongjin: **[P3] Comment essay is too long for an install script**

~20 lines here re-litigate TOCTOU, Wants=, and why this must run after stop — better suited to the PR description / a short design note than inline in `install.sh`.

Same pattern shows up in `quickcheck.sh` / `up-cube-proxy.sh` in this PR. Please compress to ~3–5 lines: what it does, why it must run after stop, and that the service-level check remains source of truth.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
```diff
@@ -112,8 +112,15 @@ locals {
...
+  # port is parametrized independently for consistency with the systemd path.
+  cubeproxy_nginx_conf_rendered = replace(
```
fslongjin: **[P3] Admin-port `replace()` is inconsistent with the main token chain**

`__CUBE_PROXY_ADMIN_PORT__` is substituted in a separate `replace()` while other tokens live in the nested chain. Functionally fine, but easy to miss when adding/removing tokens later.

Consider folding it into the main chain with `tostring(var.cube_proxy_admin_port)` for consistency. Non-blocking.

deploy/one-click/terraform/tencentcloud/tke-addons.tf
tinklone: LGTM

PR #956, opened 2026-07-14
lifecycle: Sync pause/resume state from CubeMaster to CLM

chenhengqi:
CubeMaster's /cube/sandbox/update handler is stateless — it forwards pause/resume to Cubelet but never tells CLM. When a user manually resumes a paused sandbox via connect(), CLM's Redis state key and CubeProxy's per-worker state dict stay stuck at "paused". The next dataplane request lands on /internal/resume, where CLM's resumer rejects it because auto_resume=False, and the client sees 503 until either the next sweep or a full TTL expiry lets state converge.

This change plumbs a state signal end-to-end:

  * CubeMaster: new OpState event on the existing lifecycle stream, emitted from a new UpdateSandboxSuccessHook after any RetCode==200 pause/resume. Only terminal states ("paused"/"running") are broadcast; transition markers stay private to CLM's SETNX flow. Introducing a hook (instead of calling lifecycle.PublishState directly from sandbox_update.go) keeps the sandbox→lifecycle dependency direction and mirrors the existing create/destroy hooks.

  * CLM: new statesync package consumes OpState events. Reconciles the Redis state key + CubeProxy state dict + registry.LastActiveMs (for running events, so sweeper doesn't immediately re-pause). Skips reconciliation when Redis is at a transition marker (pausing/resuming) so the CLM's own tryPause/doResume can't be clobbered by an external event.

  * Resumer fix: the AutoResume check used to run at the top of doResume, rejecting every request even when Redis already reports "running". After this change it lives inside the ownErr==nil branch — the only path that would call CubeMaster.Resume — so the errAlreadyRunning no-op path lets requests through when the state is already correct (regardless of AutoResume), and a false AutoResume with a still-paused state releases the resuming lock via ClearState so the sweeper isn't stuck for a full TTL.

### DIFF
--- a/cube-lifecycle-manager/internal/statesync/handler.go
+++ b/cube-lifecycle-manager/internal/statesync/handler.go
@@ -1,0 +1,154 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+// Package statesync reconciles the CLM's view of a sandbox's runtime
+// state with pause / resume actions driven externally through CubeMaster.
+//
+// CubeMaster is a stateless proxy: when the SDK calls Sandbox.connect() to
+// resume a paused sandbox, CubeMaster forwards the RPC to Cubelet but does
+// not touch the CLM's state key or CubeProxy's per-worker state dict.
+// Without a nudge from CubeMaster the CLM would keep thinking the
+// sandbox is paused and reject the next request from the dataplane.
+//
+// CubeMaster now emits an OpState event on the lifecycle events stream after
+// every successful pause / resume RPC (see CubeMaster/pkg/lifecycle/store.go
+// PublishState). This package consumes those events and reconciles:
+//
+//   - Redis state key (cube:v1:shared:sandbox:lifecycle:state:<id>)
+//   - CubeProxy's cube_sandbox_state dict (via proxypush.SetState)
+//   - registry.Entry.LastActiveMs when the new state is "running"
+//     (mirrors resumer.doResume: avoids the sweeper re-pausing a sandbox
+//     that just came back).
+//
+// The event source (CubeMaster) is stateless with respect to the CLM's
+// SETNX-based transition locks ("pausing", "resuming"). To avoid a state
+// event racing an in-flight sweeper.tryPause or resumer.doResume, Handle
+// skips reconciliation whenever the current state key holds a transition
+// marker; the CLM's own flow will write the terminal state moments
+// later, so the event is redundant.
+package statesync
+
+import (
+	"context"
+	"time"
+
+	"go.uber.org/zap"
+
+	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/lifecycle"
+	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/redisstream"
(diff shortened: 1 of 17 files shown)


cube-lifecycle-manager/internal/statesync/handler.go
```diff
@@ -0,0 +1,154 @@
...
+	if cur == "pausing" || cur == "resuming" {
+		log.Info("state event skipped: transition in progress",
+			zap.String("sandbox_id", ev.SandboxID),
+			zap.String("cur", cur),
+			zap.String("new", newState))
+		return
+	}
+
+	// Idempotence: already at the desired state.
+	if cur == newState {
```
ls-ggg, line 123: Consider always pushing the state to CubeProxy (even when Redis already holds the desired terminal state). That way a failed or missed proxy update can still be healed on a later OpState event, instead of leaving CubeProxy stuck with a stale paused/running view

cube-lifecycle-manager/internal/statesync/handler.go
```diff
@@ -0,0 +1,154 @@
...
+			zap.String("sandbox_id", ev.SandboxID), zap.Error(err))
+		return
+	}
+
+	// Transition-lock protection: the CLM itself is mid-flight and will
+	// write the terminal state as part of tryPause / doResume. Applying an
+	// external event here would race against that write and could clobber
+	// a "resuming" lock with a "paused" verdict — the opposite of what we
+	// want.
+	if cur == "pausing" || cur == "resuming" {
```
ls-ggg, line 114: Another CLM path can acquire the transition lock after this GetState and still get overwritten by the unconditional SET

cube-lifecycle-manager/internal/statesync/handler.go
kinwin-ustc: Alternatively, when CubeProxy receives an exception, it can detect its own state inconsistency and thus ensure eventual state consistency. For example, if CubeProxy is in a paused state while the sandbox is actually running, the CLM receives a resume request, finds that the sandbox is actually running, and returns information to correct CubeProxy's state. Or, if CubeProxy is in a running state but the sandbox is actually paused, and the data stream cannot provide service, it triggers an active state alignment for the sandbox.

zhouxianping (approved):

novahe:
@kinwin-ustc It seems like a similar issue; should we standardize on an event-driven notification mechanism—akin to the `etcd + client-go` pattern in Kubernetes—but using `Redis + xxx` instead?
ref: https://github.com/TencentCloud/CubeSandbox/issues/1047#issuecomment-5029832290

kinwin-ustc:
@novahe  Agreed. The single source of truth in the sandbox state + event synchronization of state transitions between different components + eventual consistency guarantee for the state caches of each component.

cube-lifecycle-manager/internal/statesync/handler.go
chenhengqi: > if CubeProxy is in a paused state while the sandbox is actually running, the CLM receives a resume request, finds that the sandbox is actually running, and returns information to correct CubeProxy's state.

We already have it.

> Or, if CubeProxy is in a running state but the sandbox is actually paused, and the data stream cannot provide service, it triggers an active state alignment for the sandbox.

The state is possible due to lost event. For eventual consistency, we will need reconciliation. Will keep track of this in our v0.7.0 development cycle.

PR #958, opened 2026-07-14
fix(templatecenter): make loop-mount streaming ext4 build robust

dushulin:
## Summary

`createExt4ImageStreaming` — the Phase-2 loop-mount rootfs build path gated by
`CUBEMASTER_LOOP_MOUNT_EXT4_ENABLED` — has three bugs. The first two make the
loop-mount path fail and silently fall back to the slow Phase-1
`mkfs.ext4 -d` build; the third builds fast but yields an artifact the compute
node refuses to boot.

## Bugs & fixes (all in `createExt4ImageStreaming`)

1. **Truncate before the store sub-dir exists → ENOENT.**
   The image is `truncate`d at `<store>/<artifact>/<artifact>.ext4` before that
   per-artifact directory exists (`ResolveArtifactStoreDir` only creates the
   `storage/` root; Phase-1 creates the sub-dir lazily during its `cp -a`
   relocate). The streaming path never relocates, so the truncate fails and the
   whole build falls back to Phase-1.
   **Fix:** `os.MkdirAll(filepath.Dir(ext4Path), 0o755)` before truncate.

2. **`losetup` stderr folded into the device path.**
   `losetup --find --show` is run with `CombinedOutput()`. losetup prints
   warnings to stderr (e.g. `Warning: file does not fit into a 512-byte sector;
   the end of the file will be ignored.`), which `CombinedOutput` merges into the
   returned string. The parsed `loopDevice` becomes a multiline value, so `mount`
   fails with exit 32 ("bad option"), and the same corrupt string breaks the
   deferred `losetup --detach`, leaking the loop device.
   **Fix:** capture stdout only via `cmd.Output()`, route stderr to a separate
   buffer used only for the error message.

3. **Final image not 2 MiB-aligned → boot rejected.**
   After `resize2fs -M`, the image is truncated to the filesystem length, which
   is only 512-byte aligned. The rootfs is later attached as a **virtio-pmem**
   device whose backing-file size must be a multiple of 2 MiB, otherwise the host
   device manager rejects it (`PmemSizeNotAligned`). Phase-1 avoids this for free
   because it sizes up to a 256 MiB boundary.
   **Fix:** round the apparent filesystem size **up** to 2 MiB before the final
   truncate. Rounding up never cuts into live filesystem blocks; the small tail
   is unused device space. Uses the apparent size (`os.Stat().Size()`), not
   `getFileBlockSize` (physically-allocated blocks), which can be smaller than
   the filesystem length for a minimized image and would truncate below it.

## Testing

Verified on a FUSE-mounted network filesystem artifact store, `from-image` build of a
`python:3.11-slim`-based template distributed to a bare-metal compute node:

- Phase-2 loop-mount path runs **end-to-end with no fallback** (0 "falling back
  to phase-1" warnings; `losetup`/`mount` show the loop device mounted at the
  streaming mount point).
- Build time **POST→READY ≈ 14 s** vs **~12 min** on Phase-1 for the same image
  on the same store (~50x; Phase-1 is dominated by per-small-file FUSE metadata
  round-trips during the `cp -a` relocate + `mkfs.ext4 -d` readback).
- Resulting artifact is **2 MiB-aligned** (e.g. 438304768 B = 418 MiB exactly)
  and the from-image job reaches **READY** — the compute node accepts the pmem
  device with **0** `PmemSizeNotAligned` errors.

### DIFF
--- a/CubeMaster/pkg/templatecenter/image/disk.go
+++ b/CubeMaster/pkg/templatecenter/image/disk.go
@@ -46,6 +46,12 @@ func createExt4ImageStreaming(ctx context.Context, source *PreparedSource, workD
 		return fmt.Errorf("loop mount not available")
 	}
 
+	// Ensure the per-artifact store dir exists (Phase-1 creates it lazily via relocate); otherwise truncate ENOENTs → silent fallback.
+	// 0o700 (owner-only, matching the mount point below): the finished image is a full container rootfs and must not be world-readable on a shared host.
+	if err := os.MkdirAll(filepath.Dir(ext4Path), 0o700); err != nil {
+		return fmt.Errorf("create artifact store dir for streaming: %w", err)
+	}
+
 	// 2. Create empty ext4 image using the caller-provided size estimate.
 	if err := runCommand(ctx, "", "truncate", "-s", strconv.FormatInt(estimatedSizeBytes, 10), ext4Path); err != nil {
 		return fmt.Errorf("truncate ext4 image for streaming: %w", err)
@@ -74,10 +80,18 @@ func createExt4ImageStreaming(ctx context.Context, source *PreparedSource, workD
 	}
 	defer cleanup()
 
-	// Allocate a free loop device and mount.
-	loopOut, err := exec.CommandContext(ctx, "losetup", "--find", "--show", ext4Path).CombinedOutput()
+	// Allocate a free loop device and mount. Capture stdout ONLY for the device
+	// path: losetup emits warnings (e.g. "file does not fit into a 512-byte
+	// sector") to stderr, and CombinedOutput would fold that warning into the
+	// parsed device path — corrupting the loopDevice string so every later mount
+	// and detach receives a garbage argument (mount fails "bad option", detach
+	// fails → loop device leaks).
+	losetupCmd := exec.CommandContext(ctx, "losetup", "--find", "--show", "--", ext4Path)
+	var losetupErr bytes.Buffer
+	losetupCmd.Stderr = &losetupErr
+	loopOut, err := losetupCmd.Output()
 	if err != nil {
-		return fmt.Errorf("losetup --find --show %s failed: %w: %s", ext4Path, err, string(loopOut))
+		return fmt.Errorf("losetup --find --show %s failed: %w: %s", ext4Path, err, strings.TrimSpace(losetupErr.String()))
 	}
 	loopDevice := strings.TrimSpace(string(loopOut))
 	detachLoop := func() {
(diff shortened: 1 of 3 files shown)


CubeMaster/pkg/templatecenter/image/disk.go
```diff
@@ -127,11 +141,35 @@ func createExt4ImageStreaming(ctx context.Context, source *PreparedSource, workD
...
+	// rootfs as a virtio-pmem device whose backing-file size MUST be a multiple of
+	// 2 MiB, else the host device manager rejects it ("PmemSizeNotAligned"). Phase-1
+	// sizing gets this for free (it aligns up to a 256 MiB boundary); the streaming
+	// path shrinks with resize2fs -M, which truncates the file to the exact
+	// filesystem length (not 2 MiB-aligned). Align that apparent size up here —
+	// rounding UP never cuts into live filesystem blocks; the small tail is just
+	// unused device space. Use the apparent size (fi.Size(), the fs logical length
+	// after resize2fs -M), NOT the physically-allocated block count, which can be
+	// smaller than the fs length for a minimized image and would truncate below it.
+	const pmemAlignmentBytes = int64(2 * 1024 * 1024)
```
chenhengqi, line 154: Is this arch specific?

dushulin: No, it's not architecture-specific. This 2 MiB deliberately mirrors the VMM's own hard-coded pmem check in hypervisor/vmm/src/device_manager.rs:2788:

if size % 0x20_0000 != 0 {
  return Err(DeviceManagerError::PmemSizeNotAligned);
}
// The memory needs to be 2MiB aligned in order to support hugepages.

0x20_0000 is a fixed 2 MiB literal, not gated by any #[cfg(target_arch)], so it's the same value on x86_64 and aarch64. 2 MiB is the PMD-level huge-page size on x86_64 and on aarch64 with 4 KiB base pages — the configs this VMM's pmem/DAX path targets — which is why the VMM hard-codes it. By construction the Go side then produces exactly the size the VMM's check accepts, so it can't be under- or over-aligned relative to the VMM.
If you'd prefer, I can add a short comment pointing at device_manager.rs:2788 as the source of truth so the two constants don't silently drift.

dushulin:
@chenhengqi  hi～PTAL, thanks!

CubeMaster/pkg/templatecenter/image/disk.go
chenhengqi: @lkml-likexu Please take a look here. Gemini told me that ARM64 with 64k page size does not follow this 2MB alignment here. It should be 512MB. I don't know.

fslongjin (approved):
LGTM

PR #984, opened 2026-07-16
Decoupling of CubeAPI Operations Logic and Architecture Refactoring

liciazhu:
Completely decouple the operations logic from CubeAPI into an independent Go service (CubeOps). CubeOps is exposed exclusively to the internal network and features an independent authentication mechanism. CubeAPI remains stateless and publicly accessible, retaining only the SDK / E2B compatible interfaces. All non-SDK and non-E2B compatible interfaces, except for those related to CubeSandbox SDK and E2B SDK, are migrated to CubeOps. The SDK interfaces in CubeAPI originally accessed by the WebUI are modified to retrieve data via a direct connection to CubeMaster.


CubeOps/internal/auth/handler.go
fslongjin, line 1: please use gin framework instead of manual-write http handlers

CubeOps/internal/config/config.go
```diff
@@ -0,0 +1,145 @@
...
+	// CubeAPI (for SDK endpoint proxy)
+	CubeAPIURL string
+
+	// Redis (optional)
+	RedisURL string
+
+	// Sandbox domain exposed to SDK clients; matches SDK handler's
+	// CUBE_API_SANDBOX_DOMAIN env so the /config endpoint stays in sync.
+	SandboxDomain string
+}
```
fslongjin, line 65: use a config yml is better.

CubeOps/internal/cubemaster/client.go
```diff
@@ -0,0 +1,317 @@
...
+}
+
+// ClusterVersions fetches version information from CubeMaster.
+func (c *Client) ClusterVersions(ctx context.Context) (json.RawMessage, error) {
+	return c.get(ctx, "/internal/meta/version-matrix")
+}
+
+// GetSandbox fetches sandbox detail from CubeMaster.
+func (c *Client) GetSandbox(ctx context.Context, sandboxID, instanceType string) (json.RawMessage, error) {
+	return c.get(ctx, fmt.Sprintf("/cube/sandbox/info?sandbox_id=%s&instance_type=%s", sandboxID, instanceType))
```
fslongjin: this can cause security problems

CubeOps/internal/cubemaster/client.go
```diff
@@ -0,0 +1,317 @@
...
+
+// RollbackSandbox rolls back a sandbox to a snapshot via CubeMaster.
+func (c *Client) RollbackSandbox(ctx context.Context, sandboxID string, body interface{}) (json.RawMessage, error) {
+	return c.post(ctx, fmt.Sprintf("/cube/sandbox/%s/rollback", sandboxID), body)
+}
+
+// UpdateSandbox sends a pause/resume action to CubeMaster.
+func (c *Client) UpdateSandbox(ctx context.Context, body interface{}) (json.RawMessage, error) {
+	return c.post(ctx, "/cube/sandbox/update", body)
+}
```
fslongjin, line 175: maybe we can use cubemaster's grpc endpoint?

CubeOps/internal/handler/openclaw_apply.go
fslongjin, line 1: I think that we should add a `service` layer.

fslongjin (changes requested):

kinwin-ustc:
Remove all formatting noise and retain only the substantive modifications.

fslongjin (approved):

PR #986, opened 2026-07-16
sdk/python: accept E2B envs alias

### LINKED ISSUE #966: [Bug Report] Python SDK create environment variable option differs from E2B
## Summary

CubeSandbox Python SDK uses `env_vars` when creating a sandbox with environment variables:

```python

sandbox = Sandbox.create(
    template="<template-id>",
    env_vars={"FOO": "bar"},
)
```

E2B Python SDK uses `envs` for the same user-facing option:

```python
from e2b import Sandbox

sandbox = Sandbox.create(
    template="<template-id>",
    envs={"FOO": "bar"},
)
```

This difference makes E2B-compatible examples, tests, and migration code fail or require adapter-specific branching.

## Impact

Users migrating from E2B or running SDK compatibility tests cannot use the same create options across SDKs. In particular, backend-neutral tests that pass `env_vars` must translate the option to `envs` for E2B, while E2B-style code that passes `envs` may not work with CubeSandbox Python SDK.

This is a small API compatibility gap, but it affects common sandbox creation workflows because create-time environment variables are a frequently used option.

## Expected behavior

CubeSandbox Python SDK should support the E2B-compatible `envs` parameter in `Sandbox.create()`.

For backwards compatibility, `env_vars` should continue to work. Both of these should be accepted:

```python
Sandbox.create(template="<template-id>", env_vars={"FOO": "bar"})
Sandbox.create(template="<template-id>", envs={"FOO": "bar"})
```

Both should send the same CubeAPI request body field:

```json
{
  "envVars": {"FOO": "bar"}
}
```

## Suggested fix

Add `envs: Dict[str, str] | None = None` to `Sandbox.create()` and normalize it to the existing `env_vars` field before building the request payload.

Suggested behavior:

- If only `env_vars` is provided, use `env_vars`.
- If only `envs` is provided, use `envs`.
- If both are provided and differ, raise `ValueError` to avoid silent ambiguity.
- If both are provided and equal, accept the value.

Also add unit tests for all of the above cases.

## Related compatibility note

The CubeAPI HTTP field `envVars` is already compatible with E2B's wire format. This issue is about the Python SDK public parameter name only.

tianrking:
## Summary

- accept `envs` as the E2B-compatible alias for `Sandbox.create(..., env_vars=...)`
- preserve existing `env_vars` behavior and reject ambiguous, conflicting values
- document the public alias and cover alias, equivalent dual-input, and conflict cases

Fixes #966

## Validation

- `python -m pytest` (200 passed)
- `python -m ruff check cubesandbox tests`
- `git diff --check`

Autonomously-by: Codex:GPT-5


fslongjin:
Hello~Can you add more test results/screenshots which can prove this feature can work in a **REAL** CubeSandbox cluster?

tinklone:
Thanks for the fix. The unit-level behavior looks aligned with #966, but I think two follow-ups are needed before this is ready:

1. Please fix the DCO failure. The current commit message has no `Signed-off-by:` trailer, and the `Developer Certificate of Origin` check is failing. You can amend with signoff, for example `git commit --amend -s`, then force-push the PR branch.

2. Please add an SDK compatibility E2E case under `tests/e2e/sdk_compat`, not only Python SDK unit tests. The best fit is `tests/e2e/sdk_compat/cases/lifecycle/test_create_options.py`, next to `test_create_env_vars_visible_to_command`. Add a case using `@pytest.mark.sandbox_create_options(envs={"SDK_COMPAT_E2E_VAR": "..."})` and verify the value from inside the sandbox with `sdk_sandbox.run_command('printf "%s" "$SDK_COMPAT_E2E_VAR"', ...)`. That proves the new `envs` alias works against a real CubeSandbox cluster and should address the existing request for real-cluster validation.

wbzdssm:
Thank you for your pull request. I hope you can resolve the conflicts and merge the code as soon as possible.

PR #993, opened 2026-07-16
build(builder): speed up and harden builder image builds

lkml-likexu:
This PR contains two related improvements to the builder image workflow.

1. Route Ubuntu apt packages through the China mirror (e46989d)

The MIRROR=cn opt-in previously only redirected the llvm.sh installer and clang-14 packages to a China-reachable mirror. The bulk of the Ubuntu apt packages still came from archive.ubuntu.com
/ ports.ubuntu.com, which are slow or unreachable from mainland China — leaving builds bottlenecked on exactly the part MIRROR=cn was meant to speed up.

This change extends MIRROR=cn to also source the Ubuntu apt archive from the mirror, covering:
- Both amd64 and arm64
- The builder image and the CubeEgress openresty images

When MIRROR is unset, builds keep using the upstream defaults, and CI continues to build against upstream. The LLVM GPG key still comes from apt.llvm.org. Docs (CONTRIBUTING.md /
CONTRIBUTING_zh.md) are updated to match.

2. Add --no-cache to BUILDER_FORCE_REBUILD rebuilds (e512211)

When BUILDER_FORCE_REBUILD was set, the builder-image target rebuilt the image but still let Docker reuse cached layers, so a forced rebuild could silently reproduce a stale image instead of
refreshing it.

Now --no-cache is passed only on forced rebuilds, so the image is rebuilt from scratch. A plain first build (image missing) still uses the layer cache for speed.

Test plan

- MIRROR=cn make builder-image builds successfully on amd64 and arm64
- Default (MIRROR unset) build still pulls from upstream mirrors
- BUILDER_FORCE_REBUILD=1 make builder-image rebuilds without reusing cached layers

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -44,8 +44,9 @@ Cube Sandbox provides a Docker-based builder image for a consistent build enviro
 # Build the builder image
 make builder-image
 
-# From mainland China, fetch the llvm.sh installer and clang-14 apt packages from
-# a China mirror (the LLVM GPG key still comes from apt.llvm.org)
+# From mainland China, source apt packages from China-reachable mirrors: the
+# Ubuntu apt archive plus the llvm.sh installer and clang-14 packages (the LLVM
+# GPG key still comes from apt.llvm.org)
 make builder-image MIRROR=cn
 
 # Start an interactive shell inside the builder
--- a/CONTRIBUTING_zh.md
+++ b/CONTRIBUTING_zh.md
@@ -44,8 +44,8 @@ Cube Sandbox 提供了基于 Docker 的构建镜像，以确保一致的构建�
 # 构建构建镜像
 make builder-image
 
-# 中国大陆用户可通过镜像源获取 llvm.sh 安装脚本与 clang-14 apt 软件包
-# （LLVM GPG 签名密钥仍从 apt.llvm.org 获取）
+# 中国大陆用户可通过镜像源获取 apt 软件包：包括 Ubuntu apt 仓库，以及 llvm.sh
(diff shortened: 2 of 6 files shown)


CONTRIBUTING.md
```diff
@@ -44,8 +44,9 @@ Cube Sandbox provides a Docker-based builder image for a consistent build enviro
 # Build the builder image
 make builder-image
 
-# From mainland China, fetch the llvm.sh installer and clang-14 apt packages from
-# a China mirror (the LLVM GPG key still comes from apt.llvm.org)
+# From mainland China, source apt packages from China-reachable mirrors: the
+# Ubuntu apt archive plus the llvm.sh installer and clang-14 packages (the LLVM
```
chenhengqi, line 48: Change `clang-14` to `clang` so that we don't have to update this comment when we bump clang version.

BTW, do you mind bump clang to some newer version? The newer, the better for BPF.

network-agent/internal/service/local_service_test.go
```diff
@@ -263,7 +263,7 @@ func TestRecoverCleansOrphanTapsWithoutPersistedState(t *testing.T) {
 			Name:  tap.Name,
 			Index: tap.Index,
 			IP:    tap.IP,
-			File:  os.NewFile(uintptr(1), "/dev/null"),
```
chenhengqi, line 266: You can drop changes to this file. We are refactoring it.

CONTRIBUTING.md
lkml-likexu: Let me cover it in another PR.

fslongjin (approved):

PR #995, opened 2026-07-16
fix(sdk/python): send X-API-Key on all control- and data-plane requests

wbzdssm:
When CubeAPI is started with an auth-callback URL, every request must carry
X-API-Key. Before this fix the Python SDK never sent an API key, so both
control-plane and data-plane calls were rejected with 401.

- Add Config.api_key reading CUBE_API_KEY (fallback E2B_API_KEY)
- Add _auth_headers(cfg) as single source of truth
- Attach X-API-Key to all Sandbox.* and Template.* control-plane calls
- Attach X-API-Key to data-plane requests (commands/run_code/files/pty)
- Document CUBE_API_KEY in README

Backward compatible: no key set → no header sent, behavior unchanged.

### DIFF
--- a/sdk/python/cubesandbox/_commands.py
+++ b/sdk/python/cubesandbox/_commands.py
@@ -167,7 +167,12 @@ class Commands:
 
 
 def _envd_rpc_base_url_and_headers(sandbox: "Sandbox") -> tuple[str, dict[str, str]]:
-    headers: dict[str, str] = {}
+    from ._config import _auth_headers
+
+    # The e2b-connect path uses its own httpcore pool (not sandbox._client),
+    # so it does not inherit the client's default headers. Attach the API key
+    # here so CubeAPI's auth middleware accepts the request. No-op when unset.
+    headers: dict[str, str] = dict(_auth_headers(sandbox._config))
     access_token = sandbox._data.get("envdAccessToken")
     if access_token:
         headers["X-Access-Token"] = access_token
--- a/sdk/python/cubesandbox/_commands.py
+++ b/sdk/python/cubesandbox/_commands.py
@@ -10,6 +10,8 @@ import struct
 from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
+from ._config import _auth_headers
+
 if TYPE_CHECKING:
     from .sandbox import Sandbox
 
@@ -167,12 +169,10 @@ class Commands:
 
 
 def _envd_rpc_base_url_and_headers(sandbox: "Sandbox") -> tuple[str, dict[str, str]]:
-    from ._config import _auth_headers
-
     # The e2b-connect path uses its own httpcore pool (not sandbox._client),
     # so it does not inherit the client's default headers. Attach the API key
     # here so CubeAPI's auth middleware accepts the request. No-op when unset.
-    headers: dict[str, str] = dict(_auth_headers(sandbox._config))
+    headers: dict[str, str] = _auth_headers(sandbox._config)
     access_token = sandbox._data.get("envdAccessToken")
     if access_token:
         headers["X-Access-Token"] = access_token
--- a/sdk/python/cubesandbox/_config.py
+++ b/sdk/python/cubesandbox/_config.py
@@ -17,8 +17,8 @@ class Config:
         default_factory=lambda: os.environ.get("CUBE_API_URL", "http://127.0.0.1:3000")
     )
     api_key: str | None = field(
-        default_factory=lambda: os.environ.get("CUBE_API_KEY")
-        or os.environ.get("E2B_API_KEY")
+        default_factory=lambda: os.environ.get("CUBE_API_KEY"),
+        repr=False,
     )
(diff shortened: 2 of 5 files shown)


sdk/python/cubesandbox/_config.py
```diff
@@ -14,6 +14,10 @@ class Config:
     api_url: str = field(
         default_factory=lambda: os.environ.get("CUBE_API_URL", "http://127.0.0.1:3000")
     )
+    api_key: str | None = field(
```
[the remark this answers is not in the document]
wbzdssm: Done in 286a9e0. Added `repr=False` to the `api_key` field so the key is omitted from the dataclass `__repr__`. Also dropped the `E2B_API_KEY` fallback so it now reads `CUBE_API_KEY` only, keeping the env-var naming consistent with the other `CUBE_*` settings (the fallback was introduced in this same PR and never shipped, so no impact).

sdk/python/cubesandbox/_commands.py
```diff
@@ -167,7 +167,12 @@ def _run_with_connect_fallback(
 
 
 def _envd_rpc_base_url_and_headers(sandbox: "Sandbox") -> tuple[str, dict[str, str]]:
-    headers: dict[str, str] = {}
+    from ._config import _auth_headers
```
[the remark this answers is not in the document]
wbzdssm: Done in 286a9e0. Hoisted `from ._config import _auth_headers` to module level, consistent with the rest of the package. No circular-import risk since `_config` only depends on `os`/`dataclasses`.

sdk/python/cubesandbox/_commands.py
```diff
@@ -167,7 +167,12 @@ def _run_with_connect_fallback(
...
+
+    # The e2b-connect path uses its own httpcore pool (not sandbox._client),
+    # so it does not inherit the client's default headers. Attach the API key
+    # here so CubeAPI's auth middleware accepts the request. No-op when unset.
+    headers: dict[str, str] = dict(_auth_headers(sandbox._config))
```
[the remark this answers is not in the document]
wbzdssm: Done in 286a9e0. Dropped the redundant `dict()` wrapper here and at `sandbox.py:801`. `_auth_headers` already returns a freshly allocated dict (`{"X-API-Key": key}` or `{}`), so the later in-place mutation stays safe with no extra copy.

sdk/python/cubesandbox/_config.py
```diff
@@ -30,4 +34,23 @@ class Config:
...
+    CubeAPI only enforces auth when it is started with an auth-callback URL;
+    in the default deployment no callback is set and every request passes
+    through unauthenticated. So the key is optional here: when
+    ``CUBE_API_KEY`` / ``Config.api_key`` is unset we send no
+    auth header and behavior is unchanged; when set we attach
+    ``X-API-Key: <key>`` so the SDK also works against an auth-enabled backend.
+
+    This is the single source of truth shared by the control plane
+    (``sandbox``/``template`` REST calls) and the data plane (envd requests
+    routed through CubeProxy -> CubeAPI's auth middleware).
```
chenhengqi: CubeProxy -> CubeAPI?

wbzdssm: SDK → CubeProxy → envd

sdk/python/cubesandbox/_template.py
```diff
@@ -200,7 +200,8 @@ def get(
         if next_token is not None:
             params["nextToken"] = next_token
         s = requests.Session()
-        resp = s.get(f"{cfg.api_url}/templates/{template_id}", params=params)
+        resp = s.get(f"{cfg.api_url}/templates/{template_id}", params=params,
+                     headers=_auth_headers(cfg))
```
luzhixing12345, line 204: There is a mismatch between the implementation and its unit test.

```bash
cd sdk/python
python -m pytest -q
```

```bash
=================================================== FAILURES ===================================================
___________________________ TestTemplateAPI.test_template_get_parses_network_fields ____________________________
tests/test_sandbox.py:2376: in test_template_get_parses_network_fields
    get.assert_called_once_with(
/root/.local/share/uv/python/cpython-3.13.14-linux-x86_64-gnu/lib/python3.13/unittest/mock.py:991: in assert_called_once_with
    return self.assert_called_with(*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/root/.local/share/uv/python/cpython-3.13.14-linux-x86_64-gnu/lib/python3.13/unittest/mock.py:979: in assert_called_with
    raise AssertionError(_error_message()) from cause
E   AssertionError: expected call not found.
E   Expected: get('http://localhost:3000/templates/tpl-network', params={})
E     Actual: get('http://localhost:3000/templates/tpl-network', params={}, headers={})
```

luzhixing12345: https://github.com/TencentCloud/CubeSandbox/pull/1091

Issue #855, opened 2026-07-09
[Feature] E2B-compatible persistent volumes with a customizable storage plugin framework


ls-ggg:
## Summary

Add support for persistent volumes so sandboxes can keep data across restarts and across instances, with a pluggable framework that lets operators back volumes by any storage system of their choice.

## Motivation / Use cases

E2B-compatible sandboxes are often ephemeral, but many workloads need to persist data beyond a single sandbox's lifetime, for example:

- Reusing large model weights or datasets across many sandboxes without re-downloading.
- Keeping a user's workspace/project files between sessions.
- Sharing the same dataset (read-only or read-write) across multiple sandboxes.

Today there is no first-class way to attach durable storage to a sandbox, and no standard way to integrate different storage backends.

## Proposal

- **E2B compatibility**: expose persistent storage through an API aligned with E2B (declare mounts on sandbox creation, plus a simple volume lifecycle: create / list / delete). Users and SDKs should not need to know how storage is implemented underneath.
- **Customizable plugin framework**: provide an extensible interface so operators can plug in their own storage backend (e.g. object storage, NFS, or a proprietary system) without changing the platform. The backend choice should be selectable per volume.
- **Safe lifecycle**: a volume that is still in use should not be deletable, so data cannot be removed out from under running sandboxes.

## Goals

- Sandboxes can mount durable volumes that survive restarts and are reusable across instances.
- Storage backends are pluggable and swappable by the operator.
- Volume management is safe (no deletion while in use) and transparent to end users/SDKs.

PR #997, opened 2026-07-17
feat(volume): add Volume Plugin framework with COS examples

ls-ggg:
Add e2b-compatible Volume CRUD API, CubeMaster/Cubelet plugin hooks (binary + rpc), cross-node refcount, one-click COS plugin packaging, end-to-end docs/examples, and verify_volume.py.

Implements #855

- **CubeAPI / CubeMaster**: Volume CRUD (`POST/GET/DELETE /volumes`, `GET /volumes/{volumeID}`)
- **Plugin hooks**: CubeMaster Create/Destroy + Cubelet Attach/Detach (binary + rpc)
- **Refcount**: cross-node refcount in MySQL; node-level refcount in Cubelet BoltDB
- **One-click**: COS binary/rpc plugins packaged under `CubeMaster/plugin/` and `Cubelet/plugin/`
- **Examples & docs**: EN/ZH volume plugin guide, COS binary/rpc walkthroughs, `verify_volume.py`

Assisted-by: Cursor:Composer


fslongjin (changes requested):
Thanks for putting together the Volume CRUD API, the Master/Cubelet plugin path, the COS examples, and the English and Chinese documentation. This is a broad change and the overall structure is already fairly complete. I walked through the create, attach, detach, and delete flows across `CubeAPI -> CubeMaster -> Cubelet -> plugin`, with particular attention to concurrency and failure recovery.

There are a few lifecycle and reference-counting issues that I think we should address before merging. The main concern is that the Master-side refcount can remain incorrect after a lost or replayed response, and in some interleavings a Volume that is still mounted, or about to be mounted, can be deleted. The Attach/Detach rollback paths, team isolation, and the security boundary for RPC over TCP also need a little more work. Some of these require a failure or a concurrent request to trigger, but the resulting behavior can involve data deletion or a state that cannot recover automatically.

If I have misunderstood an intended deployment or protocol constraint, please let me know. In particular, if Volume is intentionally single-tenant or TCP is only intended for a trusted local network, it would help to make that explicit in both the API contract and the documentation.

## Changes requested before merge

### 1. Refcount events can drift permanently after loss or replay

`CubeMaster/pkg/volume/refcount/refcount.go:40-99`

Could we revisit the reliability model here? Cubelet reports a delta only when its local count changes from `0 -> 1` or `1 -> 0`, and Master directly applies that delta to a single aggregate count. The event has no node ID, event ID, or version, and there is no Master-side reconciliation.

If a successful response is lost, a retry will not report `0 -> 1` again because Cubelet already holds the local reference. Master may then continue to see zero and allow deletion of a Volume that is still in use. A replayed response has the opposite problem and can leave the Volume permanently undeletable.

Would it be possible to report the current state per `(volume_id, node_id)` and use an idempotent UPSERT/DELETE, with periodic reconciliation as a fallback? A delta-only protocol seems difficult to recover after retries, restarts, or network failures.

### 2. DELETE can race with an Attach that is already in progress

`CubeMaster/pkg/service/httpservice/cube/volume.go:258-291`

There is a concurrency window here that looks worth closing before merge. Sandbox creation resolves the Volume first, but Master does not increment the refcount until Cubelet has completed Attach and returned successfully. During that interval, DELETE can still observe `refcount=0`, call the plugin's Destroy operation, and remove the database row.

If those operations interleave, Cubelet may be attaching a backend that has just been deleted. Its later `+1` also cannot be persisted because the Volume row is gone. For a backend such as COS, this could surface as data being deleted while a sandbox is being created.

Could we introduce an `active/deleting` state and a durable attachment reservation? Sandbox creation could reserve the reference before scheduling, while DELETE atomically transitions the Volume to deleting, prevents new references, and destroys it only after all reservations are gone.

### 3. A failed concurrent Create can destroy the successful request's backend

`CubeMaster/pkg/service/httpservice/cube/volume.go:193-219`

The rollback here can remove another request's newly created resource. Two requests for the same name can both pass the initial lookup and call Create with the same `volumeID=name`. One database insert succeeds; the other hits the unique constraint and unconditionally calls `Destroy(volumeID)`.

Because both operations use the same ID, the losing request cannot tell whether the backend it is destroying belongs to the successful request. Could we reserve a `creating` row in the database first, so only the winning request invokes the plugin? An operation ID or another ownership mechanism would also be needed before compensating with Destroy.

### 4. A failed Detach can leave the Master refcount stuck after retries

`Cubelet/plugins/volume/manager.go:160-194`

The current ordering appears to make this state unrecoverable from a normal retry. Cubelet releases its local reference before invoking the plugin's Detach. If Detach fails, the `1 -> 0` event is included in the failed response, but Master checks the return code first and does not consume it.

A later Detach retry can complete the actual unmount, but the local reference is already `NotHeld`, so Cubelet will not emit another `1 -> 0`. Master may therefore keep the old count indefinitely, and the user continues to receive 409 when deleting the Volume.

Could the reference release be committed only after Detach succeeds, or could pending-detach be persisted as a retryable state? Making the state update idempotent would also avoid relying on a one-shot delta in a failed response.

### 5. Failures after Attach do not run the full rollback path

`Cubelet/storage/pluginvolume.go:202-261`

After `Manager.Attach` succeeds, a failure in mkdir, bind mount, or the readonly remount returns immediately. `PluginVolumeBackendInfos` is only recorded after all those operations succeed, so the outer cleanup cannot see this Volume and will not Detach it.

This can leave a local refcount and plugin/FUSE resources behind; a readonly remount failure may also leave an existing bind mount. Could we register a rollback handle immediately after Attach, or use a deferred rollback that unmounts and Detaches until the operation is committed? Fault-injection tests at each of these points would be useful to confirm that mounts, refcounts, and metadata are restored.

### 6. The team-scoped API currently appears to use global data

`CubeAPI/src/handlers/volumes.rs:22-24`

The API description refers to team Volumes, but I could not find a path that carries the authenticated team/principal from the auth callback to Master. The middleware appears to retain only allow/deny, the Volume table has no owner/team field, and list/get/delete/name uniqueness are all global.

If one deployment accepts credentials for multiple teams, any team allowed through this route could list global Volumes, retrieve another team's token, or mount/delete another team's Volume. Could we confirm the intended boundary? For multi-tenant use, a trusted team ID needs to flow through CubeAPI, Master, the model, and every query. If the product only supports a single-tenant instance, the API text and deployment docs should say that the namespace is global instead.

### 8. The COS binary example reports success after remote deletion failures

`examples/volume/cos/binary/cube-volume-cos.sh:161-166`

`coscmd delete ... || true` treats permission errors, network failures, and partial deletion failures as success. Master then removes the database record and the user receives 204 even though objects may remain in COS, with no way to retry cleanup through the Volume API.

Could we ignore only an explicit NotFound and propagate other errors? The RPC example already preserves `DeleteMulti` errors, so keeping the binary and RPC implementations consistent would be helpful. A small coscmd mock test for deletion failure would cover this path.

## Additional fixes that would be good to include

### 9. Create allows underscores, while GET/DELETE path validation rejects them

`CubeAPI/src/cubemaster/mod.rs:612-628`

There is a small validation mismatch here. A Volume such as `my_data` can be created because the create validators allow `_`, but GET and DELETE use a path validator that allows only alphanumerics and `-`. The follow-up request then fails, and the current mapping appears to turn this into a 500.

Could these paths share the same Volume ID validator? A `my_data` create/get/delete regression test would cover the behavior and help align the documented 400 responses.

### 10. Duplicate mount names can leave asymmetric mount state

`CubeAPI/src/services/sandboxes.rs:201-245`

`volumeMounts` currently does not require unique names. Two entries with the same name and different paths are passed through to Cubelet, which performs Attach and bind operations twice. Backend info is then stored in a map keyed by name, so only one entry remains. Destroy may consequently unmount/Detach only once and leave another mount behind.

Could CubeAPI and Master reject duplicate names early, with Cubelet also validating before it creates side effects? It would also be useful to document that mount names must be unique within one sandbox and add a negative test.

### 11. The `host_path` base-directory check does not cover symlinks

`Cubelet/storage/pluginvolume.go:85-101`

This may not be a directly exploitable issue under the current trust model, but the documentation presents the base-directory check as a security boundary. `filepath.Rel/Clean` is lexical and does not resolve a symlink inside the path, so a plugin or another process that can write under the base directory could still make the final bind mount escape it.

If this is intended to defend against an untrusted plugin, it likely needs per-component symlink rejection or an `openat2`-style `RESOLVE_BENEATH/RESOLVE_NO_SYMLINKS` mechanism. If plugins are fully trusted, narrowing the documentation to say that this protects against misconfiguration, rather than a malicious plugin, would make the boundary clearer.

## Tests and documentation

The production path added by this PR is substantial, while the refcount store, manager, binary/RPC drivers, HTTP CRUD, and the important concurrency and rollback paths currently have little or no focused automated coverage. Before merge, I would suggest covering at least:

- multi-node acquire/release, lost/replayed/out-of-order events, restart, and reconciliation;
- DELETE racing with Attach, concurrent Create with the same name, and plugin success followed by a database failure;
- mkdir/bind/remount failures after Attach, plus a first Detach failure followed by a successful retry;
- duplicate mount names, Volume IDs containing underscores, unknown drivers, and missing Volumes;
- invalid/oversized binary output, timeouts, RPC disconnect/authentication, and partial COS deletion failures; and
- MySQL and PostgreSQL migrations against real databases.

The user-facing documentation should also clarify that refcount drift can currently be permanent rather than brief, the security assumptions for TCP RPC, whether the Volume namespace is global or team-scoped, the duplicate mount-name constraint, Create/Delete 409 responses, and recovery after plugin Destroy succeeds but the database delete fails.

One non-blocking architecture note: `plugin-volume-mounts` and `plugin-volume-sources` are parsed repeatedly from annotations across several layers, while Master uses a placeholder EmptyDir even though the proto now has `PluginVolumeSource`. Moving this path to the typed proto fields later would reduce duplicated parsing and silent schema drift.

Given the data-lifecycle and isolation concerns above, I am requesting changes for this revision. Once the blocking cases have either been fixed or constrained explicitly by the product contract, and the corresponding tests are in place, I would be happy to take another focused pass.

ls-ggg:
> Thanks for putting together the Volume CRUD API, the Master/Cubelet plugin path, the COS examples, and the English and Chinese documentation. This is a broad change and the overall structure is already fairly complete. I walked through the create, attach, detach, and delete flows across `CubeAPI -> CubeMaster -> Cubelet -> plugin`, with particular attention to concurrency and failure recovery.
> 
> There are a few lifecycle and reference-counting issues that I think we should address before merging. The main concern is that the Master-side refcount can remain incorrect after a lost or replayed response, and in some interleavings a Volume that is still mounted, or about to be mounted, can be deleted. The Attach/Detach rollback paths, team isolation, and the security boundary for RPC over TCP also need a little more work. Some of these require a failure or a concurrent request to trigger, but the resulting behavior can involve data deletion or a state that cannot recover automatically.
> 
> If I have misunderstood an intended deployment or protocol constraint, please let me know. In particular, if Volume is intentionally single-tenant or TCP is only intended for a trusted local network, it would help to make that explicit in both the API contract and the documentation.
> 
> ## Changes requested before merge
> ### 1. Refcount events can drift permanently after loss or replay
> `CubeMaster/pkg/volume/refcount/refcount.go:40-99`
> 
> Could we revisit the reliability model here? Cubelet reports a delta only when its local count changes from `0 -> 1` or `1 -> 0`, and Master directly applies that delta to a single aggregate count. The event has no node ID, event ID, or version, and there is no Master-side reconciliation.
> 
> If a successful response is lost, a retry will not report `0 -> 1` again because Cubelet already holds the local reference. Master may then continue to see zero and allow deletion of a Volume that is still in use. A replayed response has the opposite problem and can leave the Volume permanently undeletable.
> 
> Would it be possible to report the current state per `(volume_id, node_id)` and use an idempotent UPSERT/DELETE, with periodic reconciliation as a fallback? A delta-only protocol seems difficult to recover after retries, restarts, or network failures.
> 
> ### 2. DELETE can race with an Attach that is already in progress
> `CubeMaster/pkg/service/httpservice/cube/volume.go:258-291`
> 
> There is a concurrency window here that looks worth closing before merge. Sandbox creation resolves the Volume first, but Master does not increment the refcount until Cubelet has completed Attach and returned successfully. During that interval, DELETE can still observe `refcount=0`, call the plugin's Destroy operation, and remove the database row.
> 
> If those operations interleave, Cubelet may be attaching a backend that has just been deleted. Its later `+1` also cannot be persisted because the Volume row is gone. For a backend such as COS, this could surface as data being deleted while a sandbox is being created.
> 
> Could we introduce an `active/deleting` state and a durable attachment reservation? Sandbox creation could reserve the reference before scheduling, while DELETE atomically transitions the Volume to deleting, prevents new references, and destroys it only after all reservations are gone.
> 
> ### 3. A failed concurrent Create can destroy the successful request's backend
> `CubeMaster/pkg/service/httpservice/cube/volume.go:193-219`
> 
> The rollback here can remove another request's newly created resource. Two requests for the same name can both pass the initial lookup and call Create with the same `volumeID=name`. One database insert succeeds; the other hits the unique constraint and unconditionally calls `Destroy(volumeID)`.
> 
> Because both operations use the same ID, the losing request cannot tell whether the backend it is destroying belongs to the successful request. Could we reserve a `creating` row in the database first, so only the winning request invokes the plugin? An operation ID or another ownership mechanism would also be needed before compensating with Destroy.
> 
> ### 4. A failed Detach can leave the Master refcount stuck after retries
> `Cubelet/plugins/volume/manager.go:160-194`
> 
> The current ordering appears to make this state unrecoverable from a normal retry. Cubelet releases its local reference before invoking the plugin's Detach. If Detach fails, the `1 -> 0` event is included in the failed response, but Master checks the return code first and does not consume it.
> 
> A later Detach retry can complete the actual unmount, but the local reference is already `NotHeld`, so Cubelet will not emit another `1 -> 0`. Master may therefore keep the old count indefinitely, and the user continues to receive 409 when deleting the Volume.
> 
> Could the reference release be committed only after Detach succeeds, or could pending-detach be persisted as a retryable state? Making the state update idempotent would also avoid relying on a one-shot delta in a failed response.
> 
> ### 5. Failures after Attach do not run the full rollback path
> `Cubelet/storage/pluginvolume.go:202-261`
> 
> After `Manager.Attach` succeeds, a failure in mkdir, bind mount, or the readonly remount returns immediately. `PluginVolumeBackendInfos` is only recorded after all those operations succeed, so the outer cleanup cannot see this Volume and will not Detach it.
> 
> This can leave a local refcount and plugin/FUSE resources behind; a readonly remount failure may also leave an existing bind mount. Could we register a rollback handle immediately after Attach, or use a deferred rollback that unmounts and Detaches until the operation is committed? Fault-injection tests at each of these points would be useful to confirm that mounts, refcounts, and metadata are restored.
> 
> ### 6. The team-scoped API currently appears to use global data
> `CubeAPI/src/handlers/volumes.rs:22-24`
> 
> The API description refers to team Volumes, but I could not find a path that carries the authenticated team/principal from the auth callback to Master. The middleware appears to retain only allow/deny, the Volume table has no owner/team field, and list/get/delete/name uniqueness are all global.
> 
> If one deployment accepts credentials for multiple teams, any team allowed through this route could list global Volumes, retrieve another team's token, or mount/delete another team's Volume. Could we confirm the intended boundary? For multi-tenant use, a trusted team ID needs to flow through CubeAPI, Master, the model, and every query. If the product only supports a single-tenant instance, the API text and deployment docs should say that the namespace is global instead.
> 
> ### 8. The COS binary example reports success after remote deletion failures
> `examples/volume/cos/binary/cube-volume-cos.sh:161-166`
> 
> `coscmd delete ... || true` treats permission errors, network failures, and partial deletion failures as success. Master then removes the database record and the user receives 204 even though objects may remain in COS, with no way to retry cleanup through the Volume API.
> 
> Could we ignore only an explicit NotFound and propagate other errors? The RPC example already preserves `DeleteMulti` errors, so keeping the binary and RPC implementations consistent would be helpful. A small coscmd mock test for deletion failure would cover this path.
> 
> ## Additional fixes that would be good to include
> ### 9. Create allows underscores, while GET/DELETE path validation rejects them
> `CubeAPI/src/cubemaster/mod.rs:612-628`
> 
> There is a small validation mismatch here. A Volume such as `my_data` can be created because the create validators allow `_`, but GET and DELETE use a path validator that allows only alphanumerics and `-`. The follow-up request then fails, and the current mapping appears to turn this into a 500.
> 
> Could these paths share the same Volume ID validator? A `my_data` create/get/delete regression test would cover the behavior and help align the documented 400 responses.
> 
> ### 10. Duplicate mount names can leave asymmetric mount state
> `CubeAPI/src/services/sandboxes.rs:201-245`
> 
> `volumeMounts` currently does not require unique names. Two entries with the same name and different paths are passed through to Cubelet, which performs Attach and bind operations twice. Backend info is then stored in a map keyed by name, so only one entry remains. Destroy may consequently unmount/Detach only once and leave another mount behind.
> 
> Could CubeAPI and Master reject duplicate names early, with Cubelet also validating before it creates side effects? It would also be useful to document that mount names must be unique within one sandbox and add a negative test.
> 
> ### 11. The `host_path` base-directory check does not cover symlinks
> `Cubelet/storage/pluginvolume.go:85-101`
> 
> This may not be a directly exploitable issue under the current trust model, but the documentation presents the base-directory check as a security boundary. `filepath.Rel/Clean` is lexical and does not resolve a symlink inside the path, so a plugin or another process that can write under the base directory could still make the final bind mount escape it.
> 
> If this is intended to defend against an untrusted plugin, it likely needs per-component symlink rejection or an `openat2`-style `RESOLVE_BENEATH/RESOLVE_NO_SYMLINKS` mechanism. If plugins are fully trusted, narrowing the documentation to say that this protects against misconfiguration, rather than a malicious plugin, would make the boundary clearer.
> 
> ## Tests and documentation
> The production path added by this PR is substantial, while the refcount store, manager, binary/RPC drivers, HTTP CRUD, and the important concurrency and rollback paths currently have little or no focused automated coverage. Before merge, I would suggest covering at least:
> 
> * multi-node acquire/release, lost/replayed/out-of-order events, restart, and reconciliation;
> * DELETE racing with Attach, concurrent Create with the same name, and plugin success followed by a database failure;
> * mkdir/bind/remount failures after Attach, plus a first Detach failure followed by a successful retry;
> * duplicate mount names, Volume IDs containing underscores, unknown drivers, and missing Volumes;
> * invalid/oversized binary output, timeouts, RPC disconnect/authentication, and partial COS deletion failures; and
> * MySQL and PostgreSQL migrations against real databases.
> 
> The user-facing documentation should also clarify that refcount drift can currently be permanent rather than brief, the security assumptions for TCP RPC, whether the Volume namespace is global or team-scoped, the duplicate mount-name constraint, Create/Delete 409 responses, and recovery after plugin Destroy succeeds but the database delete fails.
> 
> One non-blocking architecture note: `plugin-volume-mounts` and `plugin-volume-sources` are parsed repeatedly from annotations across several layers, while Master uses a placeholder EmptyDir even though the proto now has `PluginVolumeSource`. Moving this path to the typed proto fields later would reduce duplicated parsing and silent schema drift.
> 
> Given the data-lifecycle and isolation concerns above, I am requesting changes for this revision. Once the blocking cases have either been fixed or constrained explicitly by the product contract, and the corresponding tests are in place, I would be happy to take another focused pass.

Thanks for the thorough review — the create/attach/detach/delete walkthrough and the concurrency notes were very helpful. Here is how we are handling each point:

1 & 2 — Refcount delta reliability / DELETE racing with Attach
We agree these are real edge cases under the current design. They are known limitations of this first version, and we are not changing the refcount/reservation model in this PR. A more complete volume management scheme (including stronger lifecycle / reconciliation) is planned for follow-up work.

3 — Concurrent Create with the same name destroying the winner’s backend
Will fix: reserve the DB row first so only the winning request invokes the plugin; losers hit the unique constraint and must not call Destroy.

4 — Detach failure leaving Master refcount stuck
Will fix: if plugin Detach fails, Cubelet rolls back the local refcount and does not report a 1→0 event to CubeMaster.

5 — Failures after Attach skipping Detach rollback
Will fix: on sandbox create failure, rollback/destroy goes through the normal Detach path so attachments are torn down the same way as a successful lifecycle.

6 — Team-scoped API language vs global data
Will fix docs/comments only: remove the “team” wording. The Volume namespace is global for this deployment model; we are not adding multi-tenant ownership in this PR.

8 — COS binary example treating delete failures as success
Will fix: stop swallowing non-NotFound delete errors so Master does not remove the DB row when COS cleanup failed.

9 — Create allows _ but GET/DELETE path validation rejects it
Will fix: align volume ID path validation with the create name rules.

10 — Duplicate mount names
Will fix: reject duplicate volumeMounts[].name so each volume can be mounted at most once per sandbox.

11 — host_path base-dir check and symlinks
No code change. Volume plugins are trusted system components; external users cannot supply or control them, so we do not treat symlink rejection as a required security boundary here.

Tests / documentation coverage
I also ran additional local test scenarios for this change (including the volume lifecycle suite and a complex/concurrent script checked into  examples/volume/cos/verify_volume.py and examples/volume/cos/volume_complex_concurrent_test.py), and they all passed on our test host. Broader automated coverage in the shared test framework is still blocked on that framework becoming available; once it is, I will add the corresponding cases there. Those framework tests are intentionally not part of this PR.

fslongjin (approved):

PR #998, opened 2026-07-17
feat: add Redis Sentinel support for one-click and Helm

xiaojunxiang2023:
## Summary

- Add optional Redis Sentinel mode across CubeMaster, cube-lifecycle-manager, CubeProxy, one-click install, and the Kubernetes Helm chart.
- When `master_name` / `redis.masterName` / `CUBE_EXTERNAL_REDIS_MASTER_NAME` is set, components resolve the current Redis master via `SENTINEL get-master-addr-by-name` instead of dialing a fixed host. Standalone Redis remains the default.
- Empty `sentinel_password` / `redis.sentinelPassword` means **no AUTH against Sentinel** (do not fall back to the Redis master password).
- One-click `install.sh` accepts external Sentinel-only Redis and skips the bundled local Redis container.
- Helm: `redis.masterName` + `redis.sentinelNodes` skips the chart-managed Redis StatefulSet and wires CubeMaster conf plus CubeProxy / CLM env vars.

### CubeMaster / CubeProxy / cube-lifecycle-manager
- Extend Redis config with `master_name`, `sentinel_nodes`, and optional `sentinel_password`.
- `wrapredis` resolves the master on each pool dial so failover is transparent.
- CubeProxy Lua (`redis_iresty.lua`): Sentinel lookup with short TTL cache; empty `redis_sentinel_pd` skips Sentinel AUTH.
- cube-lifecycle-manager: `go-redis` `FailoverClient` via `redis.UniversalClient`.

### one-click deploy
- env templates, install preflight, quickcheck, and up-scripts for external Sentinel.

### Kubernetes Helm chart
- `values.yaml`: `redis.masterName` / `sentinelNodes` / `sentinelPassword`.
- Helpers skip builtin Redis under Sentinel; validate requires `sentinelNodes` when `masterName` is set.
- CubeMaster conf + CubeProxy entrypoint + Proxy/CLM Deployment env wiring.
- Guard script: `deploy/kubernetes/chart/scripts/test-redis-sentinel-guard.sh`.

## Test plan

- [x] `cd CubeMaster && go test -count=1 -run 'TestParse|TestResolve|TestRedisDisplay' ./pkg/base/wrapredis/... -v` — passed
- [x] `cd cube-lifecycle-manager && go test -count=1 ./... -v` — passed
- [x] `cd CubeMaster && go build ./cmd/cubemaster/` — passed
- [x] `cd cube-lifecycle-manager && go build ./cmd/cube-lifecycle-manager/` — passed
- [x] `bash deploy/one-click/tests/test_runtime_file_safety.sh` — passed
- [x] `sh deploy/kubernetes/chart/scripts/test-redis-sentinel-guard.sh` — passed
- [x] `helm lint deploy/kubernetes/chart` (with test passwords) — passed
- [x] Docker (`cube-compiler:ubuntu22`): Go build/tests above — passed


xiaojunxiang2023:
Thanks for the detailed review. The latest update addresses the actionable
runtime issues in e2f0a8a:

- Added a shared-dict cache for the Sentinel-resolved master address with a
  15-second TTL. A failed connection invalidates the cached address and forces
  an immediate Sentinel lookup.
- Added cleanup of opposite-mode environment keys when switching between
  Sentinel and standalone Redis, with a regression test for key removal.
- Added DialWriteTimeout to Sentinel connections.
- Documented the intentional fail-fast behavior after all configured Sentinel
  nodes have already been exhausted.
- Added the missing RedisSentinelPassword contract documentation.

Bracketed IPv6 Sentinel endpoints are not included in the current deployment
scope, which currently uses IPv4 addresses. We will handle IPv6 consistently
across the Lua and shell parsers in a dedicated follow-up.

RedisMasterName intentionally acts as the explicit Sentinel-mode selector. A
standalone RedisAddr may still be populated by defaults, so rejecting both
values would incorrectly reject valid Sentinel configurations.

The Lua timeout applies independently to each socket operation rather than
being a shared lookup budget, so no separate timeout setting is added in this
PR. The additional unit-test coverage suggestions are useful and can be
handled separately without expanding this deployment-focused change further.

xiaojunxiang2023:
Thanks for the updated review.

Addressed in 818a493:
- Lua now connects to the Sentinel-resolved master first and only caches the
  address after a successful connect.
- The equal Sentinel/master password message is informational rather than a
  WARNING, since a shared password is a valid configuration.

IPv6 Sentinel endpoints remain out of scope for this PR (current one-click
deployments use IPv4); we will handle bracketed IPv6 in Lua and install.sh
in a follow-up.

A Go-side Sentinel address cache is deferred: Dial() mainly runs when the
connection pool creates a new connection, not on every Redis command.

RedisMasterName remains the explicit Sentinel-mode selector. A standalone
RedisAddr may still be present from defaults, so rejecting both set values
would break valid Sentinel configs.

Unit-test coverage for redisclient remains deferred to a separate change.

force-pushed in 818a493

chenhengqi (commented):
Do you use Redis Sentinel in production? At Tencent Cloud, we use Redis Cluster with a single VIP. The HA thing is behind the scene.

xiaojunxiang2023:
> Do you use Redis Sentinel in production? At Tencent Cloud, we use Redis Cluster with a single VIP. The HA thing is behind the scene.

Currently it’s only been deployed and verified in the test environment. Our team’s Redis instances all use Sentinel mode, so I create this PR.

CubeProxy/lua/init_worker_phase.lua
```diff
@@ -19,8 +19,11 @@ proxy_registry.setup({
     node_ip     = os.getenv("CUBE_PROXY_NODE_IP"),
     version     = os.getenv("CUBE_PROXY_VERSION"),
     interval_ms = tonumber(os.getenv("CUBE_PROXY_HEARTBEAT_INTERVAL_MS") or "") or 5000,
-    redis_ip    = os.getenv("CUBE_PROXY_REGISTRY_REDIS_HOST"),
+    redis_ip = os.getenv("CUBE_PROXY_REGISTRY_REDIS_HOST"),
     redis_port  = tonumber(os.getenv("CUBE_PROXY_REGISTRY_REDIS_PORT") or "") or 6379,
     redis_pd    = os.getenv("CUBE_PROXY_REGISTRY_REDIS_PASSWORD") or "",
     redis_index = tonumber(os.getenv("CUBE_PROXY_REGISTRY_REDIS_DB") or "") or 0,
+    redis_master_name = os.getenv("CUBE_PROXY_REGISTRY_REDIS_MASTER_NAME"),
```
chenhengqi, line 26: See comments below:

https://github.com/TencentCloud/CubeSandbox/blob/65b3ae15fb174b0f30a83e41498d80b2a052f852/CubeProxy/nginx.conf#L6-L7

xiaojunxiang2023: Good catch — nginx clears worker env unless explicitly whitelisted.

Added the three Sentinel registry env declarations to CubeProxy/nginx.conf:

- CUBE_PROXY_REGISTRY_REDIS_MASTER_NAME
- CUBE_PROXY_REGISTRY_REDIS_SENTINEL_NODES
- CUBE_PROXY_REGISTRY_REDIS_SENTINEL_PASSWORD

force-pushed in ecc585a

xiaojunxiang2023:
@chenhengqi @fslongjin 
Sentinel here is an optional path for external Redis, not a change to the
Tencent Cloud default (Cluster/VIP).

Self-hosted users often run Sentinel without a VIP. Supporting that opt-in
covers more of the community; VIP/standalone stays as-is when master_name
is unset.

OK to land as optional, or keep external HA as “VIP only”?

chenhengqi:
@xiaojunxiang2023 Hi, thanks for your contribution. The merge window for v0.6.0 will be closed today. We will queue this for v0.7.0.

xiaojunxiang2023:
> @xiaojunxiang2023 Hi, thanks for your contribution. The merge window for v0.6.0 will be closed today. We will queue this for v0.7.0.

@chenhengqi Thanks for the update!

Understood on the v0.6.0 freeze. This PR has been rebased onto current master, CI is green, and the latest Auto Review comments are addressed (or replied). Could you help queue / merge it for v0.7.0 when you have a moment? Happy to follow up on any remaining review feedback quickly.

CubeProxy/lua/redis_iresty.lua
```diff
@@ -51,6 +159,31 @@ end
...
+        if not ip then
+            return nil, err
+        end
+        -- Cache only after connect succeeds so a transient failure does not
+        -- leave a stale address that forces an extra failed connect cycle.
+        local ok, cerr = redis:connect(ip, port)
+        if not ok then
+            return nil, cerr
+        end
+        set_cached_master(self, ip, port)
```
fslongjin, line 231: **Blocking:** this line caches the resolved master address, but the cache is only cleared when the TCP connection itself fails. After Sentinel promotes a new master, the old master can remain reachable as a replica: connect, AUTH, and SELECT may all succeed, while writes return `READONLY`. The command path then returns the error without invalidating this cache, so writes can keep failing until the 15-second TTL expires and the Proxy does not re-query Sentinel. Please invalidate the cached master on `READONLY` and relevant connection errors, then re-resolve through Sentinel. Any retry should be limited to one attempt and only applied to commands that are safe to retry. Please add coverage for a reachable old master that rejects writes with `READONLY`, including the pipeline path.

deploy/kubernetes/chart/templates/validate.yaml
```diff
@@ -100,18 +100,21 @@
 {{- fail (printf "%s.password still equals the CHANGE_ME_* default sentinel. Override it when connecting to an external database." (include "cube.dbDriver" .)) }}
 {{- end }}
 {{- $controlPlaneNeedsRedis := and .Values.controlPlane.enabled .Values.controlPlane.master.enabled -}}
-{{- if and $controlPlaneNeedsRedis (not (eq (include "cube.redisBuiltinEnabled" .) "true")) (not .Values.redis.host) }}
-{{- fail "control plane master requires redis.host when redis.enabled=false" }}
+{{- if and (eq (include "cube.redisSentinelEnabled" .) "true") (eq (((.Values.redis).sentinelNodes) | default "") "") }}
```
fslongjin: **Blocking:** this checks only whether the raw string is empty. Values such as spaces or comma-only entries pass Helm validation, but the runtime parsers trim and discard the entries, leaving zero Sentinel endpoints. The chart then renders successfully and multiple components fail only at startup. Please split the value, trim and discard empty items, and require at least one valid endpoint before rendering. Please also add whitespace-only and consecutive-comma cases to `test-redis-sentinel-guard.sh`.

deploy/kubernetes/chart/values.yaml
```diff
@@ -470,8 +470,17 @@ redis:
     pullPolicy: IfNotPresent
   # Empty host installs and uses the chart-managed cube-redis service.
   # Non-empty host uses the configured third-party Redis service and skips cube-redis.
+  # When masterName is set (Sentinel mode), host is ignored and cube-redis is skipped.
   host: ""
   port: 6379
+  # Redis Sentinel (external). Non-empty masterName enables Sentinel for
```
fslongjin, line 476: **Suggestion:** `masterName` currently serves both as the Sentinel master identifier and as an implicit Redis connection-mode switch. Once it is set, `redis.host` and `redis.port` are silently ignored and the chart-managed Redis is skipped. This makes it difficult to tell from values which fields actually take effect, especially when upgrading an existing standalone configuration. Please consider an explicit `redis.mode: builtin|standalone|sentinel` with mode-specific fields and validation. If changing the public values structure is out of scope here, please at least reject mixed `masterName` and non-empty `host` values, and validate that `masterName` and `sentinelNodes` are either both set or both unset.

for example:

```yaml
redis:
  mode: builtin # builtin | standalone | sentinel

  standalone:
    host: "redis.example.internal"
    port: 6379
    password: "..."
   
  sentinel:
    masterName: "mymaster"
    nodes:
      - "10.0.0.11:26379"
      - "10.0.0.12:26379"
    password: ""
```

CubeProxy/lua/redis_iresty.lua
```diff
@@ -32,6 +32,114 @@ local mt = {
...
+        part = trim(part)
+        if part ~= "" then
+            table.insert(out, part)
+        end
+    end
+    return out
+end
+
+local function split_host_port(addr, default_port)
+    local host, port = addr:match("^([^:]+):(%d+)$")
```
fslongjin: **Blocking if IPv6 is supported:** this pattern only parses hostname or IPv4 `host:port` endpoints. A valid Sentinel endpoint such as `[2001:db8::10]:26379` falls through as a host value containing both brackets and port text, so CubeProxy cannot connect even though CubeMaster accepts bracketed IPv6 endpoints. Please add a `[ipv6]:port` parsing branch that passes the unbracketed IPv6 address and port to OpenResty, and make the one-click parser follow the same input contract. Add IPv4, hostname, and bracketed-IPv6 regression cases. If IPv6 is intentionally unsupported, reject it during deployment validation rather than failing at runtime.

deploy/kubernetes/chart/templates/_helpers.tpl
```diff
@@ -632,8 +632,18 @@ CUBE_SANDBOX_MYSQL_* alone is always assembled as mysql://.
 {{- if .Values.redis.host -}}{{ .Values.redis.host }}{{- else -}}{{ include "cube.redisName" . }}.{{ .Release.Namespace }}.svc.{{ include "cube.clusterDomain" . }}{{- end -}}
 {{- end -}}
 
+{{- /* Non-empty redis.masterName selects external Redis Sentinel (skips chart redis). */ -}}
```
fslongjin, line 635: Please add the reverse validation as well: setting `sentinelNodes` without `masterName` currently leaves Sentinel disabled and silently ignores the provided nodes, falling back to the fixed-address or built-in Redis path. This is a common partial configuration error and should fail during Helm rendering. Requiring these two values to be set together will make the mode-selection contract much clearer.

fslongjin (changes requested):
Thanks for adding Sentinel support across the runtime components and deployment paths. I am requesting changes for the failover recovery and configuration validation issues below.

xiaojunxiang2023:
@fslongjin Thanks for the detailed review:

1. Failover / READONLY: CubeProxy now invalidates the cached Sentinel master on READONLY and connection-class errors, then re-resolves via Sentinel with at most one retry (single commands and pipelines). Added `tests/redis_iresty_sentinel_test.lua` covering a reachable demoted master that rejects writes with READONLY, including the pipeline path.

2. Helm sentinelNodes validation: endpoints are split/trimmed; whitespace-only and comma-only values fail render. Also require masterName ↔ sentinelNodes together, and reject combining masterName with redis.host.

3. Bracketed IPv6: Lua `split_host_port` and one-click preflight now parse `[ipv6]:port` / `[ipv6]` (default 26379), aligned with the Go dial path. Documented in values.yaml / env.example.

On `redis.mode: builtin|standalone|sentinel`: I’d like to keep the current values shape for this PR (masterName as the Sentinel opt-in) to avoid a wider config migration. The new mixed-host rejection + paired masterName/sentinelNodes checks should make the contract clearer; a full mode redesign can be a follow-up if you prefer.

xiaojunxiang2023:
Hi @fslongjin  — gentle ping for a re-review when you have a moment.

You previously requested changes on this PR; those items should all be addressed on the current HEAD (`2d5d32d`). CI is green and the branch is mergeable again.

One heads-up: this PR has picked up merge conflicts with `master` a few times while waiting (we’ve rebased/resolved them repeatedly). To avoid more churn / conflict cycles, would you mind taking another look soon so I can get it approved and merged if it looks good?

Thanks a lot!

fslongjin (approved):
LGTM

PR #1000, opened 2026-07-17
fix(cubemaster): make PostgreSQL runtime paths work after dao migration

xiaojunxiang2023:
## Summary

PR #674 introduced the PostgreSQL driver and schema migrations, but three runtime paths still assumed MySQL and broke real PostgreSQL deployments:

1. **`db.Init` opened a separate MySQL pool** — `cmd/cubemaster` calls `dao.Open(postgres)` first, yet legacy callers (`templatecenter`, `nodemeta`, `instancecache`, `localcache`) still invoke `db.Init`, which ignored the global dao handle and tried to connect via MySQL DSN. This fix returns `dao.Default()` so all packages share the connection opened at startup.

2. **Artifact GC lock was MySQL-only** — `listArtifactGCCandidatesLocked` used `GET_LOCK` / `RELEASE_LOCK`, which fail on PostgreSQL. Added driver-aware `trySessionLock` / `releaseSessionLock` (`pg_try_advisory_lock(hashtext(name))` on PG, existing MySQL behavior preserved).

3. **`UpsertReplica` update path failed on PostgreSQL** — reusing the same `*gorm.DB` chain after `First` caused GORM to emit `UPDATE ... FROM t_cube_template_replica`, which PostgreSQL rejects with `SQLSTATE 42712`. The update now runs on a fresh chain with an explicit `Where` clause.

## Test plan

- [x] `go test ./pkg/base/db/ -run TestInitReturnsDaoDefault -count=1 -v` — passed (Docker PostgreSQL 16 via dockertest)
- [x] `go test ./pkg/templatecenter/ -run PostgreSQL -count=1 -v` — passed
  - `TestTrySessionLockPostgreSQL` — advisory lock acquire/release
  - `TestUpsertReplicaPostgreSQLUpdatePath` — insert then update without SQLSTATE 42712
- [x] Docker (`cube-compiler:ubuntu22` on ubuntu22, `/var/run/docker.sock` mounted): same commands above — all passed

### DIFF
--- a/CubeMaster/pkg/base/db/db.go
+++ b/CubeMaster/pkg/base/db/db.go
@@ -6,94 +6,17 @@
 package db
 
 import (
-	"context"
-	"database/sql"
-	"fmt"
-	"time"
-
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/config"
-	"github.com/tencentcloud/CubeSandbox/cubelog"
-	"gorm.io/driver/mysql"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/dao"
 	"gorm.io/gorm"
-	"gorm.io/gorm/logger"
 )
 
-func generateURL(user, password, addr, dbname string, connTimeout, readTimeout, writeTimeout int) string {
-	return fmt.Sprintf(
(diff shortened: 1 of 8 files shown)


fslongjin:
Please run `make fmt` to format the code~

xiaojunxiang2023:
Thanks for the review. Addressed in f87b847: transaction-pinned session lock,
unlock failure logging restored, panic-recovery test restored, and gofmt applied.

fslongjin:
LGTM

PR #1001, opened 2026-07-17
feat(kubernetes): add top-level persistence.storageClassName convenience key

### LINKED ISSUE #954: [Feature Request] Cube on K8s: deploy, upgrade, and canary on any K8s cluster
# [Feature Request] Cube on K8s: deploy, upgrade, and canary on any K8s cluster

---

This feature is co-developed and driven by the Cube team and the Tencent Cloud TKE team.

--- 

## Problem / Motivation

Today, Cube is delivered as a "one-click install package" — both the control plane and the compute plane run as systemd processes on the host. The issues with that are:

- Deploying, scaling, and upgrading all require SSHing into the machine to run scripts; there is no declarative K8s-style way to do it.
- There is no built-in canary release capability — upgrades are either skipped or done in one shot, which carries large risk.
- Any cloud-native integration (Prometheus, Service Mesh, etc.) has to be wrapped by hand.

Running Cube on K8s is the foundation for large-scale delivery on the cloud going forward.

## Proposed Solution

A K8s-native deployment form, in four parts:

**1. One-shot `helm install` deployment**

A single chart brings up the control plane and the compute plane (cube-node DaemonSet) together. The target cluster type is unrestricted — TKE managed / TKE Serverless / self-built / other clouds all work; CNI coverage needs to include VPC-CNI, Cilium, Flannel, and Calico-VXLAN.

**2. Nodes auto-join via label / taint**

After labeling and tainting the target nodes, there is no need to log in to the node. The chart automatically completes PVM host kernel initialization (optional) and deploys the Cube compute plane components (cubelet / network-agent / cube-egress sidecar). New nodes added later follow the same flow.

**3. Rolling upgrades that do not disrupt running sandboxes**

Both the control plane and the compute plane can be rolling-upgraded. During the upgrade, existing sandboxes are not restarted or rebuilt, their process pids stay the same, and the network stays reachable. After the upgrade, cubelet / network-agent reconnect to the existing sandboxes through shim reload / recover. The plan is to later integrate with CubeOps for canary release by node / by cluster / by version.

**4. Per-component canary upgrade for compute nodes**

A single Big Pod on a compute node runs multiple components at the same time (cubelet, network-agent, cube-egress sidecar, etc.). The upgrade should not rebuild the whole Pod at once — it should be able to do canary release at the component level, e.g., upgrade cubelet first, verify stability, then upgrade network-agent, and finally upgrade cube-egress. A problem with one component does not affect the others running in the same Pod, and this also avoids stacking the risk of multiple components within a single upgrade window.

## Alternatives Considered

- **Keep maintaining the one-click package**: works for existing customers, but production users have to SSH in for every scale-out, upgrade, or troubleshooting step — high ops cost.
- **Build a chart only for TKE**: saves effort, but does not match the scenarios of self-built K8s or other clouds — coverage is too narrow.
- **Pure Docker Compose, no K8s**: scheduling, rolling upgrade, observability, and canary all have to be re-implemented — equivalent to redoing what K8s already does.

Our intent is to reuse K8s's existing operational capabilities (scheduling, rolling upgrade, RBAC, observability, canary) so production users do not have to reinvent the wheel, and to bring ops cost down.

## Additional Context

This approach is a variant of the **Big Pod** approach discussed in [Discussion #636](https://github.com/TencentCloud/CubeSandbox/discussions/636): pack cubelet / network-agent / cube-egress sidecar and related components into a single DaemonSet Big Pod, and let K8s uniformly manage deployment and upgrades. On top of that, this proposal additionally decouples the data plane to handle overlay CNI scenarios, and adds non-disruptive rolling upgrade and per-component canary capabilities (will be updated in the PR later).

Current progress:

- ✅ Deployment and upgrade on TKE + VPC-CNI is already working — see [PR #779](https://github.com/TencentCloud/CubeSandbox/pull/779) (v0.5.0 chart); deployment / upgrade docs are at `deploy/kubernetes/chart/docs/`.
- 🚧 The next step is overlay CNI compatibility (Cilium / Flannel / Calico-VXLAN); this will also be reflected in PR #779.
- 📋 After that, integrate with CubeOps for canary upgrade — CubeOps is about to go open source.

fslongjin:
## Summary

Introduce a top-level `persistence.storageClassName` convenience key for the Helm chart, allowing users to pin all three control-plane PVCs (CubeMaster / MySQL / Redis) to a single StorageClass with one setting.

## Changes

- **New helper**: `cube.persistenceStorageClassName` template with clear three-tier precedence (component-level → top-level `persistence.storageClassName` → omit for cluster default)
- **New example file**: `runtime-values.example.yaml` as a documented starting point with inline guidance
- **PVC templates**: `master-pvc`, `mysql`, `redis` now use the new helper instead of reading component-level `storageClassName` directly
- **Documentation**: Updated README, ARCHITECTURE, FAQ, and QUICKSTART to reflect the new persistence model, clarify StorageClass creation vs PVC binding, and improve troubleshooting guidance
- **NOTES.txt**: Shows persistence mode on `helm install`
- **values.yaml**: Updated comments documenting the new key and precedence

## Use case

Before, users had to set `storageClassName` on each component individually:

```yaml
controlPlane.master.persistence.storageClassName: gp3
mysql.persistence.storageClassName: gp3
redis.persistence.storageClassName: gp3
```

Now, a single key covers all three:

```yaml
persistence.storageClassName: gp3
```

Component-level settings still take precedence when different SCs are needed per component.

## Files changed

| File | Change |
|------|--------|
| `templates/_helpers.tpl` | Add `cube.persistenceStorageClassName` helper |
| `templates/master-pvc.yaml` | Use helper for storageClassName resolution |
| `templates/mysql.yaml` | Use helper for storageClassName resolution |
| `templates/redis.yaml` | Use helper for storageClassName resolution |
| `templates/NOTES.txt` | Add persistence mode info |
| `values.yaml` | Add top-level `persistence.storageClassName`, update comments |
| `runtime-values.example.yaml` | New example values file |
| `README.md` | Update persistence documentation |
| `docs/ARCHITECTURE.md` | Update configuration table |
| `docs/FAQ.md` | Expand PVC troubleshooting, clarify reclaim policy |
| `docs/QUICKSTART.md` | Simplify storage docs, reference new convenience key |

## Testing

- Template rendering: verified `helm template` output with various combinations of top-level and component-level `storageClassName` settings
- Precedence: confirmed component-level non-empty values override top-level key correctly


deploy/kubernetes/chart/docs/ARCHITECTURE.md
```diff
@@ -328,7 +328,9 @@ externalControlPlane:
 | values 路径 | 默认 | 影响 |
 | --- | --- | --- |
 | `global.timezone` | `Asia/Shanghai` | 注入 Chart 管理容器的 `TZ` |
-| `storageClass.create` / `name` / `volumeBindingMode` | 见 values | 状态组件 StorageClass（TKE 多可用区常用 WaitForFirstConsumer） |
```
chenhengqi, line 331: No English docs?

fslongjin: Since the k8s deploy feature is not-completed, I will add all k8s-related docs in a PR later.

PR #1011, opened 2026-07-17
fix(cubebox): allow deleting paused sandboxes via shim binary delete

lf4096:
## Problem

Deleting a **paused** sandbox fails and the sandbox becomes undeletable.

`DELETE /sandboxes/<id>` on a paused sandbox returns HTTP 500:

```
CubeMaster error 130593 -> delete paused sandbox failed
  -> shim ttrpc: Others("sandbox not in normal state")
```

A paused VM cannot serve its runtime (ttrpc) API, so the shim's per-container task-delete handler rejects the request with `sandbox not in normal state`. Retries fail identically. The GC / `CleanUp` path is affected the same way: `stopTask` fails on the guard and the reconcile loops forever on `shim process still Exists`, so a paused sandbox is never reclaimed.

This is issue #219. Two earlier attempts did not land:

- #317 removed the `sb.paused()` guard in the shim's task-delete handler. Review raised state-consistency concerns (sandbox state vs container state, the `delete_exec` path).
- #480 force-deleted from Cubelet with a hand-rolled `syscall.Kill` + `/proc` PID check. In review, @fslongjin requested changes (2026-06-16): Cubelet should not reimplement the shim's teardown; the cube shim already has `delete_shim()` which SIGKILLs the shim and reclaims resources, and containerd already has a standard binary-delete flow for an unresponsive shim. The manual kill also skipped the pause-snapshot and ivshmem cleanup that `delete_shim()` does. That PR was closed without the rework.

## Approach

This PR follows @fslongjin's suggested direction: reuse the shim's own binary delete instead of any custom kill logic in Cubelet.

For a paused sandbox, `destroyPausedSandbox` runs the shim's `delete` subcommand the same way containerd reaps a dead shim — via `pkg/shim.Command` against the bundle from `l.shims.Get(id).Bundle()`:

```
containerd-shim-cube-rs -namespace <ns> -id <id> -bundle <bundle> ... delete
```

The containerd-shim framework dispatches this to the shim's `delete_shim()`, which SIGKILLs the original shim process (reaping the in-process cube-hypervisor VMM) and reclaims the VM workdir, pause snapshot, and ivshmem file through `clean_sandbox_resource`. Cubelet then detaches the shim from the runtime manager with `l.shims.Delete`. Both `Destroy` and `CleanUp` take this path when the sandbox is paused, so GC reclaims paused sandboxes instead of looping.

How this addresses the earlier review:

- **No custom kill logic in Cubelet** — the SIGKILL and all resource cleanup stay owned by the shim's `delete_shim()` / `clean_sandbox_resource`; the earlier `syscall.Kill` + `/proc` PID verification is gone.
- **Pause snapshot and ivshmem are reclaimed** — because `delete_shim()` runs, not just a bare kill.
- **Standard mechanism** — `pkg/shim.Command` with the bundle path is exactly what containerd's own `binary.Delete` builds; the bundle comes from `ShimInstance.Bundle()`, not a hand-reconstructed path.
- The addresses passed to the shim are read from the same `PropertyGRPCAddress` / `PropertyTTRPCAddress` init properties that `ShimManager` itself uses to start shims.

Detaching after the kill is best-effort: the shim and VMM are already reaped by the `delete` subcommand, so a hiccup while removing the bundle dir is logged rather than failing the delete (a leftover bundle is reconciled by GC / on restart), matching the existing `CleanUp` behavior.

## Files

- `Cubelet/services/cubebox/destroy.go` — `destroyPausedSandbox` helper; the paused branch of `Destroy` now force-deletes and cleans up the cubestore container records.
- `Cubelet/services/cubebox/local.go` — capture the containerd grpc/ttrpc addresses at plugin init; `CleanUp` takes the same force-delete path when paused.

## Testing

Verified end-to-end on a two-node cluster (both nodes on this build).

Before: `create -> pause -> DELETE` returns 500 (`sandbox not in normal state`), retries fail identically.

After, `create -> pause -> DELETE`:

| check | result |
|---|---|
| DELETE (Destroy path) | 200/204 (was 500) |
| shim process + in-process VMM | reaped |
| pause snapshot dir (`root/pausevm/<id>`) | reclaimed |
| runtime bundle (`state/io.containerd.runtime.v2.task/.../<id>`) | reclaimed |
| cubecow writable-layer volume | reclaimed |
| residual process | none |
| sandbox detail API | 404 |

GC / `CleanUp` path (exercised via a delete that routes to `engine.CleanUp`):

| check | result |
|---|---|
| result | success |
| shim / VMM / pause snapshot / bundle / volume | all reclaimed |
| log | takes the `CleanUp` force-delete branch, no `shim process still Exists` loop |

A previously stuck paused sandbox (undeletable before) also deletes cleanly once this build is running, with the same full reclamation.

### DIFF
--- a/Cubelet/services/cubebox/destroy.go
+++ b/Cubelet/services/cubebox/destroy.go
@@ -17,6 +17,7 @@ import (
 	containerd "github.com/containerd/containerd/v2/client"
 	"github.com/containerd/containerd/v2/pkg/cio"
 	"github.com/containerd/containerd/v2/pkg/namespaces"
+	shimclient "github.com/containerd/containerd/v2/pkg/shim"
 	"github.com/containerd/errdefs"
 	"github.com/containerd/errdefs/pkg/errgrpc"
 	"github.com/hashicorp/go-multierror"
@@ -117,12 +118,29 @@ func (l *local) Destroy(ctx context.Context, opts *workflow.DestroyContext) (err
 		containers = append(containers, ci)
 	}
 	if sb.GetStatus().IsPaused() {
-
-		ctx = constants.WithSkipRuntimeAPI(ctx)
-		if _, err := l.localTask.Delete(ctx, &tasks.DeleteTaskRequest{
-			ContainerID: sb.ID,
-		}); err != nil {
-			result = multierror.Append(result, fmt.Errorf("delete sandbox by binary failed: %w", err))
+		// A paused VM can't serve its runtime API, so the shim rejects the task
+		// delete. Force it down through the shim's own binary delete subcommand,
+		// which reaps the shim (and its in-process VMM) and reclaims the VM workdir,
+		// pause snapshot, and ivshmem file.
+		if err := l.destroyPausedSandbox(ctx, sb.ID); err != nil {
+			result = multierror.Append(result, fmt.Errorf("delete paused sandbox failed: %w", err))
+		}
+		// The shim reclaims VM-layer state; reclaim the host-side per-container
+		// resources it does not touch (containerd record, rootfs mount, fifos),
+		// which the normal (non-paused) path does via destroyContainer.
+		for _, cntr := range append(containers, sb.FirstContainer()) {
+			if cntr == nil {
+				continue
+			}
+			if er := l.cleanPausedContainerHost(ctx, cntr); er != nil {
+				result = multierror.Append(result, fmt.Errorf("clean paused container [%s]: %w", cntr.ID, er))
+			}
+			if er := l.cubeboxManger.Delete(ctx, &cubes.DeleteOption{
+				CubeboxID:   sb.ID,
+				ContainerID: cntr.ID,
+			}); er != nil {
+				log.G(ctx).Warnf("Delete container [%s] in cubestore fail: %v", cntr.ID, er)
+			}
 		}
 	} else {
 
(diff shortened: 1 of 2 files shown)


lf4096:
Addressed the substantive points; details below.

**Fixed — host-side per-container cleanup on the paused Destroy path** (Code Quality #2 / #7). Confirmed a real leak: the paused branch skipped `destroyContainer`, so the containerd container record (and rootfs mount / task-io fifos) were not reclaimed. The shim's `clean_sandbox_resource` only covers VM-layer state (VM workdir, pause snapshot, ivshmem), not these host-side resources. Added `cleanPausedContainerHost` — it calls `deleteContainer` + `rootfs.CleanRootfs` + `taskio.Clean` per container after the shim is reaped, with no runtime-API call (the shim's task API is already gone). Note the `CleanUp` path was already covered by its existing unconditional `cleanFn` loop, so the fix is scoped to `Destroy`.

Verified with a before/after on a two-node cluster (`create -> pause -> delete`):
- Before the fix: containerd container record remained after delete (`containers ls` still showed it).
- After the fix: record gone, no leftover.

**Fixed — CleanUp TOCTOU comment** (Security #6). Reworded the comment to state the actual consequence: a concurrent resume racing paused->running would have its shim reaped without a guest flush (power-loss semantics for that one sandbox). Kept the lock-free read: `CleanUp` runs only for a sandbox already converging to deleted, its status read is internally `RLock`-synchronized, and the effect is confined to that sandbox and never affects others. Acquiring `sb.Lock()` here would diverge from CleanUp's established lock-free teardown pattern.

**Not changing, with rationale:**

- **Context cancellation for `cmd.Run()`** (Performance, destroy.go:215). `pkg/shim.Command` builds the subprocess with `exec.CommandContext(ctx, ...)`, so it already reacts to cancellation — the caller's `destroyDeadline` (60s) kills the subprocess. The lock being held across a shim operation is the existing `Destroy` shape (the non-paused `destroyContainer` also runs shim/task calls under `sb.Lock()`), not something this PR introduces.
- **`shims.Delete` error only logged** (Code Quality #3, destroy.go:222). Intentional: at that point the shim and its VMM are already reaped by the `delete` subcommand, so a failure to detach / remove the bundle dir must not fail the delete (a leftover bundle is reconciled by GC / on restart). This mirrors CleanUp.
- **`cubeboxManger.Delete` at Warnf** (Code Quality #1). This matches the existing non-paused branch, which also logs cubestore-record delete failures at Warnf; keeping it consistent rather than diverging. The genuinely host-affecting steps (rootfs/fifo/containerd record) now go through the multierror via `cleanPausedContainerHost`.
- **Unbounded stderr** (Performance). The shim `delete` subcommand is a short-lived cleanup process whose stderr is inherently tiny; a capped writer would guard against a scenario that doesn't arise here.
- **Test coverage**. The paused path is almost entirely delegation to the shim's own `delete` subcommand plus `ShimManager` calls, which need a live containerd/shim to exercise meaningfully; there is little standalone logic to unit-test (the earlier hand-rolled kill that would have warranted unit tests is gone). Coverage is via the two-node integration runs above (Destroy and CleanUp paths, full resource reclamation, before/after leak check).

kinwin-ustc:
We have merged PR #940, supporting the ability to delete paused sandboxes

lf4096:
Verified #940 resolves this — closing in favor of it.

PR #1012, opened 2026-07-17
feat(proxy): customize Host header forwarded to sandbox services

FakeLearne:
Add E2B-compatible network.maskRequestHost so create-time config can control the Host CubeProxy sends to user services (for example localhost:${PORT}), without changing the public URL, DNS, or TLS SNI.

CubeAPI accepts camelCase/snake_case and validates the value as a host or host:port authority; only ${PORT} is supported as a placeholder. CubeMaster merges the field from templates, keeps it in sandboxspec / snapshots, and stores the unexpanded template on the sandbox proxy Redis hash. It is not forwarded to Cubelet.

CubeProxy expands ${PORT} per request after routing, rewrites upstream Host through $cube_upstream_host, and sets X-Forwarded-Host when masking applies. Envd port 49983 is exempt. Host, path, and streaming routes share the same path. Optional local-cache metadata uses a false negative sentinel so a missing MaskRequestHost key reloads from Redis instead of failing open.

Also wire the option through Go/Node/Python SDKs, docs, and a quickstart example. Redis key docs additionally list the existing AllowPublicTraffic and TrafficAccessToken fields.


sdk/go/models.go
```diff
@@ -47,9 +47,12 @@ type NetworkOptions struct {
 	// server default. AllowOut/DenyOut are L3/L4 CIDRs or hostnames; Rules are
 	// L7 host/path/SNI matches with audit and credential injection.
 	AllowPublicTraffic *bool
-	AllowOut           []string
-	DenyOut            []string
-	Rules              []Rule
+	// MaskRequestHost is the Host authority forwarded to user services.
+	// ${PORT} expands to the requested sandbox port.
+	MaskRequestHost *string
```
chenhengqi, line 55: Move this new field to last to minimize diff.

sdk/go/aligned_test.go
```diff
@@ -49,6 +51,9 @@ func TestCreateSerializesPolicyAndPublicTraffic(t *testing.T) {
 	if network["allowPublicTraffic"] != false {
 		t.Fatalf("allowPublicTraffic=%#v, want false", network["allowPublicTraffic"])
 	}
+	if network["maskRequestHost"] != maskRequestHost {
```
chenhengqi: Let's use the following packages for new tests:

```
github.com/stretchr/testify/assert
github.com/stretchr/testify/require
```

CubeAPI/src/models/mod.rs
```diff
@@ -56,7 +56,13 @@ pub struct SandboxNetworkConfig {
         skip_serializing_if = "Option::is_none"
     )]
     pub deny_out: Option<Vec<String>>,
-    #[serde(rename = "maskRequestHost", skip_serializing_if = "Option::is_none")]
+    /// Host authority forwarded to user services by CubeProxy. `${PORT}` is
+    /// expanded to the requested sandbox port; envd traffic is exempt.
+    #[serde(
+        rename = "maskRequestHost",
+        alias = "mask_request_host",
```
chenhengqi: Why adding an alias?

FakeLearne: Other parameters have also been given aliases.

sdk/go/models.go
FakeLearne: Done.

CubeProxy/lua/request_host.lua
```diff
@@ -0,0 +1,86 @@
...
+
+local utils = require "utils"
+
+local _M = { _VERSION = "0.01" }
+
+local ENVD_PORT = 49983
+local MAX_MASK_BYTES = 512
+local PORT_PLACEHOLDER_PATTERN = "%$%{PORT%}"
+
+local function validate_authority_text(value)
```
chenhengqi: This seems unnecessary.

FakeLearne: Deleted already

sdk/go/aligned_test.go
FakeLearne: Done.

CubeProxy/lua/sandbox_backend.lua
```diff
@@ -158,16 +183,26 @@ function _M.resolve_backend(ins_id, container_port)
         local k = metadata[i]
         local v = metadata[i + 1]
         metadata_map[k] = v
-        cache:set(ins_id .. ":" .. k, v, timeout)
+        if k ~= "AllowPublicTraffic"
+            and k ~= "TrafficAccessToken"
+            and k ~= "MaskRequestHost" then
+            cache:set(ins_id .. ":" .. k, v, timeout)
+        end
```
chenhengqi: This seems unnecessary, it will be override below anyway.

CubeProxy/lua/sandbox_backend.lua
```diff
@@ -124,26 +142,33 @@ function _M.resolve_backend(ins_id, container_port)
...
+    local meta_cached = cache:get(ins_id .. ":meta_cached")
+    local cached_allow_public = cache:get(ins_id .. ":AllowPublicTraffic")
+    local cached_traffic_token = cache:get(ins_id .. ":TrafficAccessToken")
+    local cached_mask_request_host = cache:get(ins_id .. ":MaskRequestHost")
     if host_ip and host_port
-        and cache:get(ins_id .. ":meta_cached") then
+        and meta_cached
+        and cached_allow_public ~= nil
+        and cached_traffic_token ~= nil
+        and cached_mask_request_host ~= nil then
```
chenhengqi: Can we omit these nil checks here?

CubeMaster/pkg/templatecenter/template_image_test.go
```diff
@@ -687,6 +649,12 @@ func TestGenerateTemplateCreateRequestClonesCubeNetworkRules(t *testing.T) {
 	if got.CubeNetworkConfig == nil {
 		t.Fatal("expected CubeNetworkConfig to be propagated")
 	}
+	if got.CubeNetworkConfig.MaskRequestHost == nil || *got.CubeNetworkConfig.MaskRequestHost != maskRequestHost {
```
chenhengqi: Use assert.

CubeProxy/lua/sandbox_backend.lua
FakeLearne: No — meta_cached alone isn’t enough. Those optional fields are separate ngx.shared.DICT keys and can be LRU-evicted independently; a missing key returns nil, which we’d treat as “no restriction / no mask” and fail open. The ~= nil checks ensure the cache entry is complete (including the false negative sentinels) before we take the hit path.

CubeProxy/lua/sandbox_backend.lua
FakeLearne: Done.

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
```diff
@@ -299,30 +303,15 @@ func mergeCubeNetworkConfigs(templateCfg *types.CubeNetworkConfig, requestCfg *t
...
-	out := cloneCubeNetworkConfigBase(in)
-	if out == nil {
-		return nil
-	}
-	out.Rules = cloneEgressRules(in.Rules)
-	return out
+	return in.DeepCopy()
 }
 
 func cloneCubeNetworkConfigBase(in *types.CubeNetworkConfig) *types.CubeNetworkConfig {
```
chenhengqi: cloneCubeNetworkConfigBase() and cloneEgressRule() can be removed now?

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
```diff
@@ -299,30 +303,15 @@ func mergeCubeNetworkConfigs(templateCfg *types.CubeNetworkConfig, requestCfg *t
 }
 
 func cloneCubeNetworkConfig(in *types.CubeNetworkConfig) *types.CubeNetworkConfig {
-	out := cloneCubeNetworkConfigBase(in)
-	if out == nil {
-		return nil
-	}
-	out.Rules = cloneEgressRules(in.Rules)
-	return out
+	return in.DeepCopy()
```
chenhengqi, line 306: nil check?

CubeMaster/pkg/templatecenter/template_image_test.go
FakeLearne: Done.

CubeMaster/pkg/service/sandbox/types/types.go
```diff
@@ -139,10 +139,13 @@ type CubeNetworkConfig struct {
...
 	// sandbox require a matching traffic-access-token header on every
 	// request hitting CubeProxy. See plan/restrict-public-access.md.
-	AllowPublicTraffic *bool         `json:"allowPublicTraffic,omitempty"`
-	AllowOut           []string      `json:"allowOut,omitempty"`
-	DenyOut            []string      `json:"denyOut,omitempty"`
-	Rules              []*EgressRule `json:"rules,omitempty"`
+	AllowPublicTraffic *bool `json:"allowPublicTraffic,omitempty"`
+	// MaskRequestHost is an ingress-only Host authority template consumed by
+	// CubeProxy. It is intentionally not forwarded to Cubelet.
+	MaskRequestHost *string       `json:"maskRequestHost,omitempty"`
```
chenhengqi: Make this the last field.

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
FakeLearne: Removed already.

CubeMaster/pkg/service/httpservice/cube/cubeboxutil.go
FakeLearne: DeepCopy() already nil-checks the receiver and returns nil, so an extra guard in this wrapper would be redundant. Call sites do pass nil when one side of the merge is missing.

CubeMaster/pkg/service/sandbox/types/types.go
FakeLearne: Done.

PR #1015, opened 2026-07-18
fix(cubelet): skip capacity admission for delete-triggered auto-resume

kinwin-ustc:
## Summary

- When `paused_resource_release_ratio > 0` and a node is overcommitted, deleting a paused sandbox was blocked by the same `admitResume` capacity check used for user-initiated resume, creating a deadlock: the user wants to free resources by deleting, but the delete is blocked because the node lacks resources.
- Add a `skipAdmission` flag to `resumeOptions` so that `resumePausedSandboxForDestroy` bypasses the capacity check. The resume is a transient step before immediate destruction, so the brief overcommit is bounded by the destroy deadline.
- User-initiated resume (`UpdateWithResume`) remains subject to the full capacity admission check.

## Changes

| File | Change |
|------|--------|
| `Cubelet/services/cubebox/update.go` | Add `skipAdmission bool` to `resumeOptions`; guard `admitResume` call with `!opts.skipAdmission` |
| `Cubelet/services/cubebox/delete_resume.go` | Set `skipAdmission: true` in `resumePausedSandboxForDestroy` |
| `Cubelet/services/cubebox/destroy_auto_resume_test.go` | Integration test: delete succeeds under capacity pressure |
| `Cubelet/services/cubebox/release_paused_resources_test.go` | Unit tests: `admitResume` rejects when exceeded; `resumeLocked` bypasses with flag; `resumeLocked` rejects without flag |

## Test plan

- [x] `TestDestroyPausedSandboxSucceedsWhenCapacityWouldRejectResume` — end-to-end: delete of a paused sandbox succeeds even when node quota would reject a normal resume
- [x] `TestAdmitResumeRejectsWhenCapacityExceeded` — confirms `admitResume` correctly rejects when post-resume demand exceeds quota
- [x] `TestResumeLockedSkipsAdmissionWhenFlagSet` — `skipAdmission: true` bypasses admission and resume succeeds
- [x] `TestResumeLockedRejectsWithoutSkipAdmissionUnderCapacityPressure` — `skipAdmission: false` (user resume path) still rejected under capacity pressure
- [x] All existing `destroy_auto_resume_test.go` and `delete_resume_test.go` tests remain passing
- [x] `GOOS=linux go test -c` compilation verified

### DIFF
--- a/Cubelet/services/cubebox/delete_resume.go
+++ b/Cubelet/services/cubebox/delete_resume.go
@@ -138,6 +138,7 @@ func (s *service) resumePausedSandboxForDestroy(ctx context.Context, sb *cubebox
 		taskDeadline:      budget.resumeDeadline(now),
 		reconcileDeadline: budget.resumeDeadline(now),
 		persist:           false,
+		skipAdmission:     true, // sandbox will be destroyed immediately after resume
 	})
 	durationMS := time.Since(now).Milliseconds()
 	if result.running {
--- a/Cubelet/services/cubebox/destroy_auto_resume_test.go
+++ b/Cubelet/services/cubebox/destroy_auto_resume_test.go
@@ -17,6 +17,7 @@ import (
 
 	"github.com/tencentcloud/CubeSandbox/Cubelet/api/services/cubebox/v1"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/api/services/errorcode/v1"
+	"github.com/tencentcloud/CubeSandbox/Cubelet/pkg/config"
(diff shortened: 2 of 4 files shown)


kinwin-ustc:
In version 0.6.0, we need to ensure that deleting paused sandboxes always works, preventing deletion failures due to insufficient single-node capacity. In a longer-term design, deleting sandboxes should be performed directly, avoiding redundant logic like recovery followed by deletion, and preventing inconsistencies caused by successful recovery but failed deletion.

Issue #1023, opened 2026-07-19
[Feature Request] Refactor control plane API architecture: unify protocol definitions, eliminate redundant types, centralize validation


kinwin-ustc:
## Background

The CubeSandbox control plane (CubeAPI → CubeMaster → Cubelet) currently suffers from significant architectural redundancy in protocol definitions, type management, serialization/deserialization, and parameter validation. This leads to an unnecessarily large blast radius for every API change and frequent inconsistencies.

## Goals

1. **Single protocol definition** — Proto as the Single Source of Truth
2. **Unified communication protocols** — CubeAPI→CubeMaster via RESTful HTTP, CubeMaster→Cubelet via gRPC
3. **Unified full-chain type definitions** — Eliminate intermediate type layers and manual type conversions
4. **Minimal API change blast radius** — Adding/removing a field should touch as few places as possible
5. **Minimize manual serialization/deserialization** — Leverage `protojson` for automatic handling
6. **Centralized parameter validation** — Declare validation rules in proto using `protovalidate`

---

### Quadruple Type Redundancy

The same concept (e.g., "create sandbox") is defined **4 times** across the chain, with manual type conversion at every boundary:

| Layer | File | Type Name | Language |
|---|---|---|---|
| CubeAPI external API | `CubeAPI/src/models/mod.rs` | `NewSandbox` | Rust |
| CubeAPI→Master wire | `CubeAPI/src/cubemaster/mod.rs` | `CreateSandboxRequest` | Rust |
| CubeMaster HTTP | `CubeMaster/pkg/service/sandbox/types/types.go` | `CreateCubeSandboxReq` | Go |
| CubeMaster→Cubelet gRPC | `CubeMaster/api/services/cubebox/v1/cubebox.proto` | `RunCubeSandboxRequest` | Protobuf |

### Duplicated Proto Files

The same proto files exist in both `CubeMaster/api/` and `Cubelet/api/` with identical semantic content, differing only in `go_package`. This means every proto change must be synchronized in two places.

| Proto File | CubeMaster Path | Cubelet Path |
|---|---|---|
| cubebox.proto | `CubeMaster/api/services/cubebox/v1/` | `Cubelet/api/services/cubebox/v1/` |
| images.proto | `CubeMaster/api/services/images/v1/` | `Cubelet/api/services/images/v1/` |
| errorcode.proto | `CubeMaster/api/services/errorcode/v1/` | `Cubelet/api/services/errorcode/v1/` |
[...]

PR #1026, opened 2026-07-19
refactor(proto): unify shared proto definitions into a standalone proto/ Go module (#1023 phase1)

kinwin-ustc:
## Summary

Phase 1 of #1023: Unify shared protobuf definitions to eliminate duplication between CubeMaster and Cubelet.

- **Created `proto/` shared Go module** (`github.com/tencentcloud/CubeSandbox/proto`) containing the canonical protobuf definitions for `cubebox`, `errorcode`, `images`, and `types`
- **Removed duplicated proto files** from `CubeMaster/api/` (entirely) and `Cubelet/api/` (shared protos only; Cubelet-specific protos like `nbi`, `cubehost`, `multimetadb`, `version` remain)
- **Migrated ~200 Go files** across both CubeMaster and Cubelet to import from the shared module
- **Extracted CRI conversion functions** from `Cubelet/api/types/v1/cri.go` into `Cubelet/pkg/critypes/convert.go` as standalone functions with nil guards, avoiding `k8s.io/cri-api` dependency pollution in the shared module
- **Added table-driven unit tests** for all CRI conversion functions (v1 and v1alpha2), including nil input and edge cases
- **Updated Makefiles** for the new proto generation layout (`proto/Makefile` for shared protos, both CubeMaster and Cubelet delegate to it)
- **Fixed `PROTOBUF_INCLUDE_DIR` detection** to prefer system dirs over `gogo/protobuf`, avoiding CI failures
- **Removed gRPC v1.67.1 pin** from Cubelet to align versions with the shared module

### Before
```
CubeMaster/api/services/cubebox/v1/cubebox.proto  ← duplicate
Cubelet/api/services/cubebox/v1/cubebox.proto      ← duplicate
(same for errorcode, images, types)
```

### After
```
proto/services/cubebox/v1/cubebox.proto            ← single source of truth
proto/services/errorcode/v1/errorcode.proto
proto/services/images/v1/images.proto
proto/types/v1/images.proto
proto/types/v1/descriptor.proto
```

## Test plan

- [x] `proto/` module: `go build ./...` passes
- [x] CubeMaster: `GOOS=linux` cross-compilation passes (all packages)
- [x] Cubelet: `GOOS=linux` cross-compilation passes (excluding CGO-only `cubemnt`/`nsenter`)
- [x] Cubelet CRI conversion unit tests: 20/20 pass (including nil guards and v1alpha2 edge cases)
- [x] CI: all builds pass (CubeMaster, Cubelet, CubeShim, CubeAPI, agent, network-agent)
- [x] CI: Code Format Check pass
- [x] CI: DCO check pass

Refs: #1023


fslongjin:
Hello— small build-path gap I noticed after the shared protos moved into `api/`.

CubeMaster’s `make proto` correctly forwards to `../api`, but Cubelet’s still only regenerates the local bits (`version` / `nbi` / `multimetadb` + the network_agent client). So root `make cubelet` still does `make proto && make build`, which looks fine at a glance, but it never actually refreshes the shared `.pb.go` under `api/`.

If someone edits `api/**/*.proto` and only builds cubelet, it’s pretty easy to compile against stale generated code without realizing it.

Could we add something like `$(MAKE) -C ../api proto` to Cubelet’s `proto` target so it stays in sync with CubeMaster? That way root Makefile / one-click / a local `make proto` all cover the shared module.

(Tiny related note: one-click still plain-`go build`s cubemaster and network-agent without running proto first — fine if generated files are always committed, but worth calling out if we want those paths to match the root Makefile.)

fslongjin (approved):
LGTM

kinwin-ustc:
We will begin merging this part of the work after the 0.6.0 release.

PR #1034, opened 2026-07-20
cubelet: store cubebox_os_image artifacts on /data by default

xiaojunxiang2023:
## Summary

- Store template rootfs artifacts under `/data/cubebox_os_image` by default instead of `/usr/local/services/cubetoolbox/cubebox_os_image`, so large ext4 images do not fill the system disk.
- Split pmem toolbox root (shared kernel under cubetoolbox) from OS-image parent (`os_image_parent_dir`, default `/data`).
- Ensure one-click and cube-node entrypoints create `/data/cubebox_os_image`; document the new default in templates troubleshooting (EN/ZH).

Upgrade note: existing nodes with artifacts under cubetoolbox can move them with:

```bash
mkdir -p /data
mv /usr/local/services/cubetoolbox/cubebox_os_image /data/
# optional compat symlink for any leftover absolute paths
ln -sfn /data/cubebox_os_image /usr/local/services/cubetoolbox/cubebox_os_image
```

## Test plan

- [x] `go test ./pkg/container/pmem/ -count=1` in `cube-compiler:ubuntu22` on ubuntu22 — passed
- [x] Path separation: shared kernel stays under toolbox; `GetPmemBasePath("cubebox")` → `<os_image_parent>/cubebox_os_image`

### DIFF
--- a/Cubelet/config/config.toml
+++ b/Cubelet/config/config.toml
@@ -119,6 +119,16 @@ dynamic_config_path = "/usr/local/services/cubetoolbox/Cubelet/dynamicconf/conf.
     # pool_trigger_interval_in_ms = 800
   [plugins."io.cubelet.internal.v1.images"]
     runtime_type = "io.containerd.cube.v2"
+    # cubetoolbox install root (shared kernel).
+    cubetool_base_dir = "/usr/local/services/cubetoolbox"
+    # Template rootfs parent defaults to /data (→ /data/cubebox_os_image).
+    # If you override this, also set CBRI image_base_path and kernel_base_path
+    # to the matching <parent>/cubebox_os_image so write/read paths stay aligned.
+    # os_image_parent_dir = "/data"
+  [plugins."io.cubelet.cbri.v1.cubebox"]
+    # Defaults match images os_image_parent_dir (/data/cubebox_os_image).
+    # image_base_path = "/data/cubebox_os_image"
+    # kernel_base_path = "/data/cubebox_os_image"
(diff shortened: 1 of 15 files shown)


kinwin-ustc:
Why not use symbolic links in your own deployment?

xiaojunxiang2023:
> Why not use symbolic links in your own deployment?

That works too, and requires fewer changes. In that case, I'll align with the community deployment scripts and create a symlink:

`/usr/local/services/cubetoolbox/cubebox_os_image -> /data/cubebox_os_image`

Would the community be open to this approach? If so, I'm happy to implement it based on the community deployment scripts (creating the symlink: `/usr/local/services/cubetoolbox/cubebox_os_image -> /data/cubebox_os_image`).

kinwin-ustc:
> > Why not use symbolic links in your own deployment?
> 
> That works too, and requires fewer changes. In that case, I'll align with the community deployment scripts and create a symlink:
> 
> `/usr/local/services/cubetoolbox/cubebox_os_image -> /data/cubebox_os_image`
> 
> Would the community be open to this approach? If so, I'm happy to implement it based on the community deployment scripts (creating the symlink: `/usr/local/services/cubetoolbox/cubebox_os_image -> /data/cubebox_os_image`).

This capability is scenario-dependent and not a universal solution; you can use it as a prerequisite for your customized deployment.

xiaojunxiang2023:
Understood. We'll keep this as part of our internal deployment.

PR #1037, opened 2026-07-20
feat(cubemaster): add CUBEMASTER_AUTO_MIGRATION switch to skip startup schema migration

dushulin:
## Summary

Adds an opt-out switch, `CUBEMASTER_AUTO_MIGRATION`, that lets a CubeMaster deployment skip the schema migration it runs at startup.

## Motivation

In a hardened deployment the CubeMaster **runtime** database account is granted DML only, no DDL. Such an account cannot run the startup migrator at all — even when the schema is already at HEAD, the fingerprint defence layer issues an unconditional `CREATE TABLE IF NOT EXISTS`, which fails with `ERROR 1142 (command denied)`. CubeMaster then refuses to boot.

The operational fix in these environments is to apply the schema **out-of-band** with a privileged account (CI job, DBA, etc.) and have CubeMaster boot without touching DDL. Today there is no way to tell CubeMaster to do that.

## Change

- New `migrate.AutoMigrationEnabled()` reads `CUBEMASTER_AUTO_MIGRATION`.
  - Defaults to **true** (migrate).
  - Returns **false** only for an explicit falsey value: `false` / `0` / `no` / `off` (case-insensitive, whitespace-trimmed).
  - Any unset, empty, or unrecognised value keeps the safe default, so a typo can never silently disable migration. (`strconv.ParseBool` is deliberately avoided because it rejects `no`/`off`.)
- `initDatabaseSchema` gates `dao.Migrate` behind the switch, between `dao.Open` and `dao.Migrate`. Because the gate sits at the top level it is dialect-agnostic (applies to both mysql and postgres).
- When the switch is off, CubeMaster logs a clear warning that DDL must be applied out-of-band.

No behaviour change for existing deployments: the default is unchanged (migrate on boot).

## Test plan

- [x] `go build ./...` in both `CubeDB` and `CubeMaster` modules
- [x] `go vet` clean on the changed packages
- [x] New table-driven unit test `TestAutoMigrationEnabled` covering unset / empty / true / 1 / True / on / yes / garbage (→ enabled) and false / 0 / no / off / FALSE / whitespace-wrapped (→ disabled); 14/14 pass

### DIFF
--- a/CubeDB/migrate/automigration_test.go
+++ b/CubeDB/migrate/automigration_test.go
@@ -1,0 +1,50 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package migrate
+
+import (
+	"os"
+	"testing"
+)
+
+func TestAutoMigrationEnabled(t *testing.T) {
+	cases := []struct {
+		name string
+		set  bool // whether to set the env var at all
+		val  string
+		want bool
+	}{
+		{name: "unset", set: false, want: true},
+		{name: "empty", set: true, val: "", want: true},
+		{name: "true", set: true, val: "true", want: true},
+		{name: "one", set: true, val: "1", want: true},
+		{name: "True mixed case", set: true, val: "True", want: true},
+		{name: "on", set: true, val: "on", want: true},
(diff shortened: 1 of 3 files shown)


fslongjin (approved):
LGTM. I think it's useful in prod environment.

PR #1038, opened 2026-07-20
fix(cubeops): remove docker dependency from store metadata endpoints

liciazhu:
fix(cubeops): remove docker dependency from store metadata endpoints, use go-containerregistry to fetch images info in template market.


ls-ggg:
Please fix the CI failures

PR #1044, opened 2026-07-20
feat(lifecycle): make proxy state push async with backoff retries

novahe:
Move the proxy state push (ProxyPush.SetState) to a background
goroutine to prevent blocking the initial dataplane request.
Add exponential backoff retries to gracefully handle transient
network failures or brief proxy unavailability.


cube-lifecycle-manager/go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager
 go 1.25.7
 
 require (
+	github.com/cenkalti/backoff/v4 v4.3.0
```
chenhengqi: Why not use the latest v7 package?

novahe: fixed

PR #1053, opened 2026-07-21
CubeProxy: Return 403 for traffic-token failures to preserve E2B comp…

chenhengqi:
…atibility

The 404-collapse in e052abd7 swept enforce_traffic_token()'s two failure paths into the "hide sandbox existence" bucket, unintentionally breaking the E2B wire contract that #639 was built to mirror. The demo, the Restrict-Public-Access docs (en + zh), the v0.5.0 changelog, and the Python SDK docstrings all promise HTTP 403 for missing / mismatched tokens — E2B does the same. 404 makes restrict_public_access.py's Step-2 verdict fail.

Add utils:respond_forbidden() (403 + {"error":"forbidden"}) and route enforce_traffic_token()'s two rejection paths through it. The comment above the function now records the deliberate tradeoff: we accept the sandbox-existence signal e052abd wanted to hide in exchange for keeping the E2B contract and already-shipped SDK/docs behavior intact. All other 404/503/400 responses e052abd introduced remain untouched.

Verified locally with restrict_public_access.py — all three probes (no-header 403, e2b header 200, cube header 200) now PASS.

Refs:  e3ecfc138a53 ("network: Restrict public access via per-sandbox traffic token")
Fixes: e052abd7c698 ("CubeProxy: Remove cube_retcode to stop leaking implementation details")

### DIFF
--- a/CubeProxy/lua/sandbox_backend.lua
+++ b/CubeProxy/lua/sandbox_backend.lua
@@ -19,10 +19,13 @@ local _M = { _VERSION = "0.01" }
 --
 -- Both args are the raw values stored in Redis (string form). expected_token
 -- being empty while allow_public is "false" indicates a server-side
--- inconsistency. Both failure paths return 404 (not 403/500) so that a
--- caller cannot distinguish "sandbox exists but access denied" from
--- "sandbox does not exist"; silently letting the request through would be
--- worse.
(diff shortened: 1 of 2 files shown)


tinklone:
Clipboard_Screenshot_1784626964
LGTM

zhouxianping (approved):

Issue #642, opened 2026-06-25
【2026犀牛鸟issue】CubeSandbox Webhook 事件通知


fslongjin:
**【产品名称/功能模块】**
CubeSandbox Webhook 事件通知

**【问题描述】**
Webhook 是在特定事件发生时，向用户指定 HTTP 端点发送的 HTTP 回调。CubeSandbox 作为面向 AI Agent 的安全沙箱服务，沙箱会高频创建、暂停、恢复与销毁，快照与模板构建也会触发关键状态变更。平台控制面虽已记录部分结构化日志事件（如 `sandbox.created`、`sandbox.deleted`），但尚未向用户提供可配置的 Webhook 订阅与异步投递能力。

缺乏 Webhook 时，上层 Agent 编排系统、运维平台或企业 IM 工具只能轮询 API 获取状态，无法做到实时联动。Webhook 可实现：沙箱生命周期实时通知、快照完成触发自动化流程、使用模式监控与分析、对接企业微信等第三方工具、关键变更的审计追踪等场景。

本 issue 为代码贡献任务，需阅读 CubeSandbox 现有代码（重点：`CubeAPI` 沙箱 Handler 与 `logging` 模块），在本地虚拟机部署的 CubeSandbox 环境上完成开发与验证。

**【期望内容/功能改进】**
1）支持配置一个或多个 Webhook 端点（配置文件或环境变量方式，可选提供 REST API 管理）；
2）支持按事件类型订阅，至少覆盖 4 类沙箱生命周期事件：`sandbox.created`、`sandbox.deleted`、`sandbox.paused`、`sandbox.resumed`；
3）事件触发时向已订阅端点异步发送 JSON 格式的 HTTP POST，请求体至少包含 `event`、`timestamp`、`sandbox_id`，并尽量携带 `template_id` 等上下文字段；
4）投递过程不得阻塞沙箱创建/销毁等主路径，须使用异步队列或后台任务；
5）支持可选的 HMAC-SHA256 签名机制，便于接收方验签；
6）投递失败时应有重试策略（建议指数退避）并记录错误日志；
7）在 `examples/` 目录下提供可运行的 Webhook 接收端示例及 README；
8）提供集成说明文档，包含配置方法、Payload 说明、签名验证示例，以及对接企业微信机器人或通用 HTTP 告警的参考做法；
9）补充必要的单元测试或集成测试（可使用 mock HTTP server），提交 Pull Request 合入上游仓库。

[...]

PR #1057, opened 2026-07-21
cubeapi: add webhook event notifications for sandbox lifecycle

NeverENG:
## What & why

Implements configurable **Webhook event notifications** for CubeSandbox, addressing #642.

Sandboxes change state at high frequency, but until now upstream agent orchestrators / ops platforms / IM tools could only poll the API. This adds asynchronous, signed HTTP delivery of sandbox lifecycle events to user-configured endpoints.

## Design

The four lifecycle events (`sandbox.created` / `sandbox.paused` / `sandbox.resumed` / `sandbox.deleted`) are **already emitted** through `state.logger.log(...)` in the sandbox handlers. Rather than re-plumbing the handlers, this PR implements `WebhookLogger` as a new `Logger` backend and registers it in the existing `MultiLogger` fan-out — so it consumes the same structured event stream with **no changes to `sandboxes.rs`**.

- **Config** — endpoints loaded from a YAML/JSON/TOML file via `--webhook-config` / `CUBE_API_WEBHOOK_CONFIG`. Per-endpoint event subscription, optional secret, timeout, retry count. Disabled (zero overhead) when unset.
- **Non-blocking** — `log()` only enqueues; a background dispatcher spawns an independent delivery task per (event, endpoint), so a slow/unreachable receiver never blocks the API request path or other endpoints. Per-request timeout.
- **Retries** — exponential backoff (1s/2s/4s…) up to `max_retries`; final failure logged at `error`.
- **Signing** — optional HMAC-SHA256 over the raw body, sent as `X-Cube-Signature: sha256=<hex>`.
- **Idempotency** — each delivery carries a stable `X-Cube-Delivery` id (unchanged across retries) so receivers can deduplicate.
- **Management API (optional per issue #1)** — `GET/POST/DELETE /webhooks` to manage endpoints at runtime (e.g. from the WebUI) without a restart. Backed by a shared `Arc<RwLock<..>>` registry; secrets are masked in responses; runtime changes are in-memory (config file provides the durable startup set).

## Payload

```json
{ "event": "sandbox.created", "timestamp": "2026-07-21T08:30:12.481Z", "sandbox_id": "sbx-abc123", "template_id": "base" }
```

## Deliverables vs. issue acceptance criteria

| # | Criterion | Status |
|---|---|---|
| 1 | Register endpoints via config, subscribe by event type | ✅ (config file + env; optional REST management API also included) |
| 2 | Create/pause/resume/delete → receiver gets correct callbacks | ✅ verified in a local deployment |
| 3 | Async / non-blocking; receiver timeout ⇒ API still succeeds | ✅ (unit-tested) |
| 4 | Optional signature verifiable by receiver | ✅ |
| 5 | Retry + error logging on failure | ✅ |
| 6 | Runnable receiver example + docs | ✅ `examples/webhook-receiver/` (EN/ZH) |
| 7 | Unit tests cover core delivery; passes fmt/clippy | ✅ 5 new tests; `cargo fmt`/`clippy` clean |

## Docs & examples

- `docs/guide/webhooks.md` + `docs/zh/guide/webhooks.md` (config, payload, signature verification, WeCom / generic alerting) + sidebar entries.
- `examples/webhook-receiver/` — stdlib-only receiver, WeCom bridge, sample config, EN/ZH README.

## Testing done

- `cargo test` — 103 pass, incl. new tests for config-file parsing, payload/fields, HMAC signing, subscription filtering, non-blocking `log()`, the `X-Cube-Delivery` id header, and the `/webhooks` management API (list/mask, create→delete, 404, validation) — real in-process HTTP receiver / `TestServer`.
- `cargo fmt --check` and `cargo clippy` clean for the changed files.
- Receiver signature verification validated in-process against the Rust `sign()` output.

## End-to-end verification

Verified end-to-end in a local CubeSandbox deployment: creating, pausing, resuming, and deleting a sandbox each delivered the corresponding webhook to the example receiver (`examples/webhook-receiver/`) with the correct payload fields.


NeverENG:
this is E2E test result

image

fslongjin (changes requested):
Hello, thank you for your contribution. I believe this feature meets the requirements of the relevant PR for the Rhino-bird activity, but it is still not sufficient for merging.

Because cubeapi is a component that requires high availability deployment, the current implementation is not conducive to high availability and may also result in some event loss. If you wish to push this PR toward merging, I think you could refine it to be truly production-ready, but be careful not to over-engineer it. 

To be honest, from the perspective of fulfilling the requirements of the Rhino-bird open source activity, this PR is already qualified.

NeverENG:
Thanks for the review! I took the "high-availability / potential event loss" feedback and hardened the webhook delivery path. All changes are confined to the CubeAPI component (plus its docs) — no changes to CubeMaster, Cubelet, or any other component — deliberately kept in scope to avoid over-engineering.

Addressing the concrete loss scenarios:

- Loss on every rolling deploy (the big one): flush() was a no-op, so graceful shutdown drained in-flight HTTP requests and then exited immediately, abandoning everything still queued, in flight, or sleeping in retry backoff. It now drains for real via a TaskTracker + CancellationToken: on shutdown it drains the remaining queue into delivery tasks, cuts retry backoff short, and waits for them to finish within a bounded grace period.
- OOM under overload losing everything at once: the unbounded channel is now bounded, fed via try_send (still non-blocking, so it never stalls the request path). On overload the newest event is dropped and logged (throttled warn) instead of growing memory without limit.
- Receivers unable to deduplicate: the delivery idempotency key is now assigned once per logical event (stable across endpoints and retries) instead of per retry attempt — exposed as X-Cube-Delivery. This is also the prerequisite for any durable at-least-once delivery later.
- Resource safety & performance: concurrent in-flight deliveries are capped with a semaphore; the payload is shared as reference-counted Bytes across endpoints and retries (zero-copy, no per-attempt re-allocation); the webhook HTTP client keeps a warm connection pool.
- Operability: queue capacity, drain grace, and max concurrency are all tunable via environment variables; event loss is surfaced in logs (full queue → warn, exhausted retries → error).
- Docs (EN + zh): delivery semantics, ordering (receivers reconcile by the payload timestamp), and the HA caveat that runtime CRUD is per-instance — the config file should be the declarative source of truth across replicas.

On the boundary, to be upfront: true at-least-once delivery across a hard crash (SIGKILL / OOM / node failure) isn't achievable inside a stateless gateway — it requires a durable event consumer in the control plane (CubeMaster already maintains a sandbox:lifecycle Redis Stream that could back this). That aligns with the "control-plane / data-plane separation" item on the roadmap, so I've left it as a natural follow-up rather than pulling it into this PR.

The work is split into 5 small commits, and cargo test is green. Feedback welcome!

fslongjin:
This implementation leads to event loss, which is unacceptable for production systems that rely on events to trigger actions.

fslongjin:
Hello, thank you for your contribution. After careful consideration, I believe this PR does not yet meet the requirements for merging. However, for the Rhino Bird project, it has already fulfilled the criteria—I’ve logged it accordingly. Therefore, I’ll temporarily close this PR for now. We warmly welcome your continued involvement in CubeSandbox’s community development—let’s build more interesting features together! 🚀

PR #1065, opened 2026-07-21
ci: add Go SDK publish workflow

### LINKED ISSUE #1016: [Feature Request] Go SDK nested module is not installable from release tags
## Problem

The Go SDK declares a nested module:

```text
module github.com/tencentcloud/CubeSandbox/sdk/go
```

However, the current root release tag `v0.5.1` is not usable for that module. Go applies nested-module tag semantics and looks for `sdk/go/v0.5.1`:

```console
$ go list -m github.com/tencentcloud/CubeSandbox/sdk/go@v0.5.1
go: github.com/tencentcloud/CubeSandbox/sdk/go@v0.5.1: invalid version: unknown revision sdk/go/v0.5.1
```

The root `v0.5.1` snapshot also predates the PTY API currently present in `sdk/go`, so pinning the root release would not cover the documented lifecycle, command, file, snapshot, clone/rollback, egress, and PTY integration surface.

## Current workaround

After validating the official SDK against a full local control/data-plane protocol suite, restart recovery, race tests, and a built-binary PTY E2E, we had to pin the exact commit as an immutable pseudo-version:

```go
require github.com/tencentcloud/CubeSandbox/sdk/go v0.0.0-20260717090631-002e4049fff0
```

We are not using a fork, `replace` directive, or floating branch. This works, but it is a temporary workaround rather than a normal consumable SDK release.

## Request

Please publish a semantic version tag for the nested Go module (for example `sdk/go/v0.5.2`, following Go nested-module conventions), including the current PTY API and recent SDK fixes. It would also help to document whether root releases and Go SDK releases are intended to share a version or move independently.

Environment: Go 1.26.5 on Linux.

wbzdssm:
Fixes #1016

Adds a GitHub Actions workflow triggered by `go-sdk-v*` tags.
When pushed, it validates the Go SDK, auto-creates the `sdk/go/v*`
tag required by Go's nested module convention, and runs unit tests.

Follows the same `*-sdk-v*` tag prefix convention as `publish-python-sdk.yml`.

Usage:
```bash
git tag go-sdk-v0.6.0 && git push origin go-sdk-v0.6.0
# Auto-creates sdk/go/v0.6.0 and runs tests
```

Also verified: all tags from v0.3.0 to v0.5.1 already contain `sdk/go/`
code but lack the corresponding `sdk/go/v*` Go module tags.
This workflow prevents the issue for future releases.

### DIFF
--- a/.github/workflows/publish-go-sdk.yml
+++ b/.github/workflows/publish-go-sdk.yml
@@ -1,0 +1,147 @@
+name: Publish Go SDK
+
+# When a go-sdk-v* tag is pushed, auto-create the nested
+# sdk/go/v* tag (Go convention) and run validation tests.
+#
+# Usage:
+#   git tag go-sdk-v0.6.0
+#   git push origin go-sdk-v0.6.0
+#
+# Automatically creates the sdk/go/v0.6.0 tag. Fixes Issue #1016.
+# Follows the same tag convention as python-sdk-v*.
+
+on:
+  push:
+    tags:
+      - 'go-sdk-v*'
+
+concurrency:
+  group: go-sdk-publish-${{ github.ref_name }}
+  cancel-in-progress: true
+
+jobs:
+  create-go-sdk-tag:
+    name: Create Go SDK tag
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+
+    steps:
+      - name: Checkout (full history, blobless)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+          filter: 'blob:none'
+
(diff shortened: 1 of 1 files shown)


kinwin-ustc:
ranslate all Chinese in the comments to English

.github/workflows/publish-go-sdk.yml
```diff
@@ -0,0 +1,126 @@
+name: Publish Go SDK
+
+# 当 go-sdk-v* 标签被推送时，为嵌套的 Go 模块
```
chenhengqi: Use English for comments.

PR #1072, opened 2026-07-22
fix(cubemaster): make dao layer cross-engine safe for
  postgres

cookiebody:
## What

Fixes three pre-existing MySQL-specific assumptions in the dao /
templatecenter code that break when CubeMaster runs against the
`postgres` driver added by the multi-driver dao layer. All three fixes
are behaviour-preserving on MySQL and additive for other dialects — no
change to existing MySQL deployments.

## Why

Running CubeMaster against PostgreSQL surfaced three failures:

1. **`UpsertReplica` — `42712 "table name specified more than once"`.**
   A single `*gorm.DB` builder was reused across `First()` and
   `Updates()`. `First()` finalizes the statement (adds `ORDER BY`/`LIMIT`
   and pins the table), so reusing it for `Updates()` emits SQL that
   references `t_cube_template_replica` twice. MySQL tolerates the
   duplicate FROM target; PostgreSQL rejects it, which **stalled template
   distribution** — replicas never reached `READY`.

2. **Artifact GC advisory lock hard-coded `GET_LOCK`/`RELEASE_LOCK`,**
   which do not exist on PostgreSQL.

3. **`base/db.Init` imported `gorm.io/driver/mysql` directly** and always
   opened a MySQL connection, bypassing the driver-agnostic dao layer.

## How

1. Build a fresh `WHERE` for each finalizing call so the statement is
   finalized exactly once — correct on both dialects.
2. Select advisory-lock statements by dialect
   (`GET_LOCK`/`RELEASE_LOCK` on MySQL; `pg_try_advisory_lock` /
   `pg_advisory_unlock` keyed by `hashtext()` on PostgreSQL) and run them
   on a pinned `*sql.Conn`, since advisory locks are session-scoped on
   both engines and acquire/release must share one connection.
3. Route `Init` through `dao.Open`, falling back to `dao.Default()`, so
   the configured driver is honoured.

## Tests

- `TestUpsertReplicaUpdateSQLReferencesTableOnce` — asserts the finalized
  `UPDATE` references the table exactly once, rendered via gorm `ToSQL`
  on a pgx `DryRun` handle (no live connection).
- `TestArtifactGCLockSQL` — asserts the per-dialect lock statements.

Both pass. Also runtime-verified end to end against an external
PostgreSQL instance: template distribution reaches `READY` with no
`42712`, and sandbox create/destroy completes.

---

Part of #517 (PostgreSQL support for the CubeMaster dao layer). This PR
is the cross-engine correctness subset and stands alone; the deployment
wiring will follow in a separate PR.

### DIFF
--- a/CubeMaster/pkg/base/db/db.go
+++ b/CubeMaster/pkg/base/db/db.go
@@ -7,93 +7,77 @@ package db
 
 import (
 	"context"
-	"database/sql"
-	"fmt"
-	"time"
 
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/config"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/dao"
 	"github.com/tencentcloud/CubeSandbox/cubelog"
-	"gorm.io/driver/mysql"
 	"gorm.io/gorm"
-	"gorm.io/gorm/logger"
 )
 
-func generateURL(user, password, addr, dbname string, connTimeout, readTimeout, writeTimeout int) string {
-	return fmt.Sprintf(
-		"%s:%s@tcp(%s)/%s?charset=utf8&parseTime=true&loc=Local&timeout=%ds&readTimeout=%ds&writeTimeout=%ds",
-		user, password, addr, dbname, connTimeout, readTimeout, writeTimeout)
-}
-
-// Init opens a *gorm.DB against the given DBConfig.
+// Init returns a *gorm.DB for the given DBConfig.
 //
-// Deprecated: prefer pkg/base/dao.Open / dao.Default. This shim is kept
-// so the v0.2.2 call-sites in templatecenter / nodemeta / instancecache /
+// Deprecated: prefer pkg/base/dao.Open / dao.Default directly. This shim is
+// kept so the v0.2.2 call-sites in templatecenter / nodemeta / instancecache /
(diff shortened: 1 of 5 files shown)


fslongjin:
hello，can you rebase it on master?

cookiebody:
Thanks for taking a look! While rebasing onto the latest `master` I found that #1000 (`fix(cubemaster): make PostgreSQL runtime paths work after dao migration`, merged 2026-07-17) already fixes the exact same three runtime paths this PR targets:

- `db.Init` now returns `dao.Default()` instead of opening a separate MySQL pool;
- artifact GC uses a driver-aware advisory lock (`pg_try_advisory_lock(hashtext(name))` on PostgreSQL);
- `UpsertReplica` runs the update on a fresh `*gorm.DB` chain to avoid the `UPDATE ... FROM ...` duplicate-table `SQLSTATE 42712` on PostgreSQL.

That makes this PR redundant, so I'm closing it — no rebase needed. I'd verified the same three fixes independently against an external PostgreSQL deployment (template distribution reaching READY, sandbox create/destroy), so I can confirm #1000's approach holds up end to end.

I'll follow up separately with the one-click **deployment** wiring for external PostgreSQL (installer flags, conf.yaml `driver`/`sslmode` patching, connectivity preflight), which #1000 doesn't cover, as part of #517. Thanks again!

PR #1084, opened 2026-07-22
hypervisor: fix virtio device livelock on snapshot restore by kicking queues and injecting interrupt on resume

lisongqian:
## Problem

After a VM is snapshotted and later restored, virtio devices can wedge in a livelock:

- A restored virtqueue may already contain pending descriptors at the moment the guest was snapshotted.
- On the host side, the queue eventfd signal is **not** persistent snapshot state — it is only re-emitted when the guest issues a fresh `kick`.
- After restore, cloud-hypervisor unparks the worker thread and then waits for a queue eventfd to fire. If the guest had already notified the queue before the snapshot, it won't notify again, so the worker sits idle.
- At the same time, the guest is waiting on an interrupt for the in-flight request to complete, so it never kicks.

Both sides end up waiting for each other. In practice this reproduces as:

- `virtio-blk` flush stalls during early boot after snapshot/restore.
- `virtio-net` traffic freezes after resume unless the old `driver_awake` workaround is used.

## Fix

Cherry-pick three upstream cloud-hypervisor patches that fix the resume-path kick properly in the shared `VirtioCommon` layer, and then retire the ad-hoc `driver_awake` workaround in `virtio-net`:

1. **`virtio-devices: signal activated queue eventfds on resume`**
   `VirtioCommon` now keeps clones of the activated queue eventfds and signals each of them once on resume, right after unparking the worker threads. This guarantees the worker re-checks the queue even if the guest already kicked before the snapshot.
   (cherry-picked from cloud-hypervisor `f7c3b775819f`)

2. **`virtio-devices: trigger interrupt into guest on resume`**
   Also inject a used-ring interrupt into the guest on resume, so the guest re-checks any request it thought was in flight instead of waiting forever for a notification that already happened before the snapshot.
   (cherry-picked from cloud-hypervisor `abf32187f342`)

3. **`virtio-devices: net: Remove "driver_awake" workaround for restore`**
   With the generic resume path above, virtio-net no longer needs its own device-specific `driver_awake` hack — the shared kick + interrupt already covers it.
   (cherry-picked from cloud-hypervisor `5d02dc7c1ddd`)

## Scope

- Files touched: `hypervisor/virtio-devices/src/device.rs`, `hypervisor/virtio-devices/src/net.rs`.
- No behavior change on the cold-boot / activate path — only affects the `resume()` path used by snapshot/restore.
- All three patches are straight cherry-picks from upstream cloud-hypervisor with proper `Signed-off-by` / `(cherry picked from commit ...)` trailers preserved.

## Testing

- Snapshot + restore a VM whose virtio-blk had a flush in flight at snapshot time — no longer stalls in early boot.
- virtio-net traffic resumes correctly after restore without the removed `driver_awake` workaround.

### DIFF
--- a/hypervisor/virtio-devices/src/device.rs
+++ b/hypervisor/virtio-devices/src/device.rs
@@ -19,6 +19,8 @@ use std::sync::{
     Arc, Barrier,
 };
 use std::thread;
+
+use anyhow::anyhow;
 use virtio_queue::Queue;
 use vm_memory::{GuestAddress, GuestMemoryAtomic, GuestUsize};
 use vm_migration::{MigratableError, Pausable};
@@ -202,6 +204,7 @@ pub struct VirtioCommon {
     pub paused_sync: Option<Arc<Barrier>>,
     pub epoll_threads: Option<Vec<thread::JoinHandle<()>>>,
     pub queue_sizes: Vec<u16>,
+    pub queue_evts: Vec<EventFd>,
     pub device_type: u32,
     pub min_queues: u16,
     pub access_platform: Option<Arc<dyn AccessPlatform>>,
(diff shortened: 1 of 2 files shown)


chenhengqi:
Just curious, where did these commit hashes come from?

For example, the third commit should be 72fc0976f113284ee229bc657926d3b712f74e6a.
https://github.com/cloud-hypervisor/cloud-hypervisor/commit/72fc0976f113284ee229bc657926d3b712f74e6a

lisongqian:
> Just curious, where did these commit hashes come from?
> 
> For example, the third commit should be 72fc0976f113284ee229bc657926d3b712f74e6a. [cloud-hypervisor/cloud-hypervisor@72fc097](https://github.com/cloud-hypervisor/cloud-hypervisor/commit/72fc0976f113284ee229bc657926d3b712f74e6a)

From this PR: https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8004

up2wing (approved):
LGTM

up2wing:
There is an CI case failed, PTAL.

lisongqian:
> There is an CI case failed, PTAL.

reruned.

PR #1087, opened 2026-07-22
docs(example): update browser example code and sync remote access guidance

luzhixing12345:
Expand the documentation regarding remote Cube invocation.

use `with` to automatically manage the sandbox lifecycle.

### DIFF
--- a/examples/browser-sandbox/README_zh.md
+++ b/examples/browser-sandbox/README_zh.md
@@ -28,12 +28,12 @@ CubeProxy ── https://<sandbox_id>-9000.<domain>/cdp?
 ## 3. 架构
 
 ```
-┌──────────────────────┐         ┌─────── Cube Sandbox ──────────────┐
+┌──────────────────────┐         ┌─────── Cube Sandbox ───────────────┐
 │                      │         │                                    │
-│  你的脚本             │  CDP WS │  ┌───────────────────────────┐    │
-│  (Playwright Python) │────────►│  │  Chromium --remote-debug  │    │
-│                      │  HTTPS  │  │  端口 9000                │    │
-│                      │         │  └───────────────────────────┘    │
+│  你的脚本             │  CDP WS │  ┌───────────────────────────┐     │
+│  (Playwright Python) │────────►│  │  Chromium --remote-debug  │     │
+│                      │  HTTPS  │  │  端口 9000                 │     │
+│                      │         │  └───────────────────────────┘     │
 └──────────────────────┘         │                                    │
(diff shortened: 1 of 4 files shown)


examples/browser-sandbox/browser.py
```diff
@@ -15,17 +15,16 @@
...
-    page = context.new_page()
-    page.goto("http://www.tencent.com", wait_until="domcontentloaded")
-    print(page.title())
+    cdp_url = f"https://{sandbox.get_host(9000)}/cdp?"
+    # use playwright to operate the browser through cdp_url
+    with sync_playwright() as playwright:
+        browser = playwright.chromium.connect_over_cdp(
+            cdp_url,
+        )
+        context = browser.new_context(ignore_https_errors=True)
```
chenhengqi, line 27: With `NODE_EXTRA_CA_CERTS` in place, why do we need to `ignore_https_errors` here?

chenhengqi: I see, this is for `page.goto()` below.

examples/browser-sandbox/README_zh.md
```diff
@@ -122,7 +124,13 @@ with sync_playwright() as playwright:
 | 3 | `connect_over_cdp(cdp_url)` | Playwright 接入已运行的 Chromium 进程 |
 | 4 | `page.goto(...)` | 完整 Playwright API：导航、点击、截图、抓取等 |
 
-## 7. 进阶用法
+## 7. 远程访问
+
+如需远程访问，需要在远程机器上设置 cube.app 的通配 dns 解析到 CubeProxy IP
```
chenhengqi, line 129: Can we use `--host-resolver-rules`?

https://medium.com/@hoppjerka/playwright-resolve-host-without-hostfile-manipulation-84c9e284b962

luzhixing12345: I just tried this method but it seems not work.

There are two separate DNS lookups:

1. Playwright resolves the CDP endpoint:
     `9000-<sandbox-id>.cube.app`

2. Remote Chromium resolves the page URL:
     www.tencent.com

--host-resolver-rules only affects the second lookup. The first happens in Playwright’s Node.js driver before
  it connects to Chromium.

connect_over_cdp() (https://playwright.dev/python/docs/api/class-browsertype#browser-type-connect-over-cdp) provides no proxy or custom DNS option. Therefore, the CubeProxy hostname must be resolvable in the
environment running Playwright.

luzhixing12345:
Does this PR require any further modifications or additional explanations?

PR #1090, opened 2026-07-22
feat(metrics): expose guest workload metrics through Task.Stats

zyl1121:
## Motivation

CubeSandbox does not currently expose guest workload CPU and memory accounting through the standard containerd task metrics contract. CubeShim does not implement `Task.Stats`, so host-side callers cannot retrieve the existing cube-agent `StatsContainer` data through containerd.

This PR adds the minimal host-initiated transport seam required by later Cubelet resource collection:

```text
containerd Task.Stats -> CubeShim -> cube-agent StatsContainer
```

The sandbox does not push metrics to the host.

## PR Series

This work is split into two PRs so the guest runtime transport and Cubelet-owned metrics semantics can be reviewed independently:

* #1090 (this PR): guest workload cgroup accounting, capability negotiation, and standard `Task.Stats` transport.
* #1122: lifecycle-aware guest and host accounting, bounded sampling, caching, Prometheus export, and operator documentation.

#1122 depends on #1090 and should be merged after it.

## What Changed

* Add strict guest workload CPU and memory accounting to `cube-agent StatsContainer` for the unified cgroup v2 layout.
* Keep the workload accounting cgroup process-free and place runtime processes in a child leaf so delegated envd cgroups remain descendants of the measured workload.
* Bind the sandbox-visible cgroup v2 mount to the workload subtree so envd-created cgroups are included in workload accounting.
* Add an internal resource-metrics capability version; incomplete or older guest layouts fail closed.
* Implement CubeShim `Task.Stats` by calling `StatsContainer` for the requested container and encoding the result as the canonical containerd cgroup metrics type.

## Contract

The current single-container sandbox model uses `container_id == sandbox_id`, while the transport keeps both identifiers explicit for later multi-container discovery.

CPU total/user/system/throttling and memory current/limit/failures are guest-kernel workload cgroup values. This PR transports raw cumulative counters and point-in-time gauges; it does not define lifecycle baselines or Prometheus semantics.

Guests without resource-metrics capability version `1` return `FailedPrecondition` instead of an all-zero or partial metric.

## Scope

This PR intentionally does not add lifecycle epochs, counter baselines, host sandbox accounting, background sampling, cache, Prometheus export, one-click packaging changes, or operator documentation. Those remain in the dependent Cubelet PR.

## Validation

* CubeShim's complete package test suite passes 43 tests, including guest capability validation, CPU/memory conversion, canonical type URL encoding, missing-field rejection, and generated-code stability.
* cube-agent/rustjail tests pass for cgroup v2 CPU and memory parsing, delegated controllers, descendant PID collection, scoped cgroup mounts, and cleanup.
* CubeShim's restored container-ID mapping test verifies that containerd task identity is translated to the guest workload identity before `StatsContainer` is called.
* A standard one-click bundle installed CubeShim and cube-agent from the exact combined commit. The reusable node-local `tests/e2e/resource_metrics/guest_workload_task_stats.sh` case passed on a fresh sandbox with nanosecond CPU counters and non-zero memory current/limit. Cluster runners without explicit node-local `CUBECLI` and `CUBELET_ADDRESS` settings skip this check successfully; SDK/API lifecycle E2E remains in #1122.
* The combined CubeAPI/CubeProxy lifecycle E2E exercised this transport through Cubelet, CubeShim, cube-agent, and the guest workload cgroup. CPU/memory load, snapshot/rollback, pause/resume, delete, and complete resource cleanup passed.

### DIFF
--- a/CubeShim/protoc/src/agent.rs
+++ b/CubeShim/protoc/src/agent.rs
@@ -6458,6 +6458,7 @@ pub struct StatsContainerResponse {
     // message fields
     pub cgroup_stats: ::protobuf::SingularPtrField<CgroupStats>,
     pub network_stats: ::protobuf::RepeatedField<NetworkStats>,
+    pub resource_metrics_version: u32,
     // special fields
     #[cfg_attr(feature = "with-serde", serde(skip))]
     pub unknown_fields: ::protobuf::UnknownFields,
@@ -6533,6 +6534,18 @@ impl StatsContainerResponse {
     pub fn take_network_stats(&mut self) -> ::protobuf::RepeatedField<NetworkStats> {
         ::std::mem::replace(&mut self.network_stats, ::protobuf::RepeatedField::new())
     }
+
+    pub fn get_resource_metrics_version(&self) -> u32 {
+        self.resource_metrics_version
+    }
+    pub fn clear_resource_metrics_version(&mut self) {
+        self.resource_metrics_version = 0;
+    }
+
+    // Param is passed by value, moved
+    pub fn set_resource_metrics_version(&mut self, v: u32) {
+        self.resource_metrics_version = v;
+    }
 }
 
 impl ::protobuf::Message for StatsContainerResponse {
@@ -6560,6 +6573,13 @@ impl ::protobuf::Message for StatsContainerResponse {
                 2 => {
                     ::protobuf::rt::read_repeated_message_into(wire_type, is, &mut self.network_stats)?;
                 },
+                3 => {
+                    if wire_type != ::protobuf::wire_format::WireTypeVarint {
+                        return ::std::result::Result::Err(::protobuf::rt::unexpected_wire_type(wire_type));
+                    }
+                    let tmp = is.read_uint32()?;
+                    self.resource_metrics_version = tmp;
+                },
                 _ => {
                     ::protobuf::rt::read_unknown_or_skip_group(field_number, wire_type, is, self.mut_unknown_fields())?;
                 },
@@ -6580,6 +6600,9 @@ impl ::protobuf::Message for StatsContainerResponse {
             let len = value.compute_size();
             my_size += 1 + ::protobuf::rt::compute_raw_varint32_size(len) + len;
         };
+        if self.resource_metrics_version != 0 {
+            my_size += ::protobuf::rt::value_size(3, self.resource_metrics_version, ::protobuf::wire_format::WireTypeVarint);
+        }
         my_size += ::protobuf::rt::unknown_fields_size(self.get_unknown_fields());
         self.cached_size.set(my_size);
         my_size
@@ -6596,6 +6619,9 @@ impl ::protobuf::Message for StatsContainerResponse {
             os.write_raw_varint32(v.get_cached_size())?;
             v.write_to_with_cached_sizes(os)?;
         };
+        if self.resource_metrics_version != 0 {
+            os.write_uint32(3, self.resource_metrics_version)?;
+        }
         os.write_unknown_fields(self.get_unknown_fields())?;
         ::std::result::Result::Ok(())
     }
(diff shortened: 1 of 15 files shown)


chenhengqi:
The changeset is way too large. Can this be separated into small PRs or at least split it into separate commits for each component to ease the review process?

zyl1121:
> The changeset is way too large. Can this be separated into small PRs or at least split it into separate commits for each component to ease the review process?

I’ve split the original submission into two PRs for easier review:

- #1090 contains the guest workload cgroup accounting, capability negotiation, and standard `Task.Stats` transport.
- #1122 contains the Cubelet lifecycle-aware guest and host accounting, bounded sampling and caching, Prometheus export, and documentation.

#1122 depends on #1090, so the intended merge order is #1090 followed by #1122. The combined changes have been validated through a one-click build and lifecycle E2E.

ls-ggg:
Let's continue this discussion in #1122 and #1090

zyl1121:
> Let's continue this discussion in #1122 and #1090

Just to confirm: #1122 depends on the `Task.Stats` transport introduced in #1090. Your comment suggests continuing the discussion in both #1122 and #1090, but #1090 was closed at the same time.

Was #1090 closed accidentally, or would you prefer the changes to be handled another way?

ls-ggg:
> > Let's continue this discussion in #1122 and #1090
> 
> Just to confirm: #1122 depends on the `Task.Stats` transport introduced in #1090. Your comment suggests continuing the discussion in both #1122 and #1090, but #1090 was closed at the same time.
> 
> Was #1090 closed accidentally, or would you prefer the changes to be handled another way?
Sorry，please rebase and add e2e results.

CubeShim/protoc/src/agent.rs
```diff
@@ -6533,6 +6056,21 @@ impl StatsContainerResponse {
     pub fn take_network_stats(&mut self) -> ::protobuf::RepeatedField<NetworkStats> {
         ::std::mem::replace(&mut self.network_stats, ::protobuf::RepeatedField::new())
     }
+
+    // uint32 resource_metrics_version = 3;
```
chenhengqi: Weird comment and extra newlines.

zyl1121:
Rebased onto the latest `master` and added reusable E2E coverage under `tests/e2e/resource_metrics/`.

The direct `Task.Stats` test creates workload activity in a fresh sandbox and verifies that the host can retrieve non-zero CPU and memory metrics through the standard containerd task interface.

The test passed on a standard one-click deployment. The complete lifecycle E2E also exercised the same host-initiated path through CPU and memory load, snapshot/rollback, pause/resume, deletion, and cleanup. This confirms that `Task.Stats -> CubeShim -> cube-agent StatsContainer` returns valid guest workload metrics during normal sandbox operation and after restore.

CubeShim/protoc/src/agent.rs
zyl1121: Fixed at the generation source. The locked rust-protobuf generator emits this scalar-field declaration comment and excess blank lines, so `CubeShim/protoc/build.rs` now normalizes that generated fragment and ensures the file ends with a newline. `cargo test --locked` passes all 43 tests and leaves `CubeShim/protoc/src/agent.rs` unchanged.

ls-ggg:
Note that cubecli is deployed on nodes, so when running e2e against the cluster build, it can't reach the cubelet directly. Please add a special case in the e2e script to just pass the test on the cluster build. Alternatively, you could add these SDK/API-initiated e2e tests in #1122 instead

zyl1121:
Note that cubecli is deployed on nodes, so when running e2e against the cluster build, it can't reach the cubelet directly. Please add a special case in the e2e script to just pass the test on the cluster build. Alternatively, you could add these SDK/API-initiated e2e tests in 1122

Updated the direct `Task.Stats` E2E for the cluster build.

The check is node-local and now runs only when both `CUBECLI` and `CUBELET_ADDRESS` are explicitly provided. Cluster runners without direct Cubelet access report `SKIP` and exit successfully, while invalid explicit configuration and invalid metrics still fail the test.

The SDK/API-initiated lifecycle E2E suggested above is already included in #1122. It creates sandboxes through CubeAPI, runs CPU and memory workloads through CubeProxy/envd, and verifies `/v1/metrics/resource` across fresh collection, snapshot/rollback, pause/resume, and deletion.

I also added a shell test for the node-local skip and failure paths. On a deployed node, the opt-in path successfully validated non-zero CPU and memory through `Task.Stats`; the temporary sandbox, template, and task were cleaned up afterward.

ls-ggg:
LGTM

ls-ggg:
@chenhengqi @fslongjin PTAL

chenhengqi (approved):

PR #1098, opened 2026-07-23
feat(volume): support read-only volume attachments

zyl1121:
## Motivation

Some workloads need to share the same Volume across sandboxes while preventing selected sandboxes from modifying it. Typical examples include shared datasets, model artifacts, reference outputs, or baseline files that should be readable by many sandboxes but writable only by designated writers.

This is naturally a per-attachment access mode: the same Volume may be attached read-write in one sandbox and read-only in another.

The official e2b SDK's Volume mount API does not provide a read-only option, so the existing e2b-compatible `volume_mounts={path: volume}` shape can only express read-write mounts. This PR adds read-only mounts as an opt-in Cube extension while preserving that existing default behavior.

Cube already has the lower-level support needed for this in the plugin-volume path and Cubelet's bind-mount / virtiofs read-only enforcement. This PR threads the access mode through the existing flow instead of changing the Volume model itself.

## What Changed

* Add optional `volumeMounts[].readOnly` support to CubeAPI and forward it through CubeMaster's existing plugin-volume mount path.
* Add the Python SDK `VolumeMount` wrapper as a Cube extension.
* Keep plain `Volume`, `VolumeInfo`, and volume ID values read-write by default.
* Omit `readOnly` when `read_only=False` to preserve the existing request shape.
* Keep the existing restriction that a sandbox cannot attach the same Volume more than once.
* Use Cubelet's existing bind-mount and virtiofs read-only enforcement.
* Update the Python SDK reference and the English and Chinese Volume Plugin guides.
* Add API, forwarding, serialization, multi-container, and opt-in `sdk_compat` data-plane coverage. Volume Plugin cases remain skipped unless the deployment explicitly enables them with `SDK_E2E_VOLUME_PLUGIN=true`.

## Example

```python
from cubesandbox import Sandbox, VolumeMount

writer = Sandbox.create(
    volume_mounts={"/dataset": dataset_volume},
)

reader = Sandbox.create(
    volume_mounts={
        "/dataset": VolumeMount(dataset_volume, read_only=True),
    },
)
```

## Validation

* Ran the CubeAPI test suite.
* Ran the CubeMaster plugin-volume forwarding tests.
* Ran the Python SDK Volume tests.
* Ran the `sdk_compat` Volume suite against a COS-backed Volume.
* Verified that the same Volume could be attached read-write and read-only to concurrent sandboxes.
* Verified that the read-only attachment could read existing data but could not create, write, rename, or delete files.
* Verified that the read-write attachment remained writable while the read-only attachment was active.

## Scope

This PR only adds an access mode to individual sandbox Volume attachments.

It does not change:

* the Volume database schema
* the plugin protocol
* create or destroy lifecycle behavior
* refcount behavior
* the default read-write behavior of existing mounts

A read-only attachment is not an immutable snapshot and may observe changes made through another read-write attachment.

### DIFF
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -156,6 +156,14 @@ impl Default for SandboxOnTimeout {
 pub struct SandboxVolumeMount {
     pub name: String,
     pub path: String,
+    /// CubeSandbox extension: mount this volume read-only for this sandbox
+    /// attachment. Defaults to false when omitted.
+    #[serde(rename = "readOnly", default, skip_serializing_if = "is_false")]
+    pub read_only: bool,
+}
+
+fn is_false(value: &bool) -> bool {
+    !*value
 }
 
 // ─── Sandbox — create request ──────────────────────────────────────────────
@@ -604,6 +612,36 @@ mod tests {
                 .and_then(|envs| envs.get("CUBE_TEST_ENV"))
                 .map(String::as_str),
             Some("value")
+        );
+    }
+
+    #[test]
+    fn new_sandbox_volume_mount_read_only_is_optional() {
+        let req: NewSandbox = serde_json::from_value(serde_json::json!({
+            "templateID": "tpl-1",
+            "volumeMounts": [
+                {"name": "dataset", "path": "/data", "readOnly": true},
+                {"name": "workspace", "path": "/workspace"},
+                {"name": "cache", "path": "/cache", "readOnly": false}
+            ]
+        }))
+        .expect("volume mounts should deserialize");
+
+        let mounts = req.volume_mounts.expect("volume mounts");
+        assert!(mounts[0].read_only);
+        assert!(!mounts[1].read_only);
+        assert!(!mounts[2].read_only);
+        assert_eq!(
+            serde_json::to_value(&mounts[0]).expect("serialize read-only mount"),
+            serde_json::json!({"name": "dataset", "path": "/data", "readOnly": true})
+        );
+        assert_eq!(
+            serde_json::to_value(&mounts[1]).expect("serialize read-write mount"),
+            serde_json::json!({"name": "workspace", "path": "/workspace"})
+        );
+        assert_eq!(
+            serde_json::to_value(&mounts[2]).expect("serialize explicit read-write mount"),
+            serde_json::json!({"name": "cache", "path": "/cache"})
         );
     }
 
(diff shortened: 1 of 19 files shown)


ls-ggg:
LGTM

ls-ggg:
Good job， We introduced the e2e test framework in v0.6.0—could you add the corresponding e2e test cases under tests/e2e/sdk_compat

ls-ggg:
> Good job， We introduced the e2e test framework in v0.6.0—could you add the corresponding e2e test cases under tests/e2e/sdk_compat

Also: if plugins aren't configured, the test should just pass

zyl1121:
> Good job, We introduced the e2e test framework in v0.6.0—could you add the corresponding e2e test cases under tests/e2e/sdk_compat
> Also: if plugins aren't configured, the test should just pass

Added. The read-only Volume data-plane case is now in `tests/e2e/sdk_compat/cases/volume/test_read_only.py`. It verifies concurrent read-write/read-only attachments, read access through the read-only attachment, rejected create/write/rename/delete operations, and that the writer remains writable.

The Volume cases remain opt-in via `SDK_E2E_VOLUME_PLUGIN=true`. When the plugin is not configured, they are skipped cleanly and the test run passes.

I also verified the updated suite against COS: the focused read-only case passed, and the full `cases/volume` suite passed with 3 tests.

lkml-likexu:
@zyl1121 @ls-ggg More test required: Pls cover the scenario where the same Volume is mounted RW in one sandbox and RO in another sandbox concurrently is only verified manually in the commit. It is recommended to add boundary tests at the SDK layer for _serialize_volume_mounts covering "duplicate same-name volume deduplication × readOnly mixing".

zyl1121:
> More test required: Pls cover the scenario where the same Volume is mounted RW in one sandbox and RO in another sandbox concurrently is only verified manually in the commit. It is recommended to add boundary tests at the SDK layer for _serialize_volume_mounts covering "duplicate same-name volume deduplication × readOnly mixing".

Thanks for the suggestion. I added follow-up SDK boundary coverage in #1224.

The concurrent read-write/read-only scenario for the same `Volume` is already covered by `tests/e2e/sdk_compat/cases/volume/test_read_only.py`. That test keeps both sandboxes active, verifies that the read-only mount rejects mutations, and confirms that the read-write mount remains writable. I also reran this case against COS successfully.

#1224 adds direct `_serialize_volume_mounts` coverage for using the same `Volume` across separate sandbox creation requests with different `readOnly` settings.

CubeAPI currently rejects mounting the same `Volume` multiple times within a single sandbox, so #1224 stays focused on the supported case.

PR #1114, opened 2026-07-23
test(integration): lower MemTotal assertion threshold from 480M to 450M

aichitudou123:
The guest kernel grew larger after enabling CONFIG_VFAT_FS for cloud-init seed disk support, causing MemTotal on 512MB VMs to drop to ~469M, just below the original 480,000 KB assertion threshold. Lower the threshold to 450,000 KB to leave a ~20M buffer for future kernel size variations.


lisongqian:
Thank you for your first contribution! The reason for the memory increase in guest is not only the CONFIG_VFAT_FS config. Please update your commit message.

lisongqian (approved):
LGTM

PR #1115, opened 2026-07-23
vmm: probe soft-dirty support with real pagemap round-trip

aichitudou123:
The original probe_soft_dirty_support() simply wrote "4" to /proc/self/clear_refs to determine whether the kernel supports the soft-dirty mechanism. However, on kernels without CONFIG_MEM_SOFT_DIRTY (ARM64, x86 with the option disabled), clear_refs_write() silently no-ops the write and returns success, causing the probe to incorrectly return true. This led three soft-dirty-dependent unit tests to enter the test body on unsupported kernels and panic because bit 55 is always 0.

Replace the probe with a real pagemap round-trip:
1. mmap one page of anonymous memory, then write one byte to trigger VM_SOFTDIRTY to set bit 55 in the PTE.
2. Read /proc/self/pagemap to verify bit 55 is actually set.
   - If bit 55 is not 1, treat the kernel as unsupported and return false directly (skip clear_soft_dirty to avoid unnecessary mmap_lock write-lock PTE traversal on incapable machines).
   - If bit 55 is 1, call clear_soft_dirty once to arm the tracker and return true.

### DIFF
--- a/hypervisor/vmm/src/soft_dirty.rs
+++ b/hypervisor/vmm/src/soft_dirty.rs
@@ -29,7 +29,7 @@ use std::fs::{File, OpenOptions};
 use std::io::{self, Read, Seek, SeekFrom, Write};
 use std::time::Instant;
 use thiserror::Error;
-use vm_memory::{GuestAddress, GuestMemory, GuestMemoryMmap};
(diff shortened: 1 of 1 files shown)


lisongqian:
Please fix the error in the code format check

lisongqian (approved):
LGTM

PR #1116, opened 2026-07-23
test(sdk_compat): cover nested RO parent + RW child host-mount

ls-ggg:
## Summary
- Add `test_nested_ro_parent_rw_child_writable_when_child_listed_first` under `tests/e2e/sdk_compat/cases/host-mount/` to lock the nested RO parent + RW child host-mount layout (child descriptor listed first).
- **Host dirs:** provisioned via `provision_host_dirs` (throwaway sandbox `mkdir -p`), so single-node e2e needs no manual prep.
- **Multi-node / missing source:** if create fails because the scheduled node lacks the `hostPath` (Cubelet `bind mount ...: No such file or directory`), the test **passes with a warning**. Shared storage (or the same layout on every node) is required to exercise the nested assertion in multi-node CI.

## Test plan
- [ ] Single-node: run the nested host-mount case without pre-creating dirs; expect real assertions (not warn-pass)
- [ ] Confirm child-first descriptors still keep RW child writable and RO parent read-only

Assisted-by: Cursor:Composer

### DIFF
--- a/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py
+++ b/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py
@@ -18,6 +18,11 @@ can prove:
 - real read-write host sharing: two sandboxes bind-mounting the same host dir
   observe each other's writes (proving it is a genuine host mount, not a
   per-sandbox overlay);
+- nested RO parent + RW child: child remains writable even when the child
+  descriptor is listed before the parent (regression for Cubelet mount-order
+  determinism; source PR #946 / issue #944). Provisions host dirs via a
+  throwaway sandbox; a missing hostPath source on multi-node local-disk
+  clusters is treated as pass-with-warning;
 - the symlink TOCTOU escape: CubeMaster validates the path lexically while
   Cubelet resolves symlinks before mounting, so a symlink planted under the
   allowed prefix can redirect the mount to an arbitrary host path.
@@ -39,6 +44,7 @@ from framework.host_mount import (
     provision_host_dirs,
     skip_if_host_mount_unavailable,
     under_prefix,
+    warn_pass_missing_hostpath_source,
 )
 from framework.lifecycle import managed_control_sandbox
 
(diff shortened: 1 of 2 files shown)


tests/e2e/sdk_compat/cases/host-mount/test_boundary.py
```diff
@@ -348,6 +354,173 @@ def test_rw_mount_shares_data_across_sandboxes(sdk_backend, sdk_e2e_config):
...
 
+# --- Nested host mounts (RO parent + RW child) --------------------------------
+#
+# Regression for https://github.com/TencentCloud/CubeSandbox/pull/946
+# (issue #944): when a read-only parent mountPath contains a writable nested
+# child, Cubelet must apply the parent before the child so the child is not
+# hidden by a later RO parent remount. Descriptor input order must not matter.
+#
+# Topology: default **single-node**. Host dirs must already exist on the
+# Cubelet (see test docstring). Do not mkdir via a throwaway sandbox — that
```
lkml-likexu: This limitation is unexpected.

In a single-node environment, this still requires the test user to manually create the `data/shared/nested-e2e/members/agent-a` directory; otherwise, an error like this will occur.

```
cases/host-mount/test_boundary.py::test_nested_ro_parent_rw_child_writable_when_child_listed_first[cubesandbox]
  /tmp/tmp.IVNQa1N5h4/CubeSandbox/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py:515: UserWarning: host-mount host_mount_nested_seed: hostPath source missing on the scheduled node (['/data/shared/nested-e2e', '/data/shared/nested-e2e/members/agent-a']); treating as pass under the single-node default. Pre-create the dirs on each Cubelet (or use shared storage) to exercise the assertion. original error: CubeMaster returned error code 130545: prepareHostDirVolume: bind mount /data/shared/nested-e2e -> /data/cubelet/hostdir/08afa278b7aa4c85aea3e7452d7f8b27/rw/hostdir-0: exit status 32: mount: /data/cubelet/hostdir/08afa278b7aa4c85aea3e7452d7f8b27/rw/hostdir-0: special device /data/shared/nested-e2e does not exist.
         dmesg(1) may have more information after failed mount system call.
  
    if warn_pass_missing_hostpath_source(

-- Docs: https://docs.pytest.org/en/stable/how-to/capture-warnings.html
```

As an e2e test target, test users should not have to worry about whether the underlying environment is multi-node or single-node (in other words, the same e2e test should pass in both cases). Clearly, support for `host-mount` in multi-node environments should be better documented rather than restricting test scenarios and requiring users to take specific actions.

```
diff --git a/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py b/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py
index 7270c33..75edeac 100644
--- a/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py
+++ b/tests/e2e/sdk_compat/cases/host-mount/test_boundary.py
@@ -361,13 +361,16 @@ def test_rw_mount_shares_data_across_sandboxes(sdk_backend, sdk_e2e_config):
 # child, Cubelet must apply the parent before the child so the child is not
 # hidden by a later RO parent remount. Descriptor input order must not matter.
 #
-# Topology: default **single-node**. Host dirs must already exist on the
-# Cubelet (see test docstring). Do not mkdir via a throwaway sandbox — that
-# can create the source on node A while a later create schedules on node B.
+# Topology: default **single-node**. The nested host dirs are provisioned via a
+# throwaway sandbox (``provision_host_dirs``) before the seed/agent creates. On
+# a single-node cluster this materializes the source on the one Cubelet; on a
+# multi-node local-disk cluster the provision may land on a different node than a
+# later create, so a missing source there is still treated as a warned pass.
 
 
-# Fixed layout under the allowed prefix (prepare on the node before e2e):
-#   mkdir -p <prefix>/nested-e2e/members/agent-a
+# Fixed layout under the allowed prefix. Provisioned per test via
+# ``provision_host_dirs`` (mkdir -p also creates the parent ``nested-e2e``):
+#   <prefix>/nested-e2e/members/agent-a
 _NESTED_TEAM_REL = "nested-e2e"
 _NESTED_MEMBER_REL = "nested-e2e/members/agent-a"
 
@@ -390,23 +393,20 @@ def test_nested_ro_parent_rw_child_writable_when_child_listed_first(
 
     Single-node default
     -------------------
-    This case assumes a single Cubelet node where the host paths already exist.
-    It does **not** call ``provision_host_dirs`` (no throwaway sandbox mkdir):
-    host-mount sources are node-local, and a provision sandbox is not pinned to
-    the same node as a later create.
-
-    Prepare once on the Cubelet host (prefix defaults to ``/data/shared``)::
-
-        mkdir -p /data/shared/nested-e2e/members/agent-a
+    This case assumes a single Cubelet node. It provisions the nested layout via
+    ``provision_host_dirs`` (a throwaway sandbox that ``mkdir -p``s the source),
+    so no manual host prep is required. ``mkdir -p .../members/agent-a`` also
+    creates the ``nested-e2e`` parent, giving both mount sources.
 
     Multi-node / missing source
     ---------------------------
-    On a cluster with node-local disks, create may fail with Cubelet's
-    ``bind mount ...: No such file or directory`` when the scheduled node lacks
-    the source. That is treated as a **pass with a warning** — topology and
-    shared-storage setup are outside this regression's scope. Use shared
-    storage at the same path on every node (see persistent-storage.md) if the
-    nested assertion must run in multi-node CI.
+    On a cluster with node-local disks, ``provision_host_dirs`` may create the
+    source on one node while a later create is scheduled on another. When the
+    scheduled node lacks the source, create fails with Cubelet's
+    ``bind mount ...: No such file or directory``. That is treated as a **pass
+    with a warning** — topology and shared-storage setup are outside this
+    regression's scope. Use shared storage at the same path on every node (see
+    persistent-storage.md) if the nested assertion must run in multi-node CI.
     """
     token = uuid.uuid4().hex[:12]
     team_host = under_prefix(_NESTED_TEAM_REL)
@@ -417,6 +417,10 @@ def test_nested_ro_parent_rw_child_writable_when_child_listed_first(
     seed_marker = f"shared-input-{token}.txt"
     child_note = f"note-{token}.txt"
 
+    # Provision the nested layout on the host. ``mkdir -p .../members/agent-a``
+    # also creates the ``nested-e2e`` parent, so both mount sources exist.
+    provision_host_dirs(sdk_backend, sdk_e2e_config, [_NESTED_MEMBER_REL])
+
     seed_mount = "/mnt/seed"
     seed_metadata = host_mount_metadata(
         [mount_option(team_host, seed_mount, read_only=False)]

```

ls-ggg: Thanks — addressed in the latest commit: the nested case calls `provision_host_dirs` again so single-node e2e needs no manual `mkdir`. Multi-node missing-source still warn-passes as discussed.

PR #1052, opened 2026-07-21
test(e2e): migrate template alias tests to sdk_compat framework

sparkzky:
## Summary

Migrates template alias e2e tests from the standalone `sdk/python/tests/test_template_e2e.py` into the `tests/e2e/sdk_compat/` framework introduced by #835.

Follow-up to #749 (template alias feature, merged).

## Changes

| File | Change |
|---|---|
| `sdk/python/tests/test_template_e2e.py` | **Deleted** — old standalone test file |
| `tests/e2e/sdk_compat/cases/templates/test_alias.py` | **New** — 6 tests using CubeSandbox SDK |
| `tests/e2e/sdk_compat/cases/templates/__init__.py` | **New** — package init |
| `tests/e2e/sdk_compat/pytest.ini` | Added `templates` marker |

No changes to `framework/config.py`. Container image (`sandbox-code:latest`) and writable layer size (`1G`) are hardcoded as module constants in the test file — no env vars needed.

## Design decisions

- **CubeSandbox SDK only** (no dual-backend parameterization): E2B SDK's template API uses a completely different endpoint format (`/v3/templates` with Dockerfile) that CubeAPI does not implement. E2B SDK compatibility is sandbox-operations-only. Template alias is a CubeSandbox feature inspired by E2B concepts, not an E2B SDK wire-compatible implementation.
- **No `CUBE_TEMPLATE_E2E_IMAGE` / `CUBE_TEMPLATE_E2E_WRITABLE_LAYER_SIZE` env vars**: hardcoded defaults in the test module. Only `CUBE_TEMPLATE_ID` (already required by the framework) is needed.

## Tests

| Test | What it covers |
|---|---|
| `test_template_list_and_get_existing` | List templates + GET by ID |
| `test_template_create_from_image_and_cleanup` | Create from image + cleanup |
| `test_template_alias_dedicated_endpoint_rejects_invalid` | Invalid alias → 400 (no template creation) |
| `test_template_alias_create_get_and_delete` | Full alias lifecycle: create → get by alias → delete by alias → 404 |
| `test_template_alias_dedicated_lookup_endpoint` | `GET /templates/aliases/:alias` endpoint |
| `test_template_alias_rebuild_reassignment` | Rebuild with same alias → alias moves to newly READY template |

Verified against a live one-click deployment: **6 passed in 55s**.

Assisted-by: OhMyPi:zai/glm-5.2

PR #1120, opened 2026-07-23
feat: set alias on an existing template

sparkzky:
### Summary

Allow setting, reassigning, or clearing the alias of an **already-existing** template via a new `PUT /templates/{templateID}/alias` endpoint, closing the gap left by #749 (alias could only be attached at create time). The endpoint accepts a template ID **or** the template's current alias as the path parameter and returns the updated template detail (same shape as `GET /templates/:id`).

### API contract

```
PUT /templates/{templateID}/alias
Content-Type: application/json
{ "alias": "<string>" }        // empty / absent / null → clear
→ 200 TemplateDetail
```

| Condition | HTTP |
|---|---|
| Invalid alias format, or target is a snapshot | **400** |
| Template not found, or `DELETING` | **404** |
| Target not `READY` (claim), or concurrent duplicate claim | **409** |
| Anything else | **500** |

- **Claim** requires the template to be `READY`; **clear** is allowed on any non-`DELETING` template (including `FAILED`), so a stuck alias can always be released.
- **Reassign semantics**: setting an alias currently held by another template transfers it (release + claim), consistent with #749's last-commit-wins behavior. This transfer happens **only** through this operator endpoint — background job finalize never steals.
- No DB migration: the existing `alias_key` STORED generated column recomputes on UPDATE.

Alias validation lives **only** in CubeMaster (`^[a-z0-9][a-z0-9-]{0,63}$`, no `tpl-`/`snap-` prefix including the bare prefixes). CubeAPI and the SDKs forward without re-validating, so the rule cannot drift between layers.

### Concurrency invariants (the core of the review rounds)

1. **Transactional set-alias**: `SetTemplateAlias` runs in a single transaction holding the definition row `FOR UPDATE` — the `display_name` change and the CREATE/REDO image-job `RequestJSON` sync (for both the target and the old alias holder) either all commit or all roll back. A sync failure aborts the request; the API can never return 200 with a stale redo blob that would later revert the operator's change. Deadlocks (MySQL `1205`/`1213`, PostgreSQL `40P01`/`55P03`) are retried once.
2. **Create/redo never steals**: at build finalize, an alias collision hits the `alias_key` unique index and the template simply finishes `READY` without the alias. Only the operator path above transfers aliases between templates.
3. **Finalize alias resolution** (`aliasToClaimAtFinalize`): a failure reading the job's stored `RequestJSON` skips the claim entirely (no fallback to the frozen submit-time alias); a non-empty `display_name` set by an operator during an in-flight build wins over the job's alias.
4. **Redo source selection**: `SubmitRedoTemplateFromImage` now picks the latest CREATE/REDO job as the request source and strips the alias when falling back to a non-CREATE/REDO job, so redoing a commit-origin template can't resurrect a foreign alias.

### Changes by layer

- **CubeMaster (Go)**
  - Store: `SetTemplateAlias` + transactional helpers (`setTemplateAliasLocked`, `claimTemplateAliasTx`, `syncCreateRedoImageJobAliasTx`); new sentinels `ErrInvalidAlias`, `ErrAliasNotApplicableToSnapshot`, `ErrTemplateNotReady`; exported `IsDuplicateAliasError`.
  - HTTP handler `PUT /cube/template/:id/alias` with the error → `ret_code` mapping above.
  - CLI: `cubemastercli tpl set-alias <id> --alias <a>` (or `--clear` / empty `--alias`).
- **CubeAPI (Rust)**: forwards the request, maps `ret_code` → HTTP via the existing `map_err`; registers the path in `openapi.rs`; `openapi.yml` hand-appended (regenerating via utoipa would drop unregistered served paths — known project-wide debt).
- **SDKs (Go / Node / Python)**: thin wrappers `SetTemplateAlias` / `setAlias` / `set_alias` returning the updated `TemplateInfo`.
- **e2e**: `tests/e2e/sdk_compat/cases/templates/test_alias.py` — 2 new cases (set on an existing template; reassign between templates) alongside #749's 6.

### Tests

- Docker-backed MySQL & PostgreSQL suites (dockertest; auto-skip without Docker, fail in CI) covering: concurrent claim mutex behavior, create/redo no-steal, `FAILED`-holder occupancy, job `RequestJSON` sync, and rollback-on-sync-failure.
- Go unit tests across store / handler / CLI (~44 alias cases in `store_test.go` + `template_test.go`), plus request-validation regression tests for bare `tpl-`/`snap-` prefixes.
- 2 unit tests per SDK (Go/Node/Python), asserting the returned `name`.
- e2e verified against a live one-click deployment.

### Compatibility note

The tightened alias rule (rejecting bare `tpl-`/`snap-` prefixes) also applies to the shared create/redo validation path. A template created before this change with such an alias would fail redo validation with 400 — to be noted in the release changelog.


lkml-likexu:
@sparkzky Thanks for your contribution. Please add the corresponding tests to the `tests/e2e/sdk_compat` directory.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1013,6 +1021,48 @@ func claimAliasAfterReady(ctx context.Context, templateID, alias string) (claimW
...
+//	alias == ""  → clear: UPDATE display_name = '' WHERE template_id = ?
+//	alias != ""  → claim: reuse claimTemplateAlias (release from any other
+//	                    holder inside the same transaction, then UPDATE on the
+//	                    target template).
+//
+// The target must be a non-deleting template. Snapshots are rejected because
+// their display_name is an informational label, not a unique alias (their
+// alias_key is always NULL per the STORED generated column). DELETING templates
+// return ErrTemplateNotFound, matching GetTemplateByAlias' behavior (store.go:921).
+//
```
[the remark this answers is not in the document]
sparkzky: Acknowledged. A concurrent DELETE that wins the race leaves the UPDATE at 0 affected rows; the handler's follow-up getTemplateInfoFn then returns 404, so user-facing behavior is correct. The DELETING guard is documented at store.go:1033. Not adding a status predicate to the shared updateDefinitionFields to avoid changing its contract.

sdk/python/cubesandbox/_template.py
```diff
@@ -449,3 +449,39 @@ def delete(cls, template_id: str, *, config: Config | None = None) -> None:
...
+        Raises:
+            ApiError: On invalid alias (HTTP 400), template not found (HTTP 404),
+                or concurrent alias conflict (HTTP 409).
+        """
+        cfg = config or Config()
+        payload = {"alias": alias or ""}
+        s = requests.Session()
+        resp = s.put(
+            f"{cfg.api_url}/templates/{template_id}/alias",
+            json=payload,
```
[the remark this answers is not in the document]
sparkzky: All six Python SDK template methods (get/rebuild/delete/build-status/build-logs/set_alias) interpolate template_id directly; quoting only this one would break internal consistency. template_id is constrained to tpl-<hex> / [a-z0-9-], so no unsafe chars can occur. A repo-wide Python SDK url-encoding pass is out of scope here.

sparkzky:
Both dependencies have landed — #1119 (squash `cb9ab44`) and #1052 (`b89d10f`).

This branch is now **rebased onto the latest `master`** as a single commit (`origin/master..HEAD` = 1 commit); the already-merged #1119 SDK commits were dropped, and the template-alias e2e tests were relocated to `tests/e2e/sdk_compat/cases/templates/test_alias.py` per #1052.

Conflicts resolved:
- `CubeAPI/src/openapi.rs` — merged the drifted import list (master-side models + `SetTemplateAliasRequest`).
- `openapi.yml` — kept both the master-added `RefreshRequest` schema and the new `SetTemplateAliasRequest` schema.
- `sdk/python/tests/test_template_e2e.py` — accepted deletion (#1052 moved it to `sdk_compat`); rewrote the 2 set-alias e2e cases in the new framework style.

All CI is green (build + CubeAPI/CubeMaster tests + DCO + Code Format Check). Marked **ready for review**.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1013,6 +1021,48 @@ func claimAliasAfterReady(ctx context.Context, templateID, alias string) (claimW
...
+// Unlike claimAliasAfterReady, this does NOT require the template to be READY —
+// it is an explicit operator action; see design §3.5.
+func SetTemplateAlias(ctx context.Context, templateID, alias string) error {
+	alias = strings.TrimSpace(alias)
+	if err := validateTemplateAlias(alias); err != nil {
```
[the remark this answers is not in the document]
sparkzky: Acknowledged — the check and the mutation are indeed separate transactions, so a concurrent deleteTemplate that flips status to DELETING in between can (claim path) attach the alias to a now-DELETING row. Two mitigating points: (1) the window is only between GetDefinition and the mutation; (2) it is self-healing — once the DELETING template is garbage-collected the row (and its alias_key) is removed, freeing the alias for re-claim. A fully atomic fix means adding a `status <> DELETING` predicate + RowsAffected check inside claimTemplateAlias's transaction (and a targeted clear-path UPDATE rather than the shared updateDefinitionFields). That's a focused follow-up to avoid regressing the shared helpers; not bundling it into this PR.

openapi.yml
```diff
@@ -780,6 +780,55 @@ paths:
...
+      requestBody:
+        content:
+          application/json:
+            schema:
+              $ref: '#/components/schemas/SetTemplateAliasRequest'
+        required: true
+      responses:
+        '200':
+          description: Alias set or cleared; returns the updated template detail
+          content:
```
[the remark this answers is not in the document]
sparkzky: No change needed. openapi.yml documents the CubeAPI public surface, whose handler uses `Json<SetTemplateAliasRequest>` — a JSON body (at minimum `{}`) is mandatory there. The bodyless `ContentLength == 0` clear is a CubeMaster-internal affordance that is never reached through CubeAPI (it always serializes a JSON body when forwarding), so `required: true` is correct for what SDK clients actually call.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1013,6 +1021,48 @@ func claimAliasAfterReady(ctx context.Context, templateID, alias string) (claimW
...
+		return fmt.Errorf("%w: %v", ErrInvalidAlias, err)
```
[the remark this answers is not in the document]
sparkzky: Addressed in 048d567 — added an early `ErrTemplateIDRequired` guard (mirrors the existing sentinel + the `templateID == ""` convention used across the package) and a unit test `TestSetTemplateAlias_RejectsEmptyID`.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1013,6 +1021,51 @@ func claimAliasAfterReady(ctx context.Context, templateID, alias string) (claimW
...
+	}
+	if alias == "" {
+		// Clear path. updateDefinitionFields (snapshot_ops.go:1031) writes
+		// "updated_at" into the caller's map, so pass a fresh map to avoid
+		// polluting any caller-held state.
+		return updateDefinitionFields(ctx, templateID, map[string]any{"display_name": ""})
+	}
+	// Claim path — claimTemplateAlias already has its own isReady() guard
+	// and release+claim transaction.
+	return claimTemplateAlias(ctx, templateID, alias)
```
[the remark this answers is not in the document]
sparkzky: This is a deliberate, documented design decision — see the comment at store.go:1038-1039 ("Unlike claimAliasAfterReady, this does NOT require the template to be READY — it is an explicit operator action; see design §3.5"). Allowing non-READY targets is intentional so an operator can set/clear an alias regardless of build state (e.g. clear an alias on a FAILED template).

The revert-by-claimAliasAfterReady race is real but narrow: it only fires when an operator re-aliases a template that is *still building* AND whose build carries a different req.Alias; on completion claimAliasAfterReady re-claims the build's req.Alias and overwrites the operator value. For already-READY templates (the normal case, and all the PR's e2e targets) there is no subsequent build-completion re-claim, so the operator change is stable.

Treating this as a known tradeoff of the documented design rather than a defect to fix here.

CubeMaster/pkg/service/httpservice/cube/template.go
```diff
@@ -79,6 +81,133 @@ func deleteTemplateGinHandler(c *gin.Context) {
...
+	}
+	rt.RetCode = int64(errorcode.ErrorCode_Success)
+	return &templateResponse{
+		Res: &types.Res{
+			Ret: &types.Ret{
+				RetCode: int(errorcode.ErrorCode_Success),
+				RetMsg:  "success",
+			},
+		},
+		TemplateID:                 info.TemplateID,
```
fslongjin: This block hand-writes the full `templateResponse` (15 fields) again — `getTemplate` (template.go:410-431) hand-writes the exact same fields. The next time a field is added to `templateResponse`, we have to remember to update both spots; miss one and GET and this endpoint silently return different shapes. Could we extract a shared `templateResponseFromInfo(info, createReq)` used by both? The Rust side already does this via `template_detail_from_cubemaster`.

CubeAPI/src/services/templates.rs
```diff
@@ -64,41 +64,54 @@ impl TemplateService {
...
+                None
+            } else {
+                Some(trimmed)
+            }
+        });
+
+        // Client-side validation mirrors CubeMaster's `validateTemplateAlias`
+        // so a bad alias surfaces as 400 without a round-trip.
+        if let Some(value) = alias {
+            if !is_valid_alias(value) {
```
fslongjin: The alias rule (lowercase alphanumeric/hyphen, max 64 chars, not starting with `tpl-`/`snap-`) is now implemented twice: `is_valid_alias` here and `validateTemplateAlias` on the CubeMaster side. Changing the rule later means updating two places. It's also inconsistent across layers — the Go/Node/Python SDKs don't validate at all. Since CubeMaster already rejects invalid aliases with 400 (mapped to BadRequest by `map_err`), this pre-check only saves a round trip. Suggest dropping the pre-check and keeping the rule in exactly one place (CubeMaster). If you'd rather keep it for fast feedback, at least add a test asserting the two implementations are equivalent — today each side has its own tests but nothing checks they agree.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1013,6 +1021,51 @@ func claimAliasAfterReady(ctx context.Context, templateID, alias string) (claimW
...
+//	alias != ""  → claim: reuse claimTemplateAlias (release from any other
+//	                    holder inside the same transaction, then UPDATE on the
+//	                    target template).
+//
+// The target must be a non-deleting template. Snapshots are rejected because
+// their display_name is an informational label, not a unique alias (their
+// alias_key is always NULL per the STORED generated column). DELETING templates
+// return ErrTemplateNotFound, matching GetTemplateByAlias' behavior (store.go:921).
+//
+// Unlike claimAliasAfterReady, this does NOT require the template to be READY —
```
fslongjin: The create path deliberately defers alias claiming until the template is READY (see the comment at store.go:957: 'so an alias never points to a broken template'). `SetTemplateAlias` explicitly skips that wait, so an operator can point an alias at a template that is still building or will never reach READY — and because it was set explicitly, the broken template keeps the alias indefinitely. This may be an intentional operator override, but it's only documented in a code comment. Could you state the intent in the PR description, and log a warning when the target isn't READY? If it's not intentional, aligning with the create path's status guard would be safer.

CubeAPI/src/handlers/templates.rs
```diff
@@ -208,6 +209,46 @@ pub async fn update_template(
...
+    )
+)]
+pub async fn set_template_alias(
+    State(state): State<AppState>,
+    Path(template_id): Path<String>,
+    Json(body): Json<SetTemplateAliasRequest>,
+) -> AppResult<impl IntoResponse> {
+    // Normalize HTTP input here; TemplateService::set_template_alias repeats the
+    // trim/empty->None pass as a guard for direct (non-HTTP) callers and unit tests.
+    let alias = body.alias.and_then(|s| {
```
fslongjin: The trim + empty-string->None normalization is done twice: here in the handler and again in `TemplateService::set_template_alias` (services/templates.rs:78-88), which even notes the handler already did it. Could we keep it in just one place — the service, as the shared boundary — and have the handler forward the raw value?

CubeMaster/pkg/service/httpservice/cube/template.go
```diff
@@ -79,6 +81,133 @@ func deleteTemplateGinHandler(c *gin.Context) {
...
+				RetMsg:  err.Error(),
+			}},
+			TemplateID: rawTemplateID,
+		}
+	}
+	if err := setTemplateAliasFn(ctx, resolvedTemplateID, bodyReq.Alias); err != nil {
+		code := int(errorcode.ErrorCode_MasterInternalError)
+		switch {
+		case errors.Is(err, templatecenter.ErrTemplateNotFound),
+			errors.Is(err, templatecenter.ErrAliasNotApplicableToSnapshot):
```
fslongjin: A snapshot exists but simply doesn't accept an alias, yet this is mapped to the same 404 as 'template not found'. Callers can't distinguish 'resource gone' from 'operation not supported on this resource'. A more accurate code (e.g. 422 or 400) would let clients give different feedback for the two cases.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -989,6 +991,12 @@ func isDuplicateAliasError(err error) bool {
 	return strings.Contains(s, "23505") || strings.Contains(s, "unique_constraint")
 }
 
+// IsDuplicateAliasError is the exported wrapper around isDuplicateAliasError
+// so the HTTP layer (separate package) can map concurrent-claim failures to
+// HTTP 409 / RetCode=ErrorCode_Conflict without duplicating the driver
+// detection logic. See design §3.3.
+func IsDuplicateAliasError(err error) bool { return isDuplicateAliasError(err) }
```
fslongjin, line 1143: This one-line exported wrapper exists so the HTTP package can map concurrent-claim failures to 409. It's a symptom that error-classification knowledge is spread across the store, the handler, and CubeAPI. Non-blocking, but a typed error (e.g. `type DuplicateAliasError struct{}` matched via `errors.As`) would remove the wrapper and keep the classification in one place.

openapi.yml
```diff
@@ -780,6 +780,55 @@ paths:
             application/json:
               schema:
                 $ref: '#/components/schemas/ApiError'
+  /templates/{templateID}/alias:
```
fslongjin, line 783: The openapi.yml entries here were hand-appended (61 lines) because regenerating via utoipa drops the served `/cluster/*` and `/nodes/*` routes that aren't registered in `paths()`. This is pre-existing debt and this PR extends the hand-written pattern, so the spec can drift from the code. Not blocking — but please file a follow-up to fix the `paths()` registration so the spec is regenerable.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1013,6 +1021,51 @@ func claimAliasAfterReady(ctx context.Context, templateID, alias string) (claimW
...
+		return ErrAliasNotApplicableToSnapshot
+	}
+	if def.Status == StatusDeleting {
+		return ErrTemplateNotFound
+	}
+	if alias == "" {
+		// Clear path. updateDefinitionFields (snapshot_ops.go:1031) writes
+		// "updated_at" into the caller's map, so pass a fresh map to avoid
+		// polluting any caller-held state.
+		return updateDefinitionFields(ctx, templateID, map[string]any{"display_name": ""})
```
fslongjin: The clear path is forced to pass a fresh map because `updateDefinitionFields` writes `updated_at` into the caller's map (snapshot_ops.go:1031). The workaround is fine, but the root cause is a helper mutating its caller's input. Could `updateDefinitionFields` copy internally instead, so callers don't have to work around it?

CubeMaster/pkg/templatecenter/store.go
fslongjin: `claimTemplateAlias` runs `UPDATE ... WHERE template_id = ?` but never checks how many rows were affected, and deleting a template hard-deletes the row (cleanupTemplateMetadata -> `Unscoped().Delete`). If the target is deleted between `GetDefinition` (1048) and the claim, the UPDATE silently matches 0 rows and the alias ends up pointing at a template that no longer exists. Low probability — deletion is heavyweight and blocks on active jobs — but a RowsAffected check is cheap.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -651,6 +652,81 @@ var TemplateDeleteCommand = cli.Command{
...
+			return errors.New("template-id is required")
+		}
+		serverList = getServerAddrs(c)
+		if len(serverList) == 0 {
+			return errors.New("no server addr")
+		}
+		port = c.GlobalString("port")
+		host := serverList[rand.Int()%len(serverList)]
+
+		// --clear takes precedence and sends alias="". Otherwise forward the
```
fslongjin, line 694: `--clear` and `--alias ""` express the same operation, and the code defines `--clear` as taking precedence. Could the `--help` text state that precedence explicitly so users don't have to guess?

fslongjin (changes requested):

CubeMaster/pkg/service/httpservice/cube/template.go
sparkzky: Addressed in d4ca9a3 — extracted `templateResponseFromInfo(info, createReq)`, now shared by `getTemplate` and `setTemplateAlias` (no more duplicated 15-field block).

CubeAPI/src/services/templates.rs
sparkzky: Kept the client-side pre-check (fast 400 at the REST boundary, unit-tested — the alternative you offered) and strengthened the comment to name CubeMaster's `validateTemplateAlias` as the single source of truth, with a keep-in-sync note if the rule changes.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in d4ca9a3 — `SetTemplateAlias` now logs a warning when the target isn't READY; intent is also stated in the PR description.

CubeAPI/src/handlers/templates.rs
sparkzky: Addressed in d4ca9a3 — normalization now lives only in the service layer; the handler forwards the raw `body.alias`.

CubeMaster/pkg/service/httpservice/cube/template.go
sparkzky: Addressed in d4ca9a3 — snapshot rejection now maps to `ErrorCode_MasterParamsError` (400), distinct from a real 404.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Acknowledged as non-blocking; leaving `IsDuplicateAliasError` as-is for now. Can convert to a typed error (`errors.As`) in a follow-up.

openapi.yml
sparkzky: Acknowledged as non-blocking pre-existing debt; will file a follow-up to register the served `/cluster/*` and `/nodes/*` routes in `paths()` so the spec is regenerable.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Acknowledged — keeping the fresh-map workaround (you noted it's fine); deferring the internal-copy change in `updateDefinitionFields` to avoid churning its other callers.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in d4ca9a3 — `claimTemplateAlias` now checks `RowsAffected` and returns `ErrTemplateNotFound` if the target was concurrently hard-deleted (no silent no-op).

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
sparkzky: Addressed in d4ca9a3 — the `--clear` help text now states it takes precedence over `--alias`.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -968,9 +970,19 @@ func claimTemplateAlias(ctx context.Context, templateID, alias string) error {
...
 		}
-		return tx.Table(constants.TemplateDefinitionTableName).
+		res := tx.Table(constants.TemplateDefinitionTableName).
 			Where("template_id = ?", templateID).
-			Update("display_name", alias).Error
+			Update("display_name", alias)
+		if res.Error != nil {
+			return res.Error
+		}
+		if res.RowsAffected == 0 {
```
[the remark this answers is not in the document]
sparkzky: Valid catch — reverted in 0b353db. Under the default MySQL driver (rows-changed, not rows-matched) an UPDATE that sets a column to its current value also reports 0 affected rows, so the check broke PUT idempotency (retrying the alias a template already holds → false 404) and regressed the redo claim path. The deleted-target case is already covered: the handler's follow-up getTemplateInfoFn returns 404, and a hard-deleted row can't be updated so no alias is left dangling.

CubeMaster/pkg/service/httpservice/cube/template.go
```diff
@@ -79,6 +81,145 @@ func deleteTemplateGinHandler(c *gin.Context) {
...
+}
+
+// setTemplateAlias sets, transfers, or clears the alias of an existing
+// template. The templateID path param may be a tpl- id or a current alias
+// (resolved via resolveTemplateIdentifierFn, matching GET/DELETE).
+//
+// Error mapping (design §3.3):
+//   - validateTemplateAlias failure → 400 MasterParamsError
+//   - ErrTemplateNotFound (missing OR DELETING) → 404 NotFound
+//   - ErrAliasNotApplicableToSnapshot → 404 NotFound (a snapshot has no
```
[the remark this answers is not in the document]
sparkzky: Fixed in 0b353db — the error-mapping doc comment now says snapshot → 400 (MasterParamsError), matching the switch, TestSetTemplateAliasHandler_400_OnSnapshot, and the PR description.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1066,60 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+		// Clear path. updateDefinitionFields (snapshot_ops.go:1031) writes
+		// "updated_at" into the caller's map, so pass a fresh map to avoid
+		// polluting any caller-held state.
+		return updateDefinitionFields(ctx, templateID, map[string]any{"display_name": ""})
+	}
+	// Claim path — claimTemplateAlias does the release+claim in one
+	// transaction. A target hard-deleted between GetDefinition and the UPDATE
+	// is caught by the handler's follow-up getTemplateInfoFn (→ 404), and a
+	// deleted row can't hold the alias, so none is left dangling.
+	return claimTemplateAlias(ctx, templateID, alias)
```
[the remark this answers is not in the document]
sparkzky: Addressed in 9f01898 — claimTemplateAlias now retries the release+claim transaction once on InnoDB deadlock / lock-wait-timeout (errno 1205/1213). The deadlocked transaction rolls back, so re-running is safe and almost always succeeds; the rare double-deadlock still surfaces as an error rather than retrying indefinitely.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1066,60 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	if err != nil {
+		return err
+	}
+	if isSnapshotDefinition(def) {
+		return ErrAliasNotApplicableToSnapshot
+	}
+	if def.Status == StatusDeleting {
+		return ErrTemplateNotFound
+	}
+	if def.Status != StatusReady {
```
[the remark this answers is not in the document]
sparkzky: The non-READY allowance is the documented design (settled with the human reviewer; see design §3.5). 9f01898 corrects the warning's accuracy: it now states that for a still-building template the create-time claim can silently overwrite the operator-set alias (the previous 'not auto-reverted' wording was wrong for that case). FAILED targets remain an intentional operator choice, flagged by the warning.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1066,60 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+// return ErrTemplateNotFound, matching GetTemplateByAlias' behavior (store.go:921).
+//
+// Unlike claimAliasForReadyTemplate, this does NOT require the template to be
+// READY — it is an explicit operator action; see design §3.5.
+func SetTemplateAlias(ctx context.Context, templateID, alias string) error {
+	if strings.TrimSpace(templateID) == "" {
+		return ErrTemplateIDRequired
+	}
+	alias = strings.TrimSpace(alias)
+	if err := validateTemplateAlias(alias); err != nil {
```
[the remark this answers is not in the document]
sparkzky: Addressed in 9f01898 — validateTemplateAlias now rejects the bare prefixes 'tpl-'/'snap-' too (previously hasValidTemplateIDPrefix only flagged prefix+suffix), so direct CubeMaster calls (e.g. cubemastercli tpl set-alias --alias tpl-) agree with CubeAPI's stricter is_valid_alias. Added unit-test cases for both.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1006,17 +1008,30 @@ func claimTemplateAlias(ctx context.Context, templateID, alias string) error {
...
+	// Two concurrent SetTemplateAlias calls swapping aliases between the same
+	// pair of templates can InnoDB-deadlock (the release locks one row, the
+	// claim locks the other, in opposite orders → errno 1213). The deadlocked
+	// transaction rolls back, so retry once — it almost always succeeds.
+	claim := func() error {
+		return store.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where("alias_key = ? AND template_id <> ?",
+					alias, templateID).
+				Update("display_name", "").Error; err != nil {
```
fslongjin: 🔴 建议合并前处理

失败的 PUT 会把别人的别名误清掉，而且对调用方伪装成「目标不存在」。

上一轮建议 claim 之后查 RowsAffected，跟进 commit 加过，后来又因为 MySQL 幂等问题回退了。回退的理由只对了一半，安全结论是错的。

现在 claimTemplateAlias 的事务顺序是：先 UPDATE 把别名从原持有者手里释放（就是这一行），再 UPDATE 挂到目标上，而且不检查目标到底更新到了没有。如果目标在 GetDefinition 和这次 claim 之间被硬删（cleanupTemplateMetadata 走的是 Unscoped().Delete），会发生：第一步已经执行，原持有者的别名被清掉；第二步匹配 0 行、不报错，事务照常提交；handler 随后 getTemplateInfoFn 返回 404，客户端只看到「目标不存在」。

作者回退时说「硬删后的行挂不上别名，所以不会悬空」——这只是在说「不会出现指向已删模板的别名」，没回答「release 的副作用谁来兜」。之前加 RowsAffected==0 → return err 其实能让整段事务（含 release）一起回滚，方向是对的；回退的真正原因是 MySQL 默认按 rows-changed 计数，同值幂等 PUT 也会报 0，会把「本来就持有该别名」的重试误判成 404。这个幂等坑是真的，但正确修法不是裸撤检查。

建议在同一事务里用一次确认查询代替裸 RowsAffected：claim 之后 SELECT COUNT(*) WHERE template_id=? AND alias_key=?，为 0 就 return ErrTemplateNotFound，让整段事务回滚、release 一并撤销。幂等 PUT 会通过确认查询，目标已删则会回滚 release。请不要再拿「handler 会 404」来证明 claim 侧没有副作用——那只能证明客户端看到了 404，证明不了别名没被动过。

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1086,61 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	if def.Status == StatusDeleting {
+		return ErrTemplateNotFound
+	}
+	if def.Status != StatusReady {
+		// Explicit operator action: aliases may target non-READY templates by
+		// design (see design §3.5). Caveat: for a still-building template the
+		// create-time claimAliasForReadyTemplate re-claims the build's original
+		// alias once it reaches a non-FAILED status, which can silently
+		// overwrite this operator-set alias.
+		log.G(ctx).Warnf("SetTemplateAlias: template %s status is %q (not READY); alias %q points at a non-ready template (may be overwritten on build completion)", templateID, def.Status, alias)
```
fslongjin: 🔴 建议合并前处理

BUILDING 期间 set-alias 返回 200，构建完成却可能用 create 时的原别名把运营刚设的值静默盖掉。

非 READY 允许设别名被标成「有意豁免」，也加了 Warn。但构建完成路径仍然用的是 create 请求里的原 req.Alias 再 claim 一次：finalizeTemplateReplicas(..., req.Alias, replicas)（image_job_runner.go:187）。

具体场景：创建时带了 alias foo；BUILDING 期间运营 set-alias 改成 bar（或者 --clear），接口返回 200；构建成功时 claimAliasForReadyTemplate(req.Alias) 又把别名写回 foo；运营只在内部日志里看到一句 warning，接口层面没有任何提示。对运营面 API 来说，「成功之后静默改回去」比「一开始就拒绝」更难排查。

代码注释和 Warn 承认了这一点，但有两个问题：第一，commit message 和 review 回复里反复引用的 docs/design-alias-for-existing-templates.md（含 §3.5），在当前仓库和本 PR 的文件列表里都不存在，审查方根本没法核对「已文档化」的说法；第二，openapi、CLI、SDK 对用户也没有「可能被覆盖」的可见说明。

建议三选一，写清楚就行：BUILDING/PENDING 直接拒绝 set-alias（最简单，和 create 路径「等 READY 再挂别名」一致）；或者 finalize 时不要用过期的 req.Alias 覆盖，改成读当前 DB 的 display_name，运营已改过就跳过 create-time claim；或者坚持现状，但把契约写进仓库内可审查的文档（至少 openapi description 加 CLI help），并让 200 响应或文档能告诉调用方「构建完成可能覆盖」。只打一句 Warn、却声称 design §3.5 已经结算——在设计文档缺失的前提下，还不够合入。

CubeAPI/src/services/templates.rs
```diff
@@ -64,41 +64,55 @@ impl TemplateService {
...
+            } else {
+                Some(trimmed)
+            }
+        });
+
+        // Client-side validation mirrors CubeMaster's `validateTemplateAlias`
+        // (the single source of truth) so a bad alias surfaces as 400 without
+        // a round-trip — keep the two in sync if the rule changes.
+        if let Some(value) = alias {
+            if !is_valid_alias(value) {
```
fslongjin: 🟡 建议处理

上一轮的 🔴 还没闭环：保留了预检，但没加一致性测试。

上一轮给的是二选一：要么删掉 CubeAPI 这层预检、规则只留 Go 一份；要么保留预检，但加测试把 Rust 和 Go 两份规则钉死成一样。跟进选了「保留预检 + 注释写 keep in sync」，并且把裸 tpl-/snap- 跟 Go 对齐了——对齐是好事。但一致性测试仍然没有，规则还是两份手写实现。

注释挡不住下次改规则只改一边。上一轮把这条标成合并前处理，就是因为「要记得改两处」的隐患还在。建议补一组测试，用同一批正反例同时覆盖 validateTemplateAlias 和 is_valid_alias 的预期；或者干脆删掉这里的预检，让 400 只来自 CubeMaster。选了保留预检却不加钉死测试，这条 🔴 就没真正闭环。

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1034,6 +1049,19 @@ func isDuplicateAliasError(err error) bool {
 	return strings.Contains(s, "23505") || strings.Contains(s, "unique_constraint")
 }
 
+// isDeadlockError reports InnoDB deadlock (1213) and lock-wait-timeout (1205) —
+// the transient errors worth one retry in the alias release+claim transaction.
+func isDeadlockError(err error) bool {
+	var mysqlErr *mysql.MySQLError
+	return errors.As(err, &mysqlErr) && (mysqlErr.Number == 1205 || mysqlErr.Number == 1213)
```
fslongjin: 🟡 建议处理

deadlock 重试只认 MySQL，PG 部署换别名死锁会直接 500。

isDeadlockError 只匹配 *mysql.MySQLError 的 1205/1213。同一份文件里 isDuplicateAliasError 已经照顾了 PG（23505），但死锁重试没有。两个 SetTemplateAlias 互换别名时的死锁在 PG 上也会发生（40P01，锁等待 55P03），MySQL 会自动再试一次，PG 部署就直接 500。

如果 PG 仍是一等公民，请把 PG 的 deadlock/lock-timeout 也纳入一次重试；否则在注释里写明「仅 MySQL 重试」，避免后人以为两端行为一致。

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
```diff
@@ -651,6 +652,81 @@ var TemplateDeleteCommand = cli.Command{
...
+		// --alias value (which may itself be "" to clear, per the shared
+		// contract).
+		alias := c.String("alias")
```
fslongjin, line 697: ⚪ 小问题

不带任何 flag 时会静默清掉别名。

既没传 --alias 也没传 --clear 时，alias 默认是空串，请求体变成 clear。cubemastercli tpl set-alias <id> 会静默清掉别名，容易手滑。建议要求二者至少提供一个，否则报用法错误。不拦合并。

openapi.yml
```diff
@@ -780,6 +780,55 @@ paths:
...
+        required: true
+      responses:
+        '200':
+          description: Alias set or cleared; returns the updated template detail
+          content:
+            application/json:
+              schema:
+                $ref: '#/components/schemas/TemplateDetail'
+        '400':
+          description: Invalid alias
```
fslongjin: ⚪ 小问题

文档和 PR 描述滞后于代码。

openapi 里 400 只写了 Invalid alias，但 snapshot 拒绝现在也是 400，文档没写。PR 描述还写着「404 not-found/snapshot」，和代码（snapshot → 400）对不上；描述里说「单 commit」，和当前多 commit 历史也对不上。建议改一下 400 描述（比如包含 invalid alias / snapshot not applicable），并把 PR 描述更新到和代码一致。不拦合并，但别让合并后文档继续骗人。

fslongjin (changes requested):
二次对抗性复核（对照 head 9f01898）。上一轮 10 条里 1/4/5/10 已修好，deadlock 单次重试和裸 tpl-/snap- 对齐也是有效加固，CI 全绿——这些值得肯定。但有两个关键路径在对抗场景下站不住，建议合并前至少处理掉这两条：

1. claimTemplateAlias 失败时会先清掉别人的别名再静默 claim 失败（store.go:1020），客户端只看到 404，第三方别名已被误清。RowsAffected 回退的安全论证只回答了「不会悬空」，没回答「release 副作用谁来兜」。
2. BUILDING 期间 set-alias 返回 200，构建完成却可能用 create 时的原别名静默盖掉运营刚设的值（store.go:1129 + image_job_runner.go:187）；而反复引用的 design 文档并不在仓库里。

另外 #3 上一轮的规则双份问题仍未闭环（保留预检但无一致性测试），建议一并收掉。其余 #4/#5/#6 不单独拦合并。

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in ebb689e — added a confirmation SELECT (template_id + alias_key) inside the claim transaction. If the target is hard-deleted between GetDefinition and the claim, the transaction now rolls back the release too, instead of silently clearing another template's alias and surfacing as 404. A SELECT (not RowsAffected) avoids the MySQL rows-changed conflation that broke idempotent re-claims under the earlier check — you were right that my revert only addressed half the problem.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in ebb689e — SetTemplateAlias now requires READY (non-READY → ErrTemplateNotReady → handler maps to 409), which removes the silent-overwrite path entirely (consistent with the create path, which claims after READY). You were right that the cited design §3.5 doc never existed in-repo — those citations are removed. Verified live: set-alias on a FAILED template returns 409.

CubeAPI/src/services/templates.rs
sparkzky: Closed in ebb689e — deleted the CubeAPI set_template_alias pre-check (and its now-obsolete tests) so the rule lives only in CubeMaster's validateTemplateAlias; invalid aliases still return 400 via map_err. is_valid_alias + its direct unit tests remain (still used by get_template_by_alias). Verified live: PUT alias=UPPER → 400.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in ebb689e — isDeadlockError now also retries on PostgreSQL 40P01 (deadlock_detected) / 55P03 (lock_not_available), matching the existing MySQL 1205/1213 handling.

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
sparkzky: Addressed in ebb689e — the CLI now requires --alias or --clear; running set-alias with neither returns a usage error instead of silently clearing the alias.

openapi.yml
sparkzky: Addressed in ebb689e + the PR description refresh — openapi 400 now notes snapshot-not-applicable and 409 notes not-READY; the description is updated (snapshot→400, multi-commit history, READY requirement).

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1034,6 +1049,19 @@ func isDuplicateAliasError(err error) bool {
 	return strings.Contains(s, "23505") || strings.Contains(s, "unique_constraint")
 }
 
+// isDeadlockError reports InnoDB deadlock (1213) and lock-wait-timeout (1205) —
+// the transient errors worth one retry in the alias release+claim transaction.
+func isDeadlockError(err error) bool {
```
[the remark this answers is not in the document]
sparkzky: Addressed in ebb689e — isDeadlockError now also retries on PostgreSQL 40P01 (deadlock_detected) / 55P03 (lock_not_available), matching the MySQL 1205/1213 handling (same concern as the human review).

CubeMaster/pkg/templatecenter/store.go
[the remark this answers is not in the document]
sparkzky: Addressed in ebb689e — SetTemplateAlias now requires READY (non-READY → ErrTemplateNotReady → 409), so this silent-overwrite path no longer exists; verified live (set-alias on a FAILED template returns 409).

CubeMaster/cmd/cubemastercli/commands/cubebox/template.go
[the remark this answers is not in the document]
sparkzky: Addressed in ebb689e — the CLI now requires --alias or --clear; running set-alias with neither returns a usage error instead of silently clearing.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1111,57 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+// of an existing template-kind definition.
+//
+//	alias == ""  → clear: UPDATE display_name = '' WHERE template_id = ?
+//	alias != ""  → claim: reuse claimTemplateAlias (release from any other
+//	                    holder inside the same transaction, then UPDATE on the
+//	                    target template).
+//
+// The target must be a READY template: non-READY is rejected with
+// ErrTemplateNotReady, so an alias never points at a building/failed template
+// and the create-time claim (claimAliasForReadyTemplate) can't silently
```
fslongjin: **[P1] rebuild/redo reclaims the create-time alias and silently reverts an operator set/clear/transfer**

`SetTemplateAlias` only updates `display_name`; it does not sync the job's `RequestJSON.Alias`. When a redo finishes, it still re-claims using the frozen create-time alias:

```go
// redo.go:364
claimWarning, err := refreshTemplateReplicaSummary(ctx, req.TemplateID, sourceReq.Alias)
```

Reproducible semantic break:
1. T1 created with alias A → READY
2. `set_alias(T2, A)` transfers A to T2 (covered by this PR's e2e)
3. rebuild/redo T1
4. T1 re-claims A from the job JSON → silently steals A back from T2

The same path re-adds A after `set_alias(..., "")` cleared it. The doc comment here ("the create-time claim ... can't silently overwrite an operator change once the build finishes") only holds for the first build completion — it does not hold for redo.

Suggested fix: in redo/rebuild, claim based on the current `display_name` (skip claim when empty), or have `SetTemplateAlias` sync `RequestJSON.Alias`. Please add a regression test: set/clear/transfer alias, then rebuild the original holder and assert the alias is not reclaimed.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1111,57 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+		return err
+	}
+	if isSnapshotDefinition(def) {
+		return ErrAliasNotApplicableToSnapshot
+	}
+	if def.Status == StatusDeleting {
+		return ErrTemplateNotFound
+	}
+	if def.Status != StatusReady {
+		return ErrTemplateNotReady
```
fslongjin, line 1211: **[P1] clear is gated on READY — an alias on a FAILED template cannot be cleared without deleting the template**

The READY guard applies to both claim and clear. But `GetTemplateByAlias` (store.go:959) only excludes `DELETING`, so an alias on a FAILED template is still resolvable. A `PUT .../alias` to clear it returns `ErrTemplateNotReady` → 409, and the only way to release the `alias_key` is to delete the template.

This also contradicts an earlier review reply stating the intent was to allow non-READY so an operator can clear an alias on a FAILED template.

Suggested fix: keep claim READY-only; allow clear for any non-DELETING template (at minimum FAILED). Branch in `SetTemplateAlias`: claim requires `status == READY`, clear requires `status != DELETING`. Add a store/e2e test: clear on a FAILED template succeeds.

CubeMaster/pkg/service/httpservice/cube/template_test.go
```diff
@@ -280,3 +281,301 @@ func TestGetTemplateIncludesDisplayMetadata(t *testing.T) {
...
+	assert.Equal(t, "my-alias", setCalledAlias)
+	assert.Equal(t, int(errorcode.ErrorCode_Success), got.Ret.RetCode)
+	assert.Equal(t, "tpl-resolved", got.TemplateID)
+	assert.Equal(t, "my-alias", got.DisplayName)
+	assert.Equal(t, int64(errorcode.ErrorCode_Success), rt.RetCode)
+}
+
+// TestSetTemplateAliasHandler_ClearPath_NoBody verifies that a PUT with no
+// body is treated as a clear (alias=""), matching the shared contract
+// ("alias 省略 / null / "" → 清除").
```
fslongjin: **[P1 / convention] Test comment contains non-English characters**

This line contains Chinese: `// ("alias 省略 / null / "" → 清除").`, which violates the project's non-English-character rule.

Please switch to English, e.g. `// ("omitted / null / empty alias → clear").`

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1006,17 +1009,47 @@ func claimTemplateAlias(ctx context.Context, templateID, alias string) error {
...
+		return store.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where("alias_key = ? AND template_id <> ?",
+					alias, templateID).
+				Update("display_name", "").Error; err != nil {
+				return fmt.Errorf("release stale alias %q fail: %w", alias, err)
+			}
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where("template_id = ?", templateID).
+				Update("display_name", alias).Error; err != nil {
```
fslongjin: **[P2] claim transaction does not re-check status — TOCTOU window**

`SetTemplateAlias` validates READY via `GetDefinition`, but the `claimTemplateAlias` transaction's UPDATE/confirm only match on `template_id` + `alias_key`, not `status`. If a concurrent redo/status flip (e.g. `deleteTemplate` flipping status to DELETING) happens in between, the alias can still be written to a non-READY row; confirm succeeds and the transaction commits. The clear path via `updateDefinitionFields` has the same gap.

Suggested fix: add `status = READY` (or the relaxed clear predicate) to the release/claim UPDATE and the confirm query; map 0 rows to `ErrTemplateNotReady` / `ErrTemplateNotFound`. Add an integration test for the in-transaction status condition and confirm-failure rollback of the release.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1006,17 +1009,47 @@ func claimTemplateAlias(ctx context.Context, templateID, alias string) error {
...
+			// the original holder keeps its alias. A SELECT (not RowsAffected)
+			// avoids MySQL's rows-changed conflation with idempotent re-claims.
+			var claimed int64
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where("template_id = ? AND alias_key = ?", templateID, alias).
+				Count(&claimed).Error; err != nil {
+				return fmt.Errorf("confirm alias claim for template %s fail: %w", templateID, err)
+			}
+			if claimed == 0 {
+				return ErrTemplateNotFound
```
fslongjin: **[P2] Several follow-up fixes lack reproducible regression tests**

The final tree has no tests encoding the behavior these fixes established:

- `0b353db6` fix(alias): revert RowsAffected claim check (MySQL idempotency) — points out that an idempotent re-claim would false-positive 404, but there is no test that `SetTemplateAlias` succeeds when the template already holds that alias.
- `9f01898a` / `ebb689e5` deadlock single retry + SELECT confirm — `RejectsNotReady` is covered, but deadlock retry and confirm (hard-deleted target rolls back the release / idempotent SELECT) are not.

Suggested tests:
1. claim path: template already holds alias → `SetTemplateAlias` succeeds (locks the MySQL rows-changed regression).
2. confirm: `Count == 0` after claim UPDATE → `ErrTemplateNotFound` and the release is rolled back.
3. deadlock: when `isDeadlockError` is true, claim is invoked twice and the second succeeds.

CubeAPI/src/models/mod.rs
```diff
@@ -830,6 +830,17 @@ pub struct RebuildTemplateRequest {
     pub extra: serde_json::Map<String, serde_json::Value>,
 }
 
+/// Body for PUT /templates/:id/alias (set / modify / clear alias).
+///
+/// `alias` is `None` / null / empty string ⇒ clear the current alias.
+/// A non-empty value is validated client-side with `is_valid_alias` and
```
fslongjin: **[P3] CubeAPI set path does not call `is_valid_alias` despite this doc comment**

The `SetTemplateAliasRequest` doc says "A non-empty value is validated client-side with `is_valid_alias`", but `TemplateService::set_template_alias` only trims and forwards to CubeMaster — it does not call `is_valid_alias`. Invalid aliases still 400 (CubeMaster `130400`), but this is inconsistent with create/lookup early validation, and the OpenAPI/PR description is inaccurate.

Pick one:
- keep early validation: call `is_valid_alias` in `set_template_alias` and add a test that an invalid alias returns 400 early; or
- drop the "client-side validated" wording and make CubeMaster the single source of truth.

Also `is_valid_alias` (CubeAPI) and `validateTemplateAlias` (CubeMaster) are two implementations — changing the rule means updating both. If both stay, add a test asserting they agree.

CubeMaster/pkg/service/httpservice/cube/template.go
```diff
@@ -79,6 +81,148 @@ func deleteTemplateGinHandler(c *gin.Context) {
...
+		case errors.Is(err, templatecenter.ErrTemplateNotFound):
+			code = int(errorcode.ErrorCode_NotFound)
+		// A snapshot exists but aliases don't apply to it — map to 400 so
+		// clients can distinguish "operation not supported on this resource"
+		// from a genuine 404 "template not found".
+		case errors.Is(err, templatecenter.ErrAliasNotApplicableToSnapshot):
+			code = int(errorcode.ErrorCode_MasterParamsError)
+		case errors.Is(err, templatecenter.ErrInvalidAlias):
+			code = int(errorcode.ErrorCode_MasterParamsError)
+		case errors.Is(err, templatecenter.ErrTemplateNotReady):
```
fslongjin, line 188: **[P3] Handler missing `ErrTemplateNotReady → 409` test; openapi/utoipa 409 description drift**

Two small gaps:

1. The handler maps `ErrTemplateNotReady` to `ErrorCode_Conflict` (here), but `template_test.go` only has `409_OnDuplicateAlias` — there is no `409_OnNotReady`. Please add a handler unit test asserting this mapping.

2. `openapi.yml:821` already says "Template is not READY, or concurrent alias claim conflict", but the utoipa annotation (`CubeAPI/src/handlers/templates.rs:225`) still says "Concurrent alias claim conflict" — regenerating via utoipa will revert to the old description. Please align both sides.

fslongjin (changes requested):
Adversarial review. CI is green and the full-stack coverage (CubeMaster → CubeAPI → SDKs → OpenAPI → e2e) is solid, and the multi-round review fixes (confirm SELECT, deadlock retry, shared response builder) are good. Requesting changes because some issues break the alias semantics this PR introduces.

Blockers (P1):
- rebuild/redo re-claims the create-time alias and silently reverts an operator set/clear/transfer
- clear is gated on READY, so an alias on a FAILED template cannot be cleared without deleting the template
- a test comment contains non-English characters (project convention)

Should-fix (P2): claim transaction TOCTOU (no in-txn status check); several follow-up fixes (reverted RowsAffected idempotency, deadlock retry, confirm rollback) lack reproducible regression tests.

Low-priority (P3): CubeAPI set path skips `is_valid_alias` despite the model doc; handler missing `ErrTemplateNotReady → 409` unit test; openapi/utoipa 409 description drift.

Details in inline comments.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in 36d5805 — the READY gate is now claim-only; clear is allowed for any non-DELETING template, so an alias stuck on a FAILED template can be released without deleting it. Added TestSetTemplateAlias_ClearAllowedOnFailed.

CubeMaster/pkg/service/httpservice/cube/template_test.go
sparkzky: Fixed in 36d5805 — the comment is now English.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Addressed in 36d5805 — the confirm SELECT now also requires status <> DELETING, so a target flipping to DELETING between GetDefinition and the claim rolls back the whole transaction (release included). I used status <> DELETING rather than status = READY because claimTemplateAlias is shared with the create/redo path, which legitimately claims at PARTIALLY_READY.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Added in 36d5805: TestSetTemplateAlias_IdempotentReclaimSucceeds (locks the rows-changed regression), TestSetTemplateAlias_ClearAllowedOnFailed, TestIsDeadlockError (PG 40P01/55P03 + negative cases). The confirm-rollback and full deadlock-retry are internal to the db transaction; the unit-test layer mocks above gorm, so they need an integration harness — I've pinned the contract at the function boundary and noted the db-level cases as a harness gap.

CubeAPI/src/models/mod.rs
sparkzky: Fixed in 36d5805 — dropped the 'validated client-side with is_valid_alias' wording; CubeMaster's validateTemplateAlias is now named as the single source.

CubeMaster/pkg/service/httpservice/cube/template.go
sparkzky: Addressed in 36d5805 — added TestSetTemplateAliasHandler_409_OnNotReady, and aligned the utoipa 400/409 descriptions with openapi.yml so regenerating the spec no longer reverts them.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1111,57 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	if err != nil {
+		return err
+	}
+	if isSnapshotDefinition(def) {
+		return ErrAliasNotApplicableToSnapshot
+	}
+	if def.Status == StatusDeleting {
+		return ErrTemplateNotFound
+	}
+	if def.Status != StatusReady {
```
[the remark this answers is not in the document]
sparkzky: Addressed via the same fix as the human review — clear is now allowed on any non-DELETING template (36d5805), so a stuck alias on a FAILED template is recoverable.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1006,17 +1009,47 @@ func claimTemplateAlias(ctx context.Context, templateID, alias string) error {
...
+				Count(&claimed).Error; err != nil {
+				return fmt.Errorf("confirm alias claim for template %s fail: %w", templateID, err)
+			}
+			if claimed == 0 {
+				return ErrTemplateNotFound
+			}
+			return nil
+		})
+	}
+	if err := claim(); err != nil {
```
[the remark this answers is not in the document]
sparkzky: Addressed in 36d5805 — the confirm now requires status <> DELETING, closing the DELETING-transition window.

CubeMaster/pkg/templatecenter/store.go
sparkzky: Agreed this is a real P1. Root cause: redo reads the frozen alias from the image job's RequestJSON (t_cube_template_image_job, redo.go:205), which SetTemplateAlias never updates — so a rebuild re-claims the original alias and reverts an operator set/clear/transfer. The correct fix is to sync that stored RequestJSON.Alias on SetTemplateAlias, for the target AND any template whose alias is released in the claim transaction (else a transfer leaves the old holder's job alias stale and it steals the alias back on redo). This is a cross-table, cross-template change in the create/redo/claim path; I've landed the other 6 threads in 36d5805 and am scoping this one as a focused change to avoid destabilizing the build/redo flow — will implement once the approach is confirmed.

sparkzky: Implemented in ee01da9. SetTemplateAlias now syncs the operator's alias change into the image job's stored RequestJSON (the blob redo reads), so a rebuild/redo re-claims the operator's current alias instead of the frozen create-time value:
- claim: syncs the target's job alias to the new value AND clears the old holder's job alias (found via GetTemplateByAlias before the claim), so a redo of either template can't steal the alias back.
- clear: syncs the target's job alias to "" so redo won't re-add it.
The sync uses a generic JSON edit (not a struct round-trip) so unrelated fields — notably RegistryPassword — are preserved; it's best-effort behind an isReady guard (display_name remains the source of truth), so a sync failure is logged, not fatal. Tests cover the rewrite+preservation and the target+old-holder sync on claim.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1006,17 +1009,48 @@ func claimTemplateAlias(ctx context.Context, templateID, alias string) error {
...
+			}
+			// Confirm the target still exists, holds the alias, and is not
+			// DELETING. If it was hard-deleted (or flipped to DELETING) between
+			// GetDefinition and this transaction the confirm matches 0 rows;
+			// returning here rolls back the release so the original holder keeps
+			// its alias. A SELECT (not RowsAffected) avoids MySQL's rows-changed
+			// conflation with idempotent re-claims.
+			var claimed int64
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where("template_id = ? AND alias_key = ? AND status <> ?", templateID, alias, StatusDeleting).
```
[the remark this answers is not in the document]
sparkzky: Addressed in 7c9e9aa — claimTemplateAlias now takes requireReady; the SetTemplateAlias path pushes status = READY into the claim/confirm transaction (not just the out-of-tx GetDefinition check), so a target that starts rebuilding between GetDefinition and the claim can't receive an alias. Existing-but-no-longer-READY → ErrTemplateNotReady (409); gone → ErrTemplateNotFound (404). The shared create/redo path keeps status <> DELETING (it claims at PARTIALLY_READY).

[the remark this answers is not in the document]
sparkzky: Same TOCTOU as above — addressed in 7c9e9aa (requireReady pushes status = READY into the claim transaction, closing the GetDefinition→claim window).

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1112,59 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	}
+	if isSnapshotDefinition(def) {
+		return ErrAliasNotApplicableToSnapshot
+	}
+	if def.Status == StatusDeleting {
+		return ErrTemplateNotFound
+	}
+	// Claim requires READY; clear is allowed for any non-DELETING template so a
+	// stuck alias (e.g. on a FAILED template) can always be released.
+	if alias != "" && def.Status != StatusReady {
```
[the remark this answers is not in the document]
sparkzky: Addressed in 7c9e9aa — the claim path now also clears the alias from OTHER templates' in-progress (PENDING/RUNNING) build jobs, so a still-building template created with that alias can't reclaim it on completion and steal it back. Together with the redo-sync in ee01da9, this covers both the create-time and redo revert paths.

CubeMaster/pkg/templatecenter/request_validation.go
```diff
@@ -127,7 +127,10 @@ func validateTemplateAlias(alias string) error {
 	if alias == "" {
 		return nil
 	}
-	if hasValidTemplateIDPrefix(alias) {
+	// Reject any alias that collides with a template/snapshot ID prefix,
+	// including the bare "tpl-"/"snap-" (hasValidTemplateIDPrefix only flags
+	// prefix+suffix). Matches CubeAPI's is_valid_alias so the two layers agree.
+	if strings.HasPrefix(alias, "tpl-") || strings.HasPrefix(alias, "snap-") {
```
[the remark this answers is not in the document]
sparkzky: Acknowledged and documented in 7c9e9aa (design §3.4): validateTemplateAlias is shared with the create/normalize path, so rejecting the bare tpl-/snap- prefixes also tightens create-time aliases. This is intentional — a bare-prefix alias collides with the tpl-/snap- ID prefixes — and is now called out in the doc. No code change; it's the correct, consistent behavior across create and set.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1133,207 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+			continue
+		}
+		newPayload, changed, err := applyAliasToRequestJSON(jobs[i].RequestJSON, alias)
+		if err != nil {
+			continue // skip unparseable jobs rather than aborting the alias op
+		}
+		if !changed {
+			continue
+		}
+		if err := updateTemplateImageJob(ctx, jobs[i].JobID, map[string]any{"request_json": newPayload}); err != nil {
```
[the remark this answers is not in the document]
fslongjin: Outdated. `syncTemplateImageJobAlias` is gone. Alias sync now runs in the same transaction as `display_name` (`syncCreateRedoImageJobAliasTx`); a per-job failure rolls back SetTemplateAlias instead of Warn+200 with a half-updated job list.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1133,207 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+			return fmt.Errorf("sync image job alias for template %s: %w", templateID, err)
+		}
+	}
+	return nil
+}
+
+// listInProgressImageJobs returns image jobs in a PENDING/RUNNING (building) state.
+func listInProgressImageJobs(ctx context.Context) ([]models.TemplateImageJob, error) {
```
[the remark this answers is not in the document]
fslongjin: Outdated. `clearAliasFromOtherInProgressJobs` and `listInProgressImageJobs` were removed. Create/redo no longer scans other in-progress jobs.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -998,25 +1001,77 @@ func ResolveTemplateIdentifier(ctx context.Context, identifier string) (string,
...
+			if err := tx.Table(constants.TemplateDefinitionTableName).
+				Where(fmt.Sprintf("template_id = ? AND alias_key = ? AND %s", statusExpr), templateID, alias, statusVal).
+				Count(&claimed).Error; err != nil {
+				return fmt.Errorf("confirm alias claim for template %s fail: %w", templateID, err)
+			}
+			if claimed == 0 {
+				// Operator path: distinguish "deleted" (404) from "exists but no
+				// longer READY" (409, retry once it rebuilds to READY).
+				if requireReady {
+					var exists int64
```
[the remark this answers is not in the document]
fslongjin: Fixed. Existence COUNT errors now propagate instead of being mapped to 404. Docker tests cover `requireReady=true` (`OperatorClaimRejectsNonReadyTarget`, `CreateRedoDoesNotStealReadyHolder`).

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1058,6 +1133,207 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	}
+	// Also clear the alias from any OTHER template's in-progress build job, so a
+	// still-building template created with this alias can't reclaim it on
+	// completion and steal it back (design §3.6).
+	clearAliasFromOtherInProgressJobsBestEffort(ctx, alias, templateID)
+	return nil
+}
+
+// applyAliasToRequestJSON returns payload with its "alias" field set to alias
+// (omitted when alias is ""), preserving every other field byte-for-byte via a
```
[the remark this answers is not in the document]
fslongjin: Outdated. The comment now says the rewrite is semantic, not byte-identical. COMMIT / SNAPSHOT / LEGACY jobs are no longer rewritten.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1037,6 +1115,226 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+// create-time value (which would silently revert a set/clear/transfer — see
+// design §3.6). display_name stays the source of truth for resolution; this
+// only prevents redo revert, so the best-effort wrapper logs failures instead
+// of failing the alias op.
+func syncTemplateImageJobAlias(ctx context.Context, templateID, alias string) error {
+	if !isReady() {
+		return nil // store not initialized (e.g. unit tests) — skip best-effort sync
+	}
+	jobs, err := listTemplateImageJobsByTemplateID(ctx, templateID)
+	if err != nil {
```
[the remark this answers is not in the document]
fslongjin: Outdated. Sync only touches CREATE/REDO (`listCreateRedoImageJobsByTemplateIDTx`). Docker tests assert COMMIT `RequestJSON` is left unchanged.

CubeMaster/pkg/templatecenter/request_validation.go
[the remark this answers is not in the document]
fslongjin: Acknowledged, intentional. Rejecting bare `tpl-` / `snap-` matches CubeAPI `is_valid_alias`. We are not reverting this.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1037,6 +1115,226 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	// silently steal it back from this template.
+	oldHolder := ""
+	if cur, err := GetTemplateByAlias(ctx, alias); err == nil && cur != nil && cur.TemplateID != templateID {
+		oldHolder = cur.TemplateID
+	}
+	if err := claimTemplateAlias(ctx, templateID, alias, true); err != nil {
+		return err
+	}
+	syncTemplateImageJobAliasBestEffort(ctx, templateID, alias)
+	if oldHolder != "" {
```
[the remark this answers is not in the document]
fslongjin: Outdated. The best-effort helpers are gone. `display_name` and CREATE/REDO JSON update in one transaction; sync failure fails the API.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1051,6 +1166,214 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	}
+	// Claim requires READY; clear is allowed for any non-DELETING template so a
+	// stuck alias (e.g. on a FAILED template) can always be released.
+	if alias != "" && def.Status != StatusReady {
+		return ErrTemplateNotReady
+	}
+	// Idempotent clear of an already-empty alias is a no-op: do not rewrite
+	// in-flight first-create job JSON (design §3.6 — PENDING DisplayName is
+	// empty until finalize claims).
+	if alias == "" && strings.TrimSpace(def.DisplayName) == "" {
```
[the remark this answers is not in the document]
fslongjin: Acknowledged, intentional. Empty `DisplayName` + `PUT alias=""` is a no-op so we do not rewrite an in-flight first-create job JSON. Claiming a non-empty alias on a non-READY template already returns 409. We are not adding “cancel pending alias” in this PR.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1051,6 +1166,214 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+				return err
+			}
+			if err := claimTemplateAliasTx(tx, templateID, alias, true); err != nil {
+				return err
+			}
+			if err := syncCreateRedoImageJobAliasTx(tx, templateID, alias); err != nil {
+				return err
+			}
+			if oldHolder != "" {
+				if err := syncCreateRedoImageJobAliasTx(tx, oldHolder, ""); err != nil {
```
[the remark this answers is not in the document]
fslongjin: Acknowledged. The race exists (we `FOR UPDATE` the target only). Finalize prefers `DisplayName` over job alias, so resolution stays correct. Not changing this in this PR.

CubeMaster/pkg/templatecenter/template_image.go
```diff
@@ -191,42 +191,59 @@ func SubmitRedoTemplateFromImage(ctx context.Context, req *types.RedoTemplateFro
...
+				if errors.Is(err, gorm.ErrRecordNotFound) {
+					return ErrTemplateNotFound
+				}
+				return err
+			}
+			if err := allowRedoResumePhase(latestJob); err != nil {
+				return err
+			}
+			sourceJob := latestJob
+			if !isCreateRedoJobOperation(latestJob.Operation) {
```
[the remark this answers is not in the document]
fslongjin: Acknowledged. Redo is the from-image pipeline (rebuild rootfs, cold-start, snapshot). COMMIT and snap- templates are not redo recipes, so we will not switch redo to the COMMIT payload. Falling back to an older CREATE spec when latest is COMMIT is leftover from alias sync; mixed CREATE-then-COMMIT on the same `tpl-` is rare. Leaving as-is for this PR.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1051,6 +1166,214 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+	return aliasFromRequestJSON(job.RequestJSON), true
+}
+
+// aliasToClaimAtFinalize is the create/redo finalize alias (design §3.6):
+//  1. currentJobAlias ok=false → skip claim (empty)
+//  2. definition.DisplayName nonempty → claim DisplayName (covers operator SET
+//     during an in-flight redo; does NOT cover CLEAR)
+//  3. otherwise claim the job alias (first create, or operator CLEAR that
+//     already synced RequestJSON to "")
+func aliasToClaimAtFinalize(ctx context.Context, templateID, jobID string) string {
```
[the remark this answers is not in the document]
fslongjin: Acknowledged. `aliasToClaimAtFinalize` is computed before the finalize transaction, so a concurrent operator CLEAR can be overwritten if the in-memory alias is stale. Unique `alias_key` still prevents two READY holders. Not changing this in this PR.

CubeMaster/pkg/templatecenter/store.go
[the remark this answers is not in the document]
fslongjin: Acknowledged, intentional. Same as the PENDING clear no-op: empty `DisplayName` + `PUT alias=""` does not rewrite first-create job JSON, so finalize still claims the create-time alias. We are not adding “cancel pending alias” in this PR.

CubeMaster/pkg/templatecenter/store.go
```diff
@@ -1051,6 +1166,214 @@ func claimAliasForReadyTemplate(ctx context.Context, templateID, alias string) (
...
+					}).Error; err != nil {
+					return err
+				}
+				return syncCreateRedoImageJobAliasTx(tx, templateID, "")
+			}
+			if def.Status != StatusReady {
+				return ErrTemplateNotReady
+			}
+			oldHolder := ""
+			if cur, err := getTemplateByAliasTx(tx, alias); err == nil && cur != nil && cur.TemplateID != templateID {
```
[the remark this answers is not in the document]
fslongjin: Acknowledged. Old-holder lookup is a non-locking read. Unique `alias_key` still prevents two holders; finalize prefers `DisplayName`. Not changing this in this PR.

fslongjin (approved):

PR #1022, opened 2026-07-18
feat(metrics): add real-time metrics for a single sandbox

emailcannotbeblank:
## Summary

Related to #1021 

This PR implements the sandbox metrics API requested in the issue by adding `GET /sandboxes/{sandboxID}/metrics`, with optional `start` and `end` Unix timestamps in seconds. The endpoint returns an E2B-compatible JSON array containing the sample time, CPU allocation and utilization, memory usage, and disk usage for the requested sandbox.

This API gives users a standard way to inspect sandbox resource usage. It supports performance analysis, CPU and memory bottleneck detection, disk capacity monitoring, operational troubleshooting, sandbox health monitoring, and resource quota evaluation.

## Changes

The request flows through the following components:

```text
Client
  -> CubeAPI: GET /sandboxes/{sandboxID}/metrics
  -> CubeMaster: GET /cube/sandbox/metrics
  -> Cubelet: gRPC GetSandboxMetrics
  -> envd: GET http://<sandbox-ip>:49983/metrics
```

CubeAPI registers the public route, validates the optional time range, calls CubeMaster, maps internal errors to public API errors, and converts the response to the E2B-compatible JSON format. The related files are:

- `CubeAPI/src/routes.rs`
- `CubeAPI/src/handlers/sandboxes.rs`
- `CubeAPI/src/models/mod.rs`
- `CubeAPI/src/cubemaster/mod.rs`
- `CubeAPI/src/services/sandboxes.rs`

CubeMaster registers the internal HTTP endpoint, resolves the Cubelet that owns the sandbox, verifies that the node is available, calls Cubelet with the inherited request context, filters the current sample using `start` and `end`, and returns empty results as `[]`. The related files are:

- `CubeMaster/pkg/cubelet/actions.go`
- `CubeMaster/pkg/server/server_test.go`
- `CubeMaster/pkg/service/httpservice/cube/cube.go`
- `CubeMaster/pkg/service/httpservice/cube/routes.go`
- `CubeMaster/pkg/service/httpservice/cube/sandbox_metrics.go`
- `CubeMaster/pkg/service/httpservice/cube/sandbox_metrics_test.go`
- `CubeMaster/pkg/service/sandbox/sandbox_metrics.go`
- `CubeMaster/pkg/service/sandbox/sandbox_metrics_test.go`
- `CubeMaster/pkg/service/sandbox/types/types.go`

[...]

PR #1122, opened 2026-07-23
feat(metrics): add lifecycle-aware sandbox resource metrics

zyl1121:
## Motivation

PR #1090 adds the standard guest `Task.Stats` transport, but raw cumulative counters are not sufficient as a stable sandbox metrics contract. Template restore and rollback can inherit or rewind guest counters, while reusable host cgroup slots can retain counters from a previous sandbox. Query-time collection would also make every Prometheus scrape fan out to active sandboxes.

This PR makes Cubelet the owner of lifecycle-aware CPU and memory semantics, bounded collection, latest-sample caching, and Prometheus export.

## What Changed

* Implement Cubelet sandbox-controller `Metrics` by querying the primary Cube task through standard containerd `Task.Stats` and converting transport failures to native containerd errors.
* Add persisted guest metrics epochs and raw counter baselines for CPU, throttling, and memory failures.
* Add host cgroup v1/v2 readers and persisted assignment baselines so reused pool slots do not expose a previous sandbox's counter history.
* Expose `guest_workload` and `host_sandbox` as independent accounting scopes; `all` returns both without adding them together.
* Add bounded guest and host samplers with independent concurrency limits, fixed per-cycle queues, request timeouts, staleness handling, and latest-sample caching.
* Add the dedicated cache-only Prometheus endpoint `GET /v1/metrics/resource`.
* Start only the samplers selected by `export_scopes`; default to `host_sandbox`, with `guest_workload` and `all` available through configuration.
* Keep the existing generic containerd cgroups monitor available under `/v1/metrics`; the Cube-native resource endpoint is additive and independent.
* Add aligned English and Chinese operator guides covering scopes, lifecycle semantics, configuration, and the Prometheus endpoint.

The PR is organized as three commits so lifecycle accounting, collection/export, and operator documentation can be reviewed in that order.

## Metrics Semantics

CPU total/user/system, throttling, periods, and memory failures are cumulative counters. Memory current and limit are point-in-time gauges. Cubelet exports base values; CPU cores and percentage are derived with Prometheus rates rather than precomputed in the runtime.

Fresh create, clone from inherited state, rollback, and workload recreation start a new `guest_workload` epoch and baseline. Snapshot/commit, pause/resume, and Cubelet restart preserve the current epoch. Paused, deleted, unavailable, or stale views disappear instead of exporting zero.

`host_sandbox` uses a separate baseline for each sandbox/cgroup assignment. New sandboxes capture this baseline when the pool slot is assigned, before the sandbox process is attached. Existing sandboxes upgraded without a persisted baseline establish a compatibility baseline from their first successful sample. If assignment-time capture fails, sandbox creation continues but host metrics fail closed for that sandbox instead of exporting an incomplete counter window. Host accounting remains continuous across guest rollback because the host process and cgroup assignment are not rewound.

Prometheus reads only the latest typed cache. A scrape never initiates `Task.Stats`, guest RPCs, or host cgroup reads.

Exact epoch-scoped memory peak remains intentionally unexported because inherited maxima cannot be normalized by subtraction and sampled peak tracking can miss values between samples.

### Rollback counter semantics

Rollback keeps the same `sandbox_id` but restores an earlier guest kernel and cgroup state. Cubelet therefore persists a prepared guest epoch before entering the shim restore, does not sample the prepared epoch, and activates a new pending baseline only after successful restore.

Task lookup and other local preflight checks complete before epoch preparation, so preflight failure preserves the current epoch. Preparation failure aborts before runtime mutation. A failed or uncertain restore leaves the new epoch prepared and `guest_workload` fail closed; a later successful rollback establishes a new epoch, while an unrecoverable sandbox must be deleted and recreated. `host_sandbox` counters remain continuous across rollback.

### Relationship to #1022

This PR provides the Cubelet-owned collection/cache layer that can be reused by a later public sandbox metrics API. It does not add `GET /sandboxes/{id}/metrics`, query-time envd collection, range history, or E2B-compatible response routing.

## Scope

This PR is limited to CPU and memory for the primary Cube workload container and the host sandbox cgroup. It does not add disk/network/filesystem metrics, OTLP, historical storage, Cube Guard, vsock metrics forwarding, eBPF collection, or a generic collector for non-Cube containerd runtimes.

## Validation

* Linux focused tests cover guest epoch persistence, rollback ordering, counter normalization, host cgroup v1/v2 reads, pool-slot reuse baselines, and Cubelet restart recovery.
* The full resource-metrics package passes under the race detector, covering bounded scheduling, concurrency saturation, stale samples, deletion, cache isolation, scope selection, and Prometheus collection.
* Generic cgroup monitor coexistence, OOM event handling, server registration, and the dedicated resource endpoint pass focused race tests.
* Reusable E2E cases are included under `tests/e2e/resource_metrics/`. They use CubeAPI for lifecycle operations, the repository SDK through CubeProxy/envd for workload execution, and `/v1/metrics/resource` for assertions.
* The exact one-click integration bundle installs Cubelet, CubeShim, cube-agent, and the guest image from one commit. The final lifecycle E2E exits 0: guest CPU and memory increase under controlled load, rollback advances the guest epoch and resets guest counters to a new baseline, host counters remain continuous, pause/resume preserves the rollback epoch, and delete removes the cached series. Sandbox, snapshot, temporary-template, containerd-task, and metric-series cleanup all complete.


Cubelet/cmd/cubelet/builtins.go
```diff
@@ -21,8 +21,6 @@
 package main
 
 import (
-	_ "github.com/containerd/containerd/v2/core/metrics/cgroups"
-	_ "github.com/containerd/containerd/v2/core/metrics/cgroups/v2"
```
fslongjin, line 25: **[P1] Why remove the generic cgroups monitor here?**

These two imports were deleted:

```go
_ "github.com/containerd/containerd/v2/core/metrics/cgroups"
_ "github.com/containerd/containerd/v2/core/metrics/cgroups/v2"
```

This feels odd to me.

`/v1/metrics/resource` is a new endpoint, and the docs say it is independent from `/v1/metrics`. Those two can coexist. I don't see why “add sandbox resource metrics” requires removing the old cgroups TaskMonitor. After this change, `container_*` series that used to show up on `/v1/metrics` may simply disappear.

The PR calls it unused, but before merge I'd like clarity on:

1. Why remove it in *this* PR, and how that relates to the PR goal?
2. Is anyone scraping `container_*`? If we keep the removal, how do upgrade notes tell users?

My preference: if this is not required, restore these side-effect imports and keep this PR focused on sandbox resource metrics. If it must go, please explain the reason and dependency check in the review thread first, then update the changelog / upgrade docs.

Cubelet/cmd/cubelet/builtins_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+import (
+	"testing"
+
+	"github.com/containerd/containerd/v2/plugins"
+	"github.com/containerd/plugin"
+	"github.com/containerd/plugin/registry"
+)
+
+func TestGenericCgroupsTaskMonitorIsNotRegistered(t *testing.T) {
```
fslongjin: **[P1 follow-up]**

This test hard-locks “the cgroups monitor must never be registered again”.

That's fine only if we have really decided to remove it permanently. Please explain the motivation together with the `builtins.go` change: if the reason is weak, this test should be reverted with those imports; if we keep the removal, please document the reason and replacement in a comment or the upgrade guide — don't leave only an assertion.

Cubelet/plugins/cube/internals/resourcemetrics/host_sandbox_sampler.go
```diff
@@ -0,0 +1,477 @@
...
+	usage, err := s.reader.UsageSnapshot(ctx, cgroupPath)
+	if err != nil {
+		s.recordError(cb.ID, token, err)
+		return
+	}
+	baseline := cb.HostMetricsBaselineCopy()
+	if baseline == nil || baseline.CGroupPath != cgroupPath {
+		previous := baseline
+		candidate := hostMetricsBaselineFromUsage(cgroupPath, usage)
+		cb.RestoreHostMetricsBaseline(&candidate)
```
fslongjin, line 271: **[P2] This is a first-sample baseline, not a cgroup-assignment baseline**

The implementation writes `HostMetricsBaseline` from the raw counters on the **first successful** `UsageSnapshot`.

But the help text / docs say *since host cgroup assignment baseline* — i.e. counting from cgroup assignment. With the current timing:

1. After the cgroup is assigned and the process is attached, CPU / throttling / memory-failure counters are already increasing;
2. The first sample can be up to one `collection_interval` later;
3. Using that moment as the baseline permanently drops the earlier delta.

Avoiding inherited history when a later sandbox reuses a cgroup path from the pool looks fine (empty baseline, or a changed cgroup path, triggers a rebuild).

What I'm worried about is the other case: default `host_sandbox` silently drops CPU / throttling / memory-failure counts between sandbox start and the first sample.

Please pick one:

1. Establish the baseline in the cgroup assignment path (e.g. after `CleanForReuse`, around process attach); the sampler should only read an existing baseline;
2. Or, if we keep this implementation for now, change the help text and docs to “since first successful sample” and stop saying assignment.

Cubelet/plugins/cube/internals/resourcemetrics/prometheus.go
```diff
@@ -0,0 +1,145 @@
...
+	guestMemoryCurrent       = prometheus.NewDesc("cubesandbox_guest_workload_memory_current_bytes", "Current memory charged to the guest workload cgroup.", guestLabels, nil)
+	guestMemoryLimit         = prometheus.NewDesc("cubesandbox_guest_workload_memory_limit_bytes", "Configured memory limit for the guest workload cgroup.", guestLabels, nil)
+	guestMemoryFailures      = prometheus.NewDesc("cubesandbox_guest_workload_memory_failures_total", "Total guest workload memory limit failures since the current metrics epoch baseline.", guestLabels, nil)
+	guestEpoch               = prometheus.NewDesc("cubesandbox_guest_workload_metrics_epoch", "Current guest workload metrics epoch generation.", guestLabels, nil)
+	guestEpochStarted        = prometheus.NewDesc("cubesandbox_guest_workload_metrics_epoch_start_time_seconds", "Start time of the current guest workload metrics epoch as Unix seconds.", guestLabels, nil)
+
+	hostCPUUsage            = prometheus.NewDesc("cubesandbox_host_sandbox_cpu_usage_seconds_total", "CPU time consumed by the current sandbox since its host cgroup assignment baseline.", hostLabels, nil)
```
fslongjin, line 31: **[P2 follow-up]**

For example:

> CPU time consumed by the current sandbox since its host cgroup assignment baseline.

Compared with the sampler, this is closer to “since first successful sample” than “since cgroup assignment completed”.

Please keep this consistent with the host-sampler comment above: either implement a real assignment-time baseline, or change this (and the EN/ZH docs) to a first-sample baseline. Otherwise users will read “from assignment” and under-count startup usage.

Cubelet/services/cubebox/rollback.go
```diff
@@ -197,11 +203,26 @@ func (s *service) RollbackSandbox(ctx context.Context, req *cubebox.RollbackSand
...
+) error {
+	setSandboxRollingBack(cb, true)
+	defer setSandboxRollingBack(cb, false)
+	if err := prepare(); err != nil {
+		return fmt.Errorf("prepare guest metrics epoch for rollback: %w", err)
+	}
+	if err := restore(); err != nil {
+		return fmt.Errorf("restore sandbox runtime: %w", err)
+	}
+	return nil
```
fslongjin, line 242: **[P2] After a failed restore, epoch stays `Prepared` — how do guest metrics recover?**

Today `prepare` first persists a new accounting window as `Prepared`, then `restore` runs; on failure it stays `Prepared` and the sampler skips it. Tests lock this in. I read the intent as: when restore outcome is uncertain, prefer exporting nothing over exporting wrong numbers.

Being conservative when restore may have already mutated the runtime is fine. There is another case though:

- `prepare` has already overwritten the previous usable baseline;
- restore fails before the runtime is actually mutated, so the sandbox may still be on the old state;
- `guest_workload` stays missing until a later successful create/rollback;
- there is no clear signal that we are stuck in `Prepared`.

Please decide the final behavior and capture it in code or docs:

1. If we can tell “runtime not mutated yet” from “uncertain”: restore the previous accounting window in the former case;
2. If we keep “on failure, never export”: document how to recover, and add searchable logs or a diagnostic metric so metrics do not disappear quietly.

Cubelet/services/cubebox/guest_metrics_epoch.go
```diff
@@ -0,0 +1,110 @@
...
+		return fmt.Errorf("prepared guest metrics epoch is required")
+	}
+	if err := cb.ActivatePreparedGuestMetricsEpoch(epoch.Generation); err != nil {
+		return err
+	}
+	setRuntimeSnapshotBindingLabels(cb, snapshotID, activatedAt)
+	setRuntimeRestoreBaseLabels(cb, snapshotID, activatedAt)
+	if err := syncer.SyncByID(ctx, cb.ID); err != nil {
+		return fmt.Errorf("persist activated guest metrics epoch for %s: %w", cb.ID, err)
+	}
```
fslongjin, line 114: **[P3]**

Here the in-memory state moves from `Prepared` to `Pending`, then `SyncByID` persists it. If persist fails:

- memory is already `Pending` (the sampler can keep going)
- disk may still be `Prepared`

If Cubelet restarts, it reloads `Prepared` from disk and guest metrics get stuck again. In the same file, prepare/ready failure paths restore in-memory state; activate does not. The caller mostly gets a warning and may still think rollback succeeded.

Not a merge blocker for me. Prefer rolling memory back or retrying persist on failure; at least mark it abnormal and emit a searchable error.

Cubelet/plugins/cube/internals/sandbox/cube_sandbox_manager.go
```diff
@@ -195,7 +203,62 @@ func (c *controllerLocal) Status(ctx context.Context, sandboxID string, verbose
...
+		return nil, fmt.Errorf("workload metrics for %s/%s are empty", sandboxID, containerID)
+	}
+	return &types.Metric{
+		Timestamp: timestamppb.Now(),
+		ID:        containerID,
+		Data:      raw,
+	}, nil
+}
+
+func guestWorkloadSnapshotFromTaskStats(
```
fslongjin: **[P3]**

`guestWorkloadSnapshotFromTaskStats` is mainly used by transport-layer tests; production collection goes through `Metrics()` and normalization in the sampler. That makes the lower-level sandbox controller depend upward on `resourcemetrics`.

Not a merge blocker. Please move it to a test helper or a separate adapter instead of leaving it on the controller.

fslongjin (changes requested):
Thanks — overall direction looks good to me.

This PR turns `Task.Stats` into Cubelet-owned sandbox CPU/memory metrics with lifecycle handling (create, rollback, and reused cgroup paths). The layering is clear: raw stats → baseline / accounting window → bounded background sampling + cache → read-only export on `/v1/metrics/resource`. I like that scrapes do not fan out per sandbox, guest/host stay separate, rollback uses prepare→restore→activate, and `host_sandbox` is the default.

Before merge, a few points I want to confirm first (details in the inline comments):

| ID | Question |
| --- | --- |
| P1 | Why remove the generic cgroups monitor in this PR? |
| P2-1 | Docs say “from cgroup assignment”, implementation is “from first sample” |
| P2-2 | After a failed restore, epoch stays `Prepared` — how do guest metrics recover? |
| P3-1 | After a failed activate persist, in-memory and on-disk state can diverge (follow-up OK) |
| P3-2 | `guestWorkloadSnapshotFromTaskStats` is unused on the production path (follow-up OK) |

I'm marking **Request changes**, mainly for these three:

1. **P1**: The new endpoint and `/v1/metrics` can coexist. I don't yet see why this PR must also remove the cgroups monitor — please explain the motivation.
2. **P2-1**: Default `host_sandbox` docs and implementation disagree; please fix the implementation or the wording.
3. **P2-2**: Preferring to export nothing on uncertain failure is fine, but please make recovery clear so metrics do not just disappear quietly.

Questions for the author:

1. Why does this PR remove the generic cgroups monitor? How does that relate to the PR goal? Is anyone scraping `container_*` from `/v1/metrics`? If we keep the removal, how will the upgrade notes call it out?
2. On the host side we can drop usage between sandbox start and the first sample (up to about one `collection_interval`). Is that acceptable? If not, can the baseline be established at cgroup assignment time?
3. After a failed restore, guest metrics may stay unavailable for a long time. Is that the intended final behavior, or should we restore the previous accounting window when the runtime was not actually mutated?

zyl1121:
@fslongjin Thanks for the detailed review. I have updated the PR to address all five points.

1. The generic containerd cgroups monitor is restored. `/v1/metrics/resource` is now strictly additive and independent from `/v1/metrics`.

2. New sandboxes now capture the raw `host_sandbox` counter baseline immediately after the cgroup pool slot is assigned and before the sandbox process is attached. The baseline is propagated through the create workflow and persisted in `CubeBox`.

   Existing sandboxes upgraded without this field retain the first-successful-sample compatibility behavior. For a newly created sandbox, if the assignment-time read fails, sandbox creation continues but `host_sandbox` fails closed rather than exporting an incomplete accounting window.

3. Rollback task lookup and local preflight now complete before the new guest epoch is prepared. A preflight failure therefore preserves the current epoch.

   Once `Task.Update` has been dispatched, Cubelet cannot safely determine whether the runtime was partially mutated. A restore failure therefore intentionally leaves the epoch `Prepared` and keeps `guest_workload` unavailable. Recovery is a later successful rollback, or delete/recreate if the sandbox cannot be recovered. The failure log now includes the sandbox context, epoch generation, failure stage, and recovery guidance.

4. If activation persistence fails, Cubelet conditionally restores the previous `Prepared` epoch and the previous snapshot/restore labels, avoiding in-memory/on-disk divergence without overwriting a concurrent state transition.

5. The transport-only normalization helper has been moved out of production controller code and into the transport test.

Validation completed on real deployment:

- focused lifecycle, cgroup, sampler, transport, and server tests passed under `-race`;
- bilingual VitePress documentation build passed;
- the full CubeAPI/CubeProxy lifecycle E2E passed for fresh CPU/memory collection, snapshot/rollback, pause/resume, deletion, and cache cleanup;
- rollback observed guest CPU `5.375190 -> 0`, guest epoch `1 -> 2`, while host CPU remained continuous at `7.391498 -> 8.836734`.

Cubelet/cmd/cubelet/builtins.go
zyl1121: > [P1] Why remove the generic cgroups monitor here?

The original removal was based on our observation that the generic monitor does not currently produce usable `container_*` series for Cube tasks and cannot provide the lifecycle semantics required by this PR.

I agree that this does not justify removing an existing containerd capability as part of the sandbox resource metrics change. I restored both side-effect imports. `/v1/metrics/resource` is now strictly additive and remains independent from `/v1/metrics`.

Any future removal of the generic monitor should be handled separately with a usage and compatibility audit.

Cubelet/cmd/cubelet/builtins_test.go
zyl1121: Agreed. Since the generic cgroups monitor is no longer being removed, I deleted this test as well.

Cubelet/plugins/cube/internals/resourcemetrics/host_sandbox_sampler.go
zyl1121: Fixed by implementing an assignment-time baseline.

Cubelet now reads the raw host counters immediately after the cgroup pool slot is assigned and before the sandbox process is attached. The baseline is carried through the cgroup create result into `CubeBox` and persisted with the sandbox metadata. The sampler uses this persisted baseline instead of dropping startup usage before its first collection.

For compatibility, a sandbox that already existed before this field was introduced may still establish a baseline from its first successful sample after upgrade.

If assignment-time capture fails for a newly created sandbox, sandbox creation continues, but the failure is persisted and `host_sandbox` fails closed for that sandbox instead of exporting an incomplete accounting window.

Cubelet/plugins/cube/internals/resourcemetrics/prometheus.go
zyl1121: The implementation now matches this wording: new sandboxes capture and persist the baseline at cgroup assignment time, before process attachment.

The first-successful-sample behavior is retained only as an upgrade compatibility path for existing sandboxes that do not have the new persisted assignment baseline. The English and Chinese documentation now describe both cases explicitly.

Cubelet/services/cubebox/rollback.go
zyl1121: I split the failure handling at the point where Cubelet can still determine whether runtime mutation has begun.

Task lookup and other local preflight checks now complete before the new epoch is prepared. A preflight failure therefore preserves the previous usable epoch and does not affect guest metrics.

Once `Task.Update` has been dispatched, Cubelet cannot reliably determine whether the restore partially mutated the runtime. In that case the new epoch intentionally remains `Prepared`, and `guest_workload` stays fail closed rather than exporting counters against a potentially invalid baseline.

Recovery is a later successful rollback, which establishes a new epoch, or deleting and recreating the sandbox if it cannot be recovered. I added a searchable error log with the sandbox context and epoch generation, and documented the recovery behavior in both operator guides.

Cubelet/services/cubebox/guest_metrics_epoch.go
zyl1121: Fixed. Before activation, Cubelet now snapshots the previous `Prepared` epoch and the existing snapshot/restore binding labels.

If persistence fails after the in-memory transition to `Pending`, it conditionally restores the previous epoch and labels. The restoration checks the expected generation and state first, so it does not overwrite a concurrent successful transition.

Cubelet/plugins/cube/internals/sandbox/cube_sandbox_manager.go
zyl1121: Fixed. `guestWorkloadSnapshotFromTaskStats` has been removed from the production sandbox controller and moved into `resource_metrics_transport_test.go` as a test-only helper.

The production controller now only exposes the standard `Metrics()` transport boundary; decoding and normalization remain owned by the resource metrics layer.

ls-ggg:
Thanks for the pr，Please supplement this PR with e2e test cases and their test results

zyl1121:
Added reusable E2E cases under `tests/e2e/resource_metrics/`.

The lifecycle test creates a sandbox through CubeAPI, runs controlled CPU and memory workloads through CubeProxy/envd, and verifies `/v1/metrics/resource` across fresh collection, snapshot/rollback, pause/resume, and deletion.

The test passed on a standard one-click deployment. CPU and memory metrics changed as expected under load, rollback started a new guest accounting window while host accounting remained continuous, pause/resume preserved the current accounting state, and deletion removed the exported series and test resources.

ls-ggg:
PR #1122 Test Results
Environment: cube-test env, Cubelet from pr-1122 (commit 3bcc510), export_scopes=["all"], collection_interval=5s.

Plugin startup: OK. Cubelet loaded resource-metrics plugin, HTTP server on port 9998, /v1/metrics/resource returns 200.

Baseline collection: OK. Created sandbox from tpl-e6c7b3ce. Guest workload emitted 16 series (cpu user/system/throttled, memory current/limit/failures, epoch). Host sandbox emitted 8 series. All populated, no empty values.

Counter increment: OK. Guest CPU 0.16→0.52→0.71 over 15s, host CPU 0.20→0.69→1.34. Monotonic, no gaps.

Rollback 1: OK. Created snapshot, waited 10s, triggered rollback. Epoch went 1→2 immediately. Guest CPU reset to 0 (new epoch baseline), then accumulated normally. Host CPU
did NOT reset — continued from pre-rollback value (1.57→2.27). This matches design intent (host baseline preserved across rollback).

Rollback 2: OK. Created second snapshot, triggered rollback again. Epoch 2→3. Guest CPU reset to 0, clean accumulation 0→0.47 over 20s. Host CPU continued accumulating (4.08→4.48). No degradation, no stuck epoch.

Error log: Zero warnings or errors in cubelet log throughout all tests. Plugin ran silently.

Conclusion: Core functionality — plugin startup, metric collection, Prometheus export, rollback epoch transitions, consecutive rollbacks — all working correctly. Feature is default-on. Safe to merge.

ls-ggg:
@fslongjin PTAL

fslongjin (approved):
LGTM

PR #1123, opened 2026-07-23
fix(volume): CubeMaster/Cubelet mixed-version volumeMounts compatibility

ls-ggg:
## Summary

- **Goal:** keep CubeMaster and Cubelet **new/old versions mutually compatible** when sandboxes use plugin `volumeMounts` (rolling upgrades / mixed clusters).
- **Bug:** newer CubeMaster injected plugin volumes as `EmptyDir(StorageMediumDefault)` placeholders plus `plugin-volume-sources`. Older Cubelets have no plugin skip path and treat every Default EmptyDir as a second rootfs derive (`sb-<id>-rootfs-gen0`), which collides with `cube_rootfs_rw` and fails create with `cubecow object already exists` (e.g. CubeAPI `130545`).
- **Fix (CubeMaster):** stop injecting Default EmptyDir for plugin volumes; wire an empty `VolumeSource` name-only entry. New Cubelets still Attach via the Volume Plugin framework + annotation; older Cubelets skip the empty source in `prepareDefaultMedium` and no longer collide on rootfs.
- **Defense (Cubelet):** keep `isPluginVolume` skip so a CubeMaster that still sends Default EmptyDir placeholders does not break a newer Cubelet.

### Compatibility matrix (intended)

| CubeMaster | Cubelet | Plugin volumeMounts create |
|---|---|---|
| new (this PR) | old | OK — no Default EmptyDir placeholder |
| new (this PR) | new | OK — annotation + plugin Attach |
| old (EmptyDir placeholder) | new | OK — Cubelet skips plugin EmptyDir |
| old (no volume plugin) | old | N/A — volumes API not used |

## Test plan

- [x] Unit: `CubeMaster/pkg/service/sandbox` plugin volume annotation test asserts no Default EmptyDir on the wire
- [x] Smoke: create sandbox with COS `volumeMounts` succeeds after deploying fixed CubeMaster
- [x] E2E: `examples/volume/tests/e2e-volume-lifecycle.sh` with `DRIVER=cos` and `DRIVER=cos-rpc`
- [x] SDK examples: `verify_volume.py` (cos / cos-rpc) and `volume_complex_concurrent_test.py`

Assisted-by: Cursor:Composer

### DIFF
--- a/CubeMaster/pkg/service/sandbox/plugin_volume_annotation_test.go
+++ b/CubeMaster/pkg/service/sandbox/plugin_volume_annotation_test.go
@@ -10,6 +10,20 @@ import (
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
 )
 
+func TestPluginVolumeWireVolume_hasNoDefaultEmptyDir(t *testing.T) {
+	t.Parallel()
+
+	v := pluginVolumeWireVolume("max-data")
+	if v.GetVolumeSource() == nil {
+		t.Fatal("VolumeSource must be non-nil so the volume is still declared")
+	}
+	if v.GetVolumeSource().GetEmptyDir() != nil {
+		t.Fatalf("plugin volumes must not use EmptyDir placeholders; old Cubelets "+
+			"derive rootfs for StorageMediumDefault EmptyDir and collide with "+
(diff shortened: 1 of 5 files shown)


tinklone:
LGTM

PR #1124, opened 2026-07-23
feat(cubeapi): add webhook event notifications

aptacc2421:
## Summary

Add asynchronous HTTP webhook delivery for sandbox lifecycle events. When a sandbox is created, paused, resumed, or deleted, CubeAPI sends a signed JSON POST to user-configured HTTP endpoints without blocking the API path.

Closes #642.

## Problem

CubeSandbox records lifecycle events internally, but external orchestration systems, audit platforms, and notification tools must poll the sandbox API to detect state transitions. Polling adds latency and unnecessary load.

## Solution

Add an `HttpLogger` backend implementing the existing `Logger` trait, plugged into `MultiLogger` alongside `FileLogger`. The design follows `FileLogger`'s proven `mpsc::unbounded_channel + tokio::spawn` pattern:

- **Zero handler changes** — lifecycle events are already emitted via `state.logger.log()`
- **Non-blocking delivery** — `log()` does a microsecond `tx.send()`; heavy work (serialization, signing, HTTP) runs in spawned tasks
- **Bounded concurrency** — `Arc<Semaphore>(64)` caps concurrent delivery tasks
- **Per-event multi-endpoint fan-out** — `join_all` sends to all configured endpoints concurrently
- **Graceful shutdown** — `flush()` drains in-flight deliveries with a 30s timeout

## Changes

| File | Description |
|------|-------------|
| `CubeAPI/Cargo.toml` | Add `hmac`, `sha2`, `hex` dependencies |
| `CubeAPI/src/config/mod.rs` | Add `webhook_urls`, `webhook_events`, `webhook_secret` fields |
| `CubeAPI/src/main.rs` | CLI flags + conditional `HttpLogger` registration |
| `CubeAPI/src/logging/http.rs` | Replace stub with full `HttpLogger` implementation |
| `CubeAPI/src/logging/mod.rs` | Update module table |
| `CubeAPI/src/handlers/sandboxes.rs` | Add `template_id` to `sandbox.resumed` event |
| `examples/webhook-receiver/` | Standalone axum receiver with HMAC verification + WeCom forwarding |
| `docs/guide/webhooks.md` | English integration guide |
| `docs/zh/guide/webhooks.md` | Chinese integration guide |

## Configuration

```bash
export CUBE_WEBHOOK_URLS="https://your-server.com/webhook"
export CUBE_WEBHOOK_EVENTS="sandbox.created,sandbox.deleted,sandbox.paused,sandbox.resumed"
export CUBE_WEBHOOK_SECRET="your-hmac-secret-key"
```

Or via CLI: `--webhook-urls`, `--webhook-events`, `--webhook-secret`.

When `CUBE_WEBHOOK_URLS` is empty, webhook delivery is completely disabled with zero overhead.

## Payload

```
POST <webhook-url>
Content-Type: application/json
X-Cube-Event: sandbox.created
X-Cube-Delivery: 550e8400-e29b-41d4-a716-446655440000
X-Cube-Signature: sha256=<hex>  (only when secret is configured)
```

```json
{
  "event": "sandbox.created",
  "timestamp": "2026-07-23T12:00:00.123Z",
  "sandbox_id": "sb-abc123",
  "template_id": "tpl-xyz"
}
```

## Validation

| Check | Result |
|-------|--------|
| `cargo test -- logging::http` | 10 passed |
| `cargo fmt --check` | passed |
| `cargo clippy` (changed files) | no issues |
| HMAC verification (correct sig → 200, wrong sig → 401) | passed |
| Sandbox lifecycle E2E (create/pause/resume/delete) | 4/4 events received |
| WeCom bot forwarding | 4/4 messages delivered |
| Retry on 5xx | confirmed (3 retries with backoff) |
| No retry on 4xx | confirmed (single attempt) |

Environment: CubeSandbox v0.5.1 on OpenCloudOS 9 (CVM, 8 vCPU / 16 GB, PVM kernel).

Full verification logs attached.

## Scope

**Included:**
- Static configuration via env vars / CLI flags
- 4 lifecycle events: `sandbox.created`, `sandbox.deleted`, `sandbox.paused`, `sandbox.resumed`
- Optional HMAC-SHA256 signing
- Exponential backoff retry (3 retries, 5xx/connection errors only)
- Runnable receiver example with HMAC verification + WeCom forwarding
- Bilingual integration documentation

**Deferred (future PRs):**
- REST API for dynamic endpoint management
- Per-endpoint event filters and secrets
- Dead-letter queue / persistent retry
- Prometheus metrics

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -67,3 +67,4 @@ __pycache__
 /tests/e2e/sdk_compat/.pytest_cache/
 
 web/tsconfig.tsbuildinfo
+examples/webhook-receiver/Cargo.lock
--- a/CubeAPI/Cargo.lock
+++ b/CubeAPI/Cargo.lock
@@ -526,6 +526,7 @@ dependencies = [
  "axum",
  "axum-extra",
  "axum-test",
+ "bytes",
  "chrono",
  "clap",
  "config",
@@ -533,9 +534,12 @@ dependencies = [
  "dotenvy",
  "futures",
  "governor",
+ "hex",
+ "hmac",
  "reqwest",
  "serde",
  "serde_json",
+ "sha2",
  "thiserror 1.0.69",
  "tokio",
  "tower 0.4.13",
@@ -589,6 +593,7 @@ checksum = "9ed9a281f7bc9b7576e61468ba615a66a5c8cfdff42420a70aa82701a3b1e292"
 dependencies = [
  "block-buffer",
  "crypto-common",
+ "subtle",
 ]
 
 [[package]]
@@ -915,6 +920,21 @@ name = "heck"
 version = "0.5.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "2304e00983f87ffb38b55b444b5e3b60a884b5d30c0fca7d82fe33449bbe55ea"
+
+[[package]]
+name = "hex"
+version = "0.4.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+
+[[package]]
+name = "hmac"
+version = "0.12.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "6c49c37c09c17a53d937dfbb742eb3a961d65a994e6bcdcf37e7399d0cc8ab5e"
+dependencies = [
+ "digest",
+]
 
 [[package]]
 name = "http"
--- a/CubeAPI/Cargo.toml
+++ b/CubeAPI/Cargo.toml
@@ -84,6 +84,14 @@ utoipa = { version = "5.4.0", features = ["chrono", "uuid", "yaml"] }
 # ── Graceful shutdown ─────────────────────────────────────────────────────
 # (tokio::signal + axum::serve().with_graceful_shutdown is sufficient)
 
+# ── Webhook delivery ─────────────────────────────────────────────────────
+bytes = "1"
+
+# ── Webhook signing ──────────────────────────────────────────────────────
+hmac = "0.12"
+sha2 = "0.10"
+hex = "0.4"
+
 [dev-dependencies]
 tokio = { version = "1", features = ["full"] }
 tower = { version = "0.4" }
--- a/CubeAPI/src/config/mod.rs
+++ b/CubeAPI/src/config/mod.rs
@@ -76,6 +76,18 @@ pub struct ServerConfig {
     /// Env var: CUBE_API_KEY
     #[serde(default)]
     pub cube_api_key: Option<String>,
+
+    /// Comma-separated webhook endpoint URLs. Env var: CUBE_WEBHOOK_URLS (default "" = disabled)
+    #[serde(default = "default_webhook_urls")]
+    pub webhook_urls: String,
+
+    /// Comma-separated event types to deliver. Env var: CUBE_WEBHOOK_EVENTS.
+    #[serde(default = "default_webhook_events")]
+    pub webhook_events: String,
+
+    /// Shared secret for HMAC-SHA256 signing. Env var: CUBE_WEBHOOK_SECRET (default "" = no signing)
+    #[serde(default = "default_webhook_secret")]
+    pub webhook_secret: String,
 }
 
 fn default_bind() -> String {
@@ -109,6 +121,17 @@ fn default_log_dir() -> String {
 fn default_log_prefix() -> String {
     "cube-api".to_string()
 }
+fn default_webhook_urls() -> String {
+    std::env::var("CUBE_WEBHOOK_URLS").unwrap_or_default()
+}
+fn default_webhook_events() -> String {
+    std::env::var("CUBE_WEBHOOK_EVENTS").unwrap_or_else(|_| {
+        "sandbox.created,sandbox.deleted,sandbox.paused,sandbox.resumed".to_string()
+    })
+}
+fn default_webhook_secret() -> String {
+    std::env::var("CUBE_WEBHOOK_SECRET").unwrap_or_default()
+}
 
 impl ServerConfig {
     pub fn from_env() -> anyhow::Result<Self> {
@@ -135,6 +158,9 @@ impl Default for ServerConfig {
             log_prefix: default_log_prefix(),
             auth_callback_url: None,
             cube_api_key: std::env::var("CUBE_API_KEY").ok().filter(|s| !s.is_empty()),
+            webhook_urls: default_webhook_urls(),
+            webhook_events: default_webhook_events(),
+            webhook_secret: default_webhook_secret(),
         }
     }
 }
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -310,7 +310,11 @@ pub async fn resume_sandbox(
     tracing::info!(sandbox_id = %sandbox_id, "resume_sandbox: success");
     state
         .logger
-        .log(LogEvent::new(LogLevel::Info, "sandbox.resumed").field("sandbox_id", &sandbox_id))
+        .log(
+            LogEvent::new(LogLevel::Info, "sandbox.resumed")
+                .field("sandbox_id", &sandbox_id)
+                .field("template_id", &sandbox.template_id),
+        )
         .await;
 
     Ok((StatusCode::CREATED, Json(sandbox)))
--- a/CubeAPI/src/logging/http.rs
+++ b/CubeAPI/src/logging/http.rs
@@ -123,7 +123,10 @@ impl HttpLogger {
                     Msg::Event(event) => {
                         let payload = build_payload(&event);
                         let body_bytes =
-                            bytes::Bytes::from(serde_json::to_string(&payload).unwrap_or_default());
+                            bytes::Bytes::from(
+                            serde_json::to_vec(&payload)
+                                .expect("serializing a freshly-built serde_json::Value never fails"),
+                        );
                         let event_name = event.event.clone();
 
                         pending_bg.fetch_add(1, Ordering::Relaxed);
--- a/CubeAPI/src/logging/http.rs
+++ b/CubeAPI/src/logging/http.rs
@@ -342,3 +342,264 @@ async fn deliver_with_retry(
 
 // ─── Tests ───────────────────────────────────────────────────────────────
 
+#[cfg(test)]
+mod tests {
+    use super::super::{LogEvent, LogLevel};
+    use super::*;
+    use axum::{routing::post, Router};
+    use std::sync::Mutex;
+
+    #[test]
+    fn build_payload_includes_event_and_timestamp() {
+        let event = LogEvent::new(LogLevel::Info, "sandbox.created")
+            .field("sandbox_id", "sb-abc")
+            .field("template_id", "tpl-xyz");
+        let payload = build_payload(&event);
+        assert_eq!(payload["event"], "sandbox.created");
+        assert!(!payload["timestamp"].as_str().unwrap().is_empty());
+        assert_eq!(payload["sandbox_id"], "sb-abc");
+        assert_eq!(payload["template_id"], "tpl-xyz");
+    }
+
+    #[test]
+    fn sign_payload_produces_deterministic_hmac() {
+        let body = b"test body";
+        let sig = sign_payload("secret", body);
+        assert!(sig.starts_with("sha256="));
+        assert_eq!(sig, sign_payload("secret", body));
+        assert_ne!(sig, sign_payload("different", body));
+    }
+
+    #[test]
+    fn redact_url_strips_path_and_query() {
+        assert_eq!(
+            redact_url("https://hooks.example.com/webhook"),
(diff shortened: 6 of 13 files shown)


aptacc2421:
## Verification

**Environment:** CubeSandbox v0.5.1 on OpenCloudOS 9 (Tencent Cloud CVM, 8 vCPU / 16 GB, PVM kernel `6.6.69-opencloudos9.cubesandbox.pvm.host`), Rust 1.85.

### 1. Unit Tests & Code Style (local)

**10/10 tests pass** — payload construction, HMAC signing, URL parsing, event filtering (pre-send), HTTP delivery (mock server), retry behavior (5xx retried, 4xx not), and graceful shutdown coordination.

**`cargo fmt --check` and `cargo clippy`** on changed files — clean.

01-cargo-test-10-passed
02-cargo-fmt-clippy

Full test output: [unit-tests.log](https://github.com/user-attachments/files/30308987/unit-tests.log)


---

### 2. Sandbox Lifecycle E2E (CVM)

All 4 sandbox lifecycle events delivered with valid HMAC signatures:

| Event | sandbox_id | template_id | HMAC |
|-------|:---:|:---:|:---:|
| `sandbox.created` | ✅ | ✅ | ✅ |
| `sandbox.paused` | ✅ | — | ✅ |
| `sandbox.resumed` | ✅ | ✅ | ✅ |
| `sandbox.deleted` | ✅ | — | ✅ |

Payload fields verified — `event`, `timestamp` (RFC 3339), `sandbox_id`, `template_id` (created/resumed only), `X-Cube-Delivery` (UUID v4), `X-Cube-Signature` (`sha256=<hex>`).

Full log (commands + output):  [webhook-e2e-verification.log](https://github.com/user-attachments/files/30309065/webhook-e2e-verification.log)

---

### 3. WeCom Bot Forwarding

All 4 events forwarded to a WeCom group bot via the `examples/webhook-receiver` adapter. Messages with `template_id` (created, resumed) show the Template field; messages without (paused, deleted) show `N/A` — matching handler behavior.

03-wecom-bot-4-messages

aptacc2421:
Thanks for the thorough review. All findings addressed:

**1. Payload field collision (Medium)** — Fixed. `build_payload` now skips reserved keys `"event"` and `"timestamp"` during field iteration.

**2. HMAC constant-time comparison (Low)** — Fixed. The example receiver now uses `hmac::Mac::verify_slice` instead of string equality. Removes the `sign_payload` helper and handles malformed signatures without panicking.

**3. Redundant event filter (Low)** — Fixed. Removed the receiver-loop filter (dead code) and the unused `subscribed_events_bg` capture.

**4. IPv6 doc (Low)** — Acknowledged. `redact_url` handles bracketed IPv6 addresses correctly; documenting every edge case inline would bloat a simple string helper.

**5. `serde_json::to_string` silent failure (Low)** — Fixed in e1e906f. Changed to `serde_json::to_vec(&payload).expect(...)` which documents the invariant (a freshly-built Value never fails to serialize) and avoids the unnecessary Value→String→Bytes round-trip.

**6. PendingGuard / Notify-based wakeup (Low)** — Intentionally deferred. The delivery path has no panic sites (all `await`, all `Result`). The 100ms poll loop only runs during graceful shutdown — not on the hot path. Adding a `Notify` primitive to eliminate an unnoticeable shutdown delay would be over-engineered for MVP.

fslongjin (changes requested):
Hello, thank you for your contribution. I believe this PR has basically met the requirements of the Rhino-bird project and is generally acceptable. However, for merging the PR, since we need to consider high availability and architectural soundness, this PR currently does not meet those standards. Our expectations are:

1. CubeAPI should support horizontal scaling and high availability (the current implementation is not friendly to this);
2. The corresponding webhook configurations should be able to integrate with the operations system.

In other words, this entire area requires further and better design.

That said, for the "Rhino-bird activity," the current PR is already sufficient. If you would like to bring it to a production-ready level, you are welcome to continue discussing under this PR.

aptacc2421:
> Hello, thank you for your contribution. I believe this PR has basically met the requirements of the Rhino-bird project and is generally acceptable. However, for merging the PR, since we need to consider high availability and architectural soundness, this PR currently does not meet those standards. Our expectations are:
> 
> 1. CubeAPI should support horizontal scaling and high availability (the current implementation is not friendly to this);
> 2. The corresponding webhook configurations should be able to integrate with the operations system.
> 
> In other words, this entire area requires further and better design.
> 
> That said, for the "Rhino-bird activity," the current PR is already sufficient. If you would like to bring it to a production-ready level, you are welcome to continue discussing under this PR.

Hi , @fslongjin ! Thank you for the clarification. I agree that the current per-instance in-memory
queue is not suitable for multi-replica production deployments, and webhook
configurations should be manageable at runtime rather than through CLI flags.

Before changing the code, I'd like to propose the following direction and
confirm the component boundaries with you.

My understanding of the requirements:

1. **High availability** — if a CubeAPI Pod crashes, pending webhook events
   should remain available and another replica should be able to continue
   delivery.

2. **Ops integration** — webhook endpoints, event subscriptions, enabled
   state, and HMAC secrets should be manageable through the operations system
   without restarting CubeAPI.

**Proposed architecture:**

```mermaid
flowchart TB
    REQ["Client / SDK"] -->|"POST /sandboxes"| API["CubeAPI"]
    API -->|"HTTP POST /cube/sandbox"| MASTER["CubeMaster"]
    MASTER -->|"lifecycle hooks\nPublishCreate / Delete / State"| LIFECYCLE["lifecycle.Store"]
    LIFECYCLE -->|"XADD"| REDIS[("Redis Stream\nlifecycle:events")]
    REDIS -->|"XREADGROUP\nXAUTOCLAIM"| CONSUMER["CubeAPI Webhook Consumer\n（每个副本一个，共享消费者组）"]
    CONSUMER -->|"HTTP POST\nHMAC-SHA256"| ENDPOINT["Webhook Endpoint"]
    CONSUMER -->|"terminal failed\ndelivery"| DL[("Webhook Dead-letter Stream")]

    subgraph CONFIG["Configuration"]
        OPS["CubeOps"] -->|"CRUD API"| DB[("DB\nt_webhook_subscription")]
        DB -.->|"config distribution TBD\n(CubeMaster / CubeOps)"| CONSUMER
    end
```

**Key changes from the current implementation:**

- **Event source**: Instead of depending on each CubeAPI handler's
  `logger.log()` call, CubeMaster lifecycle transitions become the
  authoritative event source, with Redis Stream used as the shared delivery
  transport. This decouples webhook delivery from the specific API replica
  that handled the request.

- **Delivery**: A Redis Stream consumer group runs in each CubeAPI replica.
  Replicas use unique consumer names and reclaim abandoned pending entries
  via XAUTOCLAIM when a consumer terminates unexpectedly.

- **Reuse existing logic**: The Rust HTTP client, HMAC signing, event
  filtering, concurrency control, and retry logic from the current
  HttpLogger are reused. The delivery ID will change from a random UUID
  to a stable hash (stream_entry_id + subscription_id) for deduplication.

- **Delivery semantics**: At-least-once with a stable delivery ID (derived
  from stream_entry_id + subscription_id) for receiver-side deduplication.
  A Stream entry is only acknowledged after all matched subscriptions reach
  a terminal result (success, non-retryable 4xx, or max retries). The entry
  remains pending while delivery and retry are in progress. Terminal failures
  go to a dead-letter stream and are recorded through structured logs.

- **Event schema adaptation**: The lifecycle Stream schema differs from the
  current webhook payload. In particular, the CubeMaster lifecycle hook
  currently reads `req.Annotations["template_id"]`, while CubeAPI sends the
  template identity under `cube.master.appsnapshot.template.id`. Therefore,
  create Stream events may currently omit `template_id`. The implementation
  would normalize this field using the existing annotation constant. State
  events (`paused`/`running`) would also be mapped to the public webhook
  event names (`sandbox.paused`/`sandbox.resumed`).

- **Configuration**: Webhook subscriptions are stored in a dedicated DB
  table and managed through authenticated CubeOps CRUD APIs. CubeAPI loads
  configuration at startup and refreshes it at runtime, so no restart is
  needed after configuration changes.

I would like to confirm three architectural decisions:

1. Is it acceptable for CubeAPI to connect directly to Redis for lifecycle
   event consumption?

2. Should webhook delivery reuse the existing lifecycle Stream, or use a
   dedicated Stream with its own retention policy? The existing Stream
   currently has bounded retention (`MAXLEN ~ 100000`) and best-effort
   publication (CubeMaster swallows XADD errors), so I would like to confirm
   whether its contract may be extended for webhook delivery or a dedicated
   Stream is preferred.

3. Should CubeAPI retrieve subscription configuration through CubeMaster,
   CubeOps, or another existing configuration-distribution mechanism?

After these decisions are confirmed, I will post a file-level implementation
and migration plan under this PR before changing the code. The validation
plan will include two CubeAPI replicas, consumer termination and pending-
entry reclaim, endpoint failure and retry, duplicate delivery, and runtime
configuration updates.

fslongjin:
I don’t think it’s necessary for the Cube API to connect directly to Redis to consume events. Publishing them into Redis as an event stream is a solid approach. Instead, I believe this delivery responsibility should fall to operational systems like CubeMaster or CubeOps. Also, wouldn’t it be more appropriate to implement sandbox lifecycle–related events within the Cube Lifecycle Manager itself?
@aptacc2421

aptacc2421:
> I don’t think it’s necessary for the Cube API to connect directly to Redis to consume events. Publishing them into Redis as an event stream is a solid approach. Instead, I believe this delivery responsibility should fall to operational systems like CubeMaster or CubeOps. Also, wouldn’t it be more appropriate to implement sandbox lifecycle–related events within the Cube Lifecycle Manager itself? @aptacc2421


Hi @fslongjin, thank you. I agree that CubeAPI should not connect to Redis or own Webhook delivery. I will remove it from the proposed delivery path and use successful lifecycle transitions published to Redis as the shared event source.

Before reworking the PR, I would like to confirm the intended product scope and component boundary.

[Issue #642](https://github.com/TencentCloud/CubeSandbox/issues/642) refers to “user-specified HTTP endpoints” and lists Agent orchestration systems, operations platforms, and enterprise IM tools as possible receivers. However, its baseline configuration mechanism is a configuration file or environment variables, with REST management being optional. My current PR therefore implements deployment-level, operator-managed endpoints rather than per-user or per-tenant subscriptions.

These two possible long-term directions lead me to different designs:

- If Webhooks remain operator-managed integrations, my preference is for CubeOps to manage subscriptions and run the background delivery worker.
- If user/tenant-managed Webhooks are an intended future direction, my preference is a dedicated Webhook Dispatcher as an independent delivery plane, with subscriptions managed through an appropriate user-facing control plane rather than the CubeOps admin backend.

Regarding Cube Lifecycle Manager, its current documented responsibility is auto-pause/auto-resume coordination between CubeMaster, CubeProxy, and Redis. I understand why lifecycle event semantics belong close to it, and I would treat events as actual successful transitions regardless of whether they were initiated through CubeAPI or automatically by lifecycle management.

My concern is not that CLM uses HTTP—it already communicates with internal CubeSandbox components over HTTP. The concern is that delivery to configured external endpoints is a separate responsibility from internal lifecycle coordination. Hosting it in CLM would change the component from a focused lifecycle coordinator into both a coordinator and an external integration dispatcher. It would also make CLM own endpoint URLs, secrets, retry state, and outbound HTTP to external systems, increasing its complexity and coupling two different failure and scaling profiles.

Could you confirm whether you intend CLM to own only lifecycle event processing while another component performs delivery, or whether you expect CLM itself to perform external Webhook delivery? Also, should the long-term subscription model remain operator-managed, or should the design preserve a path toward user/tenant-managed subscriptions?

Once these boundaries are confirmed, I will update the architecture and post a concrete implementation and validation plan before changing the code.

fslongjin:
@aptacc2421 I believe we should distinguish between the event producer​ and the webhook sender.

aptacc2421:
HI, @fslongjin . # Event Producer and Webhook Sender

My understanding, based on the current component responsibilities, is:

```mermaid
flowchart LR
    ACTION["API request or automatic lifecycle action"] --> MASTER["CubeMaster<br/>executes transition + produces event"]
    MASTER -->|"success hook: XADD lifecycle event"| STREAM[("Redis Stream")]

    STREAM -->|"consumer group: lifecycle"| CLM["Cube Lifecycle Manager"]
    CLM -->|"internal meta/state sync"| PROXY["CubeProxy"]

    STREAM -->|"separate consumer group (proposed)"| SENDER["Webhook sender<br/>CubeOps Worker or Dispatcher"]
    SENDER -->|"filter · sign · retry · HTTP POST"| ENDPOINT["Configured external endpoint"]
```

- CubeMaster produces the lifecycle event.
- CLM consumes it for internal lifecycle coordination; it does not relay it to the Webhook sender.
- The proposed Webhook sender independently consumes the same Stream and delivers the external callback.

## Blocking decision

The remaining implementation blocker is the intended subscription ownership model:

- **CubeOps background worker** — for deployment-level, operator-managed integrations.
- **Separate Webhook Dispatcher** — if user/tenant-managed subscriptions, independent scaling, or stronger failure isolation are expected.

The receiver may be an Agent orchestration system, operations platform, audit system, or enterprise IM endpoint in either model. The distinction is who owns and manages the subscription.

Should subscriptions remain deployment-level and operator-managed, or should the architecture preserve a path toward user/tenant-managed subscriptions? Once this is confirmed, the Webhook sender placement can be finalized.

PR #1139, opened 2026-07-24
fix(dev-sidecar): strip stale Content-Encoding after decompression

zyl1121:
## Motivation

The dev sidecar can currently forward an invalid response for gzip or deflate encoded upstream responses. `aiohttp.ClientSession` automatically decompresses the response body, but the sidecar keeps the original upstream `Content-Encoding` header. Downstream clients such as `httpx` then try to decompress the already-decoded body again. For gzip responses, this fails with:

```text
httpx.DecodingError: incorrect header check
```

This PR fixes the stale response header while keeping `aiohttp` automatic decompression enabled.

## Reproduction

Proxy a gzip-encoded response larger than CubeProxy's 1,000-byte gzip threshold through the dev sidecar, then read it with an automatically decoding client such as `httpx`.

Before this fix, the request deterministically fails with the decoding error above because the downstream client attempts to decompress bytes that `aiohttp` has already decoded.

## What Changed

* Stop forwarding stale upstream `Content-Encoding` for gzip and deflate responses that `aiohttp` has already decompressed.
* Continue omitting the upstream compressed `Content-Length`, since it no longer describes the forwarded response body.
* Preserve optional or unsupported `Content-Encoding` values when automatic decompression is not guaranteed.
* Add regression coverage for gzip- and deflate-encoded file and Connect responses larger than CubeProxy's gzip threshold.
* Verify that decoded responses have consistent headers and unsupported encodings remain intact.

## Validation

* Confirmed the regression test fails on the previous implementation with:

```text
httpx.DecodingError: incorrect header check
```

* Ran the current fixed dev-sidecar test file:

```text
12 passed
```

* Ran Python syntax compilation for the source and test files.
* Verified the updated sidecar against a K3s deployment using the official E2B SDK:

  * commands completed successfully
  * file write, existence check, read, and directory listing succeeded
  * code interpreter state and file access succeeded
  * signed download and upload URLs returned HTTP 200

## Scope

This PR only fixes HTTP response forwarding in `examples/e2b-dev-sidecar`.

### DIFF
--- a/examples/e2b-dev-sidecar/dev_sidecar.py
+++ b/examples/e2b-dev-sidecar/dev_sidecar.py
@@ -27,6 +27,9 @@ HOP_BY_HOP_HEADERS = {
     "transfer-encoding",
     "upgrade",
 }
+# ClientSession auto-decompresses these by default but preserves the
+# upstream Content-Encoding response header.
+AUTO_DECOMPRESSED_ENCODINGS = {"deflate", "gzip"}
 WEBSOCKET_REQUEST_HEADERS = {
     hdrs.CONNECTION,
     hdrs.SEC_WEBSOCKET_ACCEPT,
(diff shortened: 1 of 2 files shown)


fslongjin:
Hi there! Does #950 resolve your issue? ~

zyl1121:
> Hi there! Does #950 resolve your issue? ~

Yes, it resolves the same issue with a more general content-encoding forwarding fix. Thanks for pointing it out, I’ll close this PR as superseded.

PR #1144, opened 2026-07-24
feat(python-sdk): expose distribution scope for sandbox creation

luzhixing12345:
CubeAPI already supports restricting sandbox placement through the distributionScope field, but the Python SDK only allowed it to be passed implicitly through **kwargs.

Add an explicit distribution_scope parameter to Sandbox.create(), map it to the backend distributionScope field.

### DIFF
--- a/sdk/python/README.md
+++ b/sdk/python/README.md
@@ -106,6 +106,18 @@ sb.pause(timeout=60, interval=0.5) # custom poll params
 sb2 = Sandbox.connect(sb.sandbox_id)
 ```
 
+### Compute node placement
+
+Restrict scheduling to one or more compute node IDs or host IPs. A single
+entry pins the sandbox to that node; creation fails if the node cannot run the
+requested template.
+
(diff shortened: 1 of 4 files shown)


wbzdssm:
1. Inconsistent type annotation: `distribution_scope: List[str] | None` should be changed to `list[str] | None` for consistency with the existing `volume_mounts: list[VolumeMountArg] | None` in the same method. Also, remove the unused `typing.List` import.

2. Add a test for the missing parameter: Add a test case to verify that `distributionScope` is excluded from the request payload when `distribution_scope` is omitted (or set to `None`). This prevents regressions if the "falsiness" check is modified in the future (refer to the existing `test_create_default_lifecycle_omitted`).

3. Unrelated README change: The newly added `## Test` section is unrelated to the PR's purpose. Consider splitting this into a separate commit to keep the diff clean.

luzhixing12345:
modified as suggested

PR #1160, opened 2026-07-25
feat(CubeAPI): add lifecycle webhook delivery

bqx619:
## Summary

- Add configurable webhook subscriptions for `sandbox.created`, `sandbox.deleted`, `sandbox.paused`, and `sandbox.resumed` lifecycle events.
- Deliver webhook payloads asynchronously through a bounded queue with configurable concurrency, request timeouts, exponential-backoff retries, and optional HMAC-SHA256 signatures.
- Include `timestamp`, `sandbox_id`, and available `template_id` context, and emit events only after successful lifecycle operations.
- Support multiple independently filtered endpoints through environment configuration, including one-click deployment integration and secret redaction.
- Add a dependency-free Python receiver example with signature verification and optional WeCom forwarding, plus bilingual integration guides.
- Add unit and route-level coverage for subscriptions, fan-out, signing, retries, queue saturation, lifecycle events, and non-blocking delivery.

Ref https://github.com/TencentCloud/CubeSandbox/issues/642


fslongjin:
Hello, thank you for your contribution. I believe this PR has basically met the requirements of the Rhino-bird project and is generally acceptable. However, for merging the PR, since we need to consider high availability and architectural soundness, this PR currently does not meet those standards. Our expectations are:

1. CubeAPI should support horizontal scaling and high availability (the current implementation is not friendly to this);
2. The corresponding webhook configurations should be able to integrate with the operations system.

In other words, this entire area requires further and better design.

That said, for the "Rhino-bird activity," the current PR is already sufficient. If you would like to bring it to a production-ready level, you are welcome to continue discussing under this PR.

bqx619:
> Hello, thank you for your contribution. I believe this PR has basically met the requirements of the Rhino-bird project and is generally acceptable. However, for merging the PR, since we need to consider high availability and architectural soundness, this PR currently does not meet those standards. Our expectations are:
> 
> 1. CubeAPI should support horizontal scaling and high availability (the current implementation is not friendly to this);
> 2. The corresponding webhook configurations should be able to integrate with the operations system.
> 
> In other words, this entire area requires further and better design.
> 
> That said, for the "Rhino-bird activity," the current PR is already sufficient. If you would like to bring it to a production-ready level, you are welcome to continue discussing under this PR.

Thank you for your recognition. I will set aside time to further refine the design based on your requirements and implement it until this PR reaches a production-ready, mergeable state—this is very meaningful to me.

bqx619:
A possible direction for discussion could be:

1. Replace the per-instance in-memory queue with a shared Redis Stream. CubeAPI replicas would publish to the same stream and consume through a shared consumer group, allowing unfinished deliveries to be recovered by another replica. Delivery would remain at-least-once with a stable delivery ID for deduplication.

2. Allow CubeAPI to load versioned webhook configuration from an operations-system endpoint. All replicas would use the same configuration source, refresh it through a revision or ETag, and retain the last known good configuration during temporary failures. The existing environment configuration could remain as a local-development fallback.

```text
Operations System -> Webhook Configuration
                              |
CubeAPI Replicas -> Redis Stream -> Delivery Workers -> Webhook Endpoints
```

@fslongjin Do you think this proposal is feasible?

fslongjin:
> A possible direction for discussion could be:
> 
> 1. Replace the per-instance in-memory queue with a shared Redis Stream. CubeAPI replicas would publish to the same stream and consume through a shared consumer group, allowing unfinished deliveries to be recovered by another replica. Delivery would remain at-least-once with a stable delivery ID for deduplication.
> 2. Allow CubeAPI to load versioned webhook configuration from an operations-system endpoint. All replicas would use the same configuration source, refresh it through a revision or ETag, and retain the last known good configuration during temporary failures. The existing environment configuration could remain as a local-development fallback.
> 
> ```
> Operations System -> Webhook Configuration
>                               |
> CubeAPI Replicas -> Redis Stream -> Delivery Workers -> Webhook Endpoints
> ```
> 
> @fslongjin Do you think this proposal is feasible?

I'm wondering if it would be more reasonable to handle this within CubeMaster or the Cube Lifecycle Manager. After all, we're dealing with features like auto-suspension and resumption, which don't actually go through the Cube API.

bqx619:
> > A possible direction for discussion could be:
> > 
> > 1. Replace the per-instance in-memory queue with a shared Redis Stream. CubeAPI replicas would publish to the same stream and consume through a shared consumer group, allowing unfinished deliveries to be recovered by another replica. Delivery would remain at-least-once with a stable delivery ID for deduplication.
> > 2. Allow CubeAPI to load versioned webhook configuration from an operations-system endpoint. All replicas would use the same configuration source, refresh it through a revision or ETag, and retain the last known good configuration during temporary failures. The existing environment configuration could remain as a local-development fallback.
> > 
> > ```
> > Operations System -> Webhook Configuration
> >                               |
> > CubeAPI Replicas -> Redis Stream -> Delivery Workers -> Webhook Endpoints
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > @fslongjin Do you think this proposal is feasible?
> 
> I'm wondering if it would be more reasonable to handle this within CubeMaster or the Cube Lifecycle Manager. After all, we're dealing with features like auto-suspension and resumption, which don't actually go through the Cube API.

@fslongjin Thanks for the suggestion. I agree.A more reasonable direction could be:

- CubeMaster publishes events to the existing lifecycle event stream after successful creation, deletion, and manual pause or resume operations.
- The Cube Lifecycle Manager publishes automatic pause and resume results to the same stream.
- The Cube Lifecycle Manager runs a separate webhook consumer group and background delivery workers, so webhook delivery cannot block its existing lifecycle processing.
- The delivery workers handle subscription filtering, HMAC signing, retries, and failure logging.

Would this responsibility split be more consistent with the intended architecture?

Issue #927, opened 2026-07-12
[Bug Report] dev-env cube-autostart.sh starts legacy oneclick unit and conflicts with systemd-managed one-click deployment


xboHodx:
## Summary

`dev-env/cube-autostart.sh` still enables the legacy `cube-sandbox-oneclick.service`, which runs `up-with-deps.sh`. With the current systemd-managed one-click deployment, this conflicts with `cube-sandbox-control.target` and can leave the VM in a mixed state.

`dev-env/cube-autostart.sh` 仍然启用旧的 `cube-sandbox-oneclick.service`，该 unit 会执行 `up-with-deps.sh`。但当前 one-click 安装已经由 `cube-sandbox-control.target` 和一组分组件 systemd service 管理，这会导致两套启动路径互相冲突，并可能把 VM 留在混合状态。

## Environment

- CubeSandbox version / commit:
  - Release package: `v0.5.1-amd64`
  - Release git commit: `a164417f497234a0d787cb328b0ae96480b1569b`
- Host OS and kernel version:
  - Ubuntu 24.04.4 LTS
  - `Linux 6.8.0-134-generic x86_64`
- Guest OS:
  - OpenCloudOS 9.6
  - `Linux 6.6.119-49.21.oc9.x86_64`
- Deployment mode: single-node dev-env VM
- Relevant component: `dev-env/cube-autostart.sh`, `cube-sandbox-oneclick.service`, `cube-sandbox-control.target`, systemd-managed one-click deployment, CoreDNS/cube-proxy startup

## Steps to Reproduce

1. Start and log into a dev-env VM:

   ```bash
   cd dev-env
   sudo ./run_vm.sh
   sudo ./login.sh
   ```

2. Inside the VM, complete a current one-click installation successfully:

   ```bash
   curl -sL https://cnb.cool/CubeSandbox/CubeSandbox/-/git/raw/master/deploy/one-click/online-install.sh | MIRROR=cn bash
   ```

   After installation, the systemd-managed target is enabled/active:

   ```bash
   systemctl is-enabled cube-sandbox-control.target
   systemctl is-active cube-sandbox-control.target
   ```

3. On the host, run:

   ```bash
   cd dev-env
   sudo ./cube-autostart.sh
   ```

4. Confirm the prompt to enable autostart.

## Expected Behavior

[...]

PR #1173, opened 2026-07-27
refactor(dev-env): refactor dev-env scripts to use systemd target for CubeSandbox

xboHodx:
## 概述 / Summary

https://github.com/TencentCloud/CubeSandbox/issues/927

- 将 dev-env 自启动管理从旧的 `cube-sandbox-oneclick.service` 迁移到 `cube-sandbox-control.target`。
- 停止在 `prepare_image.sh` 中安装旧 unit，并更新 `sync_to_vm.sh` 的重启提示。
- 增加旧 unit 迁移逻辑、自动化测试及中英文文档，明确废弃的脚本和 `sudo` 使用方式。

- Migrate dev-env autostart management from the legacy `cube-sandbox-oneclick.service` to `cube-sandbox-control.target`.
- Stop installing the legacy unit during image preparation and update the restart hint in `sync_to_vm.sh`.
- Add legacy-unit migration tests and update the English and Chinese documentation.

## 测试 / Testing

- `bash dev-env/tests/test_systemd_autostart.sh`
- `bash -n dev-env/*.sh dev-env/internal/*.sh dev-env/tests/*.sh`
- Completed a one-click installation in the dev VM.
- Confirmed the control target is enabled and active, all 14 component services are active, the legacy unit is absent, and `quickcheck.sh` reports `OK`.

Assisted-by: Codex:GPT-5

### DIFF
--- a/deploy/one-click/README.md
+++ b/deploy/one-click/README.md
@@ -17,8 +17,23 @@ This directory is used to build and deliver the single-machine one-click release
 - `smoke.sh`: Runs basic health checks.
 - `env.example`: Shared environment variable template for both the build machine and the target machine.
 - `lib/common.sh`: Common shell utility functions.
-- `scripts/one-click/`: Validation and maintenance helpers used by the systemd-managed deployment after installation.
+- `scripts/one-click/`: Mixed-purpose historical directory containing deprecated aggregate entry points, compatibility scripts, and helpers still used by systemd and Terraform. See "Script support status" below.
 - `terraform/tencentcloud/`: Terraform deployer for a **clustered** CubeSandbox on Tencent Cloud (TKE control plane + CVM compute nodes). `create.sh` is the entry point; `destroy.sh` tears everything down. These files are shipped both at the release-bundle top level and inside `sandbox-package` (see "Tencent Cloud Cluster Deployment").
+
+### Script support status
+
+Do not treat every script under `scripts/one-click/` as deprecated. The current
+status is:
+
+| Status | Scripts | Usage |
+|---|---|---|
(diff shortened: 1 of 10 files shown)


fslongjin:
Thanks for the work on this — I appreciate the effort to clean up the legacy oneclick unit path.

However, I don't think we should continue investing in this direction for `dev-env`.

I've stated this before:

- On [#949](https://github.com/TencentCloud/CubeSandbox/pull/949#issuecomment-4967275339): `dev-env` is still useful for testing Cube on a personal machine, but **it only needs to spin up an OpenCloudOS 9 VM and forward ports such as SSH and cube-api to the host**. That is sufficient.
- On [#927](https://github.com/TencentCloud/CubeSandbox/issues/927#issuecomment-4989006932): treat the guest as a normal VM, and use `scp` (or similar) to copy files into it. The old `up-with-deps.sh` style orchestration is deprecated; we already manage services with systemd on a real deployment.

In other words: `dev-env` should stay a thin VM launcher + port forwarder, not a second deployment / service-management system that mirrors production one-click behavior.

With AI agents, it's already straightforward to SSH into a plain VM and perform install / sync / restart there. Building more autostart migration logic, legacy-unit handling, and systemd-target orchestration into `dev-env` is over-engineering for what should remain a simple local testing environment.

I'd prefer we **not merge** this kind of refactor. If anything, we should simplify `dev-env` toward "boot VM + SSH/port forward", and leave service lifecycle to the normal one-click / systemd path inside the guest (driven manually or by an agent over SSH).

Happy to discuss if there's a concrete developer workflow this PR unblocks that the simpler model can't cover.

PR #1174, opened 2026-07-27
feat(qos): add template-level sandbox QoS

zyl1121:
## Motivation

Multiple sandboxes can share the same host, network egress path, and storage resources. Without per-sandbox ceilings, one sandbox can consume disproportionate network bandwidth, packet-processing capacity, block throughput, or IOPS and affect other workloads.

CubeSandbox already has virtio-net and virtio-blk Token Bucket enforcement paths, but using them requires callers to construct internal annotations and low-level bucket fields. Those annotations are runtime contracts, are filtered by annotation policy, and should not be exposed as the public QoS interface.

This PR adds an opt-in, template-managed QoS contract for explicit per-sandbox network and block I/O ceilings. Templates that omit `qos` retain the existing behavior.

## What Changed

- Add the public template QoS shape:

  ```json
  {
    "qos": {
      "network": {
        "bandwidthMbps": 100,
        "packetsPerSecond": 5000
      },
      "blockIo": {
        "throughputMiBps": 64,
        "iops": 1000
      }
    }
  }
  ```

- Accept `qos` through CubeAPI template creation and `cubemastercli tpl create-from-image --qos`.

- Support both inline JSON and `--qos @qos.json` in the CLI, with strict field validation and bounded file input.

- Resolve and persist the public QoS setting in CubeMaster, then translate network limits to the versioned `cube.master.net` request and block I/O limits to `cube.master.blk.qos` for the existing Cubelet and CubeShim Token Bucket paths.

- Preserve template-managed QoS when committing a sandbox with a public `create_request` override. If the source sandbox specification cannot be loaded, fail the commit instead of silently creating a template without its ceilings.

- Keep QoS as runtime template configuration rather than a rootfs input, so templates with different limits can reuse the same rootfs artifact.

- Centralize template QoS resolution behind a resource-aware seam. The current implementation does not infer limits when `qos` is omitted; the seam allows platform defaults to be derived from template CPU and memory later, while explicit fields override inferred fields.

- Reject caller-supplied internal network, block I/O, and filesystem QoS annotations on sandbox and template requests, and remove these annotations from public template, preview, and snapshot responses.

- Tighten Cubelet validation for versioned network QoS requests by rejecting unknown or mis-cased fields, unsupported versions or modes, missing QoS, and negative or incomplete Token Bucket values. Legacy version-zero payloads remain permissive for compatibility.

- Expose `configuredQos` for templates and sandboxes, plus `qosApplied` for persisted sandbox runtime state, and update `cubemastercli` output and the English and Chinese QoS guides.

- Add opt-in SDK compatibility E2E coverage for template and sandbox state, bidirectional bandwidth enforcement, block throughput and IOPS, and node-observed PPS enforcement.

## Behavior and Scope

- The current release configures QoS on templates. Every sandbox created from a template receives the same configured ceilings, and sandbox creation cannot override or disable them.
- QoS is disabled when the template omits `qos`; existing templates keep their current behavior.
- Network bandwidth and packet-processing limits are applied independently to upload and download, using the same configured value for both directions.
- Block throughput and IOPS limits are applied independently to each virtio-blk device. Reads and writes on the same device share its limits.
- Refill and burst fields remain platform-managed implementation details.
- These settings provide ceilings, not reservations, minimum performance, priorities, or weighted fairness. Sandboxes can still contend when the underlying host, storage, or network capacity is below the sum of their ceilings.

This PR does not add separate upload/download values, separate read/write values, per-sandbox overrides, filesystem QoS, or non-empty CPU/memory-derived defaults.

## How to run E2E

Install the SDK compatibility E2E dependencies:

```bash
cd tests/e2e/sdk_compat
python -m pip install -r requirements.txt
```

The control/runtime E2E creates a QoS template and sandbox, verifies `configuredQos` and `qosApplied`, and cleans both resources. The QoS file manages its own template, so selecting it alone does not require `CUBE_TEMPLATE_ID`.

```bash
CUBE_API_URL=http://127.0.0.1:3000 \
CUBE_TEMPLATE_E2E_IMAGE=cube-sandbox-cn.tencentcloudcr.com/cube-sandbox/sandbox-code:latest \
pytest --run-e2e \
  cases/qos/test_template_qos.py::test_template_qos_control_and_runtime_state
```

For bandwidth enforcement, start an iperf3 server at an IP reachable from the sandbox:

```bash
iperf3 -s -p 5201
```

Then run the bandwidth and block I/O dataplane cases. The suite creates a temporary template with 100 Mbps network bandwidth, 32 MiB/s block throughput, and 1000 IOPS by default, creates one sandbox, and cleans up both resources. When `CUBE_QOS_IPERF_SERVER` is an IP literal, the temporary template automatically adds the corresponding host CIDR to `allow_out`.

```bash
CUBE_API_URL=http://127.0.0.1:3000 \
CUBE_TEMPLATE_E2E_IMAGE=cube-sandbox-cn.tencentcloudcr.com/cube-sandbox/sandbox-code:latest \
CUBE_QOS_IPERF_SERVER=10.0.0.10 \
CUBE_QOS_DATAPLANE_E2E=1 \
CUBE_QOS_E2E_INSTALL_TOOLS=1 \
pytest --run-e2e \
  -m 'qos and not qos_node' \
  cases/qos/test_template_qos.py
```

The pytest runner does not need `iperf3` or `fio`, but the sandbox does. If either command is missing, its case is skipped unless `CUBE_QOS_E2E_INSTALL_TOOLS=1` is set; `CUBE_QOS_E2E_INSTALL_COMMAND` can override the installation command. The block I/O case also skips unless its test file is on a writable virtio-blk-backed filesystem.

To reuse existing templates, set `CUBE_QOS_CONTROL_TEMPLATE_ID` for the control/runtime case, `CUBE_QOS_E2E_TEMPLATE_ID` for the bandwidth and block I/O cases, or `CUBE_QOS_PPS_TEMPLATE_ID` for the PPS case. A reused template must already allow the required test endpoint and match the QoS values expected by that case.

The PPS case must run on the sandbox compute node because it reads the sandbox TAP counter. `CUBE_QOS_IPERF_SERVER` identifies the compute-node IP reachable from the sandbox; this case uses its own UDP sink rather than iperf3.

```bash
CUBE_API_URL=http://127.0.0.1:3000 \
CUBE_TEMPLATE_E2E_IMAGE=cube-sandbox-cn.tencentcloudcr.com/cube-sandbox/sandbox-code:latest \
CUBE_TEMPLATE_E2E_NODE=current-node-id \
CUBE_QOS_IPERF_SERVER=10.0.0.10 \
CUBE_QOS_DATAPLANE_E2E=1 \
CUBE_QOS_NODE_PPS_E2E=1 \
CUBE_QOS_PACKETS_PER_SECOND=100 \
pytest --run-e2e -m qos_node cases/qos/test_template_qos.py
```

## Validation

- Focused CubeMaster tests passed for public QoS validation and conversion, template resolution and persistence, annotation rejection and hiding, sandbox reporting, and CLI parsing.
- Cubelet tests passed for strict version-one network QoS decoding, legacy version-zero compatibility, TAP request propagation, and block I/O annotation decoding.
- CubeAPI QoS tests and SDK compatibility regression tests passed.
- Linux builds passed for the affected CubeMaster, Cubelet, CLI, and CubeAPI paths.
- The control/runtime E2E passed with network bandwidth, PPS, block throughput, and IOPS present in `configuredQos`, and sandbox `qosApplied=true`.
- Live dataplane E2E passed for upload and download bandwidth, block throughput and IOPS, and node-observed PPS enforcement.
- A no-QoS template compatibility E2E passed: the template and sandbox reported no configured or applied QoS, no internal `cube.master.net` annotation was exposed, and a one-sandbox iperf upload was not capped by the QoS limiter.
- Five concurrent sandboxes from a 10 Mbps template each remained between 9.907 and 9.993 Mbps in two runs, including one sandbox using eight connections while the other four used one connection each.
- With the same workload and no QoS, the eight-connection sandbox consumed 88-160 Mbps of an approximately 204 Mbps sustained shared link, while normal sandboxes received an unstable 6.10-38.07 Mbps each.

These results verify the public configuration path, compatibility for templates that omit QoS, and per-sandbox blast-radius control.

### DIFF
--- a/CubeAPI/src/cubemaster/mod.rs
+++ b/CubeAPI/src/cubemaster/mod.rs
@@ -1168,6 +1168,10 @@ pub struct GetSandboxDataItem {
     #[serde(default)]
     pub labels: HashMap<String, String>,
     #[serde(default)]
+    pub configured_qos: Option<CubeQosConfig>,
+    #[serde(default)]
+    pub qos_applied: bool,
+    #[serde(default)]
     pub containers: Vec<GetSandboxContainerItem>,
     #[serde(default)]
     pub namespace: String,
@@ -1217,6 +1221,8 @@ pub struct SandboxDetail {
     pub annotations: HashMap<String, String>,
     pub labels: HashMap<String, String>,
     pub volume_mounts: Vec<CubeVolumeMount>,
+    pub configured_qos: Option<CubeQosConfig>,
+    pub qos_applied: bool,
 }
 
 fn parse_cpu_millicores(s: &str) -> i32 {
@@ -1396,6 +1402,8 @@ impl GetSandboxResponse {
             annotations: item.annotations,
             labels: item.labels,
             volume_mounts: item.volume_mounts,
+            configured_qos: item.configured_qos,
+            qos_applied: item.qos_applied,
         })
     }
 }
@@ -1851,6 +1859,8 @@ pub struct TemplateResponse {
     pub replicas: Vec<serde_json::Value>,
     #[serde(default)]
     pub create_request: Option<serde_json::Value>,
+    #[serde(default)]
+    pub configured_qos: Option<CubeQosConfig>,
 }
 
 /// Body for DELETE /cube/template.
@@ -1994,9 +2004,50 @@ pub struct CreateTemplateFromImageReq {
     /// Whether the template build sandbox should include ivshmem.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub enable_ivshmem: Option<bool>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub qos: Option<CubeQosConfig>,
     /// Whether CubeMaster bakes the CubeEgress root CA into the template rootfs.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub with_cube_ca: Option<bool>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct CubeQosConfig {
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub network: Option<CubeNetworkQosConfig>,
+    #[serde(rename = "blockIo", default, skip_serializing_if = "Option::is_none")]
+    pub block_io: Option<CubeBlockIoQosConfig>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct CubeNetworkQosConfig {
+    #[serde(rename = "bandwidthMbps", default, skip_serializing_if = "is_zero_u32")]
+    pub bandwidth_mbps: u32,
+    #[serde(
+        rename = "packetsPerSecond",
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub packets_per_second: Option<u32>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(deny_unknown_fields)]
+pub struct CubeBlockIoQosConfig {
+    #[serde(
+        rename = "throughputMiBps",
+        default,
+        skip_serializing_if = "Option::is_none"
+    )]
+    pub throughput_mibps: Option<u32>,
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub iops: Option<u32>,
+}
+
+fn is_zero_u32(value: &u32) -> bool {
+    *value == 0
 }
 
 /// Minimal container overrides for template creation.
@@ -2499,6 +2550,10 @@ mod tests {
                 "host_id": "host-1",
                 "status": 1,
                 "template_id": "tpl-1",
+                "configured_qos": {
+                    "network": { "bandwidthMbps": 100 }
+                },
+                "qos_applied": true,
                 "containers": [
                     {
                         "container_id": "workload-1",
@@ -2527,6 +2582,15 @@ mod tests {
         assert_eq!(detail.host_id, "host-1");
         assert_eq!(detail.cpu_count, 2);
         assert_eq!(detail.memory_mb, 2048);
+        assert_eq!(
+            detail
+                .configured_qos
+                .as_ref()
+                .and_then(|qos| qos.network.as_ref())
+                .map(|network| network.bandwidth_mbps),
+            Some(100)
+        );
+        assert!(detail.qos_applied);
         assert_eq!(
             detail
                 .started_at
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -361,6 +361,10 @@ pub struct SandboxDetail {
     pub state: SandboxState,
     #[serde(rename = "volumeMounts", skip_serializing_if = "Option::is_none")]
     pub volume_mounts: Option<Vec<SandboxVolumeMount>>,
+    #[serde(rename = "configuredQos", skip_serializing_if = "Option::is_none")]
+    pub configured_qos: Option<QosConfig>,
+    #[serde(rename = "qosApplied")]
+    pub qos_applied: bool,
 }
 
 // ─── Sandbox — pause/resume/connect/snapshot ──────────────────────────────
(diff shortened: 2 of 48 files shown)


chenhengqi:
Thanks for the PR.

* We should also add pps limit.
* The hypervisor also supports block io QoS. We should add it for completeness.
* Can we just imply this kind of QoS from CPU/memory allocated to sandbox? More CPU/memory, higher QoS limit. Like most cloud provider does. So that users don't have to learn how to configure these limits.

zyl1121:
> Thanks for the PR.
> 
> * We should also add pps limit.
> * The hypervisor also supports block io QoS. We should add it for completeness.
> * Can we just imply this kind of QoS from CPU/memory allocated to sandbox? More CPU/memory, higher QoS limit. Like most cloud provider does. So that users don't have to learn how to configure these limits.

Thanks. I suggest keeping this PR scoped to network QoS and extending it with an optional template-level `qos.network.packetsPerSecond` setting alongside `bandwidthMbps`, since both use the same `cube.master.net` -> Cubelet -> CubeShim virtio-net Token Bucket path.

The configuration model can support platform inference without removing explicit control:

```text
explicit template setting > inferred platform profile > legacy/default behavior
```

This PR would define and validate the explicit network fields. A future platform profile can provide defaults when a field is omitted, while an explicit template value remains the override for workloads whose needs do not correlate with CPU or memory.

Block-device and filesystem QoS use separate storage paths and should be handled in a follow-up storage QoS PR.
Does this network boundary and precedence model match the intended direction?

kinwin-ustc:
@zyl1121  I believe this is correct. The platform's default limitations can still be calculated when creating the template based on the input CPU and memory configurations, and then passed in as template Qos setting.

chenhengqi:
> Block-device and filesystem QoS use separate storage paths and should be handled in a follow-up storage QoS PR.

The change should be the same for the template side and API side.
This PR already touches 37 files, I don't like another PR doing almost the same thing, so let's complete it in this one.

chenhengqi:
Another question is, should we also add per-sandbox QoS?

chenhengqi:
BTW, we require e2e test cases covering new features for now.

zyl1121:
@chenhengqi Thanks. I’ve updated this PR to complete the end-to-end path for explicit template-level QoS, including:

* network bandwidth and PPS limits;
* block throughput and IOPS limits;
* propagation through CubeAPI, CubeMaster, Cubelet, CubeShim, the CLI, and documentation;
* control/runtime E2E tests and live dataplane E2E tests for upload/download bandwidth, PPS, block throughput, and IOPS.

Templates without `qos` continue to behave as before. When QoS values are specified explicitly, they are validated and translated to the existing virtio-net and virtio-blk token-bucket implementations.

CPU/memory-derived QoS defaults are not included in this revision because the mapping from resource configuration to QoS limits has not yet been defined. The resolver already has access to the template resource configuration, so the defaults can be added there once the policy is agreed, without changing the public API or the runtime propagation path. Explicit template values will continue to take precedence.

I’ve also kept the public API template-level in this revision. Before adding per-sandbox QoS, we need to define how sandbox-level values should interact with the template configuration: whether they may override template values, only reduce them, or disable individual limits. My preference is to treat the template values as upper bounds and allow a sandbox to request stricter limits only. Allowing unrestricted per-sandbox overrides could otherwise bypass the isolation guarantees established at the template level.

The PR description now includes the E2E setup and reproduction commands. The live tests verified a 10 Mbps bandwidth limit, 100 PPS, 8 MiB/s block throughput, and 100 IOPS, along with compatibility for templates without QoS and proper cleanup.

chenhengqi:
@ls-ggg Please consider adding this PR to the v0.7.0 release.

chenhengqi:
@zyl1121 I think this will be defered to v0.8.0. Will review this on our next development cycle. Thanks.

lkml-likexu:
@zyl1121 This one is cool, and we definitely need end-to-end tests to cover this part.

zyl1121:
> This one is cool, and we definitely need end-to-end tests to cover this part.

There are already E2E tests for this in `tests/e2e/sdk_compat/cases/qos/test_template_qos.py`, covering both the control/runtime path and the QoS dataplane behavior.

Tomorrow I’ll resolve the conflicts and rebase the PR, then rerun the E2E tests.

zyl1121:
@lkml-likexu I rebased onto the latest master and resolved the conflicts.

The E2E tests were rerun after the rebase, and both the QoS suite and the no-QoS compatibility check pass.

The rebased branch has been force-pushed, and the PR description has been updated with the latest test results.

lkml-likexu:
@zyl1121 Would you help to rebase this one so that we can review ?

PR #1178, opened 2026-07-27
fix(node sdk): omit sandbox timeout when not explicitly provided

### LINKED ISSUE #1177: [Bug Report] Node SDK sends default timeout when timeout is not explicitly provided
## Summary

The Node SDK incorrectly sends the default `timeout: 300` when the user does not explicitly provide `timeout`, overriding the server-side sandbox lifecycle policy.

## Environment

- CubeSandbox version / commit: latest `master` / `origin/master`
- Host OS and kernel version: N/A, SDK-only issue
- KVM info (`modinfo kvm`): N/A, SDK-only issue
- Deployment mode: N/A, SDK-only issue
- Relevant component: other, Node SDK (`@cubesandbox/sdk`)

## Steps to Reproduce

1. Call `Sandbox.create` with the Node SDK without explicitly passing `timeout`:

   ```ts
   await Sandbox.create({ config })
   ```

2. Inspect the request body sent to CubeAPI. It currently includes `timeout: 300`.

3. Call `Sandbox.connect` without explicitly passing `timeout`:

   ```ts
   await Sandbox.connect(sandboxId, { config })
   ```

4. Inspect the request body sent to CubeAPI. It currently includes `timeout: 300`.

5. Call `resume` on an existing sandbox instance without explicitly passing `timeout`:

   ```ts
   await sb.resume()
   ```

6. Inspect the request body sent to CubeAPI. It currently includes `timeout: 300`.

## Expected Behavior

When the user does not explicitly provide `timeout`, the Node SDK should omit the `timeout` field from the request body and let the server apply its own default lifecycle policy.

Expected request body for `Sandbox.create({ config })`:

```json
{
  "templateID": "tpl-test"
}
```

Expected request body for `Sandbox.connect(sandboxId, { config })`:

```json
{}
```

Expected request body for `sb.resume()`:

```json
{}
```

Explicit timeout values should still be preserved:

```ts
await Sandbox.create({ timeout: 0, config })
await Sandbox.create({ timeout: 600, config })
await sb.resume(0)
await sb.resume(600)
```

These calls should still send `timeout: 0` or `timeout: 600` respectively.

## Actual Behavior

The Node SDK currently sends the default `timeout: 300` even when the user does not explicitly provide `timeout`.

Actual request body for `Sandbox.create({ config })`:

```json
{
  "templateID": "tpl-test",
  "timeout": 300
}
```

Actual request body for `Sandbox.connect(sandboxId, { config })`:

```json
{
  "timeout": 300
}
```

Actual request body for `sb.resume()`:

```json
{
  "timeout": 300
}
```

This overrides the server-side lifecycle policy. For example, if the server wants to use a longer TTL, a shorter TTL, or no automatic timeout, the Node SDK forces the value to 300 seconds.

```text
No runtime error is thrown. The bug is the unexpected timeout field in the control-plane request body.
```

## Additional Context

This creates behavior inconsistency between the Node SDK, the project documentation, and the current Python / Go SDK behavior.

The expected semantics are that omitting `timeout` means the SDK should not send a timeout field, allowing the server to decide the default lifecycle policy. The current Node SDK behavior writes the SDK-side default value into the request body, which affects lifecycle-related features such as `pause/resume`, auto-resume, and long-running agent sessions.

Suggested fix:

- In `Sandbox.create()`, only set `payload.timeout` when `options.timeout !== undefined`
- In `Sandbox.connect()`, send `{}` by default instead of `{ timeout: cfg.timeout }`
- Change `sb.resume()` to `resume(timeout?: number)` and only send `payload.timeout` when `timeout !== undefined`
- Preserve explicit user-provided values such as `timeout: 0` and `timeout: 600`

Suggested regression tests:

- `Sandbox.create({ config })` should omit `timeout`
- `Sandbox.create({ timeout: 0 })` should send `timeout: 0`
- `Sandbox.connect(id, { config })` should omit `timeout`
- `sb.resume()` should omit `timeout`
- `sb.resume(0)` should send `timeout: 0`

XiaoFeiASK:
## Summary

Fixes the Node SDK sandbox lifecycle timeout behavior so `timeout` is only sent when explicitly provided, aligning it with the documentation and the Python / Go SDKs.

fix: #1177 

### Modified Files

* `sdk/node/src/sandbox.ts` — Stop sending the SDK-side default `timeout: 300` from `Sandbox.create`, `Sandbox.connect`, and `sb.resume`
* `sdk/node/src/config.ts` — Update the `DEFAULT_SANDBOX_TIMEOUT_S` comment to reflect that it is retained on `Config` only
* `sdk/node/test/sandbox.test.ts` — Add regression coverage for omitted timeout and explicit timeout values

## Features

1. `Sandbox.create({ config })` now omits `timeout` when the user does not explicitly provide it
2. `Sandbox.connect(id, { config })` now sends an empty request body instead of `{ timeout: cfg.timeout }`
3. `sb.resume()` now omits `timeout` by default instead of sending `timeout: 300`
4. Explicit timeout values are preserved, including `timeout: 0` and `timeout: 600`
5. Server-side sandbox lifecycle policy is no longer overridden by the Node SDK default timeout
6. Behavior is now consistent with the current Python and Go SDK implementations

## Testing

```bash
# Run targeted sandbox lifecycle tests
cd sdk/node
npm test -- sandbox.test.ts

# Run TypeScript type checking
npm run typecheck

# Run the full Node SDK test suite
npm test
Test results:
- npm test -- sandbox.test.ts — passed, 37 tests
- npm run typecheck — passed
- npm test — passed, 174 tests passed, 1 skipped

### DIFF
--- a/sdk/node/src/config.ts
+++ b/sdk/node/src/config.ts
@@ -65,7 +65,7 @@ function normalizeProxyScheme(scheme: string | undefined, port: number): string 
   return port === 443 ? "https" : "http";
 }
 
-/** Default sandbox TTL in seconds, shared by {@link Config} and ``resume``. */
+/** Default sandbox TTL in seconds retained on {@link Config}. */
 export const DEFAULT_SANDBOX_TIMEOUT_S = 300;
 
 export class Config {
--- a/sdk/node/src/sandbox.ts
+++ b/sdk/node/src/sandbox.ts
@@ -4,7 +4,7 @@
 import { fetch, type Dispatcher } from "undici";
 
 import { Commands } from "./commands.js";
-import { Config, DEFAULT_SANDBOX_TIMEOUT_S, resolveConfig, type ConfigOptions } from "./config.js";
+import { Config, resolveConfig, type ConfigOptions } from "./config.js";
 import {
   ApiError,
   AuthenticationError,
@@ -352,10 +352,10 @@ export class Sandbox {
       );
     }
 
-    const payload: Record<string, unknown> = {
-      templateID: tpl,
-      timeout: options.timeout ?? cfg.timeout,
-    };
+    const payload: Record<string, unknown> = { templateID: tpl };
+    if (options.timeout !== undefined) {
+      payload.timeout = options.timeout;
+    }
     if (options.envVars) {
       payload.envVars = options.envVars;
     }
@@ -416,7 +416,7 @@ export class Sandbox {
     const resp = await controlFetch(cfg, `${cfg.apiUrl}/sandboxes/${sandboxId}/connect`, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
-      body: JSON.stringify({ timeout: cfg.timeout }),
+      body: JSON.stringify({}),
     });
     await checkControlResponse(resp);
     return new Sandbox((await resp.json()) as Record<string, any>, cfg);
@@ -570,11 +570,15 @@ export class Sandbox {
    * @deprecated Use {@link Sandbox.connect} which auto-resumes and returns a
    * fresh instance.
    */
-  async resume(timeout = DEFAULT_SANDBOX_TIMEOUT_S): Promise<void> {
+  async resume(timeout?: number): Promise<void> {
+    const payload: Record<string, unknown> = {};
+    if (timeout !== undefined) {
+      payload.timeout = timeout;
+    }
     const resp = await controlFetch(this.config, `${this.config.apiUrl}/sandboxes/${this.sandboxId}/resume`, {
       method: "POST",
       headers: { "Content-Type": "application/json" },
-      body: JSON.stringify({ timeout }),
+      body: JSON.stringify(payload),
     });
     await checkControlResponse(resp);
   }
--- a/sdk/node/test/sandbox.test.ts
+++ b/sdk/node/test/sandbox.test.ts
@@ -137,6 +137,26 @@ describe("Sandbox.create", () => {
     sb.close();
   });
 
+  it("omits timeout when not explicitly provided", async () => {
+    setHandler(() => ({ status: 201, json: SANDBOX_DATA }));
+    const cfg = makeConfig();
+    cfg.timeout = 600;
+
+    const sb = await Sandbox.create({ config: cfg });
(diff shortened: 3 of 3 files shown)


fslongjin:
Regarding this issue, my understanding is as follows: To align with the behavior of the e2b SDK, having a default timeout of 300 seconds set by the SDK is fine. We provide a "never timeout" option that allows users to explicitly specify no timeout.

Changing the default timeout behavior would be a significant breaking change, and I don't think we should do that.

cc @wbzdssm

XiaoFeiASK:
I agree that changing the default timeout behavior can be a compatibility concern, but the current CubeSandbox SDK contract is already not "the SDK always sends 300 seconds". Both the Python and Go SDKs treat the sandbox lifecycle timeout as an explicit create option. If the caller does not provide a timeout, the field is omitted from the request body.

Python example:

- `Sandbox.create(..., timeout: int | None = None)` makes `timeout` optional at the API level.
- When the caller does not pass `timeout`, its value is `None`.
- The request payload is initialized as `{"templateID": tpl}`.
- The SDK only adds `payload["timeout"] = timeout` inside `if timeout is not None`, so omitted timeout is not sent on the wire.

See `sdk/python/cubesandbox/sandbox.py:185-263`.

Go example:

- `CreateOptions.Timeout` is `*time.Duration`, so the zero state for "not provided" is `nil`.
- The comment explicitly says that `nil` omits the field and lets the server decide.
- The request payload is initialized with only `"templateID"`.
- The SDK only adds `payload["timeout"]` inside `if opts.Timeout != nil`, so omitted timeout is not sent on the wire.

See `sdk/go/models.go:59-78` and `sdk/go/client.go:116-130`.

XiaoFeiASK:
Which semantic do we want CubeSandbox SDKs to follow consistently across Node, Python, and Go?

wbzdssm:
GTM, approved.

Omitting timeout when not explicitly set aligns with Python/Go behavior. The breaking change just needs a one-liner in the changelog.

XiaoFeiASK:
> GTM, approved.
> 
> Omitting timeout when not explicitly set aligns with Python/Go behavior. The breaking change just needs a one-liner in the changelog.

thanks for your review.

PR #1179, opened 2026-07-27
fix(Cubelet): cleanup host-dir mount directories

luzhixing12345:
There are uncleared host-dir directories under `/data/cubelet/hostdir`.

```bash
# ls /data/cubelet/hostdir
2cd3ec83a9ec46f5ab84822c5e41db56  6edb6cf299c8457499e65e2ff8de6a11  
cc853f00b0f04a32a14bab559e3de8be  da78bbf168e641a6854dad1f6b2713ac
```

#333 added `cleanupHostDirVolumes` to remove host-dir mount directories during sandbox destruction.

#997 accidentally replaced this cleanup call while adding plugin volume cleanup, causing host-dir directories to remain after sandboxes were destroyed.

### DIFF
--- a/Cubelet/storage/local.go
+++ b/Cubelet/storage/local.go
@@ -1281,6 +1281,10 @@ func (l *local) destroy(ctx context.Context, info *StorageInfo, opts *workflow.D
 		errs = errors.Join(errs, err)
 	}
 
+	if err := l.cleanupHostDirVolumes(ctx, info); err != nil {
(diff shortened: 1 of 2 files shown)


lkml-likexu:
@luzhixing12345 Thanks for your contribution. Could you please add a test case to verify these changes?

luzhixing12345:
Modified as suggested.

PR #1183, opened 2026-07-27
feat(net): expose dns_enforce_query end-to-end to close the DNS side-channel

sparkzky:
## Why this change

Today, sandbox DNS queries (UDP/53) are **always forwarded**, regardless of `dns_allow`. `dns_allow` only decides whether a resolved IP is *learned* (so the app can later connect to it). The consequence: **a sandbox can still resolve domains that are not in its allow-list** — it gets the A record back (can confirm a domain exists, or exfiltrate via DNS) even though `deny_out` then blocks the TCP connection to that IP.

That leaves a **DNS resolution side-channel**: non-allow-listed domains remain resolvable/discoverable even in a locked-down sandbox (`allow_internet_access=false`, `deny_out=0.0.0.0/0`).

This PR closes it. When `dns_enforce_query` is enabled, the eBPF datapath (`dns_finish`) **drops** DNS A queries whose domain is absent from `dns_allow` at the TC layer (`TC_ACT_SHOT`) — the query never reaches the upstream resolver, so no answer ever returns. Allow-listed domains still resolve and learn IPs (enforce implies learning, so allow-listed domains keep working). Default is **off** — zero behavior change for existing sandboxes.

## What's in the PR

Wires the (already-implemented in `cubevs`) `DNSEnforceQuery` flag through the full stack so it's user-configurable — previously it lived only in the `cubevs` package with no proto/SDK passthrough.

- **proto**: `optional bool dns_enforce_query = 5` added to `CubeNetworkConfig` (cubebox + network_agent); `.pb.go` regenerated with the repo's pinned toolchain (protoc-gen-go v1.36.11 / protoc-gen-go-grpc v1.6.1 / protoc 28.3).
- **CubeMaster**: threaded through `types.CubeNetworkConfig`, `mapCubeNetworkConfig`, and the cubebox template-merge path (`mergeCubeNetworkConfigs` + `CubeNetworkConfig.DeepCopy` — both hand-written copy funcs that silently dropped the new field until the E2E caught it).
- **Cubelet**: threaded through `networkagentclient` types, `plugin_tap` map/clone, and the gRPC/proto mapper.
- **network-agent**: `cubeVSTapRegistration` reads `cfg.DnsEnforceQuery`.
- **CubeAPI (Rust)**: added to `SandboxNetworkConfig` (incoming) + `CubeNetworkConfig` (outgoing) + `build_cube_network_config`.
- **SDK (Python)**: `dns_enforce_query` in the `network` param.

> Note: `CubeMaster/api/.../cubebox.pb.go` also picks up a pre-existing drift (`plugin_volume` exists in the proto but the committed pb.go was stale). Regeneration brings it back in sync — incidental, not a behavior change.

## Verified end-to-end

Tested on a live one-click cluster (all components built from this branch, deployed, exercised, then reverted to original):

| DNS A query | `dns_enforce_query=true` | default (off) |
|---|---|---|
| allow-listed domain | resolves ✓ | resolves ✓ |
| non-allow-listed domain | **timeout — dropped at BPF** ✓ | resolves (side-channel open) |

BPF map confirms `dns_policy_flags=3` / `dns_enforce_query=true` when enabled. The field was traced through every hop live (CubeAPI→CubeMaster→Cubelet→network-agent→BPF). Plus per-hop unit tests cover each conversion (proto round-trip, CubeMaster decode+map+merge+deepcopy, Cubelet map/clone/proto, network-agent proto→struct→MVMOptions), including the jsoniter live-decode path.

## Design note: could replace the current policy entirely

This ships `dns_enforce_query` as an **opt-in** flag to preserve backward compatibility. If the maintainers prefer, it can instead **become the default backend policy** — i.e. drop the separate knob and simply make "drop DNS A queries for non-allow-listed domains" the default behavior for any sandbox that has a `dns_allow` list. The datapath change is identical; only the policy decision in `netpolicy.go` (`buildNetPolicyPlan`) would flip — turn the enforce bit on whenever `dns_allow` is non-empty, instead of gating on the new flag. Happy to refactor to enforce-by-default if that matches the project's threat model better than opt-in.

---


chenhengqi:
I am afraid this breaks the E2B semantics. And real world DNS usage is quite complicated. A simple command like `apt update` involves different kinds of DNS resolutions. We can't simply block it.

chenhengqi:
Do you have any real world security concerns against the current implementation?

sparkzky:
Thanks @chenhengqi — fair pushback, and the DNS-complexity point in particular made me go measure it rather than hand-wave. Addressing each:

## On "real world security concerns with the current implementation"

Two concrete ones, both specific to the **locked-down** model (`allow_internet_access=false`, `deny_out=0.0.0.0/0`) — which is the only place this feature is meant to be used:

1. **DNS data exfiltration.** A compromised sandbox can encode outbound data in subdomain queries — e.g. `<base64-secret>.attacker.example`. The *query itself* carries the data out to the resolver (which forwards to public DNS), where the attacker's authoritative nameserver observes it. `deny_out` blocking the response doesn't help — the query already left the sandbox. It's the one egress channel that survives "default-deny IP."
2. **Domain-existence probing / recon.** The sandbox can distinguish NXDOMAIN vs a real A record for arbitrary names — mapping internal/external name presence even though it can't connect.

Both are narrow, but they're exactly the residue left once IP egress is locked down.

## On "breaks E2B semantics"

It's opt-in and **default off** — no change to the default network model or to any existing sandbox. It's an extra knob for the strict-security case, not a change to what E2B-compatible sandboxes do by default.

## On "we can't simply block it — apt update, different kinds of DNS resolutions"

This was my biggest worry too, so I measured the two things that could go wrong.

**1) No functional loss for non-allow domains.** Today a non-allow domain resolves, but its IP is never learned into `allow_out`, so `deny_out` already blocks the connection — the sandbox can't *use* it. Enforce just moves the failure from TCP-connect to DNS-resolve; net reachability is the same (zero). For apt specifically: if its mirror domains aren't allow-listed, apt **already fails today** (can't connect to the resolved IP); enforce doesn't make it worse — it fails at DNS instead of TCP.

**2) CNAME chains don't cause breakage in practice.**
- **Recursive resolver chases CNAME server-side.** Querying 119.29.29.29 (DNSpod) for a set of CNAME'd domains, every one returns `CNAME + final A` in a single response — `www.baidu.com→www.a.shifen.com`, `docs.python.org→dualstack.python.map.fastly.net`, `www.tencent.com→…eo.dnse5.com`, `download.jetbrains.com→…cloudfront.net`, `files.pythonhosted.org→…fastly.net`. Zero "bare CNAME" responses that would force a client follow-up.
- **The sandbox makes no CNAME-target follow-up queries.** I tcpdump'd the sandbox's actual DNS traffic while it resolved CNAME'd allow-listed domains: it sends A/AAAA queries **only** for the allow-listed names themselves — zero queries for the CNAME targets. glibc takes the recursive resolver's full answer directly.

Since enforce drops only A queries whose name is **not** in `dns_allow`, and the sandbox never queries a CNAME target directly, enforce can't drop anything relevant to an allow-listed (even CNAME'd) domain. So "different kinds of DNS resolutions" doesn't translate into breakage here.

## Two honest caveats (happy to document / fix)

- **A-only.** The filter sits in the A-query pipeline; AAAA/SRV/MX aren't filtered. So allow-listed domains' IPv6 resolution is unaffected (good), but non-allow domains' AAAA isn't dropped either — the measure is partial. Adding AAAA is a small follow-up if we want it airtight.
- **Resolver-dependent.** The CNAME safety relies on the recursive resolver chasing server-side, which standard resolvers (DNSpod, 8.8.8.8, …) do. Worth a one-line doc note ("use a standard recursive resolver").

## On the design direction

I shipped it as opt-in to be safe, but if the project's threat model is "locked-down sandboxes should have no DNS side-channel by default," I can flip it to **enforce-by-default whenever a sandbox has a `dns_allow` list** (one-line change in `buildNetPolicyPlan`) instead of a separate flag — the datapath is identical. Happy to go either way; just want to align on how strict the default should be.

chenhengqi:
@sparkzky I personally don't think this is a good design. I also have doubt in your `DNS data exfiltration` example.(A compromised sandbox can't change its DNS server, no?)

The simplest way to support this feature is allow adding domain names (`*` for deny all) to `deny_out`.

Issue #929, opened 2026-07-12
[Bug Report] create 成功但数据面异步激活且无就绪信号:存在慢激活(数秒~数十秒)与永不激活两种故障态


ENCHIGO:
> 环境:v0.5.1-rc5 全套 + network-agent rc11(commit 27153ec)源码构建;裸机 Ubuntu,内核 5.15,CUBE_PVM_ENABLE=0(宿主 /dev/kvm + vmlinux-bm);单节点 host-direct;消费端 e2b Python SDK 2.21.0。

**现象**:POST /sandboxes 返回成功、控制面全绿(EnsureNetwork RPC 4ms 内到达
network-agent、TAP 从池出队、路由已发布 redis、envd 经快照恢复 49-102ms 存活),
但 proxy→沙箱的 TCP SYN 无应答。实测存在三态:
- 快激活(<1s,多数);
- **慢激活**(数秒~数十秒:同一沙箱连续探测 `HTTP 000 → 200`,间隔仅数秒);
- **终生暗**(永不激活,偏发于空闲间隔后的孤立 create 与 network-agent 重启后数分钟)。

**复现**:任意模板 create 后立刻对暴露端口做 HTTP 探测,统计首响应延迟分布;
长空闲后首建/agent 重启后 5 分钟内建,可稳定观察到慢激活与终生暗。

**期望**:①create 响应或 sandbox 状态中提供数据面 ready 语义(字段/事件/可轮询状态);
②文档明确"create 返回后数据面异步激活"的事实与建议等待策略。

**影响**:所有消费端被迫自建"创建后探活+销毁重建"守卫;单次探活会把慢激活误判为
终生暗导致误杀(我们已为此实现重试退避)。

---
如需完整时间戳日志/探测脚本/远程复现,我们可以提供。

zyl1121:
Thanks for reporting this.

There are a few details that would help us understand the issue better. The current description seems a bit unclear.

You mentioned that `envd` can be restored from snapshot within `49-102ms`, which indicates that the snapshot restore path itself has completed successfully, and the data plane path for the sandbox (e.g. the `49983` endpoint for `envd`) should be working.

However, you also mentioned:

> proxy → sandbox TCP SYN gets no response

Could you please clarify what exactly this TCP SYN refers to?

* Which IP:PORT is the SYN sent to?
* Where was the packet captured (on the proxy side, inside the sandbox, or elsewhere)?
* Is this accessing a custom service port in the sandbox?

[...]

PR #1186, opened 2026-07-27
docs(templates): document the sandbox readiness contract for the probed port

ENCHIGO:
Refs #929

## Problem

`docs/guide/tutorials/template-from-image.md` explains that the probe decides when a **template** is ready, but nothing in the docs states what that implies for **sandboxes** created from it. A consumer that finds a port unreachable right after `create` cannot tell whether that is a bug or expected asynchronous behaviour, so every consumer ends up building a defensive probe-with-retry layer — and cannot safely remove it, because an observation is not a guarantee.

## Change

Adds a short **Sandbox readiness** subsection next to the probe documentation (English + `docs/zh` mirror), stating the contract exactly as @chenhengqi described it in https://github.com/TencentCloud/CubeSandbox/issues/929#issuecomment-5087160653:

> Only the probed port is guaranteed to be ready. This is done by a health check on template creation.
> [...] This holds true for all sandboxes as long as the sandbox is running and the service listening to the port is healthy.

So the section says: the probed port is served immediately by a running sandbox (fresh create, and equally after resume / auto-resume), clients need no extra wait after `create` returns, and it spells out the two limits — exposed-but-unprobed ports are not covered, and the guarantee is about platform-side reachability, not a service that crashes inside the sandbox.

No behaviour change; documentation only. Happy to reword anything that does not match the intended contract — the wording is my reading of the answer above, and you are the source of truth here.


chenhengqi (approved):

ENCHIGO:
@chenhengqi Thanks for the approval. Could someone with write access merge this when convenient?

Flagging it only because of the ordering: #929 was closed as completed a few seconds after the approval, but the wording it was closed on lives in this PR, so until this merges the readiness contract is still undocumented — which is the one thing that issue asked for. The branch is up to date with `master` and the checks are green.

Happy to rebase or reword anything first if that helps.

chenhengqi:
@fslongjin @tinklone Please take a look.

fslongjin (approved):

ENCHIGO:
@fslongjin Thanks for the review — that is two approvals now (@chenhengqi on 2026-08-04, yours today), `mergeable_state: clean`, docs-only diff. Could one of you merge it when you have a moment?

The reason I keep nudging is the ordering: #929 was closed as *completed* seconds after the first approval, and the wording it was closed on lives only in this PR. Until it merges, the readiness contract that issue asked for is still undocumented — `grep -r "Sandbox readiness" docs/` on master returns nothing. Thirteen days of a closed-but-undelivered issue is the only thing I am trying to fix here.

Happy to rebase if the base has moved in a way that matters, or to reword anything before merge.

PR #1193, opened 2026-07-28
virtio-blk: drain inflight IO and flush at pause

lisongqian:
Pause only parked the epoll loop, leaving inflight IO unaccounted
for. Under pause -> snapshot -> SIGKILL -> restore this can yield
permanently pending guest IOs and an inconsistent backend.

Add a before_pause() hook that drains inflight_requests and, in
writeback mode, issues a synchronous fsync. Bounded by 20s; on
timeout the pause itself fails so the snapshot is aborted.


up2wing (approved):
LGTM

PR #1196, opened 2026-07-28
feat(helm): support external PostgreSQL via mysql.driver

xiaojunxiang2023:
## Summary

- Add `mysql.driver` (`mysql` | `postgres`) to the Kubernetes Helm chart so deployments can use an **external PostgreSQL** database (complements the one-click path in #1082 / runtime support already in CubeMaster & CubeOps).
- When `mysql.driver=postgres`, require `mysql.host`, skip the chart-managed MySQL StatefulSet, set `driver: postgres` in CubeMaster `conf.yaml`, and pass CubeOps a `DATABASE_URL` with a `postgresql://` scheme (MYSQL_* alone always assembles `mysql://`).
- Reject invalid drivers and `CHANGE_ME_*` passwords for external `mysql.host`; document the postgres values shape in `values.yaml` / `runtime-values.example.yaml` and the chart README.

## Test plan

- [x] `helm lint deploy/kubernetes/chart --set-string mysql.password=test --set-string mysql.rootPassword=test --set-string redis.password=test` — passed
- [x] `helm template chart-check deploy/kubernetes/chart ...` (default mysql path) — passed
- [x] Existing chart guards (`test-big-pod-inplace-guard.sh`, `test-pvm-native-ds-gvk-guard.sh`, `test-master-recreate-strategy-guard.sh`, `test-cluster-resource-names.sh`) — passed
- [x] `sh deploy/kubernetes/chart/scripts/test-mysql-driver-postgres-guard.sh` — passed (default mysql; postgres without host / invalid driver fail; postgres+host sets CubeMaster driver + CubeOps `DATABASE_URL` and skips MySQL StatefulSet)

### DIFF
--- a/deploy/kubernetes/chart/values.yaml
+++ b/deploy/kubernetes/chart/values.yaml
@@ -407,7 +407,16 @@ diagnostics:
   # can collect release state without maintaining a second script copy.
   enabled: true
 
+# Control-plane database (CubeMaster / CubeOps / CubeAPI) engine selector.
+# The driver decides which section below is read; the other one is ignored.
+#   mysql    -> mysql.*    (chart-managed StatefulSet when mysql.host is empty,
+#                           or an external MySQL when mysql.host is set)
+#   postgres -> postgres.* (external only, postgres.host is required)
+database:
+  driver: mysql
+
 mysql:
+  # Read when database.driver=mysql.
   enabled: true
   image:
     repository: mysql
@@ -424,8 +433,9 @@ mysql:
   # production, provide `mysql.password` and `mysql.rootPassword` via -f
   # secret-values.yaml or --set, or use `mysql.host` to point at an externally
   # managed MySQL. templates/validate.yaml refuses to render when passwords
-  # equal the default sentinel below.
+  # equal the default sentinels below.
   password: CHANGE_ME_MYSQL_PASSWORD
+  # rootPassword is only used by the chart-managed StatefulSet.
   rootPassword: CHANGE_ME_MYSQL_ROOT_PASSWORD
   persistence:
     enabled: true
@@ -440,6 +450,17 @@ mysql:
     accessModes:
       - ReadWriteOnce
   resources: {}
+
+# External PostgreSQL, read only when database.driver=postgres.
+# The chart never deploys PostgreSQL, so host must point at an existing server.
+postgres:
+  host: ""
+  port: 5432
+  database: cube_mvp
+  user: cube
+  # SECURITY: same rule as mysql.password; templates/validate.yaml refuses to
+  # render while this equals the CHANGE_ME_* sentinel.
+  password: CHANGE_ME_POSTGRES_PASSWORD
 
 redis:
   enabled: true
--- a/.github/workflows/kubernetes-chart-check.yml
+++ b/.github/workflows/kubernetes-chart-check.yml
@@ -41,6 +41,8 @@ jobs:
         run: sh deploy/kubernetes/chart/scripts/test-master-recreate-strategy-guard.sh
       - name: Guard cluster-scoped resource names
         run: sh deploy/kubernetes/chart/scripts/test-cluster-resource-names.sh
+      - name: Guard database.driver postgres path
+        run: sh deploy/kubernetes/chart/scripts/test-database-postgres-guard.sh
       - name: Check startup-gate shell syntax
         run: |
           sh -n deploy/kubernetes/images/scripts/node-prep-lib.sh
--- a/deploy/kubernetes/chart/README.md
+++ b/deploy/kubernetes/chart/README.md
@@ -196,10 +196,30 @@ helm upgrade --install cube ./deploy/kubernetes/chart   -n cube-system   --creat
 
 > Do not store SSH passwords or node login credentials in this chart. Host mutation is performed through Kubernetes privileged Pods, not through SSH.
 
-## Use third-party MySQL or Redis
+## Use third-party MySQL, PostgreSQL, or Redis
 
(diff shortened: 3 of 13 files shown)


xiaojunxiang2023:
Thanks for the review!

1. Addressed: CubeOps now loads DATABASE_URL via secretKeyRef (`database-url` in the chart Secret) instead of a plaintext Deployment env value. Used a Secret key rather than `$(VAR)` substitution so we keep Helm `urlquery` encoding for credentials with special characters.

2. Left the defensive catch-all message as-is — it still covers `mysql.enabled=false` without host; the postgres-specific fail fires first for the postgres path.

3. Added a CHANGE_ME + external `mysql.host` failure case to the guard script.

4. Agreed — the mysql branch of `cube.databaseURL` is unused by ops today; happy to unify both drivers onto DATABASE_URL in a follow-up if useful.

force-pushed in 18e95d4

fslongjin (commented):
Hello, thank you for your contribution. Let me take a look — MySQL and PostgreSQL are two different databases, but here the MySQL driver is written as PG. I find this quite odd, haha. Perhaps we could implement it in a more reasonable way.

xiaojunxiang2023:
@fslongjin 
Thanks for the feedback — agreed that nesting postgres under mysql.driver was awkward.

Reshaped in 5cee3e6 into a clean split:
- `database.*` — sole control-plane DB config (`driver` / `host` / `port` / credentials), aligned with one-click `CUBE_EXTERNAL_DB_*`
- `mysql.*` — only the chart-managed MySQL StatefulSet (`enabled` / image / persistence)

No dual keys or aliases.

Example:
```yaml
database:
  driver: postgres
  host: "10.0.0.10"
  port: 5432
  user: postgres
  password: "..."
  name: cube_mvp

**Comment by kinwin-ustc** [comment]
We need to consider backward compatibility, such as ensuring that users with older configuration fields can still use it. Also, the PostgreSQL host is currently named cube.mysqlHost; should we use cube.dbHost instead?

**Comment by xiaojunxiang2023** [comment]
@kinwin-ustc Thanks for the review!

Addressed in 041afeb:
- Backward compatibility: deprecated `mysql.host` / `mysql.port` / `mysql.user` / `mysql.password` / `mysql.rootPassword` / `mysql.database` / `mysql.driver` still resolve when the matching `database.*` value is empty (or still `CHANGE_ME_*` for passwords). New deployments should use `database.*`.
- Renamed the resolved endpoint helper to `cube.dbHost`; `cube.mysqlHost` remains as a thin deprecated alias.

File: deploy/kubernetes/chart/values.yaml [hunk 1]
```diff
@@ -441,6 +431,35 @@ mysql:
...
+  # postgres example:
+  #   driver: postgres
+  #   host: "10.0.0.10"
+  #   user: postgres
+  #   password: "..."
+  #   name: cube_mvp
+  #   # port defaults to 5432; rootPassword can be ignored for external databases
+  password: CHANGE_ME_MYSQL_PASSWORD
+  rootPassword: CHANGE_ME_MYSQL_ROOT_PASSWORD
+
```
fslongjin: I found some of the changes here confusing. In the built-in MySQL deployment configuration, the initial password has been moved to the database section, which seems to introduce some implicit dependencies that could easily lead to user misoperation. Also, the example configuration in the database section looks a bit convoluted. For instance, `driver: mysql` is not filled in by default (it's commented out in the config file), and the automatic fallback logic here introduces some implicit inference. Users editing this section will be confused: 'How do I use the built-in MySQL? How do I switch to an external MySQL or PostgreSQL?' The design of this configuration scatters the switches around and includes some implicit deduction logic. It took me a while to understand it, and I still find it quite convoluted.

Perhaps a better approach would be to introduce separate subsections:

```
- database
   - driver: mysql/postgres

- mysql
  - host: xxxxx

- postgres
  - host: xxxxx
```

fslongjin (changes requested):

deploy/kubernetes/chart/values.yaml
xiaojunxiang2023: Thanks, that's a much clearer shape — implemented in 24b6a78.

`database.driver` is now the only switch, and each engine owns a complete,
self-contained connection section:

    database:
      driver: mysql        # the only switch

    mysql:                 # read when driver=mysql
      enabled: true
      host: ""             # empty -> built-in StatefulSet, set -> external MySQL
      port: 3306
      database: cube_mvp
      user: cube
      password: ...
      rootPassword: ...    # only used by the built-in StatefulSet

    postgres:              # read when driver=postgres, external only
      host: ""             # required
      port: 5432
      database: cube_mvp
      user: cube
      password: ...

Specifically addressing your points:
- The built-in MySQL credentials moved back to `mysql.*`, right next to the
  StatefulSet settings they belong to. No more implicit dependency on another section.
- `database.driver: mysql` is now written out explicitly instead of being commented out.
- All the fallback/inference logic is gone. The section that doesn't match the driver
  is simply ignored — `mysql.host` will not satisfy `driver: postgres`, and there's a
  guard test asserting exactly that.
- This also resolves @kinwin-ustc's backward-compatibility concern for free: existing
  `mysql.*` values files keep working unchanged, since `mysql.*` is the real location
  again rather than a deprecated alias.

Side effect: the diff shrank noticeably. The four unrelated guard scripts and the CI
workflow are byte-identical to master again, and values.yaml is down to +22/-1.

The new guard covers: default MySQL, external `mysql.host`, postgres without host,
`mysql.host` alone not satisfying the postgres driver, invalid driver, per-engine
CHANGE_ME sentinel checks, and the Secret-sourced `DATABASE_URL`.

xiaojunxiang2023:
@fslongjin I fixed it. Could you review it again?

fslongjin (approved):
LGTM

PR #1201, opened 2026-07-28
feat(cubeops): unify cubelog request and service observability

liciazhu:
- Replace scattered standard logging with structured cubelog output.
- Add rolling request and statistics logs under /data/log/CubeOps.
- Record request ID, caller, callee, path, status code, and latency.
- Propagate request-scoped logging context through auth, SDK, AgentHub, CubeMaster client, and service handlers.
- Add build-time version, commit, and build timestamp metadata.
- Add logging coverage for request tracing and log-level handling.

### DIFF
--- a/CubeOps/Makefile
+++ b/CubeOps/Makefile
@@ -6,8 +6,20 @@
 BINARY=cubeops
 BIND=127.0.0.1:3010
 
+# Version — injected at build time via ldflags.
+# Priority: env var CUBE_VERSION/CUBE_COMMIT/CUBE_BUILD_TIME > local git/date fallback.
+CUBE_VERSION ?= $(shell git describe --tags --abbrev=0 --match 'v*' 2>/dev/null || echo 0.0.0-dev)
+CUBE_COMMIT ?= $(shell git rev-parse HEAD 2>/dev/null || echo unknown)
+CUBE_BUILD_TIME ?= $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+VERSION_PKG=github.com/tencentcloud/CubeSandbox/CubeOps/internal/version
+BUILD_FLAGS=-ldflags "-s -w \
+  -X '$(VERSION_PKG).Version=$(CUBE_VERSION)' \
+  -X '$(VERSION_PKG).Commit=$(CUBE_COMMIT)' \
+  -X '$(VERSION_PKG).BuildTime=$(CUBE_BUILD_TIME)'"
+
 build:
-	go build -o bin/$(BINARY) ./cmd/cubeops
+	go build $(BUILD_FLAGS) -o bin/$(BINARY) ./cmd/cubeops
 
 run: build
 	JWT_SECRET=test-secret-dummy \
@@ -29,7 +41,11 @@ clean:
 	rm -rf bin/
 
 docker:
-	docker build -t cube-ops:latest -f Dockerfile ..
+	docker build \
+		--build-arg CUBE_VERSION=$(CUBE_VERSION) \
+		--build-arg CUBE_COMMIT=$(CUBE_COMMIT) \
+		--build-arg CUBE_BUILD_TIME=$(CUBE_BUILD_TIME) \
+		-t cube-ops:latest -f Dockerfile ..
 
 tidy:
 	go mod tidy
--- a/CubeOps/README.md
+++ b/CubeOps/README.md
@@ -82,6 +82,13 @@ tail -50 /data/log/CubeOps/cubeops-req.log
 tail -f /data/log/CubeOps/cubeops-req.log
 ```
 
+Every HTTP request receives an `X-RequestID` when one is not supplied. The
+request trace is attached to the request context, so request-scoped `cubelog`
+records include fields such as `RequestId`, `Action`, `Caller`, and
+`CalleeAction`. CubeMaster calls propagate `X-Caller: cubeops` and stamp the
+inbound `RequestID` into the outbound body and query string — that is how
+cross-service correlation works (CubeMaster ignores `X-RequestID`).
+
 Quick health check (no auth required):
 
 ```bash
--- a/CubeOps/cmd/cubeops/main.go
+++ b/CubeOps/cmd/cubeops/main.go
@@ -6,7 +6,7 @@ package main
 import (
 	"context"
 	"errors"
-	"log/slog"
+	"fmt"
 	"net/http"
 	"os"
 	"os/signal"
@@ -22,18 +22,17 @@ import (
 func main() {
 	cfg, err := config.Load()
 	if err != nil {
-		// Config load failed before logging is initialised; use a
-		// stdout-only slog so the error is still visible.
-		slog.SetDefault(slog.New(slog.NewTextHandler(os.Stdout, &slog.HandlerOptions{
-			Level: slog.LevelInfo,
-		})))
-		slog.Error("failed to load config", "error", err)
+		// Logging is not yet initialised here; write directly to stderr
+		// so a config-load failure is always visible regardless of
+		// cubelog's default-writer behaviour.
+		fmt.Fprintf(os.Stderr, "cubeops: failed to load config: %v\n", err)
 		os.Exit(1)
 	}
 
 	logging.Init(logging.Options{
 		Level:      cfg.LogLevel,
 		LogDir:     cfg.LogDir,
+		Module:     "cubeops",
 		FileNum:    cfg.LogFileNum,
 		FileSizeMB: cfg.LogFileSize,
 	})
@@ -44,7 +43,7 @@ func main() {
 	// Initialise database + migrations + master key
 	s, err := store.New(ctx, cfg.DaoConfig())
 	if err != nil {
-		slog.Error("failed to initialise database", "error", err)
+		logging.G(ctx).Errorf("failed to initialise database: err=%q", err.Error())
 		os.Exit(1)
 	}
 	defer s.Close()
@@ -53,7 +52,7 @@ func main() {
 	// auto-generate and persist to DB (zero-config deployment).
 	jwtSecret, err := s.BootstrapJWTSecret(ctx, cfg.JWTSecret)
 	if err != nil {
-		slog.Error("failed to bootstrap JWT secret", "error", err)
+		logging.G(ctx).Errorf("failed to bootstrap JWT secret: err=%q", err.Error())
 		os.Exit(1)
 	}
 	cfg.JWTSecret = jwtSecret
@@ -65,19 +64,19 @@ func main() {
 		sigCh := make(chan os.Signal, 1)
 		signal.Notify(sigCh, syscall.SIGINT, syscall.SIGTERM)
 		<-sigCh
-		slog.Info("received shutdown signal")
+		logging.G(context.Background()).Info("received shutdown signal")
 		shutdownCtx, shutdownCancel := context.WithTimeout(context.Background(), 10*time.Second)
 		defer shutdownCancel()
 		if err := srv.Shutdown(shutdownCtx); err != nil {
-			slog.Error("server shutdown error", "error", err)
+			logging.G(shutdownCtx).Errorf("server shutdown error: err=%q", err.Error())
 		}
 		cancel()
 	}()
 
 	if err := srv.Start(); err != nil && !errors.Is(err, http.ErrServerClosed) {
-		slog.Error("server error", "error", err)
+		logging.G(ctx).Errorf("server error: err=%q", err.Error())
 		os.Exit(1)
 	}
 
-	slog.Info("CubeOps stopped")
+	logging.G(ctx).Info("CubeOps stopped")
 }
--- a/CubeOps/internal/cubemaster/client.go
+++ b/CubeOps/internal/cubemaster/client.go
@@ -11,7 +11,9 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"time"
+
+	"github.com/google/uuid"
+	cubelog "github.com/tencentcloud/CubeSandbox/cubelog"
 )
 
 // CMError carries a CubeMaster business error (non-zero ret_code) out of
@@ -158,10 +160,7 @@ func (c *Client) GetTemplate(ctx context.Context, templateID string) (json.RawMe
 
 // DeleteSnapshot deletes a snapshot via CubeMaster.
 func (c *Client) DeleteSnapshot(ctx context.Context, snapshotID string) (json.RawMessage, error) {
-	body := map[string]interface{}{
-		"request_id": fmt.Sprintf("cubeops-del-snap-%d", time.Now().UnixNano()),
-	}
-	return c.deleteWithBody(ctx, fmt.Sprintf("/cube/snapshot/%s", snapshotID), body)
+	return c.deleteWithBody(ctx, fmt.Sprintf("/cube/snapshot/%s", snapshotID), map[string]interface{}{})
 }
 
 // RollbackSandbox rolls back a sandbox to a snapshot via CubeMaster.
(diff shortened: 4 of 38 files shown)


CubeOps/cmd/cubeops/main.go
```diff
@@ -53,7 +51,7 @@ func main() {
 	// auto-generate and persist to DB (zero-config deployment).
 	jwtSecret, err := s.BootstrapJWTSecret(ctx, cfg.JWTSecret)
 	if err != nil {
-		slog.Error("failed to bootstrap JWT secret", "error", err)
+		logging.G(ctx).WithFields(map[string]any{"error": err.Error()}).Error("failed to bootstrap JWT secret")
```
fslongjin: I think this PR should integrate the request trace structure directly into the `ctx` when logging. Many actions or other fields should reuse the fields inside it, instead of adding so many local maps to write fields — it always feels a bit off. You can refer to how CubeMaster/Cubelet handles this.

fslongjin (changes requested):

fslongjin:
Hi @liciazhu — thanks for this change.

Below are the issues we’d like addressed (or explicitly deferred), focused on **what’s broken / incomplete** and **how to improve**. The happy-path middleware wiring looks good; the gaps are mostly around **field reuse** and **cross-service RequestId**.

---

## Summary

| Priority | Issue | Ask |
|---|---|---|
| **P0** | Outbound CubeMaster body `requestID` ≠ inbound `RequestTrace.RequestID` | Reuse the same ID so logs can be joined |
| **P1** | Business IDs / error codes written into `LogContent` instead of `RequestTrace` | Fill `rt` fields; keep messages semantic |
| **P1** | Middleware always overwrites `rt.RetCode` with HTTP status | Don’t clobber business codes |
| **P1** | `OpenclawApplySpec` logs with `context.Background()` | Pass request `ctx` |
| **P2** | `/health` emits a full Trace every probe | Skip or tag as probe |
| **P2** | Missing `ReNewLogger` / optional `ReportExt` | Align with CubeMaster helpers |

---

## Issue 1 (P0): Cross-service `RequestId` does not actually correlate

### What’s wrong

CubeOps sets a request-scoped `RequestID` in middleware and forwards it as `X-RequestID`. That part is fine.

CubeMaster, however, **does not read** `X-RequestID` into its `RequestTrace`. It fills `rt.RequestID` from the **JSON body** (`requestID` / `request_id`). On the CubeOps side, almost every CubeMaster call still generates a **new** body ID (`sdkRequestID()`, `req-<nano>`, `cubeops-compensate-…`), so:

- `cubeops-*-stat` has RequestId = **A**
- `cubemaster-*-stat` has RequestId = **B**
- You cannot join the two with one query

`X-Caller: cubeops` correctly marks the caller, but that is not per-request correlation. The README claim about “cross-service correlation” via `X-RequestID` is currently overstated.

```mermaid
sequenceDiagram
    participant Client
    participant CubeOps
    participant CubeMaster

    Client->>CubeOps: HTTP + optional X-RequestID
    Note over CubeOps: rt.RequestID = A<br/>(header or new UUID)

    CubeOps->>CubeMaster: Header X-RequestID=A<br/>Body requestID=B (newly generated)
    Note over CubeMaster: Ignores header<br/>rt.RequestID = B from body

    Note over CubeOps,CubeMaster: A ≠ B → cannot join logs by RequestId
```

### Suggested fix

Prefer the inbound trace ID whenever you build a CubeMaster body:

```go
rt := cubelog.GetTraceInfo(ctx)
requestID := sdkRequestID() // fallback only
if rt != nil && rt.RequestID != "" {
    requestID = rt.RequestID
}
// then put requestID into the body under the correct json key
```

A client-level helper that injects / overrides `requestID` / `request_id` / `RequestID` from context would avoid scattering this in every handler. Please add a regression test that asserts outbound body ID equals inbound `RequestTrace.RequestID`.

> CubeMaster reading `X-RequestID` as a fallback would also help long-term, but that can be a follow-up outside this PR. **CubeOps can fix correlation on its own by aligning the body.**

---

## Issue 2 (P1): Prefer `RequestTrace` fields over hand-printed IDs / error codes

### What’s wrong

A lot of new logging looks like:

```go
logging.G(ctx).Errorf("sdk: cubemaster not found: sandboxID=%s retCode=%d msg=%q", ...)
```

`RequestId` / `Action` / `Caller` / `CalleeAction` from the middleware already land as structured JSON fields — good. But once the handler knows a sandbox ID or a CubeMaster `ret_code`, those values are almost never written back onto the shared `*RequestTrace`. They only appear inside `LogContent`, which forces full-text search and diverges from how CubeMaster / Cubelet use cubelog.

```mermaid
flowchart LR
    subgraph today [Current pattern]
        H1[Handler knows sandboxID / ret_code]
        H1 --> L1["Errorf(... sandboxID=%s retCode=%d)"]
        L1 --> LC[LogContent string only]
    end

    subgraph prefer [Preferred pattern]
        H2[Handler knows sandboxID / ret_code]
        H2 --> RT["rt.InstanceID / rt.RetCode / rt.ErrorCode"]
        RT --> RN[ReNewLogger]
        RN --> L2["Errorf(cubemaster not found)"]
        RT --> JSON["JSON fields: InstanceId, RetCode, ..."]
        L2 --> MSG[Semantic LogContent]
    end
```

### Suggested direction

**Rule of thumb:** if `RequestTrace` already has a field for it, put it there — don’t re-print it in the message.

| Put on `RequestTrace` | Typical source |
|---|---|
| `InstanceID` | sandbox / instance id |
| `RetCode` | CubeMaster / business error code |
| `ErrorCode` / `SubErrorCode` | classifiable failures (optional but useful) |
| `InstanceType` | e.g. `cubebox` |

Then:

1. Update `rt` first.
2. Add a CubeOps `ReNewLogger(ctx)` (same idea as CubeMaster/Cubelet) so the cached `Entry` picks up the new fields.
3. Keep `LogContent` semantic: e.g. `"cubemaster not found"`.
4. Use `WithFields` **only** when there is no matching trace field (e.g. `templateID` / `buildID` if you don’t map them elsewhere).

CubeMaster handlers already follow “dimensions on `rt`, message describes what happened.” Aligning CubeOps with that will make ELK/Loki filters and dashboards consistent across components.

---

## Issue 3 (P1): Middleware overwrites business `RetCode` with HTTP status

### What’s wrong

At the end of every request:

```go
rt.RetCode = int64(c.Writer.Status())
cubelog.Trace(rt)
```

Even if a handler later starts writing CubeMaster’s business `ret_code` onto `rt` (Issue 2), the defer will replace it with `404` / `502` / etc. Then `-stat` cannot be filtered by the real business code.

CubeMaster’s convention is the opposite: handlers own business `RetCode`; middleware only adjusts it for special cases (auth failure, client cancel).

```mermaid
flowchart TD
    A[Handler sets rt.RetCode = CM business code] --> B[Response HTTP 502]
    B --> C[Middleware defer: rt.RetCode = 502]
    C --> D[Trace loses business code]
```

### Suggested fix

- Let handlers set `rt.RetCode` from `CMError.RetCode` / service errors on failure paths.
- In the defer: **do not unconditionally overwrite** — e.g. only set HTTP status when `rt.RetCode` is still unset / zero, or store HTTP status somewhere that doesn’t erase the business code.
- Please document the chosen convention in a short comment next to `requestLogger`.

---

## Issue 4 (P1): `OpenclawApplySpec` drops request context

### What’s wrong

`ResolveGatewayToken` correctly takes `ctx` now, but `OpenclawApplySpec` still logs via `logging.G(context.Background())`, so those lines lose `RequestId` and any other trace fields.

### Suggested fix

Thread `ctx` through `OpenclawApplySpec` (and callers) and log with `logging.G(ctx)`.

---

## Issue 5 (P2): Health probes pollute `-stat`

`/health` goes through `requestLogger`, so every probe gets a new UUID and a Trace line. Under frequent k8s probes this dilutes real traffic metrics.

**Suggestion:** skip `Trace` for `/health` (and ready, if any), or set `Caller=probe` so it can be filtered cleanly.

---

## Issue 6 (P2): Small consistency / hygiene items

These are optional for this PR but worth tracking:

1. **`ReNewLogger`** — needed once you start mutating `rt` mid-request (Issue 2).
2. **`ReportExt`-style hop** — today CubeOps only Traces the end-to-end HTTP request; CubeMaster latency is only visible on the CubeMaster side. A later enhancement could DeepCopy + Trace outbound CubeMaster calls.
3. **`logging.Init` MkdirAll failure** — the log says “falling back to stdout” but the function returns without configuring any fallback; either really fall back or fail fast.
4. **`deploy/one-click/webui/nginx.conf` `Cache-Control`** — unrelated to logging; please split out or briefly explain in the PR description why it’s included.

---

CubeOps/Makefile
```diff
@@ -6,8 +6,20 @@
...
+# Priority: env var CUBE_VERSION/CUBE_COMMIT/CUBE_BUILD_TIME > local git/date fallback.
+CUBE_VERSION ?= $(shell git describe --tags --abbrev=0 --match 'v*' 2>/dev/null || echo 0.0.0-dev)
+CUBE_COMMIT ?= $(shell git rev-parse HEAD 2>/dev/null || echo unknown)
+CUBE_BUILD_TIME ?= $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+
+VERSION_PKG=github.com/tencentcloud/CubeSandbox/CubeOps/internal/version
+BUILD_FLAGS=-ldflags "-s -w \
+  -X '$(VERSION_PKG).Version=$(CUBE_VERSION)' \
+  -X '$(VERSION_PKG).Commit=$(CUBE_COMMIT)' \
+  -X '$(VERSION_PKG).BuildTime=$(CUBE_BUILD_TIME)'"
```
fslongjin, line 19: **[Should fix]** This ldflags wiring is correct, but the main build paths are not updated:

1. Root `Makefile` `make cubeops` still uses `-X main.Version/Commit/BuildTime` (those symbols no longer exist in `cmd/cubeops`).
2. `deploy/one-click/build-release-bundle-builder.sh` builds cubeops with only `-ldflags "-s -w"`.
3. `deploy/one-click/build-release-bundle.sh` does not pass a `version_pkg` for cubeops (unlike network-agent).

Please align those paths with `VERSION_PKG` here, otherwise released binaries keep reporting `0.0.0-dev` via `cubelog.SetVersion`.

CubeOps/internal/cubemaster/client.go
```diff
@@ -263,13 +261,57 @@ func (c *Client) AdoptTemplateCompatBaseline(ctx context.Context, body interface
...
+}
+
 // --- internal helpers ---
 
 func (c *Client) get(ctx context.Context, path string) (json.RawMessage, error) {
 	req, err := http.NewRequestWithContext(ctx, http.MethodGet, c.baseURL+path, nil)
 	if err != nil {
 		return nil, err
 	}
+	setTraceHeaders(req, RequestIDFromContext(ctx))
```
fslongjin: **[Should fix]** `get()` only sets `X-RequestID`. As the README notes, CubeMaster ignores that header and expects RequestID in body/query.

Callers such as `GetNodes` / `ClusterOverview` / `GetTemplate` therefore often cannot be joined to CubeMaster `-stat` lines.

Please also inject query params (reuse `EnsureRequestIDQuery`), or teach CubeMaster to read `X-RequestID`, and update the README accordingly.

CubeOps/README.md
```diff
@@ -82,6 +82,13 @@ tail -50 /data/log/CubeOps/cubeops-req.log
 tail -f /data/log/CubeOps/cubeops-req.log
 ```
 
+Every HTTP request receives an `X-RequestID` when one is not supplied. The
+request trace is attached to the request context, so request-scoped `cubelog`
+records include fields such as `RequestId`, `Action`, `Caller`, and
+`CalleeAction`. CubeMaster calls propagate `X-Caller: cubeops` and stamp the
+inbound `RequestID` into the outbound body and query string — that is how
+cross-service correlation works (CubeMaster ignores `X-RequestID`).
```
fslongjin, line 90: **[Suggestion]** This says outbound calls stamp RequestID into body and query. That is true for `post` / `deleteWithBody` / `getWithQuery`, but not for plain `get()`.

Please narrow the wording so operators are not misled when correlating cluster/meta GET traffic.

CubeOps/internal/redact/redact.go
```diff
@@ -16,12 +16,42 @@ package redact
...
+func (s Secret) String() string {
+	return redacted
+}
+
+func (s Secret) GoString() string {
+	return redacted
+}
+
+func (s Secret) MarshalJSON() ([]byte, error) {
+	return json.Marshal(string(s))
```
fslongjin: **[Suggestion]** `MarshalJSON` emits plaintext. cubelog serializes fields with jsoniter, which honors `MarshalJSON`, so any future `WithFields` that includes a `Secret` (or a map embedding one) will log the credential.

Current call sites wrap with `redact.Value(...)` and are safe, but the type is easy to misuse. Prefer redact-by-default for logs, and take plaintext only at the outbound API boundary.

CubeOps/internal/server/server.go
```diff
@@ -65,19 +69,19 @@ func (s *Server) Shutdown(ctx context.Context) error {
...
 func (s *Server) buildRouter() *gin.Engine {
 	// We use gin.New() rather than gin.Default() so we can attach our own
-	// slog-based access logger and recovery handler. gin.Default() writes to
+	// cubelog-based access logger and recovery handler. gin.Default() writes to
 	// stdout and bypasses any logger the operator has configured.
 	gin.SetMode(gin.ReleaseMode)
 	r := gin.New()
-	r.Use(gin.Recovery())
 	r.Use(requestLogger())
+	r.Use(gin.Recovery())
```
fslongjin: **[Suggestion]** The comment says we attach a cubelog recovery handler, but this still uses `gin.Recovery()`, which writes stacks to stderr and skips `cubeops-req` / `cubeops-stat`.

CubeMaster logs panics into cubelog and records a failure on the trace. Please do the same here so panics remain visible in the rolling logs.

CubeOps/internal/server/server.go
```diff
@@ -122,19 +125,71 @@ func (s *Server) buildRouter() *gin.Engine {
...
 }
 
-// requestLogger is a gin middleware that emits a structured slog line per
-// request, matching the format the CubeOps CLI tooling expects.
+// requestLogger emits request traces to the stat log.
+//
+// Convention: handlers own business RetCode (e.g. CubeMaster ret_code); the
+// middleware only falls back to HTTP status when rt.RetCode is still zero so
+// business codes aren't clobbered by 404/502.
 func requestLogger() gin.HandlerFunc {
```
fslongjin, line 134: **[Suggestion]** `requestLogger` / `requestIDFromHeader` have no unit tests. At least cover:

1. Invalid or oversized `X-RequestID` is rejected and regenerated
2. Response header echoes `X-RequestID`
3. `/health` skips the `-stat` line

The client/redact packages are better covered; this middleware is the entry point and deserves the same.

deploy/one-click/webui/nginx.conf
```diff
@@ -58,6 +58,7 @@ http {
 
         # CubeOps (ops/admin endpoints) — rewrite /opsapi → /api
         location /opsapi/ {
+            add_header Cache-Control "no-store" always;
```
fslongjin: **[Suggestion]** Adding `Cache-Control: no-store` on `/opsapi/` is unrelated to this logging change. The K8s WebUI `/opsapi/` block in chart `_helpers.tpl` does not get the same header.

Please drop this from the PR, or mirror it on the K8s side so the two deployments stay consistent.

fslongjin (commented):
## Review notes

Direction looks right (cubelog req/stat, context propagation, egress redaction). A few gaps should be fixed or clearly followed up before merge:

**Should fix**
1. Version ldflags: `CubeOps/Makefile` / `Dockerfile` are correct, but the root `Makefile` (`make cubeops`) still injects `-X main.Version=...`, and one-click builders do not pass the version package (`build-release-bundle-builder.sh` uses bare `-ldflags "-s -w"`; `build-release-bundle.sh` omits `version_pkg` for cubeops). Release binaries will stay at `0.0.0-dev`.
2. Cross-service RequestID: plain `get()` only sets `X-RequestID`, which CubeMaster ignores. See inline comments.

**Suggestions** (follow-up OK if tracked)
- `redact.Secret` MarshalJSON plaintext footgun
- Panic recovery bypasses cubelog
- Missing middleware unit tests
- Unrelated nginx `Cache-Control` change vs K8s WebUI

Inline comments below.

fslongjin (approved):
LGTM

Issue #1125, opened 2026-07-23
[Feature Request] support safe node deletion and decommissioning


luzhixing12345:
## Problem / Motivation

CubeMaster currently supports node registration, status updates, labels, and scheduling isolation, but it has no supported way to remove a node that has been permanently decommissioned. As a result, retired or replaced nodes remain visible in node lists and the version matrix, and stale registration, heartbeat, component-version, and scheduler-cache state must be cleaned up manually.

Node deletion is more than deleting a row. A node may still have running sandboxes, a Cubelet may continue sending heartbeats or re-register after the deletion, and multiple CubeMaster replicas may hold the node in memory. A safe solution should follow the control-plane-owned, database-backed, concurrent-safe, and fail-safe principles introduced by #894.

The feature is intended for permanently retiring a compute node. Temporary maintenance should continue to use node isolation.

## Proposed Solution

Add a control-plane-owned node decommission workflow with an idempotent API and CLI:

```bash
cubemastercli node delete <node-id> [node-id ...]
```

```http
DELETE /internal/meta/nodes/{node_id}
```

Deletion should use the following safety rules:

1. The node must already be isolated (`scheduling-disabled=true`), so no new sandbox can be admitted while deletion is being evaluated.
2. The request must fail if the control plane still knows of any non-terminal sandbox assigned to the node. The error should report that active workloads remain, without deleting partial state.
3. The operation must be idempotent: deleting an already-deleted node returns success.
4. The operation must be serialized with registration, heartbeat/status updates, label changes, isolation changes, and another deletion for the same node. Database row locking/transactions must protect multi-replica deployments in addition to the existing per-node in-process lock.
5. Successful deletion must persist a tombstone owned by the control plane before removing the node from scheduler-visible state. A Cubelet using the same `node_id` must not be able to resurrect a decommissioned node through register or heartbeat requests. Reusing the machine should require a new node identity (a separate explicit restore workflow can be added later if needed).
6. In the same transactional workflow, clean up node-owned metadata, including registration/status data and component-version rows. Other node-keyed records must be classified explicitly:
   - ephemeral cache/observability data may be deleted;
   - historical sandbox records should be retained;
   - template replicas, artifacts, snapshots, and other node-local resources must either be proven absent, cleaned up by their owning subsystem, or cause deletion to fail with a clear conflict.
[...]

PR #1202, opened 2026-07-28
feat(cubemaster): support safe node deletion

luzhixing12345:
Ref https://github.com/TencentCloud/CubeSandbox/issues/1125

CubeMaster had no supported way to remove retired nodes. Deleting records manually was also unsafe because it did not verify whether the node was isolated or still running sandboxes.

## Changes

This PR adds safe node deletion through:

```bash
cubemastercli node rm <node_id>
```
And the internal API:

```
DELETE /internal/meta/nodes/{node_id}
```

Deletion requires the node to be:

- Isolated.
- Free of sandboxes.

CubeMaster queries Cubelet for the current sandbox inventory.

On success, CubeMaster removes the node registration, status, component versions, metrics, and scheduler caches. Other CubeMaster replicas remove the node during metadata reload. The same node_id may register again later.

The English and Chinese documentation is also updated.

## CLI verification

Deleting a non-isolated node should fail:

```bash
$ cubemastercli node rm <node>
2026/07/28 17:04:31 delete failed: <node> node must be isolated before deletion
cubemastercli run fail: <node>: node must be isolated before deletion
```

Isolate the node:

```bash
$ cubemastercli node isolate <node>
node <node> isolated: scheduling_disabled=true
```

Deleting a node with remaining sandboxes should fail:

```bash
$ cubemastercli node rm <node>
2026/07/28 17:04:44 delete failed: <node>node still has sandboxes: 1 found
cubemastercli run fail: <node>: node still has sandboxes: 1 found
```

After removing all sandboxes:

```bash
$ cubemastercli node rm <node>
node <node> deleted
```

Verify that the node is gone:

```bash
$ cubemastercli node list
# Expected: node-1 is absent
```

Failures do not prevent later nodes from being attempted, but the command returns a non-zero exit code if any deletion fails.

### DIFF
--- a/CubeMaster/pkg/nodemeta/deletion.go
+++ b/CubeMaster/pkg/nodemeta/deletion.go
@@ -1,0 +1,95 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+//
+
+package nodemeta
+
+import (
+	"context"
+	"errors"
+	"fmt"
+
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/db/models"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/log"
+	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/localcache"
+	"gorm.io/gorm"
+	"gorm.io/gorm/clause"
+)
+
+var (
+	ErrNodeNotIsolated = errors.New("node must be isolated before deletion")
+	ErrNodeNotFound    = errors.New("node not found")
+)
+
+// DeleteNode removes the node's current control-plane metadata. A later
+// Cubelet restart may register the same node ID again. The node must first be
+// cordoned. The service layer is responsible for confirming that the Cubelet
+// has no sandboxes before calling this function.
+func DeleteNode(ctx context.Context, nodeID string) error {
+	if nodeID == "" {
+		return fmt.Errorf("node_id is required")
+	}
+	unlock := global.lockNodeLabels(nodeID)
+	defer unlock()
+
+	var reg models.NodeRegistration
+	if err := global.db.WithContext(ctx).Where("node_id = ?", nodeID).Take(&reg).Error; err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return fmt.Errorf("%w: %s", ErrNodeNotFound, nodeID)
+		}
+		return err
+	}
+	labels, err := parseLabelsJSON(reg.LabelsJSON)
+	if err != nil {
+		return fmt.Errorf("%w: %v", ErrLabelsJSONCorrupt, err)
+	}
+	if !schedulingDisabledFromLabels(labels) {
+		return ErrNodeNotIsolated
+	}
+
+	if err := global.db.WithContext(ctx).Transaction(func(tx *gorm.DB) error {
+		var reg models.NodeRegistration
+		if err := tx.Clauses(clause.Locking{Strength: "UPDATE"}).
+			Where("node_id = ?", nodeID).Take(&reg).Error; err != nil {
+			return err
+		}
+		lockedLabels, err := parseLabelsJSON(reg.LabelsJSON)
+		if err != nil {
+			return fmt.Errorf("%w: %v", ErrLabelsJSONCorrupt, err)
+		}
+		if !schedulingDisabledFromLabels(lockedLabels) {
+			return ErrNodeNotIsolated
+		}
+		if err := tx.Unscoped().Where("node_id = ?", nodeID).Delete(&models.NodeStatus{}).Error; err != nil {
+			return err
+		}
+		if err := tx.Where("node_id = ?", nodeID).Delete(&models.NodeComponentVersion{}).Error; err != nil {
+			return err
+		}
+		return tx.Unscoped().Where("node_id = ?", nodeID).Delete(&models.NodeRegistration{}).Error
+	}); err != nil {
+		if errors.Is(err, gorm.ErrRecordNotFound) {
+			return fmt.Errorf("%w: %s", ErrNodeNotFound, nodeID)
+		}
+		return err
+	}
+
+	global.mu.Lock()
+	delete(global.nodes, nodeID)
+	global.mu.Unlock()
+	evictDeletedNode(ctx, nodeID)
+	log.G(ctx).Infof("node deleted node_id=%s", nodeID)
+	return nil
+}
+
+func schedulingDisabledFromLabels(labels map[string]string) bool {
+	return labels[constants.LabelSchedulingDisabled] == constants.LabelSchedulingDisabledValue
+}
+
+func evictDeletedNode(ctx context.Context, nodeID string) {
+	localcache.EvictNode(nodeID)
+	if err := localcache.DeleteNodeMetric(ctx, nodeID); err != nil {
+		log.G(ctx).Warnf("node metric cleanup failed after deletion node_id=%s err=%v", nodeID, err)
+	}
+}
--- a/CubeMaster/pkg/nodemeta/service.go
+++ b/CubeMaster/pkg/nodemeta/service.go
@@ -209,9 +209,10 @@ type service struct {
 	// version writes or overwrite a newer in-memory hash with an older one.
 	versionWriteLocks sync.Map
 
-	// labelWriteLocks serialises label read-modify-write (register merge,
-	// admin labels, isolation) per node so DB commit and in-memory/localcache
-	// publication stay ordered.
+	// labelWriteLocks serialises the complete per-node lifecycle (register,
+	// heartbeat, labels, isolation, deletion) so DB commit and in-memory/cache
+	// publication stay ordered on a replica. Database row locks provide the
+	// corresponding cross-replica ordering.
 	labelWriteLocks sync.Map
 
 	// hostFactsWriteLocks serialises the host-facts persist per node so
@@ -390,13 +391,26 @@ func UpdateNodeStatus(ctx context.Context, nodeID string, req *UpdateNodeStatusR
 		HeartbeatUnix:      req.HeartbeatTime.Unix(),
 		Healthy:            reportedReady,
 	}
-	if err := global.db.Clauses(clause.OnConflict{
-		Columns: []clause.Column{{Name: "node_id"}},
-		DoUpdates: clause.AssignmentColumns([]string{
-			"conditions_json", "images_json", "local_templates_json",
-			"heartbeat_unix", "healthy", "updated_at",
-		}),
-	}).Create(status).Error; err != nil {
+	// Serialize status writes with the complete per-node lifecycle. The
+	// registration row lock prevents a heartbeat from recreating status after
+	// deletion has committed; a heartbeat for a retired registration therefore
+	// fails instead of leaving an orphan NodeStatus row.
+	unlock := global.lockNodeLabels(nodeID)
+	defer unlock()
+	if err := global.db.Transaction(func(tx *gorm.DB) error {
+		var reg models.NodeRegistration
+		if err := tx.Clauses(clause.Locking{Strength: "UPDATE"}).
+			Where("node_id = ?", nodeID).Take(&reg).Error; err != nil {
+			return err
+		}
+		return tx.Clauses(clause.OnConflict{
+			Columns: []clause.Column{{Name: "node_id"}},
+			DoUpdates: clause.AssignmentColumns([]string{
+				"conditions_json", "images_json", "local_templates_json",
+				"heartbeat_unix", "healthy", "updated_at",
+			}),
+		}).Create(status).Error
+	}); err != nil {
 		return nil, err
 	}
 
@@ -968,7 +982,7 @@ func GetNode(ctx context.Context, nodeID string) (*NodeSnapshot, error) {
 	defer global.mu.RUnlock()
 	snap, ok := global.nodes[nodeID]
 	if !ok {
-		return nil, gorm.ErrRecordNotFound
+		return nil, fmt.Errorf("%w: %s", ErrNodeNotFound, nodeID)
 	}
 	return cloneSnapshotWithCurrentHealth(snap, time.Now()), nil
 }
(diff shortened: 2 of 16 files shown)


CubeMaster/pkg/service/sandbox/node_deletion.go
```diff
@@ -0,0 +1,61 @@
...
+		Filter: &cubebox.CubeSandboxFilter{
+			LabelSelector: map[string]string{"io.kubernetes.cri.container-type": "sandbox"},
+		},
+	})
+	if err != nil {
+		// A deliberately stopped Cubelet cannot answer the final inventory
+		// query. Permit that documented node deletion workflow only when the
+		// node is already unhealthy and its last real metric explicitly
+		// reported zero sandboxes. A missing metric remains fail-closed.
+		if !n.Healthy && !n.MetricUpdate.IsZero() {
```
fslongjin: **[P0] Cubelet-unreachable fallback trusts a possibly-stale sandbox count**

When Cubelet is unreachable, the code uses the last reported sandbox count (`n.MvmNum`) to decide whether the node is empty.

The problem: that count can be stale. Disk-only heartbeats refresh `MetricUpdate` but do NOT update `MvmNum` (see `UpdateNodeMetricInProcess` — only `HasAllocated` writes `MvmNum`). So this sequence is possible:

1. Node once reported `MvmNum = 0`
2. Only disk heartbeats arrive after that — `MetricUpdate` is refreshed, but `MvmNum` stays at the old 0
3. Node actually has sandboxes again
4. Cubelet is stopped or the network blips, so `List` fails
5. Code sees "unhealthy + has MetricUpdate + MvmNum==0" and allows deletion

Result: control-plane records are deleted while sandboxes still exist on the node.

The documented workflow ("stop Cubelet, then delete") makes this path the most likely one to hit.

Suggestion: at minimum, add a freshness check on the metric (e.g. require the last update to be an allocated-type report, or cap the age of `MetricUpdate`). Also, don't treat any `List` failure as "Cubelet was deliberately stopped" — a transient network error hits the same branch. If a short-term fix is hard, consider removing this fallback and requiring Cubelet to be reachable, or adding a `--force` option so the operator explicitly takes responsibility.

CubeMaster/pkg/service/sandbox/node_deletion.go
```diff
@@ -0,0 +1,61 @@
...
+// deletion transaction.
+func DeleteNode(ctx context.Context, nodeID string) error {
+	count, err := countNodeSandboxes(ctx, nodeID)
+	if err != nil {
+		return fmt.Errorf("verify node sandboxes: %w", err)
+	}
+	if count != 0 {
+		return fmt.Errorf("%w: %d found", ErrNodeHasSandboxes, count)
+	}
+	return nodemeta.DeleteNode(ctx, nodeID)
```
fslongjin, line 45: **[P1] Empty-check and delete are not atomic; in-flight creates can slip through**

"Confirm the node is empty" and "actually delete it" are two separate steps with no lock tying them together.

After isolation, new admits on this replica are correctly blocked by `admitSelectedHost` (good). But two cases can still slip through:

1. A create request that already passed admit but hasn't finished `cubelet.Create` will keep going while the node is being deleted.
2. In a multi-master deployment, other replicas may not have synced the cordon yet and can still schedule to this node.

Also, the local in-flight counter (`LocalCreateNum`) is not consulted at all during deletion.

The docs suggest "wait >=60s after isolating before deleting", but that's only a recommendation — the code doesn't enforce it.

Suggestion: re-check the inventory right before the delete transaction, or move the empty-check and the delete into the same lock critical section. At minimum, refuse deletion when `LocalCreateNum > 0`.

CubeMaster/pkg/service/sandbox/node_deletion.go
```diff
@@ -0,0 +1,61 @@
...
+}
+
+// countNodeSandboxes asks the target Cubelet directly. A transport failure is
+// returned to the caller so node deletion fails closed instead of mistaking an
+// unreachable node for an empty one.
+func countNodeSandboxes(ctx context.Context, nodeID string) (int, error) {
+	n, ok := localcache.GetNode(nodeID)
+	if !ok || n == nil {
+		return 0, fmt.Errorf("node %s is not present in local cache", nodeID)
```
fslongjin: **[P2] Deleting an already-deleted node returns InternalError instead of NotFound**

When the node is already deleted or absent from this replica's local cache, this returns a generic error that ends up wrapped as `MasterInternalError` instead of `NotFound`.

Impact: batch `rm` or scripted retries will treat an already-deleted node as a failure, which is easy to misread in automation.

Suggestion: distinguish "node doesn't exist" (return `NotFound`, or treat as idempotent success) from "Cubelet unreachable, can't verify inventory".

CubeMaster/pkg/nodemeta/deletion.go
```diff
@@ -0,0 +1,84 @@
...
+// cordoned. The service layer is responsible for confirming that the Cubelet
+// has no sandboxes before calling this function.
+func DeleteNode(ctx context.Context, nodeID string) error {
+	if nodeID == "" {
+		return fmt.Errorf("node_id is required")
+	}
+	unlock := global.lockNodeLabels(nodeID)
+	defer unlock()
+
+	labels, err := readLabelsJSONForUpdate(global.db.WithContext(ctx), nodeID)
```
fslongjin: **[P3] Out-of-transaction FOR UPDATE is a no-op and misleading**

`readLabelsJSONForUpdate` uses `SELECT ... FOR UPDATE`, but it runs outside the transaction, so the lock is released immediately. The real protection is the `FOR UPDATE` inside the later transaction.

Behavior is safe because isolation is re-checked inside the transaction, but this code can mislead future readers into thinking the row is already locked here, and they might add logic on that assumption.

Suggestion: fold this read into the transaction below, or drop the `FOR UPDATE` here to avoid the misleading impression.

CubeMaster/pkg/nodemeta/service.go
```diff
@@ -974,16 +988,40 @@ func (s *service) applyReloadResult(next map[string]*NodeSnapshot) {
...
+// concurrent registration from being classified using an earlier view.
+// The caller syncs localcache outside the lock. The critical section uses
 // defer s.mu.Unlock() so a panic in the merge loop cannot leak the lock and
 // silently deadlock subsequent register/heartbeat handlers.
-func (s *service) mergeReloadResult(next map[string]*NodeSnapshot) []*NodeSnapshot {
+func (s *service) mergeReloadResult(next map[string]*NodeSnapshot) ([]*NodeSnapshot, []string) {
 	s.mu.Lock()
 	defer s.mu.Unlock()
+	evicted := make([]string, 0)
+	for nodeID := range s.nodes {
```
fslongjin, line 1403: **[P2] A stale reload snapshot can briefly re-insert a deleted node into the cache**

Consider this ordering:

1. The periodic reload reads a DB snapshot that still contains node X
2. Concurrently, `DeleteNode` removes X from the DB, from `s.nodes`, and evicts it from localcache
3. The reload continues merging with the now-stale snapshot. Since `mergeReloadResult` only holds `s.mu` (not the per-node lifecycle lock), it inserts X back into `s.nodes` and syncs it into localcache

Result: a deleted node can briefly reappear as a "ghost" in node lists, usually until the next reload clears it. This contradicts the docs, which say deletion "immediately cleans up the current instance's view".

Safety impact is lower than the issues above (the stale snapshot usually still carries the isolation label, so the node won't be rescheduled), but it can confuse monitoring and ops.

Suggestion: add a generation/read-timestamp check on the reload snapshot, or in `mergeReloadResult` refuse to re-insert IDs whose registration is gone from the DB and force-evict them instead.

fslongjin (changes requested):
Thanks for this PR — the isolate → empty → delete loop is a useful and long-overdue workflow. The direction is right: forced isolation first, live `List` when Cubelet is reachable, per-node lifecycle lock + DB row lock, replica eviction on reload, and a CLI that continues on per-node failure with a non-zero exit.

A few correctness concerns before this lands (details in the inline comments):

- **[P0]** The Cubelet-unreachable fallback trusts a possibly-stale `MvmNum`. Disk-only heartbeats refresh `MetricUpdate` without updating `MvmNum`, so the fallback can delete a node that still has sandboxes. This is exactly the path the docs recommend ("stop Cubelet, then delete").
- **[P1]** The empty-check and the delete are not atomic; in-flight creates and multi-master cordon-propagation lag can slip through. `LocalCreateNum` is not consulted.
- **[P2]** A stale reload snapshot can briefly re-insert a deleted node into the cache.
- **[P2]** Deleting an already-deleted node returns `InternalError` instead of `NotFound`, which trips up automation.
- **[P3]** The out-of-transaction `FOR UPDATE` is a no-op and misleading.

---

Feature request — add a `--force` option:

The current "fail closed by default" design is correct. But in real ops there are cases where the node is permanently gone and the remaining sandbox records don't matter, yet deletion is blocked because Cubelet is unreachable or a few stale sandbox records remain. Today the only workaround is hand-editing the DB, which is risky.

Could we add an explicit force flag, e.g.:

```bash
cubemastercli node rm <node_id> --force
# or
DELETE /internal/meta/nodes/{node_id}?force=true
```

To prevent misuse, force-delete could still require:
- prior isolation (force is not a bypass for cordon)
- an explicit `--force` with a prominent confirmation prompt in the CLI
- an audit log entry (who / when / which node)
- optionally a confirmation parameter acknowledging the node's sandboxes are disposable

This keeps the safe default and gives ops a documented "I know what I'm doing" escape hatch instead of going around the system to edit the DB. Happy to open a follow-up issue if this direction is acceptable.

CubeMaster/pkg/service/sandbox/node_deletion.go
luzhixing12345: fixed

CubeMaster/pkg/service/sandbox/node_deletion.go
luzhixing12345: modified

CubeMaster/pkg/nodemeta/service.go
luzhixing12345: not fixed in this pr, fix in the future

CubeMaster/pkg/service/sandbox/node_deletion.go
luzhixing12345: it's no need to atomic check a to-be-deleted node, node must be isolated first

CubeMaster/pkg/nodemeta/deletion.go
luzhixing12345: fixed

luzhixing12345:
Thanks for the review. I’ve updated the PR to:

- Remove the unsafe MvmNum fallback and fail closed when Cubelet is unreachable.
- Add explicit force deletion with CLI confirmation while still requiring isolation.
- Return NotFound for missing nodes.
- Remove the ineffective out-of-transaction FOR UPDATE.
- Add tests and update the documentation.

The create/delete atomicity and stale reload snapshot issues require broader lifecycle coordination and will be handled separately, rather than in this PR.

fslongjin:
## Adversarial Review — feat(cubemaster): support safe node deletion

The P0 from the previous round (trusting a stale `MvmNum` when Cubelet is unreachable) has been fixed. Fail-closed inventory check, the double isolation check, `NotFound` mapping, and en/zh docs matching the implementation are all clearly better, and CI is green. Below are the items that still need to be addressed before merge.

### Blocking

**1. Reload eviction can wrongly remove a freshly re-registered node**

`CubeMaster/pkg/nodemeta/service.go:1399-1407`

The problem is the order of operations. Today the code first removes every node absent from the snapshot from the in-memory map `s.nodes`, and only afterwards decides whether to also clear it from the scheduler cache. That decision is based on "is the node still in `s.nodes`?" — but the previous step already deleted it, so the check always sees "not present" and always clears the scheduler cache.

This ordering cannot defend the following scenario, which the PR explicitly supports ("the same `node_id` may register again later"):

1. This CubeMaster starts a metadata reload and reads a DB snapshot that does not contain node X (X was already deleted)
2. Concurrently, Cubelet re-registers X on this CubeMaster: writes DB, writes `s.nodes`, writes the scheduler cache
3. The reload merges the stale snapshot from step 1: X is in `s.nodes` but not in the snapshot, so X is deleted from `s.nodes`
4. The subsequent guard sees "X is not in `s.nodes`" and evicts X from the scheduler cache too

Effect: this CubeMaster considers node X absent until the next reload or the next heartbeat that lands on this replica (default worst case ~30s). In a single-CubeMaster deployment, creates targeting X fail during that window. This is not the harmless "an isolated node briefly still appears in the list" case — it is a live node being dropped by the control plane.

The existing test `TestApplyReloadResultDoesNotEvictReregisteredNode` only covers "registration lands after the merge". It does not cover "registration lands between the snapshot read and the merge acquiring the lock", which is the case above.

**2. PR description and commit body still describe the dangerous behavior that was removed**

PR body:

> If Cubelet is unreachable, deletion is allowed only when the node is unhealthy and its last valid metric reported zero sandboxes.

Commit body:

> handle unreachable Cubelets using the last reported sandbox metric

The actual code and the en/zh docs say the opposite: when Cubelet is unreachable, normal deletion fails closed and never trusts a cached sandbox count; only an explicit `force=true` bypasses the inventory check, and even then isolation is still required. That is exactly the previous P0 fix. But the PR description and commit still describe the old behavior. After merge, GitHub and git log will record an incorrect safety semantics as fact. At minimum the PR description must be updated to match the code and docs before merge.

**3. `CHANGES_REQUESTED` is still open**

The change requests from the previous round are still open. Deferring the create/delete TOCTOU and the stale-snapshot re-insertion is acceptable, but a follow-up issue should be opened and linked from this PR, and the original reviewer should dismiss or re-approve, otherwise these two items lose tracking after merge.

### Other issues

**4. `--force` "succeeds" with exit code 0 in non-interactive environments without actually deleting**

`CubeMaster/cmd/cubemastercli/commands/cubebox/node.go:157-160` + `commands/client.go:24-34`

The documented retirement path is "stop Cubelet → `cubemastercli node rm --force`". `--force` always prompts for confirmation. When stdin reaches EOF (scripts, CI, systemd without a TTY), the confirm function returns false, the CLI prints `force deletion cancelled` and `return nil`, i.e. exit code 0. A script will believe the command succeeded, but the node was not deleted.

**5. The documented retirement path effectively bypasses "safe deletion"**

`docs/guide/node-isolation.md:166-174` and the zh counterpart

The recommended sequence is: isolate → wait 60s → destroy sandboxes → stop Cubelet → force-delete. Once Cubelet is stopped, normal deletion is necessarily fail-closed, so in production the retirement path almost always goes through `force=true`. This is internally consistent, but it means the live inventory query is not the last gate of the recommended flow — `force` becomes the main path. The internal API is unauthenticated by default, and `DELETE ...?force=true` only requires the node to be isolated. The docs already note the internal/auth caveats, but this security model needs an explicit maintainer decision. The docs also do not give a non-`force` deletion example (Cubelet still up, inventory zero), so readers will assume `force` is the only usage.

**6. The deletion gate is stricter than the scheduling gate; a non-canonical label value dead-locks deletion**

`CubeMaster/pkg/nodemeta/deletion.go:86-88` requires `scheduling-disabled == "true"` to count as isolated, while `node.DecodeSchedulingDisabled` treats any presence of the label as cordoned (including `"false"` or garbage values). The normal isolate API only writes the canonical `"true"`, and neither the generic labels API nor Cubelet register can touch this key, so dead-locking deletion requires a manual DB edit or legacy dirty data. There is a unit test pinning "must be canonical value". The net effect: scheduling fail-closed (won't schedule) + deletion fail-closed (won't delete) leaves a dirty node unable to take work and unable to be deleted; the only recovery is to isolate once more to rewrite `"true"`.

---

Autonomously-by: Cursor:glm-5.2-high

luzhixing12345:
> Reload eviction can wrongly remove a freshly re-registered node

this is a known problem, fully addressing the issue of multi-node data synchronization in this PR would introduce significant complexity. Given that the node deletion operation typically does not trigger this kind of unusual behavior, it is acceptable. Issues related to multi-node synchronization will be fixed later.

> PR description and commit body still describe the dangerous behavior that was removed

modified



> The documented retirement path effectively bypasses "safe deletion"

true, will modify later

fslongjin (approved):

PR #1210, opened 2026-07-29
feat(CubeBlobstore): add cube-snapshot COS chunked snapshot tool

changpe1:
Add a new CubeBlobstore module hosting cube-snapshot, a chunked, deduplicated, zstd-compressed snapshot tool that saves/restores a flat directory of regular files to/from an S3/COS bucket. It provides the save/restore/ls/rm subcommands with cos.cfg + env credential resolution, zero-extent detection, per-chunk/per-file SHA256 verification and sparse restore.

Chunks use a content-addressed object layout,
<uuid>/chunks/<chunk-plaintext-sha256>, so identical chunks collapse to a single stored object and keys stay fixed-length regardless of how long or deeply-nested the source file path is (avoiding long-key and path-encoding pitfalls). The manifest records each file's absolute path plus the ordered per-chunk sha list used to fetch chunks and reassemble files on restore. Built on AWS SDK for Go v2 and klauspost/compress/zstd (pure Go, no cgo), producing a static binary.

The module uses the cmd/<name>/ layout so future storage utilities can be added alongside cube-snapshot. Wire it into the root Makefile: a cube-snapshot builder target plus BINARIES, manual-release tarball and fmt.


kinwin-ustc:
We should abstract out a general storage backend interface, with S3/COS being just one storage implementation corresponding to this abstraction

PR #1217, opened 2026-07-29
docs: add runnable OpenCode sandbox integration

blues-kun:
## Summary

Related to #644. This contribution completes the OpenCode sub-direction.

- add English and Chinese end-to-end guides under the existing integrations documentation;
- add a runnable `examples/opencode-integration` project with a pinned OpenCode 1.18.9 binary and SHA-256 verification;
- cover template construction, runtime dependencies, direct and CubeEgress-vault credential injection, deny-by-default egress, session state, pause/resume, artifact export, and troubleshooting;
- use a deterministic failing-test -> agent repair -> host verification workflow instead of a hello-world prompt;
- keep the activity Hy3 key out of the sandbox when the CubeEgress credential vault path is selected.

## Validation performed

- `13` offline unit tests pass with the declared `cubesandbox`/`e2b` SDK dependencies;
- Python compilation, JSON/shell/link/bilingual navigation checks pass;
- OpenCode 1.18.9 was run on the host against the real Hy3 endpoint for a native read-tool call.

## Honest environment boundary

A live CubeSandbox cluster with KVM, CubeEgress, and snapshot services was not available on this host, so the full in-sandbox build plus pause/resume flow is documented and unit-tested but not presented as a live cluster result. A Docker image build is therefore not used as a substitute for that missing cluster-level evidence.


blues-kun:
Thanks for the detailed review. I addressed all five observations in commit `209c8a3` and added one runtime fix found while following the review's security reasoning.

Changes:
- Removed inherited `bash`, `ca-certificates`, and `curl` from the install list, split OS/tool installation for better cache behavior, and added an explicit non-amd64 fail-fast guard for the pinned x64 artifact.
- Replaced the narrow `rm -rf *` rule with direct `rm`, `/bin/rm`, `/usr/bin/rm`, and `command rm` deny patterns. The EN/ZH docs now state clearly that wrappers and equivalent tools can bypass command-pattern guardrails, so MicroVM and egress controls remain the security boundary.
- Added the `TextIO` annotation to `stream_writer`.
- Made the JSONL summary handle both OpenCode `filePath` and URL-oriented `url` event fields.
- Changed the Chinese troubleshooting rows to preserve the exact searchable error `Model not found: tokenhub/hy3`.

Additional security fix:
- `network_policy.py` previously checked for the placeholder key without passing the process environment, which could fail as `<unset>` on a live cluster. It now performs a two-stage check: the sandbox-wide environment must contain no `HY3_API_KEY`, and the exact environment later passed to OpenCode must contain only the CubeEgress placeholder. The real key is never printed.

Validation:
- 21 unit tests passed (up from 13).
- Black, `compileall`, JSON parsing, `git diff --check`, and staged credential scan passed.
- OpenCode `1.18.9` successfully resolved the updated config via `opencode debug config --pure`, including the broader deny rule.
- Both EN and ZH VitePress pages built successfully.
- The local Buildx is `0.14.1`, so the optional no-build `--check` command was not claimed as executed; the docs now state its official `0.15.0+` requirement.

Full CubeEgress and pause/resume execution still requires an accessible CubeSandbox cluster and remains explicitly marked as pending.

blues-kun:
Follow-up review items are addressed in `03a3b76`.

- Both direct-key drivers (`run_opencode.py` and `resume_opencode.py`) now print a visible security warning to stderr before sandbox creation. The warning points users to `network_policy.py`.
- The source config is now explicitly named `opencode.v1.json`, and the Dockerfile copies it to the standard in-image `opencode.json` path. I used a version-bound filename rather than adding `$comment`, because OpenCode 1.x uses strict top-level config parsing and a non-schema key can make the example invalid. Tests bind `OPENCODE_VERSION=1.18.9`, the V1 filename, and the Docker copy path.
- `HY3_BASE_URL` is validated once; `hy3_host()` now receives that validated value instead of re-running full validation.
- The SDK fallback guard is split into two explicit checks and now has negative tests proving unrelated `TypeError`s are not masked.
- Added a deny-by-default `.dockerignore` allowlist. Only `Dockerfile` and `opencode.v1.json` enter the build context, so `.env`, virtualenvs, tests, and caches cannot be sent to the builder.
- Added `--verbose-events` to all three drivers. Unknown valid JSONL event types are reported on stderr in that mode, while default output remains concise and `--raw` remains available.

Validation after this commit:
- 27 unit tests passed.
- Black, `compileall`, JSON parsing, `git diff --check`, and credential scan passed.
- OpenCode 1.18.9 resolved the renamed V1 config successfully with `debug config --pure`.
- All three CLI help surfaces expose `--verbose-events`.
- EN and ZH VitePress pages built successfully.

The previously documented live-cluster boundary is unchanged: CubeEgress and pause/resume still need an accessible CubeSandbox deployment for end-to-end evidence.

blues-kun:
The third review round is addressed in `b79fc5f`.

Session streaming fix:
- `resume_opencode.py` no longer depends on `result.stdout` for the primary session ID path. A `SessionIdCapture` observes parsed JSONL events as chunks stream; `result.stdout` is only merged as a compatibility fallback.
- The raw renderer and concise renderer both feed the same event observer.
- The command adapter tracks whether SDK callbacks actually fire. If a compatible SDK accepts callback kwargs but does not invoke them, collected stdout/stderr is replayed through the same handlers after completion. This preserves output and session capture, with the documented tradeoff that display becomes buffered.

Defensive follow-ups:
- The `envs` -> `env` fallback now accepts only the known `unexpected keyword argument 'envs'` TypeError forms. A different TypeError that merely mentions `envs` is re-raised.
- Renamed `hy3_model()` to `require_hy3_model()` so the Hy3-only contract is explicit.
- Tightened base URL validation to require the normalized path to be exactly `/v1`. Tests prove `/v1/` is accepted while `/v1/chat` and `/proxy/v1` are rejected.

New regression coverage includes:
- session ID split across multiple callback chunks with empty `result.stdout`;
- SDK ignores callbacks but returns complete stdout;
- raw streaming still captures the session ID;
- semantic `TypeError` mentioning `envs` is not masked.

Validation now reports 33 tests passed, plus Black, `compileall`, OpenCode 1.18.9 resolved-config parsing, JSON validation, diff checks, credential scan, and EN/ZH VitePress builds. The live Cube cluster validation boundary remains unchanged and explicitly documented.

blues-kun:
Final low-risk review cleanup is in `1dc8686`:

- Removed unreachable `TypeError` branches around `json.loads()` on string lines.
- Renamed the pause/resume handle to `active_sandbox` so reconnect ownership and final cleanup are explicit.
- Added coverage for the double-quoted `unexpected keyword argument "envs"` SDK error form.
- Pinned the expected `cubesandbox-base:2026.16` value in the template invariant test, so inherited-package assertions cannot pass against an unnoticed base-image change.
- Split the callback compatibility sentence for readability.
- Kept `exit_code=None` compatibility behavior intentionally, with an explanatory comment, matching the existing `pi-agent-integration` helper; compatible transport/command failures are expected to raise or return nonzero.

Final local result: 34 tests passed, Black/compileall/diff/credential checks passed, and the bilingual VitePress site built successfully.

fslongjin:
Since this PR does not provide a working example in a real CubeSandbox cluster, I’ll go ahead and close it. Thanks for your contribution!

Issue #645, opened 2026-06-25
【2026犀牛鸟issue】CubeSandbox 沙箱模板与示例生态


fslongjin:
【产品名称/功能模块】
CubeSandbox 沙箱模板与示例生态

【问题描述】
一个沙箱平台的易用性，很大程度上取决于其开箱即用的模板与示例是否丰富——用户能否快速找到一个贴合自身场景的模板（某种语言运行时、数据科学环境、浏览器自动化、代码解释器、Web 开发环境、特定 Agent 框架的运行底座等）直接拉起沙箱，决定了上手成本。CubeSandbox 已具备从 OCI 镜像构建模板、快照、恢复等能力，并已有少量示例（如 code-sandbox-quickstart、browser-sandbox、mini-rl-training、openai-agents-example 等），但覆盖的场景与语言栈仍然有限，尚未形成系统化的模板/示例生态。

本 issue 为开放性生态贡献任务，目标是持续丰富 CubeSandbox 的模板与示例库，覆盖更多语言、框架与典型场景，让不同背景的开发者都能找到可直接复用的起点。允许多人并行参与、各自认领一个模板或示例，以增量方式合入，不要求单次提交完整 PR。

【期望内容/功能改进】（每人可认领其中一个或多个方向，渐进式贡献）
1）贡献一个或多个开箱即用的沙箱模板，覆盖但不限于：常见语言运行时（Python/Node/Go/Rust/Java 等）、数据科学与 ML（Jupyter、PyTorch、pandas 等）、浏览器自动化、代码解释器、Web 开发环境、特定 Agent 框架的运行底座等；
2）每个模板/示例应包含：模板定义与构建方式、最小可运行调用流程、README 说明（适用场景、资源建议、已知限制），并能在本地部署环境跑通；
3）鼓励覆盖差异化场景：长时任务与快照断点续跑、有状态工作区、多容器协作、出口网络策略受限下的运行等，体现 CubeSandbox 的差异化能力；
4）模板应遵循平台既有模板构建与命名约定，与出口管控、鉴权、资源限制等机制对齐，避免安全策略缺口；
5）在 docs/guide/tutorials 或 examples 体系中登记，保持风格一致、便于检索；本方向聚焦通用语言栈/场景模板，与"Claude Code / CodeBuddy / OpenCode 集成"方向互补、不重复。

【参与说明与验收】
[...]

PR #1218, opened 2026-07-29
feat(examples): add OpenHands integration (agent server pre-warmed in template snapshot)

haoranfan1:
## Problem

[OpenHands](https://www.openhands.dev/) executes every agent action (bash, file edits, scripts) through its **agent server**, and its stock remote workspace, `DockerWorkspace`, boots a fresh agent-server container for each session on a shared host kernel. So every session pays the server's cold start — and although OpenHands sessions are long-lived by nature, stopping the container loses the entire session: conversation state, shells, and in-flight processes all live inside it.

Refs #244 (claimed via `/claim openhands`), #645.

## Solution

Bake the agent server into a CubeSandbox template and take the boot snapshot only after the server itself reports ready (`--probe 8000 --probe-path /ready`), then plug in through the SDK's own extension point: `CubeSandboxWorkspace` subclasses `RemoteWorkspace` exactly like the first-party workspaces (`DockerWorkspace`, `ApptainerWorkspace`, …), so an existing OpenHands program switches by replacing one constructor. Measured on a real cluster, this yields:

- create→healthy in **182 ms** (the same image cold-starts its server in **~6 s** under plain Docker) — no per-session environment build at all;
- whole-VM `pause()` / `resume()`: since the agent loop and conversation state live in the in-VM server, pausing the VM freezes the *entire session*, and re-attach by `sandbox_id` resumes it from another process, even days later;
- KVM MicroVM isolation, plus platform inbound/egress controls (`private_traffic` per-sandbox tokens, egress allowlists, and a documented credential-injection path that keeps the raw LLM key out of the VM).

Layout follows the merged pi-agent integration (#701): a bilingual guide plus a runnable example.

## What Changed

- **Runnable example** at `examples/openhands-integration/`:
  - `Dockerfile` — OpenHands agent server (`openhands-agent-server==1.38.0`, standalone Python 3.12 via uv) on `cubesandbox-base`; envd stays on `:49983`; the server starts on `:8000` as the image CMD under the `user` account. `SSL_CERT_FILE` points at the system trust store so credential-injection egress rules work against the pre-warmed server.
  - `cubesandbox_workspace.py` — `CubeSandboxWorkspace(RemoteWorkspace)`, the same extension point upstream's `DockerWorkspace` uses: creates/attaches sandboxes via the E2B-compatible SDK, waits on the server's strict `/ready`, supports `private_traffic` (per-sandbox traffic token on every workspace request), whole-VM `pause()`/`resume()`, and cross-process re-attach by `sandbox_id` (auto-resume).
  - `smoke_test.py` — no-LLM end-to-end proof: hot-start latency, `/server_info`, bash + file round-trips.
  - `main.py` — full agent coding task; the result is verified **from inside the sandbox**, independent of the agent's own claims.
  - Bilingual `README.md` / `README_zh.md`, `.env.example`, `requirements.txt`.
- **Integration guide** in both languages, on the `_template.md` frontmatter, linked from both index tables:
  - `docs/guide/integrations/openhands.md`
  - `docs/zh/guide/integrations/openhands.md`

## Acceptance (#244)

1. One standalone guide file per language, linked from the integrations index ✅
2. Runnable example with template definition + reproducible flow + bilingual README ✅
3. Aligned with platform security mechanics (egress allowlist, traffic tokens, `SESSION_API_KEY`, credential-injection pointer) ✅
4. Caveats/troubleshooting aligned with existing template/probe/auth mechanics ✅

## Test Plan

- [x] `python -m py_compile` on all three scripts; ruff 0.16.0 `check` + `format --check` clean
- [x] `cd docs && npm run docs:build` (VitePress 1.6.4) — passes in 10.8 s
- [x] `git diff --check` clean
- [x] `docker build --no-cache` — 37.8 s, image 746 MB
- [x] Plain-Docker sanity: `GET :8000/ready` → **200** (~6 s cold start), `GET :49983/health` → **204**

## Live cluster run

All results below were produced against a **real single-node CubeSandbox v0.6.0 cluster** (arm64 host, self-built base image, deployment proxy on `:10080`), at this PR's head. Template registered with `--probe 8000 --probe-path /ready` (snapshot taken only after the agent server reports ready); the template build reached `READY` in ~16 s (one-time cost — per-session sandbox creation is the 182 ms below). Terminal and Dashboard screenshots below carry the full run output.

- [x] `smoke_test.py` — sandbox healthy in **182 ms**, all 4 checks pass:

smoke_test.py — 4/4 PASS, healthy in 182 ms

- [x] `main.py` — the agent created and ran `/workspace/fib.py`; verification executed the script **inside the MicroVM** and checked the Fibonacci recurrence over all 20 lines (exit 0):

```text
--- verification from inside the sandbox ---
# Print the first 20 Fibonacci numbers, one per line
a, b = 0, 1
for i in range(20):
    print(a)
    a, b = b, a + b

---
0
1
1
2
3
5
8
13
21
34
55
89
144
233
377
610
987
1597
2584
4181

PASS
```

- [x] Dashboard (v0.6.0): template `READY` card:

Dashboard — template READY card

Refs #244, #645

### DIFF
--- a/docs/guide/integrations/index.md
+++ b/docs/guide/integrations/index.md
@@ -48,3 +48,4 @@ lang: en-US
 | Title | Author | Date | Tags |
 | --- | --- | --- | --- |
 | [Pi Agent Integration Guide](./pi-agent.md) | chaojixinren | 2026-07-01 | integration, pi-agent, coding-agent, agent |
+| [OpenHands Integration Guide](./openhands.md) | Fan-hr | 2026-07-28 | integration, openhands, coding-agent, agent |
--- a/docs/guide/integrations/openhands.md
+++ b/docs/guide/integrations/openhands.md
@@ -1,0 +1,144 @@
+---
+title: OpenHands Integration Guide
+author: Fan-hr
+date: 2026-07-28
+tags:
+  - integration
+  - openhands
+  - coding-agent
+  - agent
+lang: en-US
+---
+
+# OpenHands Integration Guide
+
+[中文文档](../../zh/guide/integrations/openhands.md)
+
+## Integration Target and Version
+
+[OpenHands](https://www.openhands.dev/) is an autonomous software-development
+agent platform. Its [Agent SDK](https://github.com/OpenHands/software-agent-sdk)
+executes agent actions (bash, file edits, scripts) through an **agent server**
+that normally runs in a local Docker container.
+
+This guide wires the agent server into a Cube Sandbox MicroVM instead: the
+server is pre-installed in a Cube template and frozen *live* into the
+template's boot snapshot, so every sandbox hot-starts with a running agent
+server and full hardware isolation.
+
+Tested versions: `openhands-sdk` / `openhands-tools` /
+`openhands-agent-server` **1.38.0**, Cube Sandbox **v0.6.0**.
+
+## Prerequisites
+
+- Cube Sandbox deployment: any working deployment (single-node is fine) with
+  `cubemastercli` and the E2B-compatible API reachable.
+- SDK or CLI dependencies: Docker (image build), [`uv`](https://docs.astral.sh/uv/)
+  or a current pip >= 26 (host scripts — older pips such as Ubuntu 24.04's
+  stock 24.0 fail on an upstream `lmnr`/`opentelemetry` conflict).
+- Required environment variables: `E2B_API_URL`, `E2B_API_KEY`,
+  `CUBE_TEMPLATE_ID`; plus `LLM_MODEL`, `LLM_API_KEY`, optional
+  `LLM_BASE_URL` (any OpenAI-compatible endpoint) for the full agent demo.
+  To just verify the integration, `smoke_test.py` and `pause_resume.py`
+  need **no LLM configuration at all**.
+
+## Integration Steps
+
+1. **Build the template image** —
+   [`examples/openhands-integration/Dockerfile`](https://github.com/tencentcloud/CubeSandbox/blob/master/examples/openhands-integration/Dockerfile)
+   extends `cubesandbox-base` (envd on `:49983` preserved), installs a
+   standalone Python 3.12 via uv, pins `openhands-agent-server`, and starts it
+   on `:8000` as the image CMD under the unprivileged `user` account.
+
+2. **Register it as a template**, probing the agent server itself so the boot
+   snapshot is taken only after the server is ready:
+
+   ```bash
+   cubemastercli tpl create-from-image \
+     --image <registry>/openhands-sandbox:latest \
+     --writable-layer-size 2G \
+     --expose-port 8000 --expose-port 49983 \
+     --probe 8000 --probe-path /ready
+   ```
+
+3. **Connect the OpenHands SDK** through `CubeSandboxWorkspace`, a
+   `RemoteWorkspace` subclass (the same extension point the SDK's own
+   `DockerWorkspace` uses) that creates the sandbox via the E2B-compatible
+   SDK and points the workspace at the proxied agent-server URL.
+
+4. **Run demos** from `examples/openhands-integration/`: `smoke_test.py`
+   (no LLM; hot-start latency, bash/file round-trips), `pause_resume.py`
+   (no LLM; whole-VM freeze/thaw under a live server), `main.py` (full agent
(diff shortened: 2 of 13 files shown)


haoranfan1:
Thanks for the thorough automated passes — all three rounds are addressed. Point-by-point:

### Round 1 → 44bd5df

1. **`typing.Self` (Medium)** — Fixed: the import now sits behind `if TYPE_CHECKING:` with `from __future__ import annotations`, so the module imports on any interpreter (the hint is never evaluated at runtime) while keeping the `-> Self` annotation that ruff itself expects (PYI034). Worth noting the example's effective floor is **Python ≥ 3.12 regardless**: `openhands-sdk` declares `requires-python >= 3.12`, so on 3.9/3.10 the dependency install fails before any import runs. Now documented in `requirements.txt` (also covers finding 6).
2. **`resume()` host refresh (Low)** — Fixed as suggested, extracted into a `_public_url()` helper shared with the create path — plus one step further: `reset_client()` after the refresh, because the parent's cached httpx client keeps its `base_url` bound to the previous host.
3. **`except BaseException` (Low)** — Kept deliberately; justifying comments added at both sites. Narrowing to `Exception` would reintroduce the exact leak these blocks exist to prevent: Ctrl-C during the readiness wait (workspace) or the frozen-gap sleep (demo) is the likeliest interrupt, and it too must reap the sandbox before propagating. Both handlers re-raise unconditionally.
4. **Dual HTTP stacks (Info)** — Comment added to `_wait_for_ready`: it runs during `model_post_init`, before `super().model_post_init()` has set up the client; `_headers` (session key + traffic token) is honored on both stacks.
5. **`ADD` for the uv installer (Info)** — Switched to `RUN curl -fsSL … | sh`, matching the pi-agent image.
6. **Python floor (Info)** — Documented in `requirements.txt` (see 1).

### Round 2 → 6b60a80

- **Traffic token typing** — `_saved_traffic_token` is now `SecretStr | None`, unwrapped only at the header and persistence boundaries; an accidental `repr()`/debugger view shows `SecretStr('**********')`.
- **`__exit__` return** — Comment now states explicitly that returning `None` matches the parent contract (which never returns `True`), so no exception suppression is being discarded.

### Round 3 → 1b87cc8

- **`python-dotenv` lower bound** — Pinned `>=1.0.0`, matching the pi-agent example.
- **Sub-2s `health_check_timeout`** — Each readiness poll is now capped at the remaining overall budget (`max(0.1, min(2.0, deadline - now))`), so small timeouts are honored exactly.
- **`http://` scheme** — `_public_url` docstring now states it is intentional: cube-proxy serves plain HTTP; TLS (if any) terminates at an upstream reverse proxy.

### Round 4 — no code changes

- **`UnboundLocalError` in `pause_resume.py`** — not reachable: `sandbox_id = workspace.sandbox_id` executes *before* the `try:` statement, and the `except BaseException` handler only guards exceptions raised inside that block — by the time the handler can run, `sandbox_id` is always bound. A constructor failure propagates before the `try` begins, and the workspace's own `model_post_init` already reaps the sandbox on that path.
- **Org-URL casing / one long Dockerfile comment line / the `LLM_MODEL` placeholder** — cosmetic; happy to normalize all three in one pass if a human reviewer prefers.

### Re-verification (same v0.6.0 single-node cluster, after each round)

- `smoke_test.py` 4/4
- `pause_resume.py` PASS (ticks 5 → 5 → 8 across the 8 s frozen gap)
- Direct `pause()` → `resume()` → `execute_command()` round-trip exercising the refreshed-host path
- `private_traffic=True` create → token captured (masked repr) → readiness probe passed the private-traffic gate with the token header → exec → kill
- `docker build` with the curl-based installer succeeds; ruff `check` / `format --check` clean throughout

fslongjin:
Thanks for your contribution! Before merging, could you please rebase it onto the latest origin/master? Also, I don't think this example needs to include a pause/resume demo.

haoranfan1:
@fslongjin Thanks for the review. Both points are done:

- Rebased onto master at 1e3f821. The only conflicts were the two integrations index tables.
- Dropped the pause/resume demo: pause_resume.py is gone along with its walkthrough in both READMEs and both guides. The pause()/resume() methods still exist on CubeSandboxWorkspace because the guide's snippet uses them. Happy to remove those too if you'd prefer.
- Switched the guide author field to my GitHub ID.

PR #1222, opened 2026-07-30
fix(cubelet): skip creating sandboxes in DeadGC to fix pause failure

lkml-likexu:
A pause right after sandbox creation intermittently fails with:

```
__________ test_pause_and_connect_resume_preserves_files[cubesandbox] __________
cases/lifecycle/test_pause_resume.py:30: in test_pause_and_connect_resume_preserves_files
    sdk_sandbox.pause(timeout=sdk_e2e_config.default_timeout)
adapters/tracing_adapter.py:108: in pause
    return self._trace.capture(
framework/trace.py:163: in capture
    result = call()
             ^^^^^^
adapters/tracing_adapter.py:115: in <lambda>
    lambda: self._wrapped.pause(timeout=timeout),
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
adapters/cubesandbox_adapter.py:122: in pause
    self._sandbox.pause(timeout=timeout)
../../../sdk/python/cubesandbox/sandbox.py:485: in pause
    _check_response(resp)
../../../sdk/python/cubesandbox/sandbox.py:83: in _check_response
    raise ApiError(msg, code)
E   cubesandbox._exceptions.ApiError: CubeMaster returned error code 130490: sandbox is not running
```

A sandbox is saved into the store as CONTAINER_CREATED before its containerd task is bound and StartedAt is set. If the DeadGC scan runs during that window it probes the not-yet-bound task, wrongly marks the sandbox Unknown, and since that state outranks StartedAt the sandbox reads as terminated forever even after it is RUNNING. A follow-up pause then fails with TaskStateInvalid (130490) "sandbox is not running".

This is the same shim-busy race DeadGC already guards against for the PAUSING, PAUSED, and RollingBack states, now extended to the create window. The skip is bounded so a genuinely stuck, never-bound entry is still reaped.

Add regression tests for both the skip and its bound.

### DIFF
--- a/Cubelet/services/cubebox/rollback_test.go
+++ b/Cubelet/services/cubebox/rollback_test.go
@@ -190,3 +190,44 @@ func TestScanDeadContainerSkipsRollingBack(t *testing.T) {
 	assert.Equal(t, staleFinishedAt, got.FinishedAt, "scanDeadContainer must not touch a rolling-back cubebox")
 	assert.True(t, got.Unknown, "Unknown must be left as-is for the rollback path to fix")
 }
+
+// TestScanDeadContainerSkipsCreatingSandbox reproduces the create/DeadGC race:
+// while a sandbox sits in the CONTAINER_CREATED window (store entry saved before
+// runContainer binds the containerd task), DeadGC must NOT call RecoverContainer,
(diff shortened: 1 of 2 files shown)


ls-ggg:
LGTM

PR #1224, opened 2026-07-30
test(volume): cover per-sandbox access-mode serialization

zyl1121:
## Motivation

Follow up on review feedback from #1098 by covering the Python SDK serialization boundary for using the same Volume with different per-sandbox access modes.

## What Changed

* Cover independent read-write and read-only serialization for the same Volume across separate sandbox requests.
* Keep the existing restriction that a sandbox cannot attach the same Volume more than once.

## Validation

* Python SDK Volume tests passed.
* The COS-backed SDK compatibility case passed with concurrent read-write and read-only sandbox attachments to the same Volume.


ls-ggg:
LGTM

PR #1238, opened 2026-07-31
docs: add OpenCode integration via a plugin hook that redirects bash

Tantanovo:
## Related Issue

Refs #644

## Problem

OpenCode runs on the developer's own machine. Every `bash` command it issues executes there, with the developer's own privileges, against the developer's real filesystem.

MCP-based and SDK-based sandbox integrations only help when the model decides to call a sandbox tool. A plain `bash` call still lands on the host, and nothing reports that it happened. The failure mode is silent, which is the worst kind for a security boundary.

## Approach

This integration uses OpenCode's `tool.execute.before` hook to rewrite every `bash` invocation so the command runs inside a CubeSandbox MicroVM instead of on the host.

```
OpenCode bash tool
   -> tool.execute.before hook   (plugin/cubesandbox-bash.js)
   -> rewritten command          (python3 exec_backend.py --command '<original>')
   -> CubeSandbox E2B API        (http://127.0.0.1:3000)
   -> MicroVM: own kernel, own filesystem, own network namespace
```

Two properties follow from doing it at the hook layer:

- **The model changes nothing.** No new tool, no prompt edit, no workflow change. The agent still writes `bash` calls; they simply stop executing on the host.
- **`read` / `write` / `edit` are untouched.** The agent keeps editing the real project on the host, while its shell commands run in a throwaway environment. Running the whole agent inside a sandbox was considered and rejected, because it costs the agent access to the working tree, git credentials, and editor state.

## Design properties

| Property | How |
| --- | --- |
| Fail closed | If a command cannot be redirected safely, the hook throws and OpenCode blocks the call. It never silently falls back to the host. |
| Injection safe | The original command travels as a single `argv` element. Verified against real `bash` `argv` parsing, not substring matching. |
| Stateful | Working directory and exported environment are carried across calls within one session. |
| Concurrency safe | Calls within a session are serialised by a lock file that reclaims stale locks. |
| Escape hatch | An allowlist lets specific commands (for example `git`) stay on the host when they need host credentials. |

## Verification on a real deployment

Deployed and tested against a real CubeSandbox instance, not a mock.

| Item | Value |
| --- | --- |
| CubeSandbox | v0.6.0 (`8721dd15`, built 2026-07-24) |
| Host | Ubuntu 24.04.4 LTS on WSL2, kernel `6.18.33.2-microsoft-standard-WSL2`, glibc 2.39 |
| `/data/cubelet` | loopback XFS, `reflink=1` |
| Control plane | 14 `cube-sandbox-*` systemd units active |
| E2B API | `GET /sandboxes` -> `200` on `127.0.0.1:3000` |
| Template | `tpl-bce0a51ec8cc43e890a962ca`, `READY` |
| Sandbox create | 1.09 s; full create/exec/close cycle 2.39 s |

The most useful single data point:

```
guest kernel : 6.6.1199-0009-03_2.0.1
host  kernel : 6.18.33.2-microsoft-standard-WSL2
```

The guest reports a different kernel from the host. A container would report the host kernel, because it shares it. This is what distinguishes real virtualisation from namespace isolation, and it is the property the whole integration depends on.

Smoke run output:

```
[1] sandbox created in 1.09s
    sandbox_id: da3a5dd6df5b483bb898f49c4f7d1812
[2] run_code stdout: ['Hello from Cube Sandbox, safely isolated!\n']
[3] guest info:
    guest kernel : 6.6.1199-0009-03_2.0.1
    guest machine: x86_64
    guest cwd    : /workspace
[4] compute check (sum 1..100): 5050
[5] sandbox closed. total 2.39s
RESULT: PASS
```

## Tests

```bash
node examples/opencode-plugin-sandbox/tests/test_plugin.mjs
```

19 assertions, Node standard library only. No `npm install`, no CubeSandbox deployment, no network. A reviewer can run it immediately.

Coverage includes: non-`bash` tools are left alone, allowlisted commands pass through, disabling the plugin is respected, the rewrite is idempotent, blank commands are ignored, a missing backend causes a throw rather than a host fallback, and a quote-escape payload stays inside one `argv` element when parsed with POSIX single-quote semantics.

## Two defects found by running the code

Both are recorded in `AI_USAGE_REPORT.md`. Neither is visible by reading the source.

1. **ESM / CommonJS mix.** The first draft used `require()` alongside `export`. OpenCode loads plugins as ES modules, so this throws `ReferenceError: require is not defined` at load time, and `__dirname` does not exist either. Fixed with `fileURLToPath(import.meta.url)`.

2. **Configuration read at the wrong time.** `CUBE_OPENCODE_PYTHON` was resolved once at module top level. The plugin lives for the whole session, so a venv path exported after the editor started was silently ignored. Now read per call.

There was also a third correction worth naming, because it was a mistake in the test rather than the code: the original injection assertion checked that the payload text was absent after stripping escape sequences. That can never pass — the payload is data, so of course it is still there. The assertion now parses the rewritten command with POSIX single-quote semantics and asserts the original command occupies exactly one `argv` element, which is the property that actually matters.

## Not verified

Stated explicitly so reviewers do not have to guess:

- Tested only against a WSL2 deployment. Not tested on bare metal or a cloud VM with the PVM kernel.
- Tested with the `sandbox-code` template. Other templates may expose a different shell environment.
- Single-user, single-session testing. No multi-user or high-concurrency load testing.
- The stale-lock reclaim path was exercised by unit test, not by killing a live process mid-command.

## Files

```
docs/guide/integrations/opencode-plugin.md            English integration guide
docs/zh/guide/integrations/opencode-plugin.md         Chinese integration guide
docs/guide/integrations/index.md                      one index row
docs/zh/guide/integrations/index.md                   one index row

examples/opencode-plugin-sandbox/
  plugin/cubesandbox-bash.js    tool.execute.before interceptor
  plugin/install.sh             idempotent install / uninstall / status
  exec_backend.py               sandbox execution backend, session-scoped MicroVM reuse
  tests/test_plugin.mjs         19 offline assertions
  README.md / README_zh.md      bilingual usage docs
  AI_USAGE_REPORT.md            AI usage self-report
  .env.example / .gitignore
```

13 files, +2634 lines, 0 deletions. Both language variants are included, and each index gains exactly one row.

The example directory is named `opencode-plugin-sandbox` rather than `opencode-integration` so it does not collide with existing OpenCode contributions.

## Checklist

- [x] Both `docs/guide/` and `docs/zh/guide/` updated
- [x] Front matter includes `title`, `author`, `date`, `tags`, `lang`
- [x] Index row added in both languages
- [x] Runnable example with README
- [x] AI usage self-report included
- [x] Verified against a real CubeSandbox deployment, with logs
- [x] Offline test suite passes (19/19)
- [x] All files LF, `git diff --check` clean, 0 deletions


Tantanovo:
Thanks for the review — all five findings were real. Each was reproduced against the code before being changed, and every fix is in `09b8c86`.

## 1. High — state preservation did not work

Confirmed. `build_wrapper` ran the command through `bash -c`, a child shell, so `cd` / `export` mutated only that child and were gone before the state block ran.

Reproduced with two minimal scripts:

```
bash -c variant:   command printed /tmp/statetest/demo
                   wrapper shell still at /tmp/statetest      <- state lost

eval variant:      command printed /tmp/statetest/demo
                   wrapper shell now at /tmp/statetest/demo   <- state kept
```

Fixed by running the command with `eval` in the current shell. Re-verified end to end — after `cd /workspace/demo && export TOKEN=abc123`, the state block now reports both the new cwd and `TOKEN=abc123`.

Your observation about *why* it was missed is the part worth keeping: the offline suite only exercises the JavaScript hook, and `build_wrapper` was never executed. This is exactly the defect that the "end-to-end not verified" caveat was hiding. That connection is now written into the AI usage report rather than left implicit.

## 2. Medium (security) — idempotence guard was a silent host escape

Confirmed, and this was the worst of the five: a guard intended to make the plugin safe was the one thing that let commands bypass it.

Replaced the substring test with a structural check. `isOwnRewrite` parses the command with POSIX single-quote semantics and requires the exact six-token shape the plugin emits, with the backend path matching our resolved absolute path. Anything else — including a malformed or partially-quoted command — is treated as a user command, which is the safe direction to fail in.

Five escape cases from your report are now assertions:

```
ls exec_backend.py
cat exec_backend.py
echo x # exec_backend.py
touch /tmp/should-not-happen; echo exec_backend.py
grep -n session exec_backend.py
```

Plus one in the other direction: feed a genuine rewrite back through the hook and assert it is recognised and not nested (exactly one `--command`).

## 3. Medium-low — lock released without ownership

Confirmed. `__exit__` unlinked unconditionally, so the timeout path (`self.fd is None`, live peer still holding) removed someone else's lock. Now returns early unless the lock was actually acquired. The stale-lock reclaim continues to handle crashed holders, as you noted.

## 4. Medium-low — copy fallback installed a broken plugin

Confirmed, including the part that made it worse: the documented "reinstall" remedy reproduced the same broken state. The fallback now copies `exec_backend.py` to the matching relative location, and if that copy fails it removes the plugin again rather than leaving a setup that blocks all bash. Uninstall removes the copied backend, guarded by a content check so a user's own file of that name is never deleted.

Verified by running install then uninstall against a scratch directory: the backend lands at `<parent-of-plugin-dir>/exec_backend.py` and is recognised on removal.

## 5. Low — README contradicted itself

Fixed. The summary table and the diagram now say per call, matching the code and the known-limitations section.

## Additional notes

- **`resolveSessionId` fallback** — you are right that the comment understated it. Losing the session key does not merely cost reuse granularity; all sessions collapse onto one state file and lock, so cwd and environment bleed between concurrent sessions. Reworded to say that, while keeping the accurate part: it is still not a host escape, since commands run in a MicroVM either way.
- **Python-keyword env names** — the `isidentifier()` filter is intentional (the value is interpolated into an `export`), but it was undocumented. Now noted.
- **State-preservation examples in the docs** — these were the unverified claims you flagged. They now hold, and were checked rather than assumed.

Tests are 21/21 (was 19/19); the counts were updated across both READMEs and both integration guides. `AI_USAGE_REPORT.md` gains section 5.7 covering these findings.

One note on process, since this review was useful: the offline suite's blind spot was the Python backend. Nothing in the 19 original assertions could have caught finding 1, and the report said as much without drawing the conclusion. Worth stating plainly rather than glossing over.

Tantanovo:
## Verification screenshots

Screenshots complementing the logs in the PR description. All captured on the deployment used for this work (CubeSandbox v0.6.0, Ubuntu 24.04.4 on WSL2).

### 1. Host prerequisites: KVM device and XFS with reflink

`/dev/kvm` present, `/data/cubelet` on XFS with `reflink=1`. The reflink capability is what lets cubelet clone writable layers in O(1); ext4 does not implement `FICLONE`, so the installer rejects it.


Image

### 2. Template built and READY

`tpl-bce0a51ec8cc43e890a962ca`, status `READY`, built from `sandbox-code:latest` at digest `sha256:467494c3...`.

Image

### 3. Sandbox executed real code, in a different kernel from the host

This is the screenshot that matters most.

```
guest kernel : 6.6.1199-0009-03_2.0.1
host  kernel : 6.18.33.2-microsoft-standard-WSL2
```

The guest reports a different kernel from the host. A container would report the host kernel because it shares it. This distinguishes real virtualisation from namespace isolation, and it is the property the whole integration depends on: if the guest shared the host kernel, redirecting `bash` there would not be a meaningful security boundary.

Sandbox created in 1.09 s, full create/exec/close cycle 2.39 s.

Image

### 4. Control plane: all systemd units active

14 `cube-sandbox-*` units active. `cube-api` on `0.0.0.0:3000`, `cubemaster` on `8089`, `cubelet` gRPC on `9999`, `network-agent` on `127.0.0.1:19090`.

Image

### 5. Web UI

Image

### Offline test suite

Reproducible without any deployment:

```bash
node examples/opencode-plugin-sandbox/tests/test_plugin.mjs
```

```
==================================================================
passed: 21   failed: 0
==================================================================
```

Node standard library only. No `npm install`, no CubeSandbox instance, no network.

### Raw logs

Full install and run logs are on the test machine at `~/cube-work/logs/`, with collected evidence at `~/cube-work/evidence/` (`01-env.txt`, `02-services.txt`, `03-templates.txt`, `04-sandbox-run.txt`). Happy to paste any of them inline if that is easier for review.

Tantanovo:
Second round addressed in `f68914f`. All four were real, and one of them shows my previous fix was incomplete — worth stating plainly.

## 1. Medium — `exit` loses exit code and session state

You are right, and this is a defect I introduced. My previous fix swapped `bash -c` for `eval`, which solved the reported problem (cd/export becoming visible) but created this one: `eval` runs in the wrapper's own shell, so a command ending in `exit` kills that shell before the state block and exit-code line run.

Reproduced:

```
eval variant, command `echo before; exit 3`:
  stdout : before
  state  : (missing)
  rc     : 0        <- wrong; the command exited 3
```

Fixed by running `eval` inside a subshell carrying an `EXIT` trap. That satisfies both constraints at once: `eval` still executes where its mutations are observable to the trap, the trap fires however the subshell ends, and an `exit` inside only terminates the subshell.

Verified across four shapes:

```
plain command      rc=0  state captured  cwd=demo  TOKEN=v1
ends with exit 3   rc=3  state captured  cwd=demo  TOKEN=v2
failing command    rc=1  state captured
exec echo ...      no sentinel
```

**`exec` I could not cover, and I am not claiming otherwise.** It replaces the process image and discards traps with it; no shell-level wrapper can observe that. The backend keeps the previous session state when no sentinel arrives, so state is stale rather than wrong, and it is now listed as a known limitation in both READMEs and both guides.

On why the first fix missed it: I verified against the exact scenario in the report — `cd` then `pwd` — and it passed. I did not check the adjacent shapes of the same operation. A fix to control flow should be tested against the ways control flow can end, not only the reported case. That is recorded in §5.9 rather than glossed over.

## 2. Medium (security) — default `git` passthrough is an escape hatch

Confirmed against real git before changing anything:

```console
$ git -c alias.probe='!echo PWNED_VIA_GIT_ALIAS' probe
PWNED_VIA_GIT_ALIAS
```

My original reasoning — the sandbox has its own filesystem, so `git commit` there commits the wrong repository — is sound, but I applied it to the wrong threat model. The model's commands are the untrusted input this plugin exists to contain, so a default list containing `git` re-opens the host path the plugin closes.

**The default list is now empty.** Opting in is explicit:

```bash
export CUBE_OPENCODE_PASSTHROUGH=git,gh
```

Both the code comment and the docs now state that anything on the list must be treated as fully trusted with host privileges, and the git-alias mechanism is spelled out so the choice is informed rather than incidental. Tests cover both directions: nothing passes through by default (including the alias form), and opt-in works.

## 3. Medium — Chinese README diagram still said "per session"

Confirmed, and this is a plain process failure on my part: the previous round fixed the English diagram and I checked only the file the review named. A bilingual repository needs both files changed in the same edit. Both diagrams now say per call and agree with their own known-limitations sections.

## 4. Low–Medium — fail-closed skipped for non-string commands / missing args

Confirmed. Both now throw rather than return:

- missing `output.args` → `the bash tool call carried no arguments`
- non-string `command` → `expected a string command, received <type>`

A blank string still passes through, since nothing executes either way and failing there would be noise rather than safety.

## 5. Low — session-key fallback wording

Fixed in all four documents to match the code comment: losing every known key collapses all sessions onto one shared state file and lock, so cwd and exported environment bleed between concurrent sessions. Not a host escape, but more than a loss of reuse granularity.

## 6. Info — "19 assertions" stale in the PR body

Correct, and now stale twice over: the suite is at **24** after this round. The guides, both READMEs and the AI usage report are updated. I cannot edit the original PR description without losing the review thread's anchor, so treating this comment as the current count.

## On the minor observations

- **Sentinel collision** — agreed, and cheap. Not changed in this round because the sentinel is emitted on its own line and a line-anchored split is a behaviour change to `split_output` that I would rather not make without a test for the collision case. Noting it as a real hardening rather than dismissing it.
- **"Transparent" may overpromise** — fair. The model's bash now runs against a different filesystem, so `ls`, `grep` and project-local commands behave differently. The docs cover the filesystem difference, but "transparent" invites a reader to skip that. Worth rewording; not done here.
- **`--reset` races the lock** — real, and I have not fixed it. It runs outside `SessionLock`, so a reset concurrent with a running command can race the state file. Lower impact than the above (the failure is a lost state update, not a host escape) but it is a genuine gap.

I have left these three as stated rather than silently narrowing the claims. Happy to take them in a follow-up if you would prefer them addressed before merge.

Thanks for the second pass — the `exit` case in particular is one I would not have found by reading.

Tantanovo:
Third round addressed in `f30ffb5`. Thanks for the verdict — and finding #1 is a fair hit, because it is the same mistake as last round's missed Chinese diagram, one round later and wider.

## 1. Passthrough-default contradiction (Medium, security-adjacent)

You are right, and the `.env.example` case is the one that matters most: that is the file a user copies before reading anything else, and it still said `The default is "git,gh,opencode"`.

Fixed in all five spots. `.env.example` now leads with `EMPTY BY DEFAULT`, spells out the git-alias mechanism, and ships the line commented out as `# CUBE_OPENCODE_PASSTHROUGH=git,gh` so opting in is a deliberate edit.

The process lesson, since this is the second round it has come up: last time I concluded "change both language files". That was too narrow. **After inverting a default, grep the whole example for the old value** — not just the files the review names. Doing that here turned up `.env.example`, which no review round had mentioned. Recorded in §5.10.

## 2. Backend module docstring still claimed one MicroVM per session (Medium)

Confirmed, and it survived in the file a reviewer opens first, which is worse than the README version.

Reworded, and I took your framing about what actually persists, because it is a more useful statement than the one I had:

> Environment and cwd carry over because they are replayed as strings; filesystem changes do not. A directory created by one call is gone by the next.

Also added to the worked example in both READMEs, since your finding #5 is the same problem showing up in the docs: `cd /workspace/demo` then `pwd` only reproduces because `/workspace/demo` exists in the template rootfs. `mkdir /tmp/x` in one call followed by `cd /tmp/x` in the next lands back in `/workspace`.

## 3. Lock serialisation is best-effort, not a guarantee (Low–Medium)

Confirmed. `SessionLock` proceeds unlocked after 90 s while the default command timeout is 120 s, so concurrent same-session calls can routinely overlap with last-writer-wins on the state file. And `stale_after=300` can displace a holder running longer than that once `CUBE_OPENCODE_TIMEOUT` is raised.

The code comment was honest; the property table said "serialised". That gap is on me. Both READMEs now say **Concurrency safe (best effort)** and state what happens on timeout, and the module docstring covers both the timeout and the stale-reclaim window.

I chose to document rather than fail closed on the wait: refusing a command because a lock is busy seems worse for the agent than a lost state update. Open to the other choice if you disagree.

## 4. `exec` loses the exit code (Low)

Confirmed, and this one was a documentation claim the code did not support — I wrote "returns the process exit code unchanged" last round while `split_output` returned 0. That is worse than an undocumented gap, because it invites trust.

Fixed properly rather than by weakening the docstring, since `p.returncode` is right there as you noted:

```python
if '__CUBE_RC__=' not in p.stdout:
    sys.stdout.write('\n__CUBE_RC__=%d\n' % p.returncode)
```

`split_output` now reads the exit code from the body as well as from after the state block, and strips the line in both paths so it never surfaces as command output. Verified across four shapes:

```
[PASS] state block + rc          rc=3  cwd captured   rc line not leaked
[PASS] exec, rc in body          rc=1  no state       rc line not leaked
[PASS] exec success              rc=0  no state       rc line not leaked
[PASS] no rc at all              rc=0  no state       rc line not leaked
```

So `exec false` is now reported as a failure. The state loss on `exec` remains and stays documented.

## 6. Idempotence guard can be spoofed (Low, hardening)

Confirmed, and the reasoning is the part worth keeping: `read` is deliberately unsandboxed, so the model can read the plugin and reproduce the shape from it. Pinning only the backend path and flags left that open.

Every fixed position is now pinned, including `tokens[0]` against the resolved interpreter and `tokens[3]` against `resolveSessionId(input)` for this call. `resolveSessionId` moved above the guard for that. A new assertion builds a well-formed rewrite under one session and replays it under another, asserting it gets redirected rather than trusted.

## Nits

- **Stale PR metadata** — now stale three times over: 25 assertions, +3331 lines. I cannot edit the description without disturbing the review anchor, so treating these comments as the current numbers.
- **Node version** — good catch. Both READMEs now state Node 22 or newer, since the plugin is ESM in a `.js` file and relies on Node's module-syntax detection.

## Still not fixed

Same three as last round, unchanged and stated deliberately:

1. **Sentinel collision** — a command printing the sentinel would corrupt parsing. Line-anchored splitting is the right fix; I have not made it because it is a behaviour change to `split_output` I would rather land with a test for the collision case than slip into a docs round.
2. **"Transparent" wording** — the model's bash now runs against a different filesystem, which the docs cover but the word invites a reader to skip.
3. **`--reset` races the lock** — runs outside `SessionLock`, so a reset concurrent with a running command can race the state file.

Happy to take any of these if you would prefer them before merge.

Tantanovo:
Fourth round addressed in `8928662`. Thanks — finding 1 is the most useful one so far, because unlike the previous rounds it is not stale prose, it is a mode that never worked in any version.

## 1. `--reset` unusable as documented — fixed

Reproduced exactly as the READMEs describe it:

```console
$ python3 exec_backend.py --session testsess --reset
usage: exec_backend.py [-h] [--session SESSION] --command COMMAND
                       [--timeout TIMEOUT] [--reset]
exec_backend.py: error: the following arguments are required: --command
```

The `--reset` branch itself was correct. `--command` was declared
`required=True`, so argparse rejected the invocation before that branch could be
reached.

`--command` is no longer required; the two modes are validated against each other
after parsing:

```console
$ python3 exec_backend.py --session testsess --reset
[cubesandbox-exec] session state cleared: testsess

$ python3 exec_backend.py --session x
exec_backend.py: error: --command is required unless --reset is given
```

**Why this survived four rounds of review of the same file.** Nothing in the
offline suite touches `main()` — it exercises the JavaScript hook. And during
development I cleared session state by deleting the JSON file directly, never by
running the documented command. So a code path that only a user would take was
never taken, by either the tests or by me. The previous three rounds were stale
prose describing code that had moved on; this one is code that was never correct,
in a mode both language variants document.

## 2. Stale docstring in the plugin header — fixed

Round three corrected this in the backend docstring and both READMEs and missed
the plugin's own header. Now corrected, and it also states the consequence rather
than only the mechanism:

> Session-scoped state, not a session-scoped VM. [...] Consecutive commands
> therefore observe each other's `cd` and `export` [...] but filesystem changes
> do not survive: a file written by one call is gone by the next.

## 3. Test that could not fail — fixed

You are right that `path.basename(BACKEND) === "exec_backend.py"` is
unconditionally true for a constant-derived path. That test would have passed on
a checkout with the backend deleted.

It now checks the filesystem, which is what the plugin does at call time:

```js
assert.ok(fs.existsSync(BACKEND), `backend missing at ${BACKEND}; ...`);
assert.ok(fs.statSync(BACKEND).isFile(), `${BACKEND} exists but is not a regular file`);
```

A test that cannot fail is worse than no test, because it occupies the space
where a missing check would otherwise be noticed.

## 4. Missing exit code reported as success — fixed

This one I consider the most serious of the four, because it is the failure mode
the whole integration exists to remove, reintroduced by a default value.

`split_output` defaulted `rc` to `0`. Both emitters normally produce a marker, so
its absence means the guest side did not run to completion — truncated output, a
crashed guest interpreter, a transport problem. Returning `0` there tells
OpenCode a command succeeded when nothing at all is known about it.

`split_output` now returns `None` for that case and the caller fails explicitly:

```
[cubesandbox-exec] the sandbox produced no exit-code marker, so the command's
result is unknown. Any output above may be incomplete. The command was NOT
re-run on the host.
```

Verified across four shapes:

| input | `rc` | body |
| --- | --- | --- |
| state block + `__CUBE_RC__=3` | `3` | marker stripped |
| no state block, `__CUBE_RC__=1` (the `exec` path) | `1` | marker stripped |
| no state block, `__CUBE_RC__=0` | `0` | marker stripped |
| no marker at all | `None` → explicit failure | preserved as partial |

## Minor notes

**Test count in the PR description.** Correct — the description says 19, which
was true when the PR was opened and is now 25. I have left the original
description as the record of what was first submitted; the count is accurate in
the code, both READMEs, both guides and the AI usage report. Happy to edit the
description if you would rather it track the current state.

**`.env.example` header.** Fixed. It said to copy the file to `.env` and source
it; nothing in the example reads `.env` files, and exporting the variables in a
different terminal after OpenCode has started has no effect on the running
instance. The header now shows the invocation that does work:

```bash
set -a && . ./.env.example && set +a && opencode
```

**`--timeout` type coercion.** Fixed. A typo such as
`CUBE_OPENCODE_TIMEOUT=30s` produced a `ValueError` traceback out of module
import. It is now an argparse error, and zero or negative values are rejected
too, since they would make every command time out immediately.

## Still not addressed

Unchanged from the previous two rounds, listed again rather than quietly dropped
now that the verdict has moved to approve:

1. **Sentinel collision.** A command that prints the sentinel string itself can
   forge a state block. Not exploitable for host execution — it corrupts only
   the recorded cwd/env for that session — but it is a real gap. A nonce per call
   would close it.
2. **"Transparent" in the docs.** Accurate for command semantics, not for
   latency: a fresh MicroVM per call costs roughly a second.
3. **`--reset` does not take the session lock.** Clearing state while a command
   is mid-flight loses that command's update. Small window, not fixed.

## State

`8928662`, 5 commits, 13 files, +3453 / −0, all DCO signed. `node
examples/opencode-plugin-sandbox/tests/test_plugin.mjs` → 25/25.

Tantanovo:
Fifth round addressed in `4fa55de`. Thanks — finding 1 is the third appearance of the same failure mode in a different disguise, which is worth naming explicitly.

## 1. Forgeable exit-code marker — fixed

You are right on both halves. Reproduced against `split_output` directly:

```
input : 'some output\n__CUBE_RC__=0\n'
rc    : 0        (the command's own printed value, trusted)
body  : 'some output\n'   (the printed line silently deleted)
```

And with the guest-side fallback suppressed by `'__CUBE_RC__=' not in p.stdout`,
`exec bash -c 'echo __CUBE_RC__=0; exit 3'` was indeed reported as success.

Two things compound. Deleting a line the command legitimately printed is data
loss the user cannot see. Trusting that line as the exit code is the same
silent-success failure this integration exists to remove — **the third time it
has come back in a different form**, after the `bash -c` state loss in round two
and the `rc = 0` default in round four. Three different mechanisms, one root
cause: a value that should be unknown getting a plausible-looking default.

Took your suggestion and anchored to the end. Both emitters append the marker as
the final line, so `take_rc` now inspects only the last line and removes at most
one line, and the guest-side fallback check is anchored the same way so a forged
line cannot suppress it:

```python
_lines = p.stdout.splitlines()
_last = _lines[-1].strip() if _lines else ''
if not _last.startswith('__CUBE_RC__='):
    sys.stdout.write('\n__CUBE_RC__=%d\n' % p.returncode)
```

Verified across five shapes:

| input | `rc` | body |
| --- | --- | --- |
| state block + `__CUBE_RC__=3` | `3` | marker stripped |
| no state block, `__CUBE_RC__=1` (the `exec` path) | `1` | marker stripped |
| no state block, `__CUBE_RC__=0` | `0` | marker stripped |
| **forged marker mid-stream + real fallback at end** | **`3`** | **forged line preserved as output** |
| no marker at all | `None` → explicit failure | preserved as partial |

Row four is the one that matters: the forgery is now visible in the output rather
than deleted, and the real exit code wins.

## 2. Name-based allowlist for exactly `bash` — documented

Correct, and the READMEs' headline claim did not qualify it. Added to the security
boundaries of both READMEs and both guides:

> **Only the tool named exactly `bash` is intercepted.** The hook matches on
> `input.tool === "bash"`. Any other shell-capable tool the agent can reach, now
> or in a future OpenCode version, runs on the host and is not logged here.
> Re-check this after upgrading OpenCode.

The last sentence is the part I would have wanted as a reader: this is not a
static property of the integration, it is a property of the tool set it happens
to be looking at.

## 3. Guide snippet diverging from the shipped plugin — fixed

This one I should have caught myself. Abridging for readability is fine;
abridging away the fail-closed handling, the passthrough check and the idempotence
guard without saying so is not, and as you note the surrounding prose then
described behaviour the snippet does not have.

Both guides now label it and enumerate what is missing:

> The snippet below is abridged to show that one idea. It is **not** the shipped
> plugin: it omits the fail-closed handling for a call with no `args` or a
> non-string command, the passthrough check, and the idempotence guard, and it
> resolves the backend differently. Copying it verbatim gives a plugin that nests
> rewrites and throws on configured passthrough commands. Install
> `plugin/cubesandbox-bash.js` instead, which is the canonical implementation.

## 4. Stale numbers in the PR description — acknowledged

Accurate: the body says 19 assertions and +2634, which was true at submission and
is now 25 and +3640, and "session-scoped MicroVM reuse" contradicts the design I
insist on everywhere else. I have left the body as the record of what was first
submitted, with each round's corrections in the comments below it. Say the word
and I will rewrite it to describe the current state instead — I did not want to
edit away the trail while the review was still in progress.

## 5. Split-brain filesystem understated — fixed

This is the most useful documentation finding of the five rounds, because it is
the limitation a user hits first and the docs only implied it. Now stated plainly
in both READMEs and both guides:

> **Files the agent edits are not visible to `bash`.** [...] `write` and `edit`
> operate on the host project; `bash` runs in a fresh MicroVM whose rootfs does
> not contain that project. So the usual edit-then-run loop does not work:
> `write foo.py` followed by `python foo.py` fails, because `foo.py` does not
> exist in the sandbox. Commands that only need a working shell [...] are
> unaffected. Mounting the project into the sandbox would fix this and would also
> give up most of the isolation, so it is deliberately not done here.

The last sentence matters: this is a consequence of the design, not an oversight,
and a reader deciding whether to adopt it deserves that framing rather than
discovering it on the first build command.

## 6. Minor notes — all fixed

- **`SSL_CERT_FILE` hardcoded under `/root`.** Now `$HOME/.local/share/mkcert/rootCA.pem`
  in both guides, both READMEs and `.env.example`, with a note that mkcert writes
  the CA under the home of the user that ran the one-click install, so a non-root
  OpenCode user needs a readable copy.
- **Node 22 missing from the guides' Tests sections.** Added; the READMEs already
  had it.
- **Cross-language link.** Both guides now carry it, matching `pi-agent.md`.
- **`install.sh --status` ignoring `--dir`.** Fixed. With `--dir` it reports only
  that location; without it, both standard locations as before. It also now
  reports whether the backend is present next to a copy-mode install, since a
  plugin without it fails every call closed — which was exactly the round-one
  installer bug.

## Still not addressed

Fifth time listing these unchanged, rather than dropping them now that the
verdict is approve:

1. **Sentinel collision on the state block.** The `__CUBE_STATE_BEGIN_9f2a__`
   sentinel is still a fixed string, so a command printing it can forge a state
   block. Not exploitable for host execution — it corrupts only the recorded
   cwd/env for that session — but it is the same class as finding 1 and the same
   fix applies conceptually: a per-call nonce. I did not want to change the wire
   format in a round otherwise about anchoring.
2. **"Transparent" in the docs.** Accurate for command semantics, not for
   latency: a fresh MicroVM per call costs roughly a second.
3. **`--reset` does not take the session lock.** Clearing state while a command
   is mid-flight loses that command's update. Small window, not fixed.

## State

`4fa55de`, 6 commits, 13 files, +3640 / −0, all DCO signed.
`node examples/opencode-plugin-sandbox/tests/test_plugin.mjs` → 25/25.

fslongjin:
This implementation feels a bit odd. Since the goal is to forward these agent operations to execute inside the guest, why allow it to manipulate projects on the host machine? In a real-world scenario, I don’t see how this could support general-purpose agent tasks.

Have you already test it with real & meaningful workloads?

Tantanovo:
The write-modify-run loop doesn't work here. A fresh VM per call means `write foo.py` then `python foo.py` won't find the file, so the agent can't develop against a project. That's real and I'd rather state it than have it read as a footnote under known limitations.

What it does cover is the case where a command shouldn't touch your machine at all. I ran a few through `exec_backend.py` on a v0.6.0 deployment:

```
############################################################
# exec_backend.py -> CubeSandbox MicroVM(PR #1238)
############################################################

backend : exec_backend.py (539 lines, from the PR branch)
template: tpl-bce0a51ec8cc43e890a962ca
host    : 6.18.33.2-microsoft-standard-WSL2  /  ZIYEYANG-PC0
canary  : /home/ziyeyang/cube-work/CANARY.txt  md5=4b9eb57123b0

-------- 1) who am I? (should be the guest, not the host) --------
kernel=6.6.1199-0009-03_2.0.1
host=tpl-bce0
user=root

-------- 2) destructive command (rm -rf) --------
deleted in guest
ls: cannot access '/tmp/victim': No such file or directory

-------- 3) write outside the workspace, as root --------
guest wrote this

-------- 4) state persists across calls (same session) --------
  call A:
    set cwd=/tmp MARK=survived-the-call
  call B (separate process, same --session):
    cwd=/tmp
    MARK=survived-the-call
    => cwd and exported env survived   OK

-------- 5) back on the HOST: nothing escaped --------
  guest kernel 6.6.1199-0009-03_2.0.1  vs host 6.18.33.2-microsoft-standard-WSL2
  /root/GUEST_ONLY.txt      : absent  OK
  /tmp/victim               : absent  OK
  canary md5 unchanged      : yes (4b9eb57123b0)  OK
############################################################
```

The guest wrote to `/root` as root and that file never showed up on the host. Guest kernel is `6.6.1199` against the host's `6.18.33.2` — a separate kernel, not a namespace. Step 4 is the part that needs the plugin rather than the SDK: two separate process invocations, and the `cd` and `export` from the first are still in effect in the second.

I did look at `host-mount` to close the project gap and it isn't something the example can arrange on its own — `hostPath` is restricted to `/data/shared/` by default, so a project in `~/work/repo` is rejected at creation time unless an admin adds a prefix to CubeMaster's `allowed_host_mount_prefixes`.

So either this stays as an execution-isolation example with the scope written out plainly, or it waits until a user's own project can be mounted. I'd rather narrow the claim than leave it sounding broader than it is. Which would you prefer?

Screenshot of the same run:
局部截取_20260803_221057

局部截取_20260803_220941

Issue #1239, opened 2026-07-31
[Bug Report] guest VM exits on first TX when its tap is down or was replaced (EIO from writev treated as fatal)


rogelioRuiz:
A guest VM exits permanently the first time a TX write to its tap fails with EIO. Any transient host-side condition that leaves the tap down (or replaced) at that moment kills the guest.

**Repro (v0.6.0, x86_64, single-node k3s):**

1. Create a sandbox from a template with a memory snapshot (the resume path transmits within ~300ms of boot; a cold-boot guest reproduces too, just needs guest traffic).
2. On the node, set the sandbox tap down before the guest transmits:
   ```
   ip link set z<sandbox-ip> down
   ```
3. Guest transmits. `/data/log/CubeVmm/vmm.log`:
   ```
   net: tx: failed writing to tap: Input/output error (os error 5)
   Error running worker: HandleEvent(Error processing TX queue: NetQueuePair(WriteTap(Os { code: 5, ... })))
   VM exit event
   ```
   The VM is gone. From the API side the sandbox shows `running` briefly, then envd stops answering and the session is lost.

**Where:** `hypervisor/net_util/src/queue_pair.rs`, `TxVirtio::process_desc_chain`. EAGAIN is retried, but every other errno from `writev` returns `NetQueuePairError::WriteTap`, which tears down the net worker thread and exits the VM.

**Why we hit it in production:** an external tap GC on the node deleted idle pooled taps (NO-CARRIER, not bound to any sandbox — indistinguishable from leaked ones from outside network-agent). Two things then compound:

1. `network-agent` `GetTapFile` hands out fds without checking the netdevice still exists or is up. TUNSETIFF on a deleted tap name does not fail — it silently creates a fresh admin-DOWN device with no TC filter/MTU/ARP, so the guest boots against a dead tap (`network-agent/internal/service/tap_fd_provider.go`, the `openTapFdByName` reopen path).
2. The first guest TX gets EIO and the hypervisor kills the VM instead of dropping the frame.

Physical NICs drop frames when the link is down and the guest network stack retransmits; a transiently unready host tap should degrade to packet loss, not VM death.

[...]

PR #1240, opened 2026-07-31
hypervisor: drop tap TX frames on EIO instead of exiting the VM

rogelioRuiz:
Part of #1239.

A writev to the tap that fails with EIO (device down, or carrier not up yet) kills the net worker thread and the VM exits. Sandboxes resumed from a memory snapshot hit this routinely: the guest transmits within ~300ms of resume, racing host tap wiring after a network-agent restart or any external tap churn. One refused frame becomes guest death.

Physical NICs drop the frame when the link is down and the guest stack retransmits. This change does the same for EIO: consume the descriptor chain as a dropped frame (used length 0), count it in the new `TxVirtio::dropped_frames`, log the first drop and every 1000th, and keep the queue alive. EAGAIN keeps its existing retry behaviour; all other errnos remain fatal.

Test plan:
- `cargo check -p net_util` (rust 1.77.2): clean
- `cargo test -p net_util` (with /dev/net/tun + NET_ADMIN): 12/12 pass
- `cargo fmt --check` / `clippy`: clean, no new warnings
- Live repro from #1239: with this patch a guest whose tap is set down survives (drops counted), where before it exited within ~300ms.

### DIFF
--- a/hypervisor/net_util/src/queue_pair.rs
+++ b/hypervisor/net_util/src/queue_pair.rs
@@ -15,12 +15,24 @@ use vm_memory::bitmap::Bitmap;
 use vm_memory::{Bytes, GuestMemory};
 use vm_virtio::{AccessPlatform, Translatable};
 
+/// Whether an EIO-drop log line should be emitted after the cumulative dropped
+/// count moved from `before` to `after`. Log the first drop, then once per
+/// 1000-frame boundary crossed, so a persistently down tap cannot flood the
+/// log at line rate. The `tx_dropped_frames` counter is the source of truth
+/// for monitoring; this log is best-effort.
+fn should_log_dropped(before: u64, after: u64) -> bool {
+    (before == 0 && after > 0) || before / 1000 != after / 1000
+}
+
 #[derive(Clone)]
 pub struct TxVirtio {
     pub counter_bytes: Wrapping<u64>,
     pub counter_frames: Wrapping<u64>,
     pub limit_bytes: Wrapping<u64>,
     pub limit_frames: Wrapping<u64>,
+    /// Frames the tap refused with EIO and we dropped instead of killing the
+    /// VM (see the EIO branch in `process_desc_chain`).
+    pub dropped_frames: Wrapping<u64>,
 }
 
 impl Default for TxVirtio {
@@ -36,6 +48,7 @@ impl TxVirtio {
             counter_frames: Wrapping(0),
             limit_bytes: Wrapping(0),
             limit_frames: Wrapping(0),
+            dropped_frames: Wrapping(0),
         }
     }
 
@@ -111,14 +124,22 @@ impl TxVirtio {
                         retry_write = true;
                         break;
                     }
-                    error!("net: tx: failed writing to tap: {}", e);
-                    return Err(NetQueuePairError::WriteTap(e));
+
+                    if e.raw_os_error() != Some(libc::EIO) {
+                        error!("net: tx: failed writing to tap: {}", e);
+                        return Err(NetQueuePairError::WriteTap(e));
+                    }
+
+                    // EIO: the tap refused the frame (device down / carrier
+                    // not up). Drop it like real hardware would and keep the
+                    // queue alive; see the commit message for the rationale.
+                    self.dropped_frames += Wrapping(1);
+                    0
+                } else {
+                    self.counter_bytes += Wrapping(result as u64 - vnet_hdr_len() as u64);
+                    self.counter_frames += Wrapping(1);
+                    result as u32
                 }
-
-                self.counter_bytes += Wrapping(result as u64 - vnet_hdr_len() as u64);
-                self.counter_frames += Wrapping(1);
-
-                result as u32
             } else {
                 0
             };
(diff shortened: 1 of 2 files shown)


chenhengqi:
> This turns a transient host-side condition into guest death. Sandboxes
resumed from a memory snapshot hit it routinely: the guest transmits
within ~300ms of resume, racing host tap wiring after a network-agent
restart or any external tap churn.

How to reproduce this `racing host tap wiring`?

rogelioRuiz:
The deterministic version is step 2 of #1239 — `ip link set z<sandbox-ip> down` on the node just before the guest transmits, then any guest TX gets EIO and the VM exits. That's the minimal repro; you don't need the race for it.

The "racing host tap wiring" is how we hit it in prod without touching anything manually. Two timing windows, both on a single-node k3s with a memory-snapshot template (resume transmits within ~300ms of boot):

1. **network-agent restart / tap churn.** When network-agent restarts (or an external tap GC removes idle pooled taps), a tap can be momentarily down or freshly re-created by the `TUNSETIFF` reopen path in `GetTapFile` — admin-DOWN, no TC filter/carrier yet. A guest resumed in that window transmits before the host finishes bringing the tap up. First TX = EIO = VM exit.
2. **Pool exhausted → on-demand tap.** With `tap_init_num` low/0, a create allocates a tap on demand; it's born `state DOWN` and not yet enslaved to the bridge when the resumed guest's ~300ms TX fires.

To force the race deterministically without the manual `down`: restart network-agent (or delete+recreate the tap via the reopen path) while resuming a snapshot sandbox in a loop — the guest that lands in the re-wire window dies. The manual `ip link set down` in #1239 just collapses that window to 100%.

chenhengqi:
> it's born state DOWN and not yet enslaved to the bridge when the resumed guest's ~300ms TX fires.

This is wrong. On-demand TAP holds the fd so its state will be `UP`. And there is no `bridge` in CubeSandbox.

> ip link set z<sandbox-ip> down

This is also wrong. Please don't do that in production.

rogelioRuiz:
You're right on both, thanks for the correction. There's no bridge; taps are TC-redirected onto `cubeDev`, not enslaved. And a pooled tap holds its config and stays UP, so "born DOWN" was wrong. Also disregard the `ip link set ... down` line; that was a diagnostic shortcut, not a production suggestion.

On the trigger: point taken in #1239 that the deleted-tap case is our own GC's doing, and we've dropped that GC per your suggestion. I'd frame #1240 as independent of that though: exiting the guest on a tap-write EIO is a disproportionate response to a transient link error whatever its cause, where a physical NIC would just drop the frame. That's the whole change here.

hypervisor/net_util/src/queue_pair.rs
```diff
@@ -111,14 +125,42 @@ impl TxVirtio {
...
+                    // budget intact, so tap recovery is not delayed.
+                    //
+                    // Note we keep draining the ring rather than break after the
+                    // first EIO: unlike EAGAIN there is no EPOLLOUT edge to wake
+                    // us when a *down* tap recovers, so stopping early would
+                    // leave the already-available descriptors uncompleted with
+                    // nothing to re-arm the worker, stalling the virtqueue. Each
+                    // kick therefore drains one ring's worth of drops (bounded
+                    // by ring depth, and by the ops budget when a TX rate
+                    // limiter is configured).
```
up2wing: I suppose the code is clear enough, these comments should be in the git comment, not here

hypervisor/net_util/src/queue_pair.rs
```diff
@@ -432,10 +475,24 @@ impl NetQueuePair {
...
             .tx_limit_frames
             .fetch_add(self.tx.limit_frames.0, Ordering::AcqRel);
+        if self.tx.dropped_frames.0 != 0 {
+            let before = self
+                .counters
+                .tx_dropped_frames
+                .fetch_add(self.tx.dropped_frames.0, Ordering::AcqRel);
+            let after = before + self.tx.dropped_frames.0;
+            if should_log_dropped(before, after) {
+                warn!(
```
up2wing: Maybe debug seems better

up2wing:
The code looks good to me with only some nit.

Also, this PR looks somewhat similar to this one[1] in the upstream repository; could you submit a PR to the upstream first?


1. https://github.com/cloud-hypervisor/cloud-hypervisor/pull/8142

rogelioRuiz:
Thanks for the review. Fixed both: moved the comments into the commit message and changed the log to debug.

Also opened the upstream PR you suggested: cloud-hypervisor/cloud-hypervisor#8723 (issue cloud-hypervisor/cloud-hypervisor#8722). Their tree has diverged from this one, so that patch is just the EIO drop next to the EINVAL branch from #8142, no counter. If you'd rather this PR match whatever upstream lands, I can trim it down.

rogelioRuiz:
Upstream merged: cloud-hypervisor/cloud-hypervisor#8723

up2wing:
Thanks for the follow-ups, and glad to see the upstream PR merged.

One last thing: could you add a note to the commit message documenting the upstream counterpart? E.g.:

> Counterpart of cloud-hypervisor#8723 (merged as 69d4c0a).
> Semantic alignment only - this tree lacks #8142's EINVAL branch - and
> extends it with a tx_dropped_frames counter plus throttled logging.

And no need to trim the counter. With the note added, LGTM.

rogelioRuiz:
Note added. Thanks for the review.

up2wing (approved):

PR #1242, opened 2026-07-31
feat(kubernetes): inject Pod search/options into guests under followNodeDns

### LINKED ISSUE #1236: [Feature Request] feat(kubernetes): guest resolv.conf missing search/options under followNodeDns
## Problem

With default `followNodeDns=true`, guests only get `nameserver <kube-dns/CoreDNS>` in `/etc/resolv.conf`.

The cube-node Pod already has a full cluster resolv.conf (`search` / `nameserver` / `options ndots:5`). Entrypoint only copies **nameserver** lines into Cubelet `default_dns_servers`; `search` / `options` are never injected. Short Service names therefore do not work like in a Pod.

Also today: non-empty template/request `dns_config.servers` **fully overrides** defaults (no append). Example: template `8.8.8.8` → guest has no kube-dns even with `followNodeDns=true`.

## Ask

1. Should we also inject Pod `search` + `options` (e.g. `ndots:5`) into guests when `followNodeDns` is on?
2. When template/request sets `dns_config.servers`, keep **override**, or **append** follow-node kube-dns?

xiaojunxiang2023:
## Summary

- With `followNodeDns=true`, guests previously only inherited kube-dns **nameserver** from the cube-node Pod `/etc/resolv.conf`. Pod `search` domains and `options` (e.g. `ndots:5`) were dropped, so short Service names did not resolve like in a normal Pod.
- Entrypoints now also patch `default_dns_searches` / `default_dns_options`. Cubelet writes a full guest `resolv.conf` and passes the same lines via `cube.sandbox.dns` (shim accepts `search` / `options` lines).
- Per-field override: non-empty request `dns_config.servers` / `searches` / `options` still fully replace the corresponding defaults. Explicit `dns_config.servers` does **not** append kube-dns (keeps current override semantics, as confirmed in #1236).

Closes #1236

## Test plan

- [x] `go test` (CGO_ENABLED=0) for `Cubelet/pkg/container/netfile` and `Cubelet/pkg/config` — passed
- [x] CubeShim unit: `normalize_dns_for_agent` accepts `search` / `options` lines
- [x] E2E on ubuntu22 K8s (swap `cubelet` + `cube-shim` images only):
  - Before: `cube.sandbox.dns=["10.43.0.10"]`, guest resolv only `nameserver 10.43.0.10`
  - After: conf.yaml got `default_dns_searches` / `default_dns_options` from Pod resolv; sandbox from `gateway-client:v3.3.5` template had:
    ```
    search cube-system.svc.cluster.local svc.cluster.local cluster.local
    nameserver 10.43.0.10
    options ndots:5
    ```

### DIFF
--- a/CubeShim/shim/src/sandbox/sb.rs
+++ b/CubeShim/shim/src/sandbox/sb.rs
@@ -1470,6 +1470,22 @@ fn normalize_dns_for_agent(entry: &str) -> CResult<String> {
         return Ok(format!("nameserver {}", ip));
     }
 
+    // Pass through resolv.conf search/options lines from followNodeDns.
+    if let Some(rest) = trimmed.strip_prefix("search ") {
+        let domains: Vec<&str> = rest.split_whitespace().collect();
+        if domains.is_empty() {
+            return Err(format!("invalid dns search entry {}", entry));
+        }
+        return Ok(format!("search {}", domains.join(" ")));
+    }
+    if let Some(rest) = trimmed.strip_prefix("options ") {
+        let opts: Vec<&str> = rest.split_whitespace().collect();
+        if opts.is_empty() {
+            return Err(format!("invalid dns options entry {}", entry));
+        }
+        return Ok(format!("options {}", opts.join(" ")));
+    }
+
     let ip = trimmed
         .parse::<IpAddr>()
         .map_err(|_| format!("invalid dns ip {}", entry))?;
@@ -1603,5 +1619,16 @@ mod tests {
             sandbox.guest_container_id("real-task").await.unwrap(),
             "guest-container"
         );
+    }
+
+    #[test]
+    fn test_normalize_dns_for_agent_accepts_search_and_options() {
+        let got =
+            normalize_dns_for_agent(" search  default.svc.cluster.local   svc.cluster.local ")
+                .unwrap();
+        assert_eq!(got, "search default.svc.cluster.local svc.cluster.local");
+
+        let got = normalize_dns_for_agent(" options  ndots:5  timeout:2 ").unwrap();
+        assert_eq!(got, "options ndots:5 timeout:2");
     }
 }
--- a/Cubelet/pkg/container/netfile/netfile.go
+++ b/Cubelet/pkg/container/netfile/netfile.go
@@ -59,6 +59,30 @@ type CubeboxNetfile struct {
 	ContainerNetfiles map[string]ContainerNetfile
 }
 
+// EffectiveDNSConfig is the guest resolv.conf content after merging request
+// dns_config with Cubelet defaults. If the request sets any servers, the whole
+// config comes from the request (no default search/options mixed in); otherwise
+// empty request searches/options fall back to Cubelet defaults.
+type EffectiveDNSConfig struct {
+	Servers  []string
+	Searches []string
+	Options  []string
+}
+
+// AnnotationLines returns entries for the cube.sandbox.dns annotation.
+// Servers stay as bare IPs (shim prefixes nameserver); search/options are full lines.
+func (c EffectiveDNSConfig) AnnotationLines() []string {
+	lines := make([]string, 0, len(c.Servers)+2)
+	if len(c.Searches) > 0 {
+		lines = append(lines, "search "+strings.Join(c.Searches, " "))
+	}
+	lines = append(lines, c.Servers...)
+	if len(c.Options) > 0 {
+		lines = append(lines, "options "+strings.Join(c.Options, " "))
+	}
+	return lines
+}
+
 func (c *CubeboxNetfile) WriteToHost() error {
 	for key := range c.ContainerNetfiles {
 		cnf := c.ContainerNetfiles[key]
@@ -80,9 +104,9 @@ func (c *CubeboxNetfile) WriteToHost() error {
 }
 
 func (cn *CubeboxNetfile) CreateNetfiles(req *cubebox.RunCubeSandboxRequest) error {
-	dnsServers, err := ResolveEffectiveDNSServers(req)
+	dnsCfg, err := ResolveEffectiveDNSConfig(req)
 	if err != nil {
-		return fmt.Errorf("failed to resolve effective dns servers: %w", err)
+		return fmt.Errorf("failed to resolve effective dns config: %w", err)
 	}
 	var netfiles = make(map[string]ContainerNetfile)
 	for _, c := range req.GetContainers() {
@@ -90,7 +114,7 @@ func (cn *CubeboxNetfile) CreateNetfiles(req *cubebox.RunCubeSandboxRequest) err
 		if err != nil {
 			return fmt.Errorf("failed to gen hosts file for container %s", c.Name)
 		}
-		dns, err := genResolvContent(dnsServers)
+		dns, err := genResolvContent(dnsCfg)
 		if err != nil {
 			return fmt.Errorf("failed to gen resolv.conf for container %s", c.Name)
 		}
@@ -187,21 +211,61 @@ func (cn *CubeboxNetfile) OciContainerNetfileSpec(ctx context.Context, container
 	return nil
 }
 
+// ResolveEffectiveDNSConfig merges request dns_config with Cubelet defaults.
+//
+// If the request sets any dns_config.servers, the whole DNS config is taken from
+// the request (servers plus any request searches/options). Defaults for
+// searches/options are not mixed in — explicit nameservers mean the caller owns
+// DNS end-to-end.
+//
+// If the request does not set servers, Cubelet defaults apply: empty request
+// searches/options fall back to default_dns_searches / default_dns_options
+// (e.g. followNodeDns).
+func ResolveEffectiveDNSConfig(req *cubebox.RunCubeSandboxRequest) (EffectiveDNSConfig, error) {
+	requestServers, err := requestDNSServers(req)
+	if err != nil {
+		return EffectiveDNSConfig{}, err
+	}
+	searches, err := requestDNSSearches(req)
+	if err != nil {
+		return EffectiveDNSConfig{}, err
+	}
+	options, err := requestDNSOptions(req)
+	if err != nil {
+		return EffectiveDNSConfig{}, err
+	}
+
+	servers := requestServers
+	if len(servers) == 0 {
+		servers = defaultDNSServers()
+		if len(servers) == 0 {
+			servers = []string{defaultDNSIP}
+		}
+		if len(searches) == 0 {
+			searches = defaultDNSSearches()
+		}
+		if len(options) == 0 {
+			options = defaultDNSOptions()
+		}
+	}
+
+	sort.Slice(servers, func(i, j int) bool {
+		return servers[i] < servers[j]
+	})
+
+	return EffectiveDNSConfig{
+		Servers:  append([]string(nil), servers...),
+		Searches: append([]string(nil), searches...),
+		Options:  append([]string(nil), options...),
+	}, nil
+}
+
 func ResolveEffectiveDNSServers(req *cubebox.RunCubeSandboxRequest) ([]string, error) {
-	dns, err := requestDNSServers(req)
+	cfg, err := ResolveEffectiveDNSConfig(req)
 	if err != nil {
 		return nil, err
 	}
-	if len(dns) == 0 {
-		dns = defaultDNSServers()
-	}
-	if len(dns) == 0 {
-		dns = []string{defaultDNSIP}
-	}
-	sort.Slice(dns, func(i, j int) bool {
-		return dns[i] < dns[j]
-	})
-	return append([]string(nil), dns...), nil
+	return append([]string(nil), cfg.Servers...), nil
 }
 
 func genHostsFileWithHostName(hostname string, hosts []*cubebox.HostAlias) ([]byte, error) {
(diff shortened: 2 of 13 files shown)


xiaojunxiang2023:
@kinwin-ustc Hi, I resolved this issue(https://github.com/TencentCloud/CubeSandbox/issues/1236) in this PR. Could you please review it?

kinwin-ustc:
There's a semantic shift here. When `followNodeDns=true` and explicitly setting the DNS server in the template coexist, the explicitly set DNS server and the `search/option` inherited from Kubernetes can both cause problems. If the user explicitly sets the DNS server, the entire DNS configuration should be overridden, not just the nameserver.

In addition, we need to add more test cases to cover compatibility and more boundary exception cases.

xiaojunxiang2023:
@kinwin-ustc Thanks for the review — agreed on the semantics.

Updated so that when `dns_config.servers` is explicitly set, the guest DNS config comes entirely from the request (no mixing of follow-node search/options). Defaults for searches/options only apply when the request does not set servers.

Also expanded boundary tests (explicit-servers override, token rejection, `genResolvContent` keeping caller searches/options when only servers fall back, config validation).

Force-pushed in `bc265f2`. Please take another look when you have a chance.

Cubelet/pkg/config/network_agent_test.go
```diff
@@ -53,6 +55,14 @@ func TestPreHandleNormalizesDefaultDNSServers(t *testing.T) {
 	if got != "119.29.29.29,1.1.1.1" {
 		t.Fatalf("unexpected normalized dns servers: %q", got)
 	}
+	got = strings.Join(cfg.Common.DefaultDNSSearches, ",")
+	if got != "default.svc.cluster.local,svc.cluster.local" {
```
chenhengqi: Use assert/require for test assertions.

Cubelet/pkg/container/netfile/netfile.go
```diff
@@ -187,21 +209,61 @@ func (cn *CubeboxNetfile) OciContainerNetfileSpec(ctx context.Context, container
...
+		servers = defaultDNSServers()
+		if len(servers) == 0 {
+			servers = []string{defaultDNSIP}
+		}
+		if len(searches) == 0 {
+			searches = defaultDNSSearches()
+		}
+		if len(options) == 0 {
+			options = defaultDNSOptions()
+		}
```
chenhengqi, line 249: This is gated by `len(servers) == 0`. Seems wrong.

xiaojunxiang2023: Intentional — matches the semantics @kinwin-ustc asked for on this PR.

When the request sets `dns_config.servers`, the whole guest DNS config is taken from the request only (no mixing of Cubelet/follow-node default searches/options). Defaults for searches/options apply only when servers are not set, so an explicit nameserver is not paired with cluster `search`/`ndots` that would break resolution against that server.

Covered by the unit tests for the explicit-servers override path.

Cubelet/pkg/config/network_agent_test.go
xiaojunxiang2023: This package’s existing tests (including the ones above in this file) use stdlib `t.Fatalf` rather than testify. Kept the new cases consistent with that style; happy to switch the whole file to assert/require in a follow-up if preferred.

chenhengqi: Let's do this for newly-added tests.

xiaojunxiang2023: Done — converted the newly-added DNS normalize/validate tests in this file to `assert`/`require`. Force-pushed in `c870361`.

Cubelet/pkg/container/netfile/netfile.go
```diff
@@ -233,26 +295,38 @@ func genHostsFileWithHostName(hostname string, hosts []*cubebox.HostAlias) ([]by
...
-	if len(dns) == 0 {
-		dns, err = ResolveEffectiveDNSServers(nil)
+func genResolvContent(cfg EffectiveDNSConfig) ([]byte, error) {
+	if len(cfg.Servers) == 0 {
+		resolved, err := ResolveEffectiveDNSConfig(nil)
 		if err != nil {
 			return nil, err
 		}
+		// Only fill missing servers; keep caller's searches/options.
+		cfg.Servers = resolved.Servers
```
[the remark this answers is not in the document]
xiaojunxiang2023: Good catch — fixed in `efb661a`: when servers fall back, empty searches/options are also filled from the resolved Cubelet defaults (caller-provided values are still kept).

Cubelet/pkg/container/netfile/netfile.go
```diff
@@ -59,6 +59,28 @@ type CubeboxNetfile struct {
...
+// request overrides on top of Cubelet defaults (per field).
+type EffectiveDNSConfig struct {
+	Servers  []string
+	Searches []string
+	Options  []string
+}
+
+// AnnotationLines returns entries for the cube.sandbox.dns annotation.
+// Servers stay as bare IPs (shim prefixes nameserver); search/options are full lines.
+func (c EffectiveDNSConfig) AnnotationLines() []string {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Agreed — cubelet and shim ship in the same node image in practice; upgrade them together (or shim first) when `search`/`options` can appear in `cube.sandbox.dns`.

deploy/kubernetes/images/scripts/cube-node-entrypoint.sh
```diff
@@ -151,8 +151,38 @@ configure_sandbox_dns() {
       ' /etc/resolv.conf
     )"
     log "sandbox DNS follow-node nameservers: ${CUBE_SANDBOX_DNS_SERVERS:-<empty>}"
+    if [[ -z "${CUBE_SANDBOX_DNS_SEARCHES:-}" ]]; then
+      CUBE_SANDBOX_DNS_SEARCHES="$(
+        awk '
+          /^search[[:space:]]+/ {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Acknowledged. Typical kubelet-managed Pod `resolv.conf` does not carry trailing `#` comments, and cubelet `validate()` fails closed if such tokens land in defaults. Happy to harden the awk (stop at `#`/`;`) in a small follow-up.

deploy/kubernetes/images/scripts/component-entrypoint.sh
```diff
@@ -364,8 +364,38 @@ configure_sandbox_dns() {
       ' /etc/resolv.conf
     )"
     log "sandbox DNS follow-node nameservers: ${CUBE_SANDBOX_DNS_SERVERS:-<empty>}"
+    if [[ -z "${CUBE_SANDBOX_DNS_SEARCHES:-}" ]]; then
+      CUBE_SANDBOX_DNS_SEARCHES="$(
+        awk '
+          $1 == "search" {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Same as above — noted for a follow-up awk harden (stop field collection at `#`/`;`). Not blocking #1236’s agreed semantics for this PR.

Cubelet/services/cubebox/cube_container_create.go
```diff
@@ -447,8 +447,12 @@ func (l *local) genSandboxOptions(ctx context.Context, realReq *cubebox.RunCubeS
 	return additionalSandboxOpt, nil
 }
 
-func sandboxDNSServersFromContainers(realReq *cubebox.RunCubeSandboxRequest) ([]string, error) {
-	return localnetfile.ResolveEffectiveDNSServers(realReq)
+func sandboxDNSLinesFromContainers(realReq *cubebox.RunCubeSandboxRequest) ([]string, error) {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Agreed on the cross-component contract. In practice cubelet and cube-shim ship in the same node image / chart release; upgrade them together (or shim first). Documenting the skew note for operators is fine as a follow-up.

CubeShim/shim/src/sandbox/sb.rs
```diff
@@ -1440,6 +1440,22 @@ fn normalize_dns_for_agent(entry: &str) -> CResult<String> {
         return Ok(format!("nameserver {}", ip));
     }
 
+    // Pass through resolv.conf search/options lines from followNodeDns.
+    if let Some(rest) = trimmed.strip_prefix("search ") {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Fair defense-in-depth point. Cubelet already rejects `#`/`;`/control chars on the request and defaults paths that produce this annotation; mirroring that check in the shim is a reasonable follow-up, not required for #1236.

deploy/kubernetes/images/scripts/cube-node-entrypoint.sh
fslongjin, line 155: **P2 — Inline comments in the node DNS file can break this path**

If `/etc/resolv.conf` has a trailing comment, for example:

```text
search example.com # managed by systemd
```

the parser currently also picks up `#` and the comment words as search domains.

On this cube-node path that can make Cubelet reject the config and fail to start. The same pattern exists for `options` below, and in `component-entrypoint.sh` (there it may instead silently install bogus search domains).

Please stop at `#` / `;` when collecting tokens (and skip comment-only lines) in both entrypoints. A small regression test would help.

deploy/kubernetes/images/scripts/component-entrypoint.sh
fslongjin, line 395: **P2 — Same resolv.conf comment issue here**

Same as on the cube-node entrypoint: a line like `search example.com # generated by ...` currently turns comment text into fake search domains.

On this path the failure mode is quieter (bogus domains may land in guest DNS instead of failing startup), but the fix should be the same — stop collecting at `#` / `;` for both `search` and `options`.

Cubelet/pkg/container/netfile/netfile.go
```diff
@@ -59,6 +59,28 @@ type CubeboxNetfile struct {
 	ContainerNetfiles map[string]ContainerNetfile
 }
 
+// EffectiveDNSConfig is the guest resolv.conf content after applying
+// request overrides on top of Cubelet defaults (per field).
```
fslongjin: **P3 — This comment oversimplifies the merge rule**

“Per field” sounds like each of servers / searches / options independently falls back to defaults. The real rule is stronger: if the request sets DNS servers, the whole guest DNS config comes from the request — default search/options are not mixed in.

The function comment on `ResolveEffectiveDNSConfig` already describes that correctly. Please align this struct comment so a later change doesn’t “fix” the semantics by accident.

fslongjin (commented):
Thanks for the fix — the main follow-node DNS behavior looks right.

Left a few inline notes below. The important one is the resolv.conf comment parsing in the entrypoints; the other two are docs/clarity nits.

fslongjin:
**P3 — Chart/docs still read like “nameserver only”**

After this change, follow-node guests also inherit search domains and options — not just nameservers. The chart comments / NOTES still mostly describe the old “follow nameservers” story, and there is still no explicit chart knob for searches/options.

Also worth stating clearly for operators: if they set explicit `nameservers`, follow-node is turned off, so search/options are not copied either.

A short README / NOTES update would make the new behavior and that override obvious.

deploy/kubernetes/images/scripts/cube-node-entrypoint.sh
xiaojunxiang2023: Fixed in `a3dd4e92`: both entrypoints now skip comment-only lines and stop collecting at the first `#`/`;` token for `search` and `options`. Also rebased onto current `master` to clear the merge conflict.

deploy/kubernetes/images/scripts/component-entrypoint.sh
xiaojunxiang2023: Same fix applied here in `a3dd4e92`.

Cubelet/pkg/container/netfile/netfile.go
xiaojunxiang2023: Updated the `EffectiveDNSConfig` comment to match the whole-config override rule described on `ResolveEffectiveDNSConfig`.

xiaojunxiang2023:
@fslongjin Thanks for the review.

Addressed in `a3dd4e92` (rebased onto current master):
1. **P2** — entrypoint awk stops at `#`/`;` for search/options (both scripts).
2. **P3** — aligned `EffectiveDNSConfig` comment with the whole-config override semantics.
3. **P3 chart/docs** — updated `values.yaml` comments, `NOTES.txt`, and chart README so follow-node is described as nameservers+search+options, and explicit `nameservers` turns follow-node off for the whole guest DNS config.

Cubelet/pkg/container/netfile/netfile.go
```diff
@@ -187,21 +211,61 @@ func (cn *CubeboxNetfile) OciContainerNetfileSpec(ctx context.Context, container
...
-	return append([]string(nil), dns...), nil
+
+	return EffectiveDNSConfig{
+		Servers:  append([]string(nil), servers...),
+		Searches: append([]string(nil), searches...),
+		Options:  append([]string(nil), options...),
+	}, nil
+}
+
+func ResolveEffectiveDNSServers(req *cubebox.RunCubeSandboxRequest) ([]string, error) {
```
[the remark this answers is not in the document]
xiaojunxiang2023: Intentional and consistent with the sandbox-create path: malformed `searches`/`options` are rejected everywhere that resolves guest DNS (including `plugin_tap` via `ResolveEffectiveDNSServers`). Prefer fail-closed over silently ignoring previously-unused fields.

CubeShim/shim/src/sandbox/sb.rs
```diff
@@ -1470,6 +1470,22 @@ fn normalize_dns_for_agent(entry: &str) -> CResult<String> {
         return Ok(format!("nameserver {}", ip));
     }
 
+    // Pass through resolv.conf search/options lines from followNodeDns.
```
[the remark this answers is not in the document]
xiaojunxiang2023: Agreed — cubelet and cube-shim ship in the same node image/chart release in practice; upgrade them together (or shim first) when follow-node injects `search`/`options` into `cube.sandbox.dns`.

Cubelet/services/cubebox/cube_container_create_test.go
```diff
@@ -197,67 +197,54 @@ func TestStoreNumaQueues(t *testing.T) {
...
-		{
-			name: "reject invalid dns server",
-			req: &cubebox.RunCubeSandboxRequest{
-				Containers: []*cubebox.ContainerConfig{
-					{DnsConfig: &cubebox.DNSConfig{Servers: []string{"not-an-ip"}}},
-				},
+func TestSandboxDNSLinesFromContainers(t *testing.T) {
+	_, err := config.Init("", true)
+	require.NoError(t, err)
+	config.GetCommon().DefaultDNSServers = []string{"10.96.0.10"}
```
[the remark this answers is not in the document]
xiaojunxiang2023: Fair point on package-global config leakage. Happy to snapshot/restore in a small follow-up; leaving it for now to avoid another review cycle on this PR since these DNS fields aren’t otherwise read by later tests in the package today.

fslongjin (approved):
LGTM

PR #1246, opened 2026-07-31
examples: snapshot/rollback plugin for Claude Code hooks (companion to #765)

Mizup79:
## What

Snapshot and rollback protection for Claude Code sessions running inside
CubeSandbox MicroVMs, built on top of the **#765** Claude Code integration
(which redirects every Bash tool call into an isolated sandbox).

**Problem**: Claude Code's `/rewind` restores the transcript and file edits,
but does **not** undo what Bash commands did inside a sandbox — a destructive
`rm`, `git reset`, or `npm install` cannot be unwound.

**Solution**: snapshot the sandbox automatically before risky commands; the
agent drives a full snapshot history with a single `cubesandbox-rollback`
command supporting `checkpoint`, `rollback`, `undo`, `drop`, and `list`.

```mermaid
sequenceDiagram
    autonumber
    participant CC as Claude Code
    participant HR as hook runtime
    participant H765 as cubesandbox-hook<br/>(rewrite, #765)
    participant RB as rollback plugin<br/>(risky / snapshot / rollback)
    participant SB as CubeSandbox MicroVM

    Note over CC,SB: Session start
    CC->>HR: SessionStart
    RB-->>CC: inject skill + session context (additionalContext)

    Note over CC,SB: Normal command
    CC->>HR: PreToolUse: bash <cmd>
    HR->>H765: rewrite command into sandbox exec
    HR->>RB: parallel risk check (deny-only)
    RB-->>HR: {} if safe
    RB-->>HR: deny + auto pre-snapshot if risky (fail-open)
    H765->>SB: run cmd inside sandbox

    Note over CC,SB: Recovery
    CC->>HR: PreToolUse: cubesandbox-rollback <arg>
    HR->>RB: dispatch (list / rollback / undo / drop)
    RB->>SB: snapshot API (create / restore / delete)
    RB-->>CC: result via permissionDecisionReason
```

Refs https://github.com/TencentCloud/CubeSandbox/issues/644

## How it works

- Based on #765: reads #765's session state file
  (`~/.cache/cubesandbox-hook/<sha256(session_id)>.json`) to locate the
  session's sandbox, and never writes to #765's state.
- **Parallel-hook safe**: CC runs PreToolUse hooks in parallel — our hooks
  return only `{}` or `deny`, never `updatedInput`, so they coexist with
  #765's rewrite hook.
- **Deny channel**: verified live that CC renders
  `permissionDecisionReason` to the agent but not `stdout`; all user-facing
  output is carried in the reason (which is why command output appears with
  an `Error:` prefix in the transcript — it is the deny reason, not a
  failure).
- **Fail-open snapshot / fail-closed rollback**.
- **Discoverability**: SessionStart `additionalContext` + installed skill so
  the agent knows the commands.

### Command set (v2)

```
cubesandbox-rollback [last|<N>|<snapshot-id>]   restore any prior snapshot
cubesandbox-rollback list                       show snapshot history
cubesandbox-rollback checkpoint <name>          explicit milestone, exempt from eviction
cubesandbox-rollback undo                       revert the last rollback
cubesandbox-rollback drop <last|N|snapshot-id>  delete snapshots (real backend deletion)
```

- `rollback <N>` restores snapshot N steps back; snapshots after it are
  discarded (the discarded list is printed) but an undo point is kept, so a
  mistaken rollback can be undone.
- **Circular eviction**: automatic snapshots are capped at 30
  (`CUBE_ROLLBACK_MAX_AUTO_SNAPSHOTS`, default 30); the oldest is evicted
  with real backend deletion when the cap is exceeded. The baseline snapshot
  is permanent, and named checkpoints are never evicted.

## Files

`examples/claude-code-integration/rollback/`: 6 hook scripts
(lib / risky / snapshot / rollback / session / poststart) + idempotent
`install_rollback.sh` + `SKILL.md` + README + `.env.example` +
`requirements.txt`, plus unit tests (`test_lib.py`, `test_risky.py`).

```
examples/claude-code-integration/rollback/
├── cubesandbox_lib.py         shared logic: session state, env loading, snapshot list/prune/evict
├── cubesandbox_risky.py       PreToolUse gate: auto-snapshot before risky commands
├── cubesandbox_snapshot.py    explicit snapshot command entry
├── cubesandbox_rollback.py    rollback/checkpoint/undo/drop/list command entry
├── cubesandbox_session.py     SessionStart hook: announce plugin to the agent
├── cubesandbox_poststart.py   PostToolUse hook: sandbox bookkeeping
├── install_rollback.sh        idempotent install/uninstall (project-level only)
├── SKILL.md                   agent-facing skill with command reference
├── README.md                  user documentation
├── .env.example               config template (CUBE_* variables)
├── requirements.txt           python-dotenv dependency
├── test_lib.py                unit tests for shared lib logic
└── test_risky.py              unit tests for the risky-command gate
```

Project-level install only — never touches user-level
`~/.claude/settings.json`

## Verification

Tested end-to-end on a bare-metal CubeSandbox v0.6.0 deployment
(11 services, KVM) with Claude Code 2.1.226. One continuous session
(snapshots cross-checked against CubeAPI):

1. `echo v1 > /tmp/demo.txt` → sandbox has v1 (host-mount warning 130400
  shown, command still runs — see Known limitations #1)
2. `cubesandbox-rollback checkpoint milestone1` → named checkpoint created
3. `echo v2 > /tmp/demo.txt` → sandbox has v2; auto pre-snapshot taken
4. `cubesandbox-rollback list` → 4 snapshots: baseline + 2 checkpoints + auto
5. `cubesandbox-rollback 2` → restores the checkpoint state (v1), prints the
  discarded auto snapshot
6. `cat /tmp/demo.txt` → v1, rollback confirmed
7. `cubesandbox-rollback undo` → restores the pre-rollback state (v2)
8. `cat /tmp/demo.txt` → v2, undo confirmed
9. `cubesandbox-rollback list` → 3 snapshots (undo point consumed)
10. `cubesandbox-rollback drop 1` → deletes a snapshot by index, confirmed
  via CubeAPI
11. `cubesandbox-rollback list` → 2 snapshots remain (baseline + checkpoint)

[SCREENSHOT_1: session start — v1 + checkpoint create]
屏幕截图_20260810_072736

[SCREENSHOT_2: list + rollback 2 + discarded snapshot]
屏幕截图_20260810_072810

[SCREENSHOT_3: undo + list + drop]
屏幕截图_20260810_072833

[SCREENSHOT_4: final list — baseline + checkpoint]
屏幕截图_20260810_073053

Notes from this verification round:

- Env values wrapped in quotes (`CUBE_API_URL='http://...'`) no longer
  break config loading (python-dotenv).
- The intermittent sandbox-domain resolution failures seen mid-verification
  were traced to the bare-metal deploy's CoreDNS config
  (`deploy/one-click/coredns/Corefile.template` lacks an AAAA template,
  causing NXDOMAIN negative caching) — unrelated to this plugin; the
  session continued over direct proxy connection
  (`CUBE_PROXY_NODE_IP`/`CUBE_PROXY_PORT_HTTP`).

## Known limitations (verified during testing)

1. **Host-mount and snapshot are mutually exclusive** (platform constraint,
  Cubelet `validateCommitSandboxTarget`): adding the project dir to
  `allowed_host_mount_prefixes` breaks snapshot/rollback with error 130400.
  Keep the default prefix list — no host mount, snapshots work.
2. **Rollback on long-active sandboxes can hit a TAP `EBUSY` race** (VM
  restart rebuilds virtio-net devices; related fix in v0.3.1 changelog
  #442). Retry after a few seconds or start a fresh session.

Depends on the #765 Claude Code hooks.

Assisted-by: OPENCODE:deepseek-v4-flash

Signed-off-by: Mizup79 <1321709320a@gmail.com>

### DIFF
--- a/examples/claude-code-integration/rollback/.env.example
+++ b/examples/claude-code-integration/rollback/.env.example
@@ -3,5 +3,10 @@
 CUBE_ROLLBACK_SAFE=echo,cat,ls,pwd,cd,export,env,which,whoami
 # Redirects to /dev/null and &1/&2 are automatically excluded from risk
 
-# Uses same API endpoint and template as #765 (from ../hooks/.env)
-# CUBE_API_URL and CUBE_TEMPLATE_ID are read from the parent hook's .env
+# Max auto snapshots kept per session (ring buffer; oldest evicted from the
+# backend when exceeded). Baseline and explicit checkpoints are never evicted.
+CUBE_ROLLBACK_MAX_AUTO_SNAPSHOTS=30
+
+# Uses the same API endpoint and template as cubesandbox-hook. CUBE_API_URL
+# and CUBE_TEMPLATE_ID are read from `~/.claude/hooks/cubesandbox.env` at hook
+# startup (cubesandbox-hook's env file), or from the shell environment
--- a/examples/claude-code-integration/rollback/.env.example
+++ b/examples/claude-code-integration/rollback/.env.example
@@ -1,0 +1,7 @@
+# Rollback plugin configuration
+# Safe command whitelist (comma-separated) — these are NEVER snapshotted
+CUBE_ROLLBACK_SAFE=echo,cat,ls,pwd,cd,export,env,which,whoami
+# Redirects to /dev/null and &1/&2 are automatically excluded from risk
+
+# Uses same API endpoint and template as #765 (from ../hooks/.env)
+# CUBE_API_URL and CUBE_TEMPLATE_ID are read from the parent hook's .env
--- a/examples/claude-code-integration/rollback/README.md
+++ b/examples/claude-code-integration/rollback/README.md
@@ -1,0 +1,150 @@
+# CubeSandbox Rollback for Claude Code
+
+Snapshot and rollback protection for Claude Code sessions that run inside
+CubeSandbox MicroVMs. Before risky commands a sandbox snapshot is taken
+automatically; if the command breaks the environment, the agent runs
+`cubesandbox-rollback last` to restore the previous filesystem state. This
+example is a companion layer to the CubeSandbox Claude Code hooks
+(`examples/claude-code-integration/hooks/`, below: **#765**), which redirect
+every Bash tool call into an isolated sandbox.
+
+## Why
+
+Claude Code's `/rewind` restores the conversation and file edits but not Bash
+side effects in a sandbox: a destructive `rm`, `git reset`, or `npm install`
+cannot be unwound. This plugin snapshots before risky commands and lets the
+agent roll back to the last good state on demand.
+
+## Architecture
+
+- **Companion layer to #765.** It never imports #765 modules and never writes
+  #765's state. It only *reads* `~/.cache/cubesandbox-hook/<sha256(session_id)>.json`
+  (fields `sandbox_id`, `mount`, `state_token`) to find the session's sandbox.
+  If #765 is absent or the sandbox does not exist yet, the hooks degrade
+  gracefully (skip / no-op). The digest function is duplicated to match #765
+  exactly, so both plugins address the same session.
+- **Parallel-hook safety.** Claude Code runs multiple PreToolUse hooks in
+  parallel; each receives the original tool input, `deny` has highest
+  precedence, and `updatedInput` is not seen by other hooks. Our hooks return
+  only `{}` (no opinion) or `deny` — never `updatedInput` — so they coexist
+  with #765's rewrite hook without fighting over the command.
+- **Deny channel.** Verified with live sessions: Claude Code renders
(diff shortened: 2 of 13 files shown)


fslongjin:
Nice work! I’ll get to this PR after handling #765. At that point, I might need to ask you to rebase your branch, or we can look into integrating your changes into a future example.

Mizup79:
Thanks for the quick response! Both options work for me.

Since my branch is based on master and the rollback/ directory is self-contained, a rebase will be clean and conflict-free whenever you're ready. I'm also happy with integrating into the #765 example if that's easier for you.

Meanwhile I'll keep improving the example — I'm currently addressing the review bot's findings (the risk-classifier false negatives, backend snapshot cleanup for drop, etc.) and will update this PR accordingly.

Issue #992, opened 2026-07-16
[Bug Report] Post-rollback data-plane requests can stall on stale CubeProxy upstream connections


mon3stera:
## Summary

After rolling a running sandbox back to a snapshot, the first data-plane request may intermittently stall for several minutes after rollback. 

This is reproducible with `sandbox.files.read()`. Evidence indicates that CubeProxy may reuse an upstream keep-alive connection whose TCP state is newer than the TCP state restored inside the guest.

## Environment

  - CubeSandbox version: `v0.5.1-amd64`
  - Git commit: `a164417f497234a0d787cb328b0ae96480b1569b`
  - Python SDK: `0.5.0`
  - Python: `3.14.6`
  - Host OS: Arch Linux
  - Kernel: `7.1.3-2-cachyos`
  - Architecture: `x86_64`
  - KVM: VT-x, `kvm.ko.zst`
  - Deployment mode: single-node one-click deployment
  - Relevant components: CubeProxy, CubeShim, Python SDK rollback/data-plane handling

## Steps to Reproduce

1. Use any ready template that includes the standard envd service.

2. Set the normal SDK environment variables required by the deployment:

```bash
export CUBE_TEMPLATE_ID=<ready-template-id>
export CUBE_PROXY_NODE_IP=<cube-proxy-node-ip>  # if local DNS is not used
```

3. Run the following script:

```python
import os
from contextlib import suppress
from queue import Queue
from threading import Thread

(FILE_PATH, BASELINE) = ("/tmp/rollback-repro.txt", "before snapshot\n")
(ROUNDS, READ_TIMEOUT) = (
    int(os.getenv("CUBE_REPRO_ROUNDS", "100")),
    float(os.getenv("CUBE_REPRO_READ_TIMEOUT", "10")),
)


def read_with_timeout(sandbox, path, timeout):
    outcome = Queue(maxsize=1)

    def read():
        try:
            outcome.put((sandbox.files.read(path), None))
        except BaseException as error:
            outcome.put((None, error))

    thread = Thread(target=read, daemon=True)
    thread.start()
    thread.join(timeout)

    if thread.is_alive():
        return None

    (content, error) = outcome.get_nowait()
    if error is not None:
        raise error

    return content


def main():
    from cubesandbox import Sandbox

    template_id = os.getenv("CUBE_TEMPLATE_ID")
    if not template_id:
        raise SystemExit("CUBE_TEMPLATE_ID is required")

[...]

PR #1252, opened 2026-08-01
fix(cubeproxy): invalidate deleted sandbox route cache

zyl1121:
## Motivation

CubeProxy should stop using stale local routes after a sandbox is deleted and expose a clear response when Redis confirms that no route metadata exists.

Scanning the entire shared cache during invalidation makes lifecycle operations proportional to the global cache size and blocks the Nginx worker performing the scan. After invalidation, a confirmed route miss and a Redis lookup failure also both return HTTP 503, so clients cannot distinguish missing metadata from a temporary metadata-store failure.

## What Changed

* Replace shared-cache scans with bounded invalidation of `meta_cached` and the fixed route metadata mirrored from Redis.
* Invalidate `meta_cached` first so new data-plane requests reload route metadata instead of immediately using stale per-port entries.
* Reuse the bounded helper for lifecycle deletion and explicit backend-cache invalidation without enumerating dynamic per-port keys.
* Return HTTP 404 when both supported Redis route lookups complete successfully and neither contains route metadata.
* Preserve HTTP 503 when any Redis route lookup fails and no valid route is found.
* Log confirmed route misses at warning level and add focused regression coverage.

## Validation

* `make -C CubeProxy test` passed with the target OpenResty LuaJIT runtime.
* Regression coverage confirms that invalidation never calls `get_keys()`, removes the cache-hit sentinel before the fixed route keys, leaves dynamic per-port entries to TTL cleanup, and does not affect other sandboxes.
* Route lookup coverage confirms HTTP 404 for authoritative misses and HTTP 503 for Redis failures and mixed miss/error outcomes.

## Scope

This PR keeps cache invalidation bounded to known per-sandbox keys. It intentionally does not scan or eagerly delete dynamic per-port entries; those entries cannot form an immediate cache hit after `meta_cached` is removed and remain subject to their existing TTL.

This PR does not add negative caching, change the lifecycle event protocol, or modify Redis route publication ordering. Requests for nonexistent sandbox IDs continue to perform the same Redis lookups as before.

## Note: E2B Compatibility

Official E2B's data-plane proxy returns HTTP 502 when the requested sandbox cannot be found, while this PR currently returns HTTP 404 for a confirmed missing sandbox route.

Maintainer confirmation is needed on whether CubeProxy should keep HTTP 404 or use HTTP 502 for closer E2B data-plane compatibility. Redis lookup failures remain HTTP 503 in either case.

### DIFF
--- a/CubeProxy/lua/admin_phase.lua
+++ b/CubeProxy/lua/admin_phase.lua
@@ -106,20 +106,19 @@ local function handle_meta_delete()
     local sid, e2 = require_string(obj, "sandbox_id")
     if not sid then return reply_error(ngx.HTTP_BAD_REQUEST, e2) end
 
+    -- Invalidate fixed route metadata so the next data-plane request reloads
+    -- the sandbox route from Redis.
+    local deleted = backend_cache.invalidate_sandbox(sid)
     META:delete(sid)
     STATE:delete(sid)
     LAST:delete(sid)
-    -- Drop stale SandboxIP / host-port routing cache too; otherwise a later
-    -- recreate/resume with the same sandbox_id can keep serving the old
-    -- backend until TTL expires (and hits renew TTL).
-    local deleted = backend_cache.delete_sandbox(sid)
     return reply(ngx.HTTP_OK, { ok = true, backend_cache_deleted = deleted })
 end
 
 -- POST /admin/backend_cache/delete
 --   body: {"sandbox_id": "..."}
---   semantics: drop ngx.shared.local_cache routing entries for this sandbox
---              ({sid}:meta_cached, {sid}:{port}:backend_*, Redis field mirrors).
+--   semantics: invalidate fixed ngx.shared.local_cache route metadata for this
+--              sandbox without scanning dynamic per-port entries.
 --              Used by CubeMaster after Resume rewrites Redis SandboxIP/ports.
 local function handle_backend_cache_delete()
     local obj, err = read_json_body()
@@ -127,7 +126,7 @@ local function handle_backend_cache_delete()
     local sid, e2 = require_string(obj, "sandbox_id")
     if not sid then return reply_error(ngx.HTTP_BAD_REQUEST, e2) end
 
-    local deleted = backend_cache.delete_sandbox(sid)
+    local deleted = backend_cache.invalidate_sandbox(sid)
     return reply(ngx.HTTP_OK, { ok = true, deleted = deleted })
 end
 
--- a/CubeProxy/lua/sandbox_backend.lua
+++ b/CubeProxy/lua/sandbox_backend.lua
@@ -111,11 +111,6 @@ local function load_sandbox_proxy_metadata(ins_id)
             last_err = err
         elseif value and #value > 0 then
             return value, nil
-        else
-            -- This Redis command succeeded but the key is empty/missing. Clear
-            -- any previous key's transport error so the final result is a
-            -- truthful "not found" instead of a stale connectivity error.
-            last_err = nil
         end
     end
 
@@ -123,8 +118,7 @@ local function load_sandbox_proxy_metadata(ins_id)
         return nil, string.format("request %s using keys for %s get redis err: %s",
             ngx.var.http_x_cube_request_id, ins_id, last_err)
     end
-    return nil, string.format("request %s using keys for %s get redis nil",
-        ngx.var.http_x_cube_request_id, ins_id)
+    return nil, nil
 end
 
 --[[
@@ -177,6 +171,12 @@ function _M.resolve_backend(ins_id, container_port)
     if err then
         ngx.log(ngx.ERR, "LEVEL_ERROR||", err)
         utils:respond_unavailable()
+    end
+    if not metadata then
+        ngx.log(ngx.WARN, "LEVEL_WARN||",
+            string.format("request %s using instance %s has no Redis route",
+                ngx.var.http_x_cube_request_id, ins_id))
+        utils:respond_not_found()
     end
 
     cache:set(ins_id .. ":meta_cached", "1", timeout)
--- a/CubeProxy/Makefile
+++ b/CubeProxy/Makefile
@@ -50,11 +50,13 @@ test:
 	@if command -v lua >/dev/null 2>&1; then \
 		lua tests/redis_iresty_sentinel_test.lua; \
 		lua tests/backend_cache_test.lua; \
+		lua tests/sandbox_route_lookup_test.lua; \
 	elif command -v luajit >/dev/null 2>&1; then \
 		luajit tests/redis_iresty_sentinel_test.lua; \
 		luajit tests/backend_cache_test.lua; \
+		luajit tests/sandbox_route_lookup_test.lua; \
 	else \
-		echo "skip lua unit tests (no lua/luajit): redis_iresty_sentinel_test.lua backend_cache_test.lua"; \
+		echo "skip lua unit tests (no lua/luajit): redis_iresty_sentinel_test.lua backend_cache_test.lua sandbox_route_lookup_test.lua"; \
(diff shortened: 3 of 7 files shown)


chenhengqi:
Thanks, we do have to invalidate route cache for use-case like cross-node pause/resume.

chenhengqi:
> For clients, these responses incorrectly suggest an infrastructure or backend failure even when CubeProxy has confirmed that no route exists.

I am curious about the motivation here.
* A client can check sandbox existence via control plane API.
* There are security concerns like every request of non-existence sandbox now hit Redis.

zyl1121:
> I am curious about the motivation here. A client can check sandbox existence via control plane API.

For example, an existing sandbox may be removed asynchronously when an agent operation outlives the configured auto-kill timeout. The client may still hold the sandbox’s data-plane endpoint and continue sending command, file, or service requests without checking the control plane before each request.

The problem today is that CubeProxy may return 504 while a stale local route still exists, and 503 after the route-cache TTL expires. Both look like retryable infrastructure failures, even though the sandbox no longer exists.

Once CubeProxy receives the lifecycle delete notification, invalidating the stale route and returning 404 makes that terminal state explicit.

> There are security concerns like every request of non-existence sandbox now hit Redis.

On the security concern: nonexistent sandbox requests already query the  Redis route keys today and return 503. This change only changes the response for a confirmed miss from 503 to 404, negative caching or rate limiting for arbitrary IDs would be a separate concern.

CubeProxy/lua/admin_phase.lua
```diff
@@ -97,14 +110,37 @@ end
...
 
+    local timeout_max = tonumber(ngx.var.timeout_max)
+    if not timeout_max then
+        return reply_error(ngx.HTTP_INTERNAL_SERVER_ERROR,
+            "timeout_max is not configured")
+    end
+    -- Outlive the longest route-cache TTL so dynamic per-port entries can age
+    -- out without a shared-dict scan. The grace period also covers a request
+    -- that entered the cache-hit path just before this delete notification.
+    local tombstone_ttl = timeout_max + CACHE_TOMBSTONE_GRACE_SECONDS
```
chenhengqi: I don't understand why a tombstone is needed. Deleting the HostIP key already invalidates cache.

zyl1121: Cache hits use the per-port `backend_ip` / `backend_port` entries plus `meta_cached`, so I think the key invalidation is probably `meta_cached` rather than `HostIP`, and deleting `meta_cached` does handle the normal case.

The issue is that deleting `meta_cached` only invalidates the cache at that moment. A concurrent cache-hit request may have already read the stale route and can refresh `meta_cached` and the per-port entries after deletion.

The tombstone is used to prevent that stale route from becoming valid again. It keeps those dynamic entries unusable until their TTL expires, without scanning the shared dict to find and delete every per-port key, which would be expensive.

chenhengqi: I consider this is kind of overdesigned.

As you said, `A concurrent cache-hit request may have already read the stale route` will cause a 504 instead of 404. I think the tombstone does not handle this well. That is to say, there is always a race window.

zyl1121: The race I’m trying to cover is: a request reads the cached route, the delete handler clears the cache, and that request then writes the stale route back. The tombstone cannot change that in-flight request, but it prevents subsequent requests from using the refreshed stale entry.

The chance of this race is very low. If you consider it not worth guarding against, I can remove the tombstone and keep the simpler fixed-key invalidation.

chenhengqi: Up to you. I tend to keep the diff minimal.

zyl1121: Makes sense. I’ll keep the change minimal and remove the tombstone. The delete handler will clear `meta_cached` together with the known fixed route-cache keys.

Separately, for a confirmed route miss / sandbox-not-found case, would you prefer aligning with E2B’s `502`, or using `404` for explicit not-found semantics?

chenhengqi: For status code, I have no preference myself. But if we reset stale connections from CubeVS, the status code would be 502. See #992.

chenhengqi:
@zyl1121 Sorry for the delay. Please rebase and resolve conflict.

zyl1121:
> Sorry for the delay. Please rebase and resolve conflict.

Rebased and resolved the conflict. I also removed the delete-side cache invalidation from this PR, since #1322 now covers it through the shared backend-cache helper.

The PR currently returns a 404 when no sandbox route exists, while Redis lookup failures remain 503. I’m considering changing a confirmed route miss to a 502, which would align with E2B and with the expected behavior when CubeVS resets a stale upstream connection, as discussed in #992. Which behavior would you prefer?

chenhengqi:
@zyl1121 Do you have AI Agent update this PR automatically? It seems like it refer to #1322 which clean route cache in a bad way(Scan the entire cache in a loop).

zyl1121:
> Do you have AI Agent update this PR automatically? It seems like it refer to #1322 which clean route cache in a bad way(Scan the entire cache in a loop).

No, it wasn’t an automatic update. While rebasing, I saw that the delete-side invalidation was already in the merged code, so I checked the discussion in #1322.

That discussion initially pointed to reusing the scoped invalidation from this PR, but #1322 ultimately merged with `backend_cache.delete_sandbox()` handling the deletion path. I thought that means the merged implementation had superseded the overlapping logic here, so I removed it during the rebase and noted that change in my rebase update.

Looks like I misread the intent of that merge. Since the helper is now also shared with the cross-node pause/resume path, should I update the shared helper itself to use scoped invalidation, or only restore it for the delete path here?

chenhengqi:
I think we can have a single `handle_meta_delete()` for both sandbox deletion and pause/resume(cross-node or not). What do you think?

zyl1121:
> I think we can have a single `handle_meta_delete()` for both sandbox deletion and pause/resume(cross-node or not). What do you think?

Yes, that makes sense. I’ve updated it that way: both request paths now dispatch to the same `handle_meta_delete()`, and `handle_backend_cache_delete()` has been removed. The route-cache invalidation remains in the shared `backend_cache.invalidate_sandbox()` helper.

chenhengqi (approved):
LGTM. Thanks.

CubeProxy/lua/sandbox_backend.lua
zhouxianping: When the sandbox listens on multiple container ports, accessing the first port will result in a cache miss and set a cache in meta_cached. Subsequent accesses to other ports will use the old cache information because meta_cached already has a cache, and the old caches of backend_ip and backend_port have not been deleted.

CubeProxy/lua/sandbox_backend.lua
```diff
@@ -178,6 +172,12 @@ function _M.resolve_backend(ins_id, container_port)
...
         utils:respond_unavailable()
     end
+    if not metadata then
+        ngx.log(ngx.WARN, "LEVEL_WARN||",
+            string.format("request %s using instance %s has no Redis route",
+                ngx.var.http_x_cube_request_id, ins_id))
+        utils:respond_not_found()
+    end
 
     cache:set(ins_id .. ":meta_cached", "1", timeout)
```
zhouxianping: meta_cached should not be written immediately. Instead, it should be written after other data has been written to the cache. This ensures that concurrent worker requests do not have a short window where meta_cached's cache value is 1, while other data's cache is in a state of being about to be written but has not yet been written.

CubeProxy/lua/sandbox_backend.lua
zhouxianping: The meta_cached cache should be read first, and then the cache of other data should be read to ensure that the cache of other data is valid when meta_cached is 1.

PR #1253, opened 2026-08-01
fix(templatecenter): fix snapshot-create request ID idempotency

### LINKED ISSUE #1105: [Bug Report] Snapshot-create request_id idempotency never matches on retry: stored spec fingerprint is not reproducible
## Summary

`SubmitSandboxSnapshot`'s request_id dedup ("reuse existing job on duplicate request payload; reject on payload drift", introduced for snapshots in #360 following the #336 pattern) can never take the reuse branch for a real retry: a client that resends the identical snapshot-create request with the same `request_id` always gets `ErrTemplateAttemptInProgress` ("payload does not match"). The stored `SpecFingerprint` is unreproducible by construction, for two independent reasons.

## Mechanism

**Layer 1 — the fingerprint is computed after stamping the generated snapshot id.** `snapshot_ops.go:165-168`: `generateSnapshotID()` -> stamped into `storedReq.Annotations[cube.master.appsnapshot.template.id]` -> `buildCommitTemplateSpecFingerprint(storedReq)`. The hash covers a value a retry cannot know.

**Layer 2 — rebuilding the spec is nondeterministic anyway.** The retry path re-derives the spec via `buildSnapshotRequests(originReq, "")`, which unconditionally overwrites the template-id annotation with the empty parameter (`snapshot_ops.go:835`); `NormalizeRequest` then generates a **random** `tpl-` id for the empty annotation and stamps it into the hashed clone (`store.go:329-341`). Two rebuilds of the same input hash differently.

The legacy escapes in `snapshotCreateRequestMatches` (`raw == ""` / `SpecFingerprint == ""`) are unreachable for rows written by current code, and jobs are looked up by request_id regardless of status — so both in-flight re-attach and post-completion result recovery land on the fingerprint check and fail.

## Reproduction

Drop this into `CubeMaster/pkg/templatecenter` and run — both subtests, asserting the documented contract against the real functions, fail on master:

```go
package templatecenter

import (
	"testing"

	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
	sandboxtypes "github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/service/sandbox/types"
)

func TestSnapshotCreateRequestIDIdempotency(t *testing.T) {
	build := func() *sandboxtypes.CreateCubeSandboxReq {
		_, storedReq, err := buildSnapshotRequests(&sandboxtypes.CreateCubeSandboxReq{
			Request:      &sandboxtypes.Request{RequestID: "req-1"},
			InstanceType: "cubebox",
		}, "")
		if err != nil {
			t.Fatalf("buildSnapshotRequests: %v", err)
		}
		return storedReq
	}

	t.Run("rebuild is deterministic", func(t *testing.T) {
		if buildCommitTemplateSpecFingerprint(build()) != buildCommitTemplateSpecFingerprint(build()) {
			t.Fatal("two identical rebuilds produced different fingerprints")
		}
	})

	t.Run("identical retry matches stored job", func(t *testing.T) {
		// Store side, exactly as SubmitSandboxSnapshot does it: stamp the
		// generated snapshot id, then fingerprint.
		submitted := build()
		submitted.Annotations[constants.CubeAnnotationAppSnapshotTemplateID] = "snap-generated"
		raw, err := marshalSnapshotCreateRequest(snapshotCreateJobRequest{
			RequestID:       "req-1",
			SandboxID:       "sb-1",
			SnapshotID:      "snap-generated",
			NodeID:          "node-1",
			NodeIP:          "10.0.0.1",
			SpecFingerprint: buildCommitTemplateSpecFingerprint(submitted),
		})
		if err != nil {
			t.Fatalf("marshal: %v", err)
		}
		if !snapshotCreateRequestMatches(raw, "req-1", "sb-1", "node-1", "10.0.0.1", "", build()) {
			t.Fatal("byte-identical retry did not match the stored job")
		}
	})
}
```

```
--- FAIL: TestSnapshotCreateRequestIDIdempotency/rebuild_is_deterministic
--- FAIL: TestSnapshotCreateRequestIDIdempotency/identical_retry_matches_stored_job
```

(The package's existing dedup tests don't catch this because they patch `snapshotCreateRequestMatches` out.)

## Impact

Bounded but real: the failure is fail-closed (rejection, no duplicate job, no corruption), and a client that switches to a fresh `request_id` proceeds under the sandbox single-flight guard. What is lost is exactly what the mechanism promises: same-request_id re-attach to an in-flight job, and exactly-once result recovery after completion — a client that timed out and retries can never converge.

Scope: only SNAPSHOT_CREATE is affected. Rollback/delete matchers compare plain fields (no fingerprint). The template COMMIT flow's byte-equality reuse works within `SubmitTemplateCommit`, though its HTTP handler generates a fresh `tpl-` id per arrival — there the "retry should generate a new request id" contract is documented, so arguably by design.

## Suggested fix

Hash the spec with the template-id annotation stripped, on both the store side (`snapshot_ops.go:168`) and the match side (`:1403`) — the generated id is not part of the caller's payload. This restores idempotent matching while still detecting genuine drift; existing rows keep today's behavior (old-style fingerprints keep rejecting, empty-fingerprint legacy rows keep their escape), so no migration is needed. #1101's import flow contains an equivalent helper (`importSpecFingerprint`) for reference.

Becks723:
## Summary

Fixes #1105

`SubmitSandboxSnapshot` previously called `buildSnapshotRequests` before taking the snapshot write lock and passed an empty snapshot ID. `NormalizeRequest` filled that empty value with a random `tpl-*` ID. The first-time path later overwrote it with the generated snapshot ID, but the retry path compared a newly randomized request against the stored fingerprint, so an identical retry was rejected as payload drift.

This PR:

- Moves snapshot request construction inside the snapshot write lock.
- Rebuilds retry requests with the existing job's `TemplateID` before normalization.
- Generates the snapshot ID before request construction on the first-time path and uses that same ID throughout job and definition creation.
- Preserves the existing full-request fingerprint semantics, including the snapshot ID, and leaves the template COMMIT flow unchanged.
- Adds a regression test that exercises the real `SubmitSandboxSnapshot` retry matcher without patching it.

## Impact

An identical snapshot-create retry with the same `request_id` can reattach to the existing job or recover its stored result. Genuine request drift is still rejected, and no fingerprint fields are excluded.

## Test plan

The following tests passed in the repository-provided builder environment:

```bash
make builder-run BUILDER_CMD='cd /workspace/CubeMaster && go test -count=1 -v -run TestSubmitSandboxSnapshot ./pkg/templatecenter/'
make builder-run BUILDER_CMD='cd /workspace/CubeMaster && go test -count=1 -v -run TestBuildSnapshotRequestsUsesSnapshotID ./pkg/templatecenter/'
make builder-run BUILDER_CMD='cd /workspace/CubeMaster && go test -count=1 -v -run TestBuildCommitTemplateSpecFingerprintIgnoresRequestID ./pkg/templatecenter/'
```

`make cubemaster-test` is not fully green locally because of pre-existing ARM64/gomonkey failures that are reproducible on the unmodified `master` branch.

Assisted-by: Codex:GPT-5

### DIFF
--- a/CubeMaster/pkg/templatecenter/snapshot_ops.go
+++ b/CubeMaster/pkg/templatecenter/snapshot_ops.go
@@ -128,10 +128,7 @@ func SubmitSandboxSnapshot(ctx context.Context, requestID, sandboxID, hostID, ho
 	} else {
 		originReq.Request.RequestID = requestID
 	}
-	createReq, storedReq, err := buildSnapshotRequests(originReq, "")
-	if err != nil {
-		return nil, err
-	}
+
 	var jobID string
 	reusedExistingJob := false
 	if err := withSnapshotWriteLocks([]string{
@@ -141,6 +138,10 @@ func SubmitSandboxSnapshot(ctx context.Context, requestID, sandboxID, hostID, ho
 		if existing, err := getTemplateImageJobByRequestID(ctx, requestID); err == nil {
 			if existing.Operation != JobOperationSnapshotCreate {
 				return fmt.Errorf("%w: request %s is already bound to %s", ErrTemplateAttemptInProgress, requestID, existing.Operation)
+			}
+			_, storedReq, err := buildSnapshotRequests(originReq, existing.TemplateID)
+			if err != nil {
+				return err
 			}
 			if !snapshotCreateRequestMatches(existing.RequestJSON, requestID, sandboxID, nodeID, nodeIP, displayName, storedReq) {
 				return fmt.Errorf("%w: request %s payload does not match existing snapshot create job", ErrTemplateAttemptInProgress, requestID)
@@ -163,8 +164,10 @@ func SubmitSandboxSnapshot(ctx context.Context, requestID, sandboxID, hostID, ho
 		}
 
 		snapshotID := generateSnapshotID()
-		createReq.Annotations[constants.CubeAnnotationAppSnapshotTemplateID] = snapshotID
-		storedReq.Annotations[constants.CubeAnnotationAppSnapshotTemplateID] = snapshotID
+		createReq, storedReq, err := buildSnapshotRequests(originReq, snapshotID)
+		if err != nil {
+			return err
+		}
 		fingerprint := buildCommitTemplateSpecFingerprint(storedReq)
 		requestJSON, err := marshalSnapshotCreateRequest(snapshotCreateJobRequest{
 			RequestID:       requestID,
--- a/CubeMaster/pkg/templatecenter/snapshot_ops_test.go
+++ b/CubeMaster/pkg/templatecenter/snapshot_ops_test.go
@@ -86,6 +86,68 @@ func TestSubmitSandboxSnapshotReusesExistingRequest(t *testing.T) {
 	}
 }
 
+func TestSubmitSandboxSnapshotRetryMatchesExistingSnapshotID(t *testing.T) {
+	oldDB := store.db
+	store.db = &gorm.DB{}
+	defer func() { store.db = oldDB }()
+
+	_, storedReq, err := buildSnapshotRequests(&sandboxtypes.CreateCubeSandboxReq{
+		Request:      &sandboxtypes.Request{RequestID: "req-existing"},
+		InstanceType: "cubebox",
+		Annotations:  map[string]string{},
+	}, "snap-existing")
+	if err != nil {
+		t.Fatalf("buildSnapshotRequests returned error: %v", err)
+	}
+	requestJSON, err := marshalSnapshotCreateRequest(snapshotCreateJobRequest{
+		RequestID:       "req-existing",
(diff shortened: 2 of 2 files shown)


fslongjin:
Hello, thanks for your contribution! Could you please attach some screenshots and logs (both are needed) showing the verification process and results (In a real CubeSandbox cluster/instance)  for this PR? Thank you!

lkml-likexu:
@Becks723 Thanks for contribution. Any e2e test or current test results to cover this bug ?

luzhixing12345:
@Becks723 Hi, I tested the reproduction code mentioned in https://github.com/TencentCloud/CubeSandbox/issues/1105#issue-4956223873, and this fix successfully resolves the issue. Please rebase onto the latest code.

PR #1254, opened 2026-08-01
cubeapi: accept e2b flattened autoPause/autoResume on sandbox create

### LINKED ISSUE #965: [Bug Report]CubeAPI should accept E2B lifecycle wire fields for auto-pause
## Summary

CubeAPI currently accepts the SDK-facing lifecycle object:

```json
{
  "lifecycle": {
    "onTimeout": "pause",
    "autoResume": true
  }
}
```

However, the E2B Python SDK does not send this object directly on the wire. When users call:

```python
Sandbox.create(
    template=template_id,
    timeout=30,
    lifecycle={"on_timeout": "pause", "auto_resume": True},
)
```

E2B converts it into top-level fields before calling `POST /sandboxes`:

```json
{
  "templateID": "<template-id>",
  "timeout": 30,
  "autoPause": true,
  "autoResume": {"enabled": true}
}
```

CubeAPI's `NewSandbox` model currently does not accept `autoPause` or `autoResume`, so serde ignores those unknown fields. As a result, CubeAPI treats the request as if lifecycle was absent and forwards `auto_pause=false` / `auto_resume=false` to CubeMaster.

## Impact

E2B SDK users can request auto-pause successfully at the Python API layer, but CubeSandbox-compatible CubeAPI ignores the setting. A sandbox created with `timeout=30` and `lifecycle={"on_timeout": "pause"}` is deleted after the timeout instead of entering the `paused` state.

This also causes SDK compatibility E2E auto-lifecycle tests to skip or fail because the sandbox never reaches `paused`.

## Evidence

In E2B Python SDK:

- `Sandbox.create(..., lifecycle=...)` is the public user API.
- Internally, `SandboxApi._create_sandbox()` maps `lifecycle["on_timeout"] == "pause"` to `NewSandbox(auto_pause=True)`.
- It maps `lifecycle["auto_resume"]` to `NewSandbox.auto_resume = SandboxAutoResumeConfig(enabled=...)`.
- The generated request body serializes these as `autoPause` and `autoResume`.

In CubeAPI:

- `NewSandbox` accepts `lifecycle: Option<SandboxLifecycleConfig>`.
- `SandboxService::create_sandbox()` derives CubeMaster flags only from `body.lifecycle`.
- Unknown top-level `autoPause` and `autoResume` fields are ignored.

## Expected behavior

CubeAPI should accept both shapes:

1. Current CubeSandbox SDK shape:

```json
{"lifecycle": {"onTimeout": "pause", "autoResume": true}}
```

2. E2B SDK wire shape:

```json
{"autoPause": true, "autoResume": {"enabled": true}}
```

Both should result in CubeAPI forwarding:

```json
{"auto_pause": true, "auto_resume": true}
```

to CubeMaster.

## Suggested fix

Add E2B-compatible fields to CubeAPI's create request model, for example:

```rust
#[serde(rename = "autoPause", default)]
pub auto_pause: Option<bool>,

#[serde(rename = "autoResume", default)]
pub auto_resume: Option<AutoResumeConfig>,
```

Then update the lifecycle translation logic:

- Prefer the explicit `lifecycle` object when present.
- Otherwise fall back to top-level `autoPause` and `autoResume.enabled`.
- Add unit tests for both payload shapes.

## Test case

A compatibility test should create a sandbox through the E2B SDK with:

```python
Sandbox.create(
    template="<template-id>",
    timeout=30,
    lifecycle={"on_timeout": "pause", "auto_resume": False},
)
```

After the idle timeout, the sandbox should become `paused`, not disappear from `GET /sandboxes/{sandbox_id}`.

Arpit-Ahuja-293:
## Problem

The e2b Python SDK does not put `lifecycle` on the wire. `Sandbox.create(lifecycle={"on_timeout": "pause", "auto_resume": True})` is flattened client-side into top-level fields before `POST /sandboxes`:

```json
{ "templateID": "<id>", "timeout": 30, "autoPause": true, "autoResume": { "enabled": true } }
```

`NewSandbox` bound neither field, so serde dropped them silently. CubeAPI then resolved `lifecycle = None` and forwarded `auto_pause=false` to CubeMaster — the sandbox was **killed** at timeout instead of paused. The setting appeared to succeed at the SDK layer and did nothing.

## Fix

- `NewSandbox` binds `autoPause` and `autoResume`.
- New `SandboxAutoResume` untagged enum accepts both `{"enabled": true}` (current SDK) and bare `true` (older releases, hand-rolled clients) — same meaning, so no reason to 400 on shape.
- Lifecycle derivation extracted from the inline closure in `create_sandbox` into `resolve_lifecycle_flags()`, a pure function that unit tests can reach.

## Design decision worth reviewing

**When a request carries both `lifecycle` and the flattened fields, the nested `lifecycle` object wins.**

Reasoning: the e2b SDK never sends both, so this only disambiguates direct API callers — and an explicit `lifecycle` can only come from someone who meant it. Happy to invert this if maintainers prefer the flat fields taking precedence.

## Tests

7 lifecycle tests, 5 of them new, written test-first:

- `new_sandbox_deserializes_e2b_flat_lifecycle_fields` — the exact e2b wire body
- `new_sandbox_accepts_bare_bool_auto_resume`
- `flat_lifecycle_fields_enable_pause_and_resume`
- `nested_lifecycle_wins_over_flat_lifecycle_fields`
- `absent_lifecycle_fields_keep_kill_behaviour` — no behaviour change for callers that don't opt in

## Verification

Run in the project builder image (`rustc 1.89`, `aarch64-unknown-linux-musl`):

- `make cube-api-test` → 108 passed, 0 failed
- `make cubeapi` → clean release build
- `cargo fmt --check` → clean
- `cargo clippy --all-targets` → warning set identical to the pre-change baseline; none in new code
- `cargo build --locked` passes — no dependency changes, `Cargo.lock` untouched

## Also updated

- `openapi.yml` — regenerated via `cube-api --export-openapi`, applying only the additive schema hunks so the hand-maintained license header and `servers:` block survive.
- `docs/guide/lifecycle.md` and `docs/zh/guide/lifecycle.md` — new "Wire format" section covering both accepted shapes and the precedence rule, EN and ZH in sync.

Closes #965

### DIFF
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -138,6 +138,25 @@ pub struct SandboxLifecycleConfig {
     pub auto_resume: bool,
 }
 
+/// Wire shape of the top-level `autoResume` field the e2b SDK sends. Current
+/// SDK releases send `{"enabled": true}`; older ones and hand-rolled clients
+/// send a bare `true`. Both mean the same thing, so accept either rather than
+/// rejecting on shape.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize, ToSchema)]
+#[serde(untagged)]
+pub enum SandboxAutoResume {
+    Enabled(bool),
+    Config { enabled: bool },
+}
+
+impl SandboxAutoResume {
+    pub fn enabled(&self) -> bool {
+        match self {
+            Self::Enabled(enabled) | Self::Config { enabled } => *enabled,
+        }
+    }
+}
+
 #[derive(Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize, ToSchema)]
 #[serde(rename_all = "lowercase")]
 pub enum SandboxOnTimeout {
@@ -190,6 +209,17 @@ pub struct NewSandbox {
     /// (None) means today's behaviour: idle sandboxes are killed.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub lifecycle: Option<SandboxLifecycleConfig>,
+
+    /// e2b SDK compatibility: the SDK flattens its user-facing `lifecycle`
+    /// object into top-level `autoPause` / `autoResume` before it hits the
+    /// wire, so `lifecycle` never arrives from an SDK caller. Ignored when
+    /// `lifecycle` is present.
+    #[serde(rename = "autoPause", alias = "auto_pause", default)]
+    pub auto_pause: Option<bool>,
+
+    /// See `auto_pause`. Only meaningful alongside `autoPause = true`.
+    #[serde(rename = "autoResume", alias = "auto_resume", default)]
+    pub auto_resume: Option<SandboxAutoResume>,
 
     #[serde(skip_serializing_if = "Option::is_none")]
     pub secure: Option<bool>,
--- a/CubeAPI/src/services/sandboxes.rs
+++ b/CubeAPI/src/services/sandboxes.rs
@@ -1675,22 +1675,17 @@ mod tests {
     /// The inbound API mirrors the e2b `lifecycle` object (camelCase nested
     /// struct). CubeAPI then translates it to the two CubeMaster-side bools
     /// when constructing the create-sandbox RPC. Verify the translation
-    /// covers each meaningful combination.
+    /// covers each meaningful combination via the real resolver.
     #[test]
     fn lifecycle_object_translates_to_cubemaster_bools() {
-        use crate::models::{NewSandbox, SandboxOnTimeout};
+        use crate::models::NewSandbox;
 
-        // Helper that mimics services::create_sandbox's lifecycle decoding.
-        fn translate(body: &NewSandbox) -> (bool, bool) {
-            body.lifecycle
-                .as_ref()
-                .map(|lc| {
-                    (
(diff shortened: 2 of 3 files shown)


luzhixing12345:
Thanks for your contribution.
LGTM. Some local tests have passed.

luzhixing12345:
please squash all commits into a single commit. @Arpit-Ahuja-293

docs/guide/lifecycle.md
```diff
@@ -169,6 +169,16 @@ sandbox = Sandbox.create(
 - **`auto_resume=True`**: when any request next arrives for a `paused` sandbox (HTTP, `run_code`, file I/O, …), the platform wakes it up before the request lands. Callers never see the pause; typical resume latency is sub-second to a few seconds.
 - If `auto_resume=False` (or unset), the sandbox stays paused until you explicitly `Sandbox.connect(sandbox_id=...)`. Useful for "wait for the user" workflows.
 
+### Wire format
```
chenhengqi: I think there is no need to update the docs. This is implementation detail, not user-facing APIs.

luzhixing12345:
@Arpit-Ahuja-293 please remove the lifecycle doc change commit

fslongjin (approved):

PR #1255, opened 2026-08-01
cubeops,web: WebUI web terminal for running sandboxes (#643)

naivehath:
## What this does

Adds a **web terminal** to the WebUI: an interactive shell inside a running sandbox, opened from the dashboard with no SSH, SDK or CLI.

Refs #643

## Approach

Terminals are an operations capability, so the endpoint lives in **CubeOps** and reuses the **envd `process.Process` PTY API** that the Python/Node/Go SDKs already drive through cube-proxy. **No new RPC is added to CubeMaster, Cubelet, CubeShim or the in-guest agent.**

```
Browser (xterm.js)
   │  wss://<dashboard>/opsapi/v1/terminal/ws?ticket=<one-time ticket>
   ▼
CubeOps :3010            ← JWT auth, ticket issuance, session registry, audit log
   │  Connect-JSON over HTTP (streaming Start/Connect + unary SendInput/Update/SendSignal)
   ▼
CubeProxy ──► envd :49983 inside the sandbox
```

Two paths were considered and rejected:

- **Browser connecting straight to `/sandbox/<id>/49983/`** — that route has no authentication end to end, so anyone holding a sandbox ID gets a root shell, and the traffic never passes CubeOps so nothing can be audited.
- **Cubelet Exec** — it is fire-and-forget with dummy IO: no streaming output, no attach semantics. It cannot carry an interactive session.

## Authentication

Browsers cannot attach an `Authorization` header to a WebSocket handshake. Putting the login JWT in the query string would leak a long-lived credential into access logs and browser history, so instead:

1. `POST /api/v1/sdk/sandboxes/:id/terminal` — authenticated with the normal login JWT, on the existing authed route group. Verifies the sandbox exists and is `running`, then issues a **one-time ticket**: a 30-second JWT with its own audience (`cubeops:terminal`), consumed on first use.
2. `GET /api/v1/terminal/ws?ticket=…` — the upgrade. The ticket is the only credential, and it is useless against any other endpoint.

## Sessions

- **Multiple terminals per sandbox**, each with its own PTY and PID.
- **Closing the dialog kills the shell** — the client sends an explicit close frame and CubeOps sends `SIGKILL`.
- **A dropped connection does not.** The session registry keeps the PTY, and the UI offers a reconnect that reattaches to the same PID via envd `Connect`, so foreground work survives a network blip.
- **Idle sessions are reaped** after 30 minutes without input, configurable via `terminal_idle_timeout` / `CUBE_OPS_TERMINAL_IDLE_TIMEOUT`.

## Auditing

`requestLogger()` records no identity, so the terminal writes its own lines to the CubeOps log (`/data/log/CubeOps/cubeops-req.log`). These are **real lines captured from the test cluster**, not illustrations — sandbox ID, username, client IP, PID, duration and the reason the session ended:

One uninterrupted stretch of that log, covering two concurrent sessions, a drop-and-reconnect and both ways a session can end (sandbox ID abbreviated for width, everything else verbatim):

```
22:59:40 terminal_session_start sessionID=931fffc8-f529-41ad-ba0c-a4ebc65eaa74 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=182 reconnect=false
23:00:02 terminal_session_start sessionID=75b6cdc4-aac8-435c-b763-1fdabf17ea75 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=192 reconnect=false
23:00:07 terminal_session_end   sessionID=75b6cdc4-aac8-435c-b763-1fdabf17ea75 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=192 reason=detached     durationMs=5007
23:00:08 terminal_session_start sessionID=75b6cdc4-aac8-435c-b763-1fdabf17ea75 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=192 reconnect=true
23:00:11 terminal_session_end   sessionID=931fffc8-f529-41ad-ba0c-a4ebc65eaa74 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=182 reason=pty_exit     durationMs=31174
23:00:11 terminal_session_end   sessionID=75b6cdc4-aac8-435c-b763-1fdabf17ea75 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=192 reason=client_close durationMs=9602
```

Reading it: two terminals run side by side on the same sandbox with distinct PIDs (182, 192); 192's socket drops abnormally at 23:00:07 and the reconnect a second later comes back on the **same session ID and the same PID**, so the shell was never killed; 182 ends because the user typed `exit`, 192 because the dialog was closed.

Separately, the idle sweeper firing on a terminal left open and untouched for ~33 minutes:

```
22:01:24 terminal_session_end sessionID=a82193f4-f5d1-4116-a018-18f579299bd5 sandboxID=fe57331d… username=admin clientIP=127.0.0.1 pid=94 reason=idle_timeout durationMs=1975625
```

`reason` distinguishes `client_close`, `pty_exit`, `detached`, `idle_timeout`, `stream_error` and `reconnect_failed`.

## Deployment

`/opsapi/` forwards no `Upgrade` header and caps reads at 300s, so the WebSocket path gets its own nginx location with the upgrade headers and the same 7206s timeout the `/sandbox/` proxy already uses. `vite.config.ts` sets `ws: true` on the dev proxy for the same reason.

## Testing

**Go — 26 tests in `CubeOps/internal/terminal/`** (32 including table-driven subtests), passing under `-race`, with a fake envd HTTP server that speaks the Connect framing:

- Streaming client: PID extraction, output decoding, the `<port>-<sandboxID>.<domain>` proxy Host, non-zero exit, interrupted transport, and that unary endpoints get plain `application/json` (envd answers 415 to a streaming content-type).
- Tickets: issue/redeem, single use, expiry, foreign audience rejection, and that the consumed-ticket map is pruned.
- WebSocket bridge: end-to-end through `httptest` + a gorilla client — input, resize, ping/pong, the requested initial window size, exit reporting, ticket replay rejection, two independent sessions, reconnect to a detached session, and rejection of a PID the registry does not know.
- Registry: idle reaping, activity deferring it, detached sessions still reaped, and reattach requiring a sandbox+PID match.

**Frontend — vitest + jsdom introduced** (`web/` had no test framework), 11 tests covering the socket state machine: ticket flow, UTF-8-safe base64, reconnect-by-PID, exit vs. drop, and the two races described below.

**Real cluster** — verified against a running sandbox on a single-node KVM cluster (Ubuntu 24.04 arm64, nested virtualisation), driving the browser side of the protocol with a standalone WebSocket client:

```
=== A. open a terminal and run commands
  [PASS] handshake upgraded to 101 — HTTP/1.1 101 Switching Protocols
  [PASS] ready frame with pid — {'type': 'ready', 'pid': 197}
  [PASS] echo command executed
  [PASS] runs as root inside sandbox
  [PASS] sandbox is aarch64
=== B. UTF-8 and ANSI colour
  [PASS] multi-byte UTF-8 round-trips
  [PASS] ANSI escape preserved
=== C. resize propagates to the PTY
  [PASS] stty reports the new size
=== D. ping/pong keepalive
  [PASS] pong received — [{'type': 'pong'}]
=== E. full-screen program (top)
  [PASS] top produced output
=== F. ticket is single-use
  [PASS] replayed ticket rejected — upgrade failed: HTTP/1.1 401 Unauthorized
=== G. second concurrent terminal is independent
  [PASS] second session has a different pid — 197 vs 207
  [PASS] second session runs its own shell
  [PASS] sessions do not cross-talk
=== H. reconnect after an abnormal drop
  [PASS] reattached to the same pid — {'type': 'ready', 'pid': 207}
  [PASS] shell survived the drop
=== I. shell exit is reported
  [PASS] exit frame received — [{'type': 'exit', 'exitCode': 0}]
=== J. explicit close kills the shell

ALL CHECKS PASSED
```

The `top` case covers the issue's `ls` / `top` / `ping` acceptance criterion; the resize case asserts on `stty size` inside the PTY rather than on the UI alone.

Browser verification found two real bugs, both fixed with regression tests:

- **xterm never mounted.** Radix renders portal children on a second pass, so the ref was still null when the mount effect ran, and the effect never re-fired — the dialog was permanently blank. Fixed by tracking the node in state.
- **One dialog leaked two PTYs.** A shared "closing" boolean was reset by a newer connect attempt while an older ticket request was still in flight, so the abandoned attempt opened a socket nobody owned. Confirmed in the backend log (two sessions, pids 59 and 60, for one dialog). Fixed with a per-attempt generation counter.

A third issue surfaced from real calls: a deleted sandbox returned 502 instead of 404, because the CubeMaster business error code was not mapped.

## Screenshot

An interactive session against a running sandbox on a real cluster — ANSI colour, UTF-8, the window size reported by `stty size` after a resize, and the sandbox's own process tree:

console

terminal

## Known limitations

- **No container selector.** One sandbox is one MicroVM running one envd; the repo has no notion of a container list. Documented.
- **Authorization is coarse.** Any authenticated dashboard user can open a root terminal into any sandbox — CubeOps has no per-sandbox ownership model today. This is pre-existing rather than introduced here, but the terminal makes it consequential, so it is called out in the docs with the recommendation to treat dashboard login as cluster-wide root. Happy to add an ownership check as a follow-up if maintainers want it.
- **Scrollback lives in the browser**, capped at 5000 lines and lost on close.
- **No file transfer** — the SDK filesystem API covers that.

## Docs

`docs/guide/web-terminal.md` and `docs/zh/guide/web-terminal.md`, registered in both the EN "Operations" and ZH "安全与运维" sidebars.


fslongjin:
Thank you for your contribution. I believe this PR generally meets the requirements for the Rhino Bird open-source project. However, since it relies on envd, it cannot access various generic sandboxes where envd isn't installed. Consequently, I find it lacks practical utility. However, for the Rhino Bird project, it has already fulfilled the criteria—I’ve logged it accordingly. Therefore, I’ll temporarily close this PR for now. We warmly welcome your continued involvement in CubeSandbox’s community development—let’s build more interesting features together! 🚀

PR #1256, opened 2026-08-01
one-click: make the CubeProxy admin port configurable

### LINKED ISSUE #945: [Feature Request] Make the CubeProxy admin port configurable and detect port conflicts during one-click installation

Arpit-Ahuja-293:
## Problem

CubeProxy's `/admin/*` listener is pinned to `8082`. cube-proxy runs with host networking, so if anything else on the node already holds that port, nginx exits with:

```
nginx: [emerg] bind() to <node-ip>:8082 failed (98: Address in use)
```

`cube-sandbox-control.target` then fails to come up, and `install.sh` gives no indication of why. The public HTTP/HTTPS/gRPC ports have all been configurable for a while; the admin port was the gap.

Two checks made this worse rather than catching it:

- The pre-start conflict check in `up-cube-proxy.sh` covered `CUBE_PROXY_HTTP_PORT` / `HTTPS` / `GRPC`, but not the admin port.
- `cube-proxy-postcheck.sh` only probed HTTP and gRPC, so a failed admin bind left the unit looking healthy while every lifecycle operation (pause/resume via cube-lifecycle-manager) was dead.

## Fix

Adds `CUBE_PROXY_ADMIN_PORT`, defaulting to `8082` so existing deployments are unaffected, and threads it through everything that referenced the port:

| Layer | Change |
|---|---|
| `up-cube-proxy.sh` | new var; `__CUBE_PROXY_ADMIN_PORT__` nginx placeholder; `CUBE_PROXY_ID` / `CUBE_PROXY_ADMIN_URL` |
| `up-cube-proxy.sh` preflight | admin port added to the existing conflict loop |
| `cube-proxy-postcheck.sh` | probes the admin listener |
| `build-release-bundle.sh` | emits the new placeholder |
| `create.sh` / `tke-addons.tf` / `variables.tf` | `cube_proxy_admin_port` variable across 5 sites |
| `env.example`, `README.md`, `README_zh.md` | documented, EN + ZH |

The Kubernetes chart already supported `CUBE_PROXY_ADMIN_PORT` (`chart/files/cube-proxy/cube-proxy-entrypoint.sh`), so this brings the one-click and Terraform paths in line with naming that was already settled rather than inventing any.

## Worth a maintainer's eye

The Terraform `cube_proxy_admin_port` variable is the only piece here that adds new public surface rather than plumbing an existing value. The issue asked for Terraform coverage, so it is included — happy to drop it into a follow-up if you would rather keep this PR to the one-click path.

## Tests

Written test-first in `deploy/one-click/tests/test_package_layout.sh` (already run in CI by `terraform-validate.yml`). 8 assertions covering the default, the discovery URLs, the preflight loop, the postcheck, `env.example` and the bundle placeholder.

Plus a **round-trip test** that goes beyond static grepping: it renders the generated template with a deliberately non-default port and asserts the result is `listen 10.0.0.7:9082;`, with no `127.0.0.1:8082` binding left behind. Verified it actually bites by temporarily changing `CubeProxy/nginx.conf` to listen on `8085` — three assertions failed — then reverting.

## Verification

- `test_package_layout.sh` → `package layout tests OK`
- Full `deploy/one-click/tests/` suite → 9/10 pass
- `terraform fmt -check -recursive` → clean
- `terraform validate` → `Success! The configuration is valid.`
- `bash -n` clean on every modified script

The one failing suite file is `test_runtime_file_safety.sh`, which **fails identically on unmodified `master`** (an unrelated `cube-sandbox-webui.service` `After=` expectation) and is not run by CI. Untouched here.

Closes #945

### DIFF
--- a/deploy/one-click/README.md
+++ b/deploy/one-click/README.md
@@ -294,6 +294,9 @@ Other common parameters:
 CUBE_PROXY_HTTPS_PORT=443
 CUBE_PROXY_HTTP_PORT=80
 CUBE_PROXY_GRPC_PORT=9090
+# cube-proxy's /admin/* listener. cube-proxy uses host networking, so this
+# port must be free on the node; change it when something else holds 8082.
+CUBE_PROXY_ADMIN_PORT=8082
 # Deprecated: CUBE_PROXY_HOST_PORT is ignored; configure CUBE_PROXY_HTTP_PORT instead.
 CUBE_PROXY_CERT_DIR=/usr/local/services/cubetoolbox/cubeproxy/certs
 CUBE_PROXY_DNS_ANSWER_IP="${CUBE_SANDBOX_NODE_IP}"
--- a/deploy/one-click/README_zh.md
+++ b/deploy/one-click/README_zh.md
@@ -281,6 +281,9 @@ CUBE_PROXY_DNS_ENABLE=1
 CUBE_PROXY_HTTPS_PORT=443
 CUBE_PROXY_HTTP_PORT=80
 CUBE_PROXY_GRPC_PORT=9090
+# cube-proxy 的 /admin/* 监听端口。cube-proxy 使用 host 网络，该端口必须在
+# 节点上空闲；当 8082 已被其它进程占用时，请修改此项。
+CUBE_PROXY_ADMIN_PORT=8082
 # 已废弃：CUBE_PROXY_HOST_PORT 会被忽略；如需调整启动后检查端口，请配置 CUBE_PROXY_HTTP_PORT。
 CUBE_PROXY_CERT_DIR=/usr/local/services/cubetoolbox/cubeproxy/certs
 CUBE_PROXY_DNS_ANSWER_IP="${CUBE_SANDBOX_NODE_IP}"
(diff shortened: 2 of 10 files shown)


zhouxianping:
#955 Thank you for your contribution, Does this meet your requirements? It has already been scheduled for the next release.

Issue #311, opened 2026-05-19
[FAQ] Why must /data/cubelet be on XFS, and how to deploy on ext4-rooted hosts (Ubuntu / Debian / WSL)


fslongjin:
## TL;DR

- `cubelet` stores **container writable layers** under **`/data/cubelet`** and relies on **XFS reflink** to clone them in O(1) time.
- Therefore `/data/cubelet` **must live on an XFS filesystem** with reflink enabled (the `mkfs.xfs` default). Otherwise the `one-click install.sh` pre-flight check refuses to start.
- **Production**: mount a **dedicated XFS data disk** at `/data/cubelet` (we suggest at least **100–300 GiB**).
- **Testing / evaluation**: as a workaround, use a loopback `.img` file formatted as XFS and mounted at `/data/cubelet`.
- **Fresh installs**: prefer **OpenCloudOS 9** or another RHEL-family distro (RHEL 9 / Rocky Linux 9 / AlmaLinux 9). Their root filesystem is XFS by default — no extra mount steps needed.

## Affected scenarios

Ubuntu / Debian-family distros ship with an ext4 root filesystem. When deploying on the whole machine without a dedicated data disk, `/data/cubelet` ends up under `/`, and the one-click pre-flight check refuses to proceed:

```
[one-click] ERROR: The filesystem that will host /data/cubelet is on '/' (type: ext4), which is not XFS.
  Cube Sandbox requires the /data/cubelet directory to reside on an XFS filesystem.
  Options:
    1. Mount a dedicated XFS-formatted partition at /data/cubelet:
         mkfs.xfs /dev/<your-partition>
         mount /dev/<your-partition> /data/cubelet
    2. Ensure the parent path (/) itself is on XFS.
```

WSL and most cloud VMs also default to ext4, so this is a very common situation.

## Why XFS?

`cubelet` clones every sandbox's writable layer with `cp --reflink=always <base> <target>` against a shared base ext4 image, in **O(1) wall time**. `--reflink=always` requires kernel-level **CoW block sharing** via the `FICLONE` ioctl from the **underlying filesystem**.

On Linux, the filesystem we officially support and exercise on the reflink fast path is **XFS (`reflink=1`)**. **ext4 does not implement `FICLONE`**, so placing `/data/cubelet` on ext4 prevents cubelet from taking the fast path — hence the pre-flight refuses up front.

### Option A — Dedicated XFS data disk (recommended for production)

[...]

PR #53, opened 2026-04-22
docs: add .md extension to internal documentation links in index.md

Xinlong-Chen:
Fix broken internal links in index.md (EN/ZH) by adding .md extension

PR #1260, opened 2026-08-03
docs(troubleshooting): add WSL2 DNS troubleshooting for install preflight and sandbox data plane

Tantanovo:
## What

Adds a bilingual troubleshooting page for two DNS problems that are specific to WSL2, requested by @fslongjin.

- `docs/guide/troubleshooting/wsl-dns.md`
- `docs/zh/guide/troubleshooting/wsl-dns.md`
- One index row in each language

`quickstart.md` lists WSL as a supported platform, and [#311](https://github.com/TencentCloud/CubeSandbox/issues/311) covers the XFS loopback workaround, but neither of the two DNS failures below is documented anywhere. Both were hit while bringing up v0.6.0 on WSL2 for [#644](https://github.com/TencentCloud/CubeSandbox/issues/644).

## The two problems

**1. Preflight exits 3 before anything is downloaded**

`online-install.sh` requires `resolvectl` or a loaded `NetworkManager`. A default Ubuntu-on-WSL install has neither.

The generic prerequisite *is* documented in `self-build-deploy.md` ("DNS routing: `systemd-resolved` (preferred) or `NetworkManager + dnsmasq`"), so this is not a missing requirement — it is a discoverability gap:

- `quickstart.md` lists WSL as supported but its requirements table only mentions glibc and XFS.
- The error names the *command* `resolvectl`, while the command is shipped by the **`systemd-resolved` package**. `apt install resolvectl` does not exist, so the next step is not obvious.

Fix: `apt-get install -y systemd-resolved && systemctl enable --now systemd-resolved`.

**2. Install succeeds, sandbox is created, but `run_code` cannot resolve**

```
[1] sandbox created in 0.99s
    sandbox_id: 9e0b35c6183f4e969e1a6a8f6bdc5629     <- control plane OK
httpx.ConnectError: [Errno -2] Name or service not known   <- data plane fails
```

The control plane is fine because it goes through `127.0.0.1:3000`. The data plane needs wildcard resolution for `*.cube.app`, and WSL regenerates `/etc/resolv.conf` on every start, discarding the CoreDNS nameserver written by `dns-host-route-up.sh`.

This one is the more valuable of the two for a troubleshooting page: the error mentions neither DNS nor `cube.app`, and it can appear on a deployment that worked yesterday, because the trigger is a WSL restart rather than a config change.

Fix requires both steps — either alone does not survive a restart:

1. `/etc/wsl.conf` → `[network]` / `generateResolvConf = false`
2. `/etc/resolv.conf` is usually a symlink into `/run/...`, so `rm` it before writing; point the first nameserver at CoreDNS and keep an upstream resolver as a fallback (with only the CoreDNS address, `*.cube.app` resolves but general internet resolution breaks).

The page also notes which CoreDNS address applies to which backend (`169.254.254.53` on the `systemd-resolved` path, `127.0.0.54` on the `dnsmasq` fallback), points to the two DNS-free alternatives already in the repo (path-based access and `examples/e2b-dev-sidecar`), and lists the two pieces of state that do not survive `wsl --shutdown` (the loopback XFS mount and `resolv.conf`).

## Environment used for verification

| Item | Value |
|---|---|
| Cube Sandbox | v0.6.0 (`cubemastercli` `8721dd15`, built 2026-07-24) |
| Deployment | one-click, single node |
| Distro | Ubuntu 24.04.4 LTS on WSL2 |
| Host kernel | `6.18.33.2-microsoft-standard-WSL2` |
| glibc | 2.39 |
| `/data/cubelet` | loopback XFS, `reflink=1` |

Deployment screenshots and logs from that run are attached to [#1238](https://github.com/TencentCloud/CubeSandbox/pull/1238) (14 systemd units active, template `READY`, guest kernel `6.6.1199` vs host `6.18.33.2`).

## Checklist

- [x] Both language files added with the same slug
- [x] Index row added in both `docs/guide/troubleshooting/index.md` and `docs/zh/guide/troubleshooting/index.md`
- [x] Follows `_template.md` section structure (Symptom / Environment / Root Cause / Resolution / References)
- [x] Frontmatter keys aligned across both files (`title`, `author`, `date`, `tags`, `lang`)
- [x] All six relative links verified to exist
- [x] The quoted `online-install.sh` snippet matches upstream (marked as an excerpt)
- [x] Files are LF-only; index diffs are one added line each
- [x] DCO signed off

Happy to adjust the wording, section split, or file placement if you would prefer this merged into `deployment.md` instead of a standalone page.

### DIFF
--- a/docs/guide/troubleshooting/wsl-dns.md
+++ b/docs/guide/troubleshooting/wsl-dns.md
@@ -53,7 +53,7 @@ triggered by restarting WSL rather than by anything you changed.
 - Cube Sandbox version: v0.6.0 (`cubemastercli` `8721dd15`, built 2026-07-24)
 - Deployment mode: one-click, single node (control + compute on the same host)
 - Host OS / kernel: Ubuntu 24.04.4 LTS on WSL2, `6.18.33.2-microsoft-standard-WSL2`, glibc 2.39
-- Related components: `online-install.sh` preflight, CoreDNS (`cube-sandbox-coredns`), CubeProxy, `scripts/systemd/dns-host-route-up.sh`
+- Related components: `online-install.sh` preflight, CoreDNS (`cube-sandbox-coredns`), CubeProxy, `deploy/one-click/scripts/systemd/dns-host-route-up.sh`
 
 Other WSL preconditions were already satisfied and are not the subject of this
 page: `/dev/kvm` exists (recent WSL enables nested virtualization by default),
@@ -97,13 +97,20 @@ stuck on WSL anyway:
 The E2B-compatible SDK reaches a sandbox through a per-sandbox hostname of the
 form `<port>-<sandboxId>.cube.app`. Because the sandbox ID changes every time,
 the client needs wildcard resolution for `*.cube.app`; the bundled CoreDNS
-provides it, and `scripts/systemd/dns-host-route-up.sh` points
-`/etc/resolv.conf` at it (`169.254.254.53` on the `systemd-resolved` path,
-`127.0.0.54` on the `dnsmasq` fallback path). See
+provides it, and `deploy/one-click/scripts/systemd/dns-host-route-up.sh` routes
+`cube.app` queries to it. On both DNS backends the client-facing nameserver is
+`169.254.254.53` (the dummy-link address): on the `systemd-resolved` path the
+script attaches that address to the `cube-dns0` link via `resolvectl`, and on
+the `dnsmasq` fallback path it writes the same address into `/etc/resolv.conf`.
+`127.0.0.54` is only CoreDNS's internal loopback bind, which `dnsmasq` forwards
+`cube.app` queries to; it never appears in `/etc/resolv.conf`. See
 [HTTPS and domain](../https-and-domain.md) for the full addressing scheme.
 
-WSL, by default, regenerates `/etc/resolv.conf` on every start. That silently
-discards the nameserver the installer wrote, so:
+WSL, by default, regenerates `/etc/resolv.conf` on every start. On the
+`dnsmasq` fallback path that silently discards the nameserver the installer
+wrote; on the `systemd-resolved` path the routing lives on the `cube-dns0`
+dummy link and is dropped too, because WSL tears the link down. Either way the
+client loses the wildcard resolution, so:
 
 - The control plane keeps working, because it is reached over `127.0.0.1:3000`
   and needs no name resolution.
@@ -156,10 +163,14 @@ Two details matter here:
 - Keep an upstream resolver as the second entry. With only the CoreDNS address,
   `*.cube.app` resolves but general internet resolution breaks.
 
-Use the address that matches your DNS backend. After the fix for Problem 1 you
-are on the `systemd-resolved` path, so `169.254.254.53` is the one to use; on
-the `dnsmasq` fallback path it is `127.0.0.54`. You can confirm which one
-CoreDNS is bound to from the running Corefile.
+The nameserver to write is `169.254.254.53` regardless of which DNS backend
+you are on: it is the dummy-link address that both the `systemd-resolved` and
+the `dnsmasq` paths expose to clients. (`127.0.0.54` is only CoreDNS's internal
+loopback bind, which `dnsmasq` forwards `cube.app` queries to; it is not a
+client-facing resolver and must not be written to `/etc/resolv.conf` — a
+loopback address there would be unreachable from inside Docker containers,
+which is exactly the problem the dummy-link address was introduced to solve.)
+You can confirm the live address from the running Corefile.
 
 Verify both directions before moving on:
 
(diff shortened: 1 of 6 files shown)


Tantanovo:
Thanks for the review — both blocking issues are confirmed and fixed in `35aad91`, in both languages.

1. **DNS address mapping inverted** — you are right. `write_host_resolv_conf "${RESOLVED_COREDNS_BIND_ADDR}"` (`dns-host-route-up.sh:270`) is the only place resolv.conf is written and it always uses the dummy-link address `169.254.254.53`; `127.0.0.54` is only CoreDNS's internal loopback bind in the `server=/cube.app/${COREDNS_BIND_ADDR}#53` forwarder rule. The page now states the client-facing nameserver is `169.254.254.53` on both backends, and explicitly warns not to write `127.0.0.54` (unreachable from inside Docker containers).

2. **Wrong script path** — corrected to `deploy/one-click/scripts/systemd/dns-host-route-up.sh` in all four places (Environment + Root Cause in both languages).

Also adopted the two non-blocking suggestions:
- Root Cause now notes that on the `systemd-resolved` path the installer does not rewrite resolv.conf — the routing lives on the `cube-dns0` dummy link, which WSL also tears down on restart.
- "Notes for WSL Restarts" now lists the dummy link / `server=/cube.app/...` forwarder rule as the third piece of state lost across `wsl --shutdown`.

Tantanovo:
All three suggestions adopted in `591d6af`, both languages.

1. **Duplicate `[network]` section** — the code comment and the notes list now say to put `generateResolvConf = false` inside an existing `[network]` section rather than appending a second one.

2. **Hardcoded upstream** — `223.5.5.5` is now explicitly marked as an example, with a note that the installer's own `write_host_resolv_conf` reuses the host's existing upstream while this file is written by hand.

3. **`dig` prerequisite** — the verification block now leads with `getent hosts foo.cube.app`, which needs no extra package, and notes that `dig` comes from `dnsutils` / `bind9-dnsutils`.

Thanks for the thorough verification against the base tree — the `coredns-start.sh:24-25` and `env.example:176-177` cross-checks on the two bind addresses were useful confirmation that the corrected wording is right.

Tantanovo:
Pushed `2f7b4d6` with one more improvement I noticed while re-reading the page.

The restart notes said the dummy link and host routing are lost after `wsl --shutdown`, but did not say how to rebuild them — which is the part a reader actually needs at that moment. Added:

- **The Corefile path** — `/usr/local/services/cubetoolbox/coredns/Corefile` (`TOOLBOX_ROOT` from `scripts/systemd/common.sh:7`), instead of only saying the address can be confirmed from the running Corefile.
- **A recovery sequence** — remount the XFS volume, `systemctl start cube-sandbox-control.target`, and if `*.cube.app` still fails, restart `cube-sandbox-coredns.service` then `cube-sandbox-dns.service`, verifying with `ip addr show cube-dns0` and `getent hosts foo.cube.app`.
- **Why that order** — `cube-sandbox-dns.service` is `Type=oneshot` and declares `Requires=` / `BindsTo=` on `cube-sandbox-coredns.service`, so CoreDNS has to be listening before host resolution is switched over.

This is the sequence I actually used to recover the deployment after a WSL restart, so it is verified rather than inferred from the unit files alone.

docs/guide/troubleshooting/wsl-dns.md
```diff
@@ -0,0 +1,275 @@
...
+are in [Environment](#environment).
+
+## Symptom
+
+### Problem 1: preflight exits 3 before anything is downloaded
+
+`online-install.sh` stops immediately with:
+
+```text
+[online-install] ERROR: DNS setup requires resolvectl or NetworkManager.
```
chenhengqi: Consider adding `apt install systemd-resolved` in error message instead.

Tantanovo: Done in `3f566e9`.

Covered all three sites that reject a host for missing DNS tooling, so the hint shows up regardless of which path fails:

- `install.sh:232` (`check_dns_preflight`)
- `online-install.sh:295` (NetworkManager not loaded)
- `online-install.sh:299` (no systemctl at all)

The error now reads:

```text
[online-install] ERROR: DNS setup requires resolvectl or NetworkManager. resolvectl is provided by the systemd-resolved package: apt install systemd-resolved (or dnf install systemd-resolved).
```

Two follow-on edits since the page quotes this error verbatim: both language versions now match the new text, and the note about the package name being absent is scoped to releases predating this change rather than stated as current behaviour.

lkml-likexu:
@Tantanovo Any status update ?

Tantanovo:
@lkml-likexu Hi @lkml-likexu, thanks for checking in — the PR is ready from my side since Aug 3.

All review comments addressed (both blocking issues + all nits) in 35aad91 / 591d6af / 2f7b4d6, EN/ZH in sync
Bot verdict: "Approve with minor nits"
It's been waiting on a human review/merge since then. Happy to rebase if the base has moved. Could you take a look when you get a chance?

fslongjin:
Hello, after careful consideration, we have decided not to natively support CubeSandbox within WSL. If you do need to use Cube in WSL,  you might consider https://cubesandbox.com/guide/dev-environment

PR #1262, opened 2026-08-03
Fix Kubernetes node probe portability

backwind1233:
## Summary

- strip carriage returns from rendered CubeEgress readiness and liveness probe commands
- treat any HTTP response from CubeProxy as transport reachability in the Helm health test

## Why

A Windows checkout can inject CRLF into the included Bash probe and make `set -e` fail with an invalid option. CubeProxy also validly returns HTTP 400 for `/` without a sandbox `Host` header, so `curl --fail` incorrectly reports a reachable proxy as unhealthy.

## Testing

- `helm lint deploy/kubernetes/chart` with non-default test credentials
- `helm template` renders without carriage returns
- all eight Helm tests passed on AKS Kubernetes v1.35.6

Assisted-by: GitHub Copilot:GPT-5.4

### DIFF
--- a/deploy/kubernetes/chart/templates/node-daemonset.yaml
+++ b/deploy/kubernetes/chart/templates/node-daemonset.yaml
@@ -261,7 +261,7 @@ spec:
               command:
                 - /bin/bash
                 - -ec
-                - {{ include "cube.egressNetProbeCommand" . | quote }}
+                - {{ include "cube.egressNetProbeCommand" . | replace "\r" "" | quote }}
             initialDelaySeconds: {{ .Values.cubeEgress.network.probes.readiness.initialDelaySeconds }}
             periodSeconds: {{ .Values.cubeEgress.network.probes.readiness.periodSeconds }}
             failureThreshold: {{ .Values.cubeEgress.network.probes.readiness.failureThreshold }}
@@ -272,7 +272,7 @@ spec:
               command:
                 - /bin/bash
                 - -ec
-                - {{ include "cube.egressNetProbeCommand" . | quote }}
+                - {{ include "cube.egressNetProbeCommand" . | replace "\r" "" | quote }}
             initialDelaySeconds: {{ .Values.cubeEgress.network.probes.liveness.initialDelaySeconds }}
             periodSeconds: {{ .Values.cubeEgress.network.probes.liveness.periodSeconds }}
             failureThreshold: {{ .Values.cubeEgress.network.probes.liveness.failureThreshold }}
--- a/deploy/kubernetes/chart/templates/tests/node-health.yaml
+++ b/deploy/kubernetes/chart/templates/tests/node-health.yaml
@@ -262,9 +262,9 @@ spec:
         - -ec
         - |
           echo '[health-test] check CubeProxy via ClusterIP Service'
-          curl --connect-timeout 5 --max-time 15 -fsS -o /dev/null \
+          curl --connect-timeout 5 --max-time 15 -sS -o /dev/null \
             "http://{{ include "cube.proxyServiceFQDN" . }}:{{ .Values.cubeProxy.ports.http.containerPort }}/" \
-            || curl --connect-timeout 5 --max-time 15 -k -fsS -o /dev/null \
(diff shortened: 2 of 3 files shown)


deploy/kubernetes/chart/templates/node-daemonset.yaml
```diff
@@ -261,7 +261,7 @@ spec:
               command:
                 - /bin/bash
                 - -ec
-                - {{ include "cube.egressNetProbeCommand" . | quote }}
+                - {{ include "cube.egressNetProbeCommand" . | replace "\r" "" | quote }}
```
fslongjin, line 264: **Suggestion (non-blocking):** This `replace "\r" ""` is a correct spot-fix, but the root cause is that the repo has no `.gitattributes` to enforce LF line endings. A Windows contributor (or anyone with `core.autocrlf=true`) will hit this same class of bug on any future multi-line template rendered through `include ... | quote`.

Consider adding a `.gitattributes` at the repo root with:

```
* text=auto
*.yaml text eol=lf
*.tpl text eol=lf
```

This would prevent CRLF injection at checkout time rather than relying on each call site to remember `replace "\r" ""`. The `replace` guard here is still valuable as defense-in-depth — keeping it even after `.gitattributes` is fine.

deploy/kubernetes/chart/templates/tests/node-health.yaml
```diff
@@ -262,9 +262,9 @@ spec:
         - -ec
         - |
           echo '[health-test] check CubeProxy via ClusterIP Service'
-          curl --connect-timeout 5 --max-time 15 -fsS -o /dev/null \
+          curl --connect-timeout 5 --max-time 15 -sS -o /dev/null \
```
fslongjin, line 265: **Suggestion:** This fix correctly drops `-f` for the proxy-control-test pod, but the **main health-test pod** at lines 99–101 still uses `-fsS`:

```bash
curl --connect-timeout 5 --max-time 15 -fsS -o /dev/null \
  "http://{{ include "cube.proxyName" . }}...:{{ .Values.cubeProxy.ports.http.containerPort }}/" \
  || echo '[health-test] cube-proxy Service HTTP soft-failed (may require Host header / TLS path)'
```

That curl hits the same `/` endpoint without a sandbox `Host` header, so CubeProxy will return 400 every time. The `|| echo` makes it a soft failure (the test still passes), but the log message is guaranteed noise on **every** successful run — it's no longer "may require", it's "definitely requires".

Two options:
1. Drop `-f` there too (`-fsS` → `-sS`), matching this PR's semantics — connectivity check, not HTTP-status check.
2. Send a valid sandbox `Host` header so the check actually validates proxy *functionality*, not just reachability.

Either way, the current `|| echo` log line is misleading and should be updated to reflect that 400 is expected without a Host header.

fslongjin (changes requested):
ping~

deploy/kubernetes/chart/templates/node-daemonset.yaml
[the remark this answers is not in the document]
fslongjin: Addressed: added a root `.gitattributes` with `*.yaml` / `*.tpl` `eol=lf` to prevent CRLF at checkout. Kept the existing `replace "\r" ""` as defense-in-depth. Skipped a separate CRLF render guard for now.

fslongjin: Done — added the suggested `.gitattributes` at the repo root (`* text=auto`, `*.yaml`/`*.tpl` `eol=lf`). Left the `replace "\r" ""` call-site guard in place as defense-in-depth.

deploy/kubernetes/chart/templates/tests/node-health.yaml
fslongjin: Went with option 1: the main health-test CubeProxy check now uses `-sS` (connectivity only, matching proxy-control-test) and hard-fails with `cube-proxy Service unreachable` instead of the misleading soft-fail log.

fslongjin (approved):

PR #1264, opened 2026-08-03
fix(kubernetes): make cubecli discoverable on compute nodes

zyl1121:
## Motivation

Kubernetes deployments already stage `cubecli` in the Cubelet toolbox, but operators cannot use it through the normal Kubernetes entry point.

By default, `kubectl exec` into a `cube-node` Pod targets the `network-agent` container, while `cubecli` requires access to the local Cubelet and containerd sockets. Even when operators explicitly select the `cubelet` container, the staged `Cubelet/bin` directory is not on `PATH`.

This makes an already-installed node-local CLI hard to use. Operators need to know both the `cubelet` container name and the absolute toolbox path before they can run basic node-local checks.

Since `cubecli` uses the selected node's local Cubelet and containerd sockets, this PR makes the `cubelet` container the supported Kubernetes entry point.

## What Changed

* Set the default container annotation for `cube-node` Pods to `cubelet`, so `kubectl exec` and logs use the container that owns the local Cubelet tools.
* Preserve custom default-container annotations through `cubeNode.podAnnotations`.
* Add the staged Cubelet `bin` directory to the Cubelet image `PATH`, so `kubectl exec` into the `cubelet` container resolves `cubecli` after toolbox staging without a container-side symlink.
* Keep the existing staging validation so Cubelet startup fails clearly when `Cubelet/bin/cubecli` is missing.
* Document how to select a `cube-node` Pod and run `cubecli`, including the network-namespace requirement for `cubecli container taps`.
* Add regression checks for Helm rendering, default-container overrides, and Cubelet staging PATH exposure.
* Publish the Cubelet runtime image built from the updated Dockerfile together with the Chart, using the matching `images.cubelet.tag`.

## Validation

* Ran Helm lint.
* Ran the CubeCLI discovery, Cubelet PATH, and shell syntax guards.
* Ran the complete Kubernetes chart/image guard suite.
* Rendered the Chart and verified the `cubelet` default-container annotation and custom annotation override behavior.
* Installed the Chart on the K3s test cluster and verified:

  * `kubectl exec` without `-c` entered the `cubelet` container.
  * `command -v cubecli && cubecli ls` succeeded inside the selected `cube-node` Pod.
  * the Cubelet image PATH guard covers the staged `Cubelet/bin` directory, and a standalone staging run fails clearly if `Cubelet/bin/cubecli` is missing.

## Scope

This PR only makes the existing staged `cubecli` available inside the `cubelet` container of Kubernetes compute-node Pods.

It does not add host login-shell integration, a standalone `cubecli` workload, expose node-local sockets to another Pod, or change how `cubecli` connects to Cubelet and containerd.

Existing custom default-container annotations remain supported through `cubeNode.podAnnotations`.

Because the default-container annotation changes the `cube-node` Pod template, upgrading an existing release triggers a RollingUpdate of the `cube-node` DaemonSet. Recreated Pods receive new Pod IPs and network namespaces, so running sandboxes may be interrupted; schedule the upgrade during a maintenance window. To preserve the previous no-`-c` behavior, set `cubeNode.podAnnotations["kubectl.kubernetes.io/default-container"]` to `network-agent` before upgrading.

### DIFF
--- a/.github/workflows/kubernetes-chart-check.yml
+++ b/.github/workflows/kubernetes-chart-check.yml
@@ -4,12 +4,14 @@ on:
   push:
     branches: [master]
     paths:
+      - 'Cubelet/Dockerfile'
       - 'deploy/kubernetes/chart/**'
       - 'deploy/kubernetes/images/scripts/**'
       - 'scripts/bump-image.sh'
       - '.github/workflows/kubernetes-chart-check.yml'
   pull_request:
     paths:
+      - 'Cubelet/Dockerfile'
       - 'deploy/kubernetes/chart/**'
       - 'deploy/kubernetes/images/scripts/**'
       - 'scripts/bump-image.sh'
--- a/Cubelet/Dockerfile
+++ b/Cubelet/Dockerfile
@@ -83,7 +83,7 @@ ARG CUBE_VERSION=unknown
 ENV IMAGE_ROOT=/opt/cube-image
 ENV TOOLBOX_ROOT=/usr/local/services/cubetoolbox
 ENV CUBE_COMPONENT=cubelet
-ENV PATH="/usr/local/services/cubetoolbox/cube-shim/bin:${PATH}"
+ENV PATH="/usr/local/services/cubetoolbox/Cubelet/bin:/usr/local/services/cubetoolbox/cube-shim/bin:${PATH}"
 
 RUN apt-get update \
     && apt-get install -y --no-install-recommends \
--- a/deploy/kubernetes/chart/README.md
+++ b/deploy/kubernetes/chart/README.md
@@ -157,6 +157,14 @@ can set it to `false`.
 - `cubeNode.network.autoDetectEthName=true` auto-detects the primary host NIC and patches Cubelet `eth_name`;
 - `cubeNode.network.cidr` patches Cubelet cubevs/sandbox CIDR (default `172.16.0.0/18`, chosen to avoid common cluster Service CIDR `192.168.0.0/16` while keeping a /18 pool). A Helm `pre-install`/`pre-upgrade` Hook fails fast when this range overlaps the cluster Service CIDR or existing ClusterIPs; set `cubeNode.network.cidrSkipConflictCheck=true` only if you accept that risk.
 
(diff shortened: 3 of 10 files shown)


fslongjin:
Hi there, I feel this implementation might be a bit overly complex. As an alternative approach, perhaps we could simply add the PATH environment variable inside the cubelet container's Docker configuration? Wouldn't that meet your requirements?

zyl1121:
> Hi there, I feel this implementation might be a bit overly complex. As an alternative approach, perhaps we could simply add the PATH environment variable inside the cubelet container's Docker configuration? Wouldn't that meet your requirements?

That makes sense. To keep this PR focused, would you prefer to limit `cubecli` discoverability to the cubelet container in the Big Pod by adding it to the image `PATH`, and drop host login-shell support from this PR?

I can remove the host symlink and keep this PR limited to the container `PATH` change.

fslongjin:
> > Hi there, I feel this implementation might be a bit overly complex. As an alternative approach, perhaps we could simply add the PATH environment variable inside the cubelet container's Docker configuration? Wouldn't that meet your requirements?
> 
> That makes sense. To keep this PR focused, would you prefer to limit `cubecli` discoverability to the cubelet container in the Big Pod by adding it to the image `PATH`, and drop host login-shell support from this PR?
> 
> I can remove the host symlink and keep this PR limited to the container `PATH` change.

Yes, I believe adding the `PATH` environment variable to the cubelet container image so that `cubecli` can be executed directly inside the container should resolve the issue. From my perspective, in a Kubernetes deployment scenario, there is no need to run `cubecli` directly on the host machine of the compute nodes.

zyl1121:
> Yes, I believe adding the PATH environment variable to the cubelet container image so that cubecli can be executed directly inside the container should resolve the issue. From my perspective, in a Kubernetes deployment scenario, there is no need to run cubecli directly on the host machine of the compute nodes.

I’ve simplified the PR accordingly.

It now only does the following:

* Defaults `kubectl exec` to the `cubelet` container.
* Adds the staged `Cubelet/bin` directory to the Cubelet image `PATH`.
* Keeps the missing-`cubecli` staging guard and focused regression tests.

I removed the host symlink, cleanup integration, and `cube-node-init` changes.

I also documented that the new default-container annotation updates the Pod template and may trigger a `cube-node` RollingUpdate during upgrade. The PR body now notes that the updated Cubelet image must be published and referenced via the matching `images.cubelet.tag`.

fslongjin (approved):

PR #1271, opened 2026-08-04
fix(k8s): use stable StatefulSet DNS for CubeProxy registry heartbeat

try-agaaain:
## Summary

The CubeProxy registry heartbeat (`proxy_registry.lua`, published from an
`ngx.timer` handler) connects to Redis via `resty.redis` → `ngx.socket.tcp`.
`proxy.yaml` pinned that target to a concrete IP captured at render or startup
time:

- chart template: `lookup "v1" "Service"/"Endpoints"` baked the redis
  ClusterIP or headless Endpoints IP into the env value;
- entrypoint: `getent ahostsv4` resolved the hostname to an IP before nginx
  started.

Both pin a stale address once the redis Pod restarts (new Pod IP) while the
cube-proxy Pod keeps running. The heartbeat then silently fails, CLM stops
discovering the proxy, and `auto_resume` gate state is never pushed →
`504 Gateway Timeout`.

## Fix

Pass the Redis **hostname** through to `CUBE_PROXY_REGISTRY_REDIS_HOST` and let
nginx resolve it per connect. The chart already renders `resolver` at http
scope (`start.sh` → `conf/includes/resolver.inc`), so every Lua cosocket path —
including `ngx.timer` handlers — resolves through it. The entrypoint no longer
pre-resolves the host to an IP (`getent` block removed).

For the builtin redis, the host is the headless Service DNS name
(`<redis>.<ns>.svc.<domain>`), which always resolves to the live Pod IP and
self-heals after a redis restart — the same mechanism
`cube-lifecycle-manager` already relies on.

## Additional hardening

- **`checksum/entrypoint` annotation** on the cube-proxy Deployment hashes the
  entrypoint ConfigMap body so a `helm upgrade` that only changes the
  entrypoint rolls the pod — otherwise the old script keeps running until the
  pod is manually deleted.
- **`cubeProxy.resolver.valid` default 30s → 5s**: DNS answers are cached for
  `valid` seconds, so heartbeat recovery takes roughly `valid` + one heartbeat
  interval after a redis restart. 30s exceeded `lifecycleManager.heartbeatTTL`
  (15s), leaving the proxy judged dead for ~20s in that window.
- **Render guard test** `test-proxy-registry-host.sh`: asserts the builtin
  host renders as the Service DNS name, external host / IP literals pass
  through, sentinel mode leaves the host empty, and `checksum/entrypoint`
  tracks entrypoint content. `cube-proxy-entrypoint.sh` added to the CI
  `sh -n` syntax check.

## Compatibility notes

- External `redis.host` / `cubeProxy.redis.host` overrides must be an FQDN or
  IP literal: the nginx `resolver` does not apply `/etc/resolv.conf` search
  domains (documented in README and values.yaml comments).
- Sentinel mode is unchanged (registry host stays empty; the master is
  resolved via SENTINEL).

### DIFF
--- a/.github/workflows/kubernetes-chart-check.yml
+++ b/.github/workflows/kubernetes-chart-check.yml
@@ -53,6 +53,7 @@ jobs:
           sh -n deploy/kubernetes/images/scripts/pvm-startup-gate-reconcile.sh
           sh -n deploy/kubernetes/images/scripts/pvm-host-bootstrap.sh
           sh -n deploy/kubernetes/images/scripts/wait-node-prep.sh
+          sh -n deploy/kubernetes/chart/files/cube-proxy/cube-proxy-entrypoint.sh
       # Cloud-free chart/image guard suite (deploy/kubernetes/*/scripts/
       # test-*.sh). Helm is installed above; every script is self-contained.
       # Loop so new test-*.sh are picked up automatically instead of drifting
--- a/deploy/kubernetes/chart/README.md
+++ b/deploy/kubernetes/chart/README.md
@@ -231,6 +231,8 @@ postgres:
 
 The chart installs `cube-redis` StatefulSet only when `redis.enabled=true` and `redis.host` is empty.
 Set `redis.host` to use an existing Redis service; the chart will not install `cube-redis`.
+CubeProxy registry heartbeats resolve Redis by hostname, so a redis Pod IP change self-heals.
+Use an FQDN or IP literal: nginx's `resolver` does not apply search domains.
 
 ## CubeMaster configuration
 
@@ -464,14 +466,16 @@ connections, the chart renders an nginx `resolver` into
 discovers resolver addresses from the Pod `/etc/resolv.conf`, which resolves the
 chart-managed `cube-redis.<namespace>.svc.cluster.local` Service name and
 third-party Redis DNS names. Override only when the cluster requires explicit
-DNS servers:
+DNS servers. Keep `cubeProxy.resolver.valid` well below
+`lifecycleManager.heartbeatTTL` (defaults `5s` / `15s`) so a redis restart
+does not look like a dead proxy.
 
 ```yaml
 cubeProxy:
   resolver:
     addresses:
       - 172.18.0.10
-    valid: 30s
+    valid: 5s
     timeout: 5s
     ipv6: false
 ```
--- a/deploy/kubernetes/chart/files/cube-proxy/cube-proxy-entrypoint.sh
+++ b/deploy/kubernetes/chart/files/cube-proxy/cube-proxy-entrypoint.sh
@@ -96,21 +96,6 @@ sidecar_addr="${CUBE_SIDECAR_LISTEN_ADDR:-}"
   exit 1
 }
 
-# proxy_registry.lua publishes from ngx.timer, which has no nginx resolver.
-# Resolve a hostname REDIS target to an IP up-front when possible.
-# Sentinel mode does not need a fixed registry host.
-if [ -z "${CUBE_PROXY_REGISTRY_REDIS_MASTER_NAME:-}" ] && [ -n "${CUBE_PROXY_REGISTRY_REDIS_HOST:-}" ]; then
-  case "${CUBE_PROXY_REGISTRY_REDIS_HOST}" in
-    *[!0-9.]* )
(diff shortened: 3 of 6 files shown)


try-agaaain:
Addressed all review comments with commit `1b64488` (and the earlier two rounds
`2e926c1` / `10055df`). Mapping each comment to the final design:

**1. "the getent resolution happens once at container startup — it does not survive a redis Pod IP change" (entrypoint `cube-proxy-entrypoint.sh`)**
Removed the `getent ahostsv4` pre-resolution entirely. The entrypoint now passes
the Redis **hostname** through untouched, and `proxy_registry.lua`'s
`resty.redis` → `ngx.socket.tcp` resolves it through nginx's **http-scope
resolver** on every connect. Each timer tick re-resolves, so a redis Pod IP
change self-heals without restarting cube-proxy.

**2. "the premise (ngx.timer cosockets have no resolver) contradicts the chart's own nginx config" (proxy.yaml comment)**
Correct — this is the load-bearing claim and it is supported by the repo:
`nginx.conf:32-33` keeps `resolver` at http scope via the `start.sh`-rendered
`conf/includes/resolver.inc` explicitly so "every Lua cosocket path inherits it",
and `init_worker_phase.lua` defers the first registry publish into
`ngx.timer.at(0)` because cosocket *creation* is illegal in `init_worker`, not
because DNS resolution in timers is special. The comments in both the entrypoint
and `proxy.yaml` now state this accurately.

**3. "global.conf is included at location scope; attribute the http-scope include correctly" (entrypoint comment)**
Fixed — the comment now references `nginx.conf` / `conf/includes/resolver.inc`
at http scope, not `global.conf`.

**4. "hardcoded `-0` ordinal couples to StatefulSet internals" (proxy.yaml)**
Removed the `-0` ordinal and the `lookup()` branches entirely. For the builtin
redis, `cube.redisHost` already defaults to the headless Service DNS name
(`<redis>.<ns>.svc.<domain>`), which resolves to the live Pod IP for the
single-replica StatefulSet and self-heals on restart — the same address
`cube-lifecycle-manager` connects to through the system resolver. No ordinal
assumption, no dead `lookup` code (the builtin redis Service is always headless,
so the ClusterIP branch could never fire).

**5. "bare/short-name external redis targets lose search-domain resolution"**
Documented in `chart/README.md`: `redis.host` / `cubeProxy.redis.host` must be
an FQDN or IP literal because nginx `resolver` does not apply `/etc/resolv.conf`
search domains. The builtin FQDN is unaffected.

Verified with `helm lint` and `helm template`: `CUBE_PROXY_REGISTRY_REDIS_HOST`
renders to the headless Service DNS name.

try-agaaain:
Follow-up commits `752931f` (fix) and `3b55dfa` (test) address a review round
across all dimensions. Summary:

**Design/perf — resolver.valid vs CLM TTL (validated)**
Lowered `cubeProxy.resolver.valid` default from 30s to 5s. After a redis Pod
IP change, the registry heartbeat recovers within roughly `valid` + one
heartbeat interval; with 30s that exceeded CLM's `heartbeatTTL: 15s`, so CLM
would still judge the proxy dead for ~20s after a redis restart — the exact
auto-resume 504 window this fix targets. 5s keeps the worst-case recovery
under ~10s. Documented the coupling in values.yaml and README.

**Rollout — checksum/entrypoint (new mechanism)**
Added a `checksum/entrypoint` annotation to the cube-proxy Deployment,
hashing the entrypoint ConfigMap body (same pattern as webui.yaml). Without
it, an external-redis deployment (where the registry env value is unchanged)
would keep running the *old* entrypoint after `helm upgrade` — the old
entrypoint still pinned the Redis host to an IP, re-introducing the bug.

**Shell syntax in CI**
Added `cube-proxy-entrypoint.sh` to the kubernetes-chart-check `sh -n` step;
it was edited by this fix but previously unlisted.

**Render guard test**
New `test-proxy-registry-host.sh`: asserts the registry env renders to the
headless Service DNS name for builtin redis (never a pinned IP), passes
external `redis.host` (hostname or IP literal) through untouched, and that
`checksum/entrypoint` tracks the entrypoint file content.

**DCO**
All six commits now carry `Signed-off-by: tsingyue`.

The self-heal claim in the README was also reworded to be precise about the
recovery window (`valid` + one heartbeat interval) rather than implying
instant recovery.

fslongjin (changes requested):
Thank you for your contribution—this looks really good. However, your code changes introduce quite a few new comments that seem redundant. Could you remove them?

fslongjin (approved):

PR #1278, opened 2026-08-04
test(cubelet): avoid wall-clock race in multilock lifetime test

zyl1121:
## Motivation

`TestMultiLock_Lifetime` can fail intermittently because it asserts asynchronous expiration cleanup after a fixed wall-clock sleep.

Lock activity is recorded with integer-second timestamps, expiration uses a strict greater-than comparison, and cleanup runs on a background ticker. As a result, a fixed four-second sleep can wake before the cleanup tick observes the lock as expired, leaving the lock in the map even though the production behavior is still within its configured timing.

The test also starts concurrent lock activity without waiting for those goroutines to finish before beginning the expiration wait, so the last activity timestamp is not a stable boundary for the assertion.

## What Changed

* Wait for all concurrent lock activity to finish before testing expiration.
* Set the package-internal activity timestamp to an unambiguously expired value.
* Wait for asynchronous cleanup with a bounded deadline instead of asserting at one exact instant.
* Keep the production lock lifetime and cleanup implementation unchanged.

## Validation

```bash
cd Cubelet
go test -short ./pkg/multilock -count=100
go test -race -short ./pkg/multilock -run '^TestMultiLock_Lifetime$' -count=20
go test -short ./pkg/...
```

All tests passed in the builder container.

## Scope

This PR only updates `Cubelet/pkg/multilock/multilock_test.go`.

It does not change production multilock behavior, lock lifetime configuration, or cleanup scheduling.

### DIFF
--- a/Cubelet/pkg/multilock/multilock_test.go
+++ b/Cubelet/pkg/multilock/multilock_test.go
@@ -8,6 +8,8 @@ import (
 	"fmt"
 	"math/rand"
 	"runtime"
+	"sync"
+	"sync/atomic"
 	"testing"
 	"time"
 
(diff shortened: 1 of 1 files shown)


luzhixing12345:
LGTM

PR #1279, opened 2026-08-04
test(cubelet): avoid FIFO open race in taskio test

luzhixing12345:
Open the consumer before starting the writer so the writer cannot close all FIFO write handles before the reader is connected, which could block the test until timeout.


lkml-likexu:
@luzhixing12345 Are there any real-world failure scenarios that point to this defect? Looking at the diff, this is just to improve the stability of the unit tests.

luzhixing12345:
@lkml-likexu This is just to improve the stability of the unit tests; an error occurred in the previous CI run: https://github.com/TencentCloud/CubeSandbox/actions/runs/30886594367/job/91919097279, and this PR is simply to fix that.

PR #1283, opened 2026-08-04
fix(agenthub): fix gateway token auth race and improve template UX

liciazhu:
CubeOps: resolve OpenClaw gateway token from openclaw.json.last-good instead of openclaw.json to prevent 401 auth failures caused by token rotation during startup.

WebUI: remove system default template option, add empty-state prompt with store link when no templates exist, and enforce template selection before submission.


lkml-likexu:
@liciazhu Any status update or rebase ?

PR #1288, opened 2026-08-05
feat(helm): allow fixed NodePort values for CubeAPI and CubeProxy

xiaojunxiang2023:
## Summary

- Chart Services already accepted `service.type=NodePort`, but fixed host ports could not be set via values.
- Add optional `controlPlane.api.service.nodePort` and `cubeProxy.service.nodePorts.{http,https,grpc,admin}` (Kubernetes range `30000-32767`). Empty keeps auto-allocation; default remains `ClusterIP` + Ingress.
- `validate.yaml` rejects explicit `nodePort` when `type` is still `ClusterIP`, and rejects out-of-range values.
- Document the bare-metal / no-Ingress profile in chart README and `runtime-values.example.yaml`.

## Test plan

- [x] `helm lint deploy/kubernetes/chart` (with test passwords) — passed
- [x] `helm template` default render — no `nodePort` fields
- [x] `bash deploy/kubernetes/chart/scripts/test-service-nodeport.sh` — passed (fixed ports render; ClusterIP+nodePort and out-of-range fail)
- [x] Existing chart guards (`test-database-postgres-guard.sh`, `test-master-recreate-strategy-guard.sh`, `test-pvm-native-ds-gvk-guard.sh`, …) — passed on ubuntu22 with synced chart tree

### DIFF
--- a/deploy/kubernetes/chart/README.md
+++ b/deploy/kubernetes/chart/README.md
@@ -443,6 +443,8 @@ CubeProxy runs on the **Pod network** (no `hostNetwork`). Traffic path:
 
 TLS for `cube.app` / wildcards still terminates **inside CubeProxy**. The default Ingress annotations enable nginx-ingress SSL passthrough + HTTPS backend; override `cubeProxy.ingress.className` / `annotations` for TKE CLB or other controllers. Set `cubeProxy.ingress.enabled=false` if you manage the entrypoint yourself (keep the Service as backend).
 
+Without an Ingress / cloud LB, set `cubeProxy.service.type` / `controlPlane.api.service.type` to `NodePort` (or `LoadBalancer`) and optionally pin host ports via `cubeProxy.service.nodePorts.*` / `controlPlane.api.service.nodePort` (Kubernetes range `30000-32767`; empty keeps auto-allocation). Explicit `nodePort` values are rejected when `type` is still `ClusterIP`.
+
 When the sandbox owner is on a compute node, CubeProxy still uses Redis routing metadata to connect to the owner `HostIP:hostPort`. The chart patches the image's default nginx listeners to the configured `cubeProxy.ports.*.containerPort` values (default `80` / `443`).
 
 CubeProxy admin is reachable in-cluster at each Pod IP:`adminPort` (default `8082`) for cube-lifecycle-manager discovery; probes use the admin token header.
--- a/deploy/kubernetes/chart/scripts/test-service-nodeport.sh
+++ b/deploy/kubernetes/chart/scripts/test-service-nodeport.sh
@@ -1,0 +1,97 @@
+#!/bin/sh
+# Guard: optional fixed Service nodePorts render only when type allows them,
+# and validate.yaml rejects ClusterIP + explicit nodePort / out-of-range values.
+set -eu
+
+SCRIPT_DIR="$(CDPATH= cd -- "$(dirname "$0")" && pwd)"
+CHART_DIR="$(dirname "$SCRIPT_DIR")"
+TMP_DIR="$(mktemp -d)"
+trap 'rm -rf "$TMP_DIR"' EXIT
+
+COMMON_SETS="
+  --set-string mysql.password=test
+  --set-string mysql.rootPassword=test
+  --set-string redis.password=test
+"
+
+fail_msg() {
+  echo "FAIL: $1" >&2
+  exit 1
+}
+
+# Default ClusterIP render must not emit nodePort fields.
+helm template nodeport-default "$CHART_DIR" $COMMON_SETS > "$TMP_DIR/default.yaml"
+if grep -E '^\s+nodePort:' "$TMP_DIR/default.yaml" >/dev/null; then
+  fail_msg "default ClusterIP render unexpectedly includes nodePort"
+fi
+
+# Explicit NodePort + fixed ports must appear on api and proxy Services.
+helm template nodeport-fixed "$CHART_DIR" $COMMON_SETS \
+  --set controlPlane.api.service.type=NodePort \
+  --set controlPlane.api.service.nodePort=30030 \
+  --set cubeProxy.service.type=NodePort \
+  --set cubeProxy.service.nodePorts.http=30080 \
+  --set cubeProxy.service.nodePorts.https=30443 \
+  --set cubeProxy.service.nodePorts.grpc=30090 \
+  --set cubeProxy.service.nodePorts.admin=30082 \
+  --set cubeProxy.ingress.enabled=false \
+  > "$TMP_DIR/fixed.yaml"
+
+awk '
+  /^kind: Service$/ { in_svc=1; name=""; component=""; next }
+  in_svc && /^  name: / { name=$2 }
+  in_svc && /app.kubernetes.io\/component: / { component=$2 }
+  in_svc && /^---$/ { in_svc=0 }
+  in_svc && name != "" && component == "api" && /nodePort: 30030/ { api_np=1 }
+  in_svc && name != "" && component == "cube-proxy" && /nodePort: 30080/ { proxy_http=1 }
+  in_svc && name != "" && component == "cube-proxy" && /nodePort: 30443/ { proxy_https=1 }
+  in_svc && name != "" && component == "cube-proxy" && /nodePort: 30090/ { proxy_grpc=1 }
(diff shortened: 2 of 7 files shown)


xiaojunxiang2023:
Thanks for the thorough review! Agreed all findings are non-blocking. We'll keep the current fail-fast / default-range behavior for this PR and consider the admin test coverage + docs nits in a follow-up if useful.

deploy/kubernetes/chart/templates/validate.yaml
```diff
@@ -129,6 +141,20 @@
 {{- if and (eq (.Values.cubeProxy.service.type | default "ClusterIP") "LoadBalancer") (hasPrefix "CHANGE_ME" $clbSecurityGroups) }}
 {{- fail "cubeProxy.service.annotations[\"service.cloud.tencent.com/security-groups\"] still equals the CHANGE_ME_* sentinel from values-tke.yaml. Override it with your VPC security group id(s) (comma-separated), or set it to \"\" if you manage CLB security groups outside Helm." }}
 {{- end }}
+{{- $proxySvcType := .Values.cubeProxy.service.type | default "ClusterIP" -}}
+{{- $proxyNodePorts := ((.Values.cubeProxy).service).nodePorts | default dict -}}
+{{- range $name, $raw := $proxyNodePorts }}
```
fslongjin, line 147: `range` validates every key under `cubeProxy.service.nodePorts`, but `proxy-service.yaml` only reads `http` / `https` / `grpc` / `admin`.

A typo like `nodePorts.htttp: 30080` (with `type=NodePort`) currently passes validate and renders successfully — yet none of the Service ports get a fixed `nodePort`, so the pin silently becomes auto-allocation.

Suggest either:
- `fail` on unknown keys (allow-list the four names), or
- only run the type/range checks for the known keys the template actually consumes.

A small guard case in `test-service-nodeport.sh` for an unknown key would lock this in.

deploy/kubernetes/chart/scripts/test-service-nodeport.sh
```diff
@@ -0,0 +1,83 @@
...
+
+awk '
+  /^kind: Service$/ { in_svc=1; name=""; component=""; next }
+  in_svc && /^  name: / { name=$2 }
+  in_svc && /app.kubernetes.io\/component: / { component=$2 }
+  in_svc && /^---$/ { in_svc=0 }
+  in_svc && name != "" && component == "api" && /nodePort: 30030/ { api_np=1 }
+  in_svc && name != "" && component == "cube-proxy" && /nodePort: 30080/ { proxy_http=1 }
+  in_svc && name != "" && component == "cube-proxy" && /nodePort: 30443/ { proxy_https=1 }
+  in_svc && name != "" && component == "cube-proxy" && /nodePort: 30090/ { proxy_grpc=1 }
```
fslongjin, line 48: `values.yaml` / `proxy-service.yaml` already support `cubeProxy.service.nodePorts.admin`, but this fixed-port assertion only checks api + http/https/grpc.

Worth adding `--set cubeProxy.service.nodePorts.admin=30082` (or similar) and an awk check for `nodePort: 30082` on the cube-proxy Service, so admin pinning cannot regress unnoticed.

deploy/kubernetes/chart/README.md
```diff
@@ -443,6 +443,8 @@ CubeProxy runs on the **Pod network** (no `hostNetwork`). Traffic path:
 
 TLS for `cube.app` / wildcards still terminates **inside CubeProxy**. The default Ingress annotations enable nginx-ingress SSL passthrough + HTTPS backend; override `cubeProxy.ingress.className` / `annotations` for TKE CLB or other controllers. Set `cubeProxy.ingress.enabled=false` if you manage the entrypoint yourself (keep the Service as backend).
 
+Without an Ingress / cloud LB, set `cubeProxy.service.type` / `controlPlane.api.service.type` to `NodePort` (or `LoadBalancer`) and optionally pin host ports via `cubeProxy.service.nodePorts.*` / `controlPlane.api.service.nodePort` (Kubernetes range `30000-32767`; empty keeps auto-allocation). Explicit `nodePort` values are rejected when `type` is still `ClusterIP`.
```
fslongjin, line 446: Related nit: `templates/validate.yaml` still fails `cubeProxy.hostNetwork` with “behind a ClusterIP Service + Ingress”. That wording conflicts with this new NodePort / no-Ingress guidance — worth updating the fail string in the same change or a quick follow-up.

fslongjin (commented):
Additional review notes beyond the earlier auto-review — focused on validate/template consistency and test coverage. The NodePort+admin co-exposure on the shared Service is acknowledged and intentionally not blocking for this PR.

Note: the hostNetwork fail-message nit is anchored on the new README NodePort paragraph because the stale string itself is outside this PR’s diff.

deploy/kubernetes/chart/templates/validate.yaml
xiaojunxiang2023: Fixed — validate now allow-lists `http` / `https` / `grpc` / `admin` and fails on unknown keys (e.g. `htttp`). Added a guard case in `test-service-nodeport.sh`. Force-pushed in e99344d.

deploy/kubernetes/chart/scripts/test-service-nodeport.sh
xiaojunxiang2023: Done — the fixed-port render now pins `admin=30082` and asserts it. Force-pushed in e99344d.

deploy/kubernetes/chart/README.md
xiaojunxiang2023: Updated the `cubeProxy.hostNetwork` removal fail-string to mention Service (`ClusterIP` / `NodePort` / `LoadBalancer`; Ingress optional) so it matches the NodePort guidance. Force-pushed in e99344d.

deploy/kubernetes/chart/templates/validate.yaml
```diff
@@ -106,6 +106,18 @@
 {{- if and (eq (include "cube.redisBuiltinEnabled" .) "true") (hasPrefix "CHANGE_ME" (default "" .Values.redis.password)) }}
 {{- fail "redis.password still equals the CHANGE_ME_* default sentinel. Override it before deploying (or set redis.host to use external Redis)." }}
 {{- end }}
+{{- /* Fixed Service nodePorts are only valid for NodePort / LoadBalancer. */ -}}
+{{- $apiSvcType := .Values.controlPlane.api.service.type | default "ClusterIP" -}}
+{{- $apiNodePort := .Values.controlPlane.api.service.nodePort | default "" | toString -}}
+{{- if and (ne $apiNodePort "") (not (has $apiSvcType (list "NodePort" "LoadBalancer"))) }}
```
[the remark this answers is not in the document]
xiaojunxiang2023: Thanks — leaving the api `nodePort` guard unconditional intentionally so a leftover value in a shared values file still fails fast when api is toggled off (external control plane). Happy to gate it later if that proves noisy in practice.

deploy/kubernetes/chart/templates/validate.yaml
```diff
@@ -106,6 +106,18 @@
...
+{{- fail (printf "controlPlane.api.service.nodePort requires service.type NodePort or LoadBalancer (got %s)" $apiSvcType) }}
+{{- end }}
+{{- if ne $apiNodePort "" }}
+{{- $apiNP := $apiNodePort | int -}}
```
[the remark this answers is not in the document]
xiaojunxiang2023: Good catch on Sprig `int` / base-0. Agreed it's Low polish — will consider a `^[0-9]+$` pre-check in a follow-up rather than re-rolling this PR for it.

deploy/kubernetes/chart/templates/proxy-service.yaml
```diff
@@ -20,16 +20,28 @@ spec:
...
       targetPort: grpc
       protocol: TCP
+      {{- with (((.Values.cubeProxy).service).nodePorts).grpc }}
+      nodePort: {{ . }}
+      {{- end }}
     - name: admin
       port: {{ .Values.cubeProxy.adminPort }}
       targetPort: admin
       protocol: TCP
+      {{- with (((.Values.cubeProxy).service).nodePorts).admin }}
```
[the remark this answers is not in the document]
xiaojunxiang2023: Acknowledged — `nodePorts.admin` is opt-in and token-protected; we'll call out the host exposure in docs if we expand the bare-metal profile notes later.

fslongjin (approved):
LGTM

PR #1291, opened 2026-08-05
[Bugfix] Honor snake_case CubeMaster log rotation settings

### LINKED ISSUE #1289: [Bug Report] CubeMaster ignores file_size/file_num log rotation settings
## Summary

The shipped CubeMaster configuration uses snake_case log rotation keys, but the Go config tags use camelCase. With yaml.v3, file_size and file_num are not unmarshaled into log.Conf, so the rotation writer receives zero values and the active log files grow without a bound.

## Environment

- CubeSandbox version / commit: v0.6.0 / 8721dd151971ce3c2966482bbd32904ad98f378e
- Deployment mode: single-node
- Relevant component: CubeMaster

## Steps to Reproduce

1. Start CubeMaster with the shipped configuration:

   ~~~yaml
   log:
     module: "cubemaster"
     path: "/data/log/CubeMaster"
     file_size: 100
     file_num: 10
     level: "info"
   ~~~

2. Inspect CubeMaster/pkg/base/log/Conf: the fields currently use yaml:"fileSize" and yaml:"fileNum".
3. Observe that FileSize and FileNum remain zero after unmarshaling the shipped keys.

## Expected Behavior

The shipped file_size and file_num settings should be honored. The active file should rotate at the configured size and retain the configured number of historical files.

## Actual Behavior

The rotation writer is constructed with zero size and file-count values. In a long-running node, the active files remain unrotated (for example, cubemaster-req.log grows while no .log.N files are created).

## Proposed Fix

Use yaml:"file_size" and yaml:"file_num" on log.Conf, and add a regression test that unmarshals the shipped keys. This matches the existing configs under CubeMaster/conf.yaml, configs/single-node/cubemaster.yaml, and the Kubernetes chart.

ZhengkaiWang:
## Summary
Fix CubeMaster log rotation settings being ignored when using the shipped configuration.

## Root cause
The shipped YAML uses `file_size` and `file_num`, but `log.Conf` used `fileSize` and `fileNum` YAML tags. `yaml.v3` therefore leaves both values at zero, so `cubelog` receives no rotation size or count.

## Changes
- use the shipped snake_case keys in the Go struct tags
- add a regression test for unmarshalling the rotation settings

Closes #1289

## Validation
- `go test ./pkg/base/log` (not run locally: the Go toolchain is unavailable in this environment)

### DIFF
--- a/CubeMaster/pkg/base/config/config_test.go
+++ b/CubeMaster/pkg/base/config/config_test.go
@@ -15,6 +15,7 @@ import (
 	"github.com/stretchr/testify/assert"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/log"
+	"gopkg.in/yaml.v3"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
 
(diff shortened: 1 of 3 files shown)


luzhixing12345:
Good catch. Please squash all commits into a single commit @ZhengkaiWang.

Issue #1290, opened 2026-08-05
[Bug Report] CubeVmm and CubeShim logs need bounded rotation and safe reopen


ZhengkaiWang:
## Summary

CubeVmm and CubeShim write to long-lived shared host log files, but their current reopen paths do not provide a bounded rotation and retention contract. A normal rename-based logrotate job can leave one or more processes writing to the old inode until they explicitly reopen it; copytruncate is unsafe for concurrent writes.

## Environment

- CubeSandbox version / commit: v0.6.0 / 8721dd151971ce3c2966482bbd32904ad98f378e
- Deployment mode: single-node
- Relevant components: CubeShim and hypervisor (CubeVmm)

## Steps to Reproduce

1. Run CubeSandbox with multiple live sandboxes and inspect:

   ~~~text
   /data/log/CubeShim/cube-shim-req.log
   /data/log/CubeShim/cube-shim-stat.log
   /data/log/CubeVmm/vmm.log
   ~~~

2. In CubeShim/shim/src/log/mod.rs, observe that LogType::Rotate only breaks the writer loop; the next loop opens the same active pathname and does not archive or enforce a size limit.
3. In hypervisor/logging/src/lib.rs, observe that PlainDecorator reopens on the time bucket changing, but does not detect that the active pathname was replaced by an external rotator.
4. Apply a standard size-based logrotate policy and rename the active file. With an already-open writer, subsequent writes can continue to the old inode until that process reopens.

## Expected Behavior

CubeVmm and CubeShim should have an explicit, documented rotation contract that keeps active files bounded, retains a configured number of archives, and safely reopens after rename-based rotation.

## Actual Behavior

The current code only reopens the same pathname. It does not perform archival, size checks, or retention, and there is no documented host-side configuration that is safe for all long-lived writers. On one v0.6.0 development node after about one month, vmm.log reached about 20.9 GiB and cube-shim-req.log about 6.6 GiB with no rotated archives.

## Proposed Fix

1. Make CubeShim and the VMM detect an active-file inode replacement and reopen the newly-created pathname.
2. Document and/or ship a logrotate policy using rename + create, compression, and retention; explicitly reject copytruncate.
3. Add focused tests that rename the active file, create a new active file, write again, and verify that the old and new records land in the expected files.

Related: #1136 proposes a broader per-Sandbox stdout/stderr log-source lifecycle. This issue is narrower: it covers the host-wide CubeShim request/stat files and the CubeVmm log.

Issue #1136, opened 2026-07-24
[Feature Request] Proposal: bounded Sandbox logs and a log-source lifecycle extension


mj37yhyy:
## Background

CubeShim already forwards the init process stdout/stderr from the guest to node-local files. This is a suitable source for external collectors, but two production capabilities are missing:

1. stdout/stderr files have no built-in size limit, rotation, or retention policy;
2. a single cluster-wide collector configuration cannot route different Sandboxes to different destinations.

This proposal is generic: it does not introduce a log vendor, cloud provider, or sidecar dependency.

Related work: [#642 — CubeSandbox Webhook Event Notifications](https://github.com/TencentCloud/CubeSandbox/issues/642) already proposes asynchronous `sandbox.created` and `sandbox.deleted` events. This proposal does not replace that work. It adds the log-source information and pre-cleanup timing required by node-local collectors.

## 1. Managed stdout/stderr rotation

CubeShim should manage Sandbox stdout and stderr files itself.

Suggested node configuration:

```toml
[sandbox_log]
max_size = "100Mi"
max_files = 5
```

When either stream reaches `max_size`, CubeShim should close the current file, retain at most `max_files` historical files, rename the old file, create a new current file, and reopen it for append. stdout and stderr must remain independent.

External `logrotate` is not sufficient: CubeShim keeps the file descriptor open, so rename-based rotation continues writing to the old inode; `copytruncate` can lose or duplicate data during concurrent writes.

The resulting node-side source should be documented as a versioned contract:

```text
type: cube.sandbox.init-process-log.v1
streams: stdout, stderr
scope: container init process only
```

The contract should define the source location/access method, rotation and retention behavior, pause/resume behavior, deletion cleanup, and best-effort/drop semantics under guest-to-host pipe congestion.

## 2. Extend lifecycle Webhooks with log-source events

The lifecycle Webhook proposed in #642 should be extended so an external control plane can configure collection for each Sandbox independently.

| Event | Trigger | Use |
|---|---|---|
| `sandbox.log_source.ready` | Sandbox creation succeeds and the source is available | Create the Sandbox-specific collector assignment |
| `sandbox.log_source.delete_requested` | Before log files and bundle are removed | Stop or drain collection, then remove the assignment |
| `sandbox.log_source.removed` | Source cleanup completes | Confirm external cleanup |

[...]

PR #1292, opened 2026-08-05
[Bugfix] Reopen CubeShim and VMM logs after rename-based rotation

ZhengkaiWang:
## Summary

Keep CubeShim and VMM log files bounded with host-side hourly rotation, retention, and periodic reopen without a service restart.

## Root cause

Both writers keep log files open. Host-side `rename + create` rotation replaces the pathname while the process still holds the old file descriptor, so each writer needs a safe reopen path that does not discard writes when the replacement is temporarily unavailable.

## Changes

- keep CubeShim's existing internal `Rotate` event and reopen both log files
- preserve the VMM's existing wall-clock-hour reopen on the first log write after the hour changes
- compare the VMM pathname and held descriptor by device/inode at most once per second, so rename-based rotation is detected on a subsequent write without waiting for the next hour
- let the VMM control thread own a monotonic `timerfd` and emit the existing `LOG_CTRL_REOPEN` control record once per hour
- keep the current descriptor usable when reopen fails; retry on later writes with a one-second cooldown
- open replacement files before swapping descriptors and let host-side `logrotate create` own replacement-file creation
- keep writer-created files subject to the process umask
- document an hourly `/etc/logrotate.d/cubesandbox` policy using `rename + create`, `delaycompress`, compression, and host-owned retention
- explicitly disallow `copytruncate`

The VMM timer is created and consumed by the VMM control thread after its VMM-specific seccomp filter is installed; it is not created from deferred logger initialization or a vCPU/API thread. The pathname/device/inode check remains a write-triggered fast path for arbitrary external rotations. This PR does not impose a fixed byte-size limit such as 100M; host-side rotation and retention remain the deployment policy.

Related to #1290
Related to #1136

## Validation

- `cargo fmt --all -- --check`
- `cargo test -p logging --lib`
- `cargo check --target x86_64-unknown-linux-gnu -p cube-hypervisor --lib`
- `cargo check --target x86_64-unknown-linux-gnu -p containerd-shim-cube-rs --tests`

### DIFF
--- a/hypervisor/vmm/src/lib.rs
+++ b/hypervisor/vmm/src/lib.rs
@@ -31,7 +31,8 @@ use crate::vm_config::{
 };
 use anyhow::anyhow;
 use event_notifier::{event_notify, NotifyEvent};
-use libc::{EFD_NONBLOCK, SIGINT, SIGTERM};
+use libc::{EFD_NONBLOCK, SIGINT, SIGTERM, TFD_NONBLOCK};
+use logging::LOG_CTRL_REOPEN;
 use memory_manager::MemoryManagerSnapshotData;
 use pci::PciBdf;
 use seccompiler::{apply_filter, SeccompAction};
@@ -51,7 +52,7 @@ use std::rc::Rc;
 use std::sync::atomic::AtomicBool;
 use std::sync::mpsc::{Receiver, RecvError, SendError, Sender};
 use std::sync::{Arc, Barrier, Mutex};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use std::{result, thread};
 use thiserror::Error;
 use tracer::trace_scoped;
@@ -118,6 +119,14 @@ pub enum Error {
     /// Cannot read from EventFd.
     #[error("Error reading from EventFd: {0}")]
     EventFdRead(#[source] io::Error),
+
+    /// Cannot create the periodic log-reopen timer.
+    #[error("Error creating log-reopen timer: {0}")]
+    LogReopenTimerCreate(#[source] io::Error),
+
+    /// Cannot consume the periodic log-reopen timer event.
+    #[error("Error reading log-reopen timer: {0}")]
+    LogReopenTimerRead(#[source] io::Error),
 
     /// Cannot create epoll context.
     #[error("Error creating epoll context: {0}")]
@@ -197,6 +206,7 @@ pub enum EpollDispatch {
     Api = 2,
     ActivateVirtioDevices = 3,
     Debug = 4,
+    LogReopen = 5,
     Unknown,
 }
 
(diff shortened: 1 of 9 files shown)


lisongqian:
Thank you for your contribution. Please fix your commit messages first. Here is some info:  https://github.com/TencentCloud/CubeSandbox/blob/master/CONTRIBUTING.md#commit-messages
And some similar commits can be squashed.

lisongqian (changes requested):
In fact, `hypervisor` supports passing `LOG_CTRL_REOPEN` to notify logger thread to reopen file. Perhaps we can simplify the existing plan. How about using a timefd if we just need to reopen per hour?

lisongqian:
IMO, the changes to `shim` LGTM but `hypervisor` can do better. How about creating a timerfd and listening for this event in the vmm thread? Please let me know if you have a better idea. @ZhengkaiWang

ZhengkaiWang:
Thanks for the suggestion. I checked the seccomp layout against the earlier timerfd prototype, and the risk is real for that implementation:

- `timerfd_create` and `timerfd_settime` are allowed in `vmm_thread_rules` (`hypervisor/vmm/src/seccomp_filters.rs:612-613`), but are not allowed in `vcpu_thread_rules` or `api_thread_rules`.
- With deferred logger initialization, `build_logger_thread()` can be reached from whichever thread emits the first log after `vcpu_started` becomes true (`hypervisor/src/common.rs:129-157`). If the timer thread is spawned from that logger path and the caller is a vCPU/API thread, it inherits the restricted seccomp filter: under the default `KillProcess` action this can terminate the process; under `Trap`/`Log` the timer setup can fail and log reopening stops.

The current PR therefore does not add a timer thread. Your proposed design is different and can avoid this issue if the timerfd is created and its loop is spawned from the known VMM thread (or before per-thread filters are applied), rather than from the first logging caller. I agree that this is a safer place for a timerfd. Please confirm that VMM-thread ownership is the intended design; then I can rework the hypervisor side around that boundary.

lisongqian:
@ZhengkaiWang  I think it's okay to use VMM threads as controllers for the behaviour of `hypervisor` : )

zhuangel (commented):
There already has the timerfd path to reopen logfile, so the log reopen logic in `reopen_if_needed` checks in `RawPlainDecorator` and `PlainRecordDecorator` could be deleted, let `PlainRecordDecorator` just do log write works.

Could you please help to verify this.

zhuangel (commented):
There already has the timerfd path to reopen logfile, so the log reopen logic in `reopen_if_needed` checks in `RawPlainDecorator` and `PlainRecordDecorator` could be deleted, let `PlainRecordDecorator` just do log write works.

Could you please help to verify this.

hypervisor/vmm/src/lib.rs
```diff
@@ -209,11 +218,46 @@ impl From<u64> for EpollDispatch {
             2 => Api,
             3 => ActivateVirtioDevices,
             4 => Debug,
+            5 => LogReopen,
             _ => Unknown,
         }
     }
 }
 
+const LOG_CTRL_REOPEN: &str = "LogReopen";
```
zhuangel: use logging::{LOG_CTRL_REOPEN};

ZhengkaiWang:
@zhuangel Thanks for the suggestion. I verified the timerfd-only simplification against the current logging path.

The timerfd remains the only scheduled hourly reopen mechanism, and I removed the duplicate wall-clock check. However, deleting `reopen_if_needed` entirely is not safe in the current design:

- With the default `-v 0`, the original `info!(target: LOG_CTRL_REOPEN, ...)` is filtered by `LevelFilter::Warn` before it reaches the logger.
- The control record uses the `slog_async` queue, configured with `OverflowStrategy::Drop`, so it can be lost during a log burst.

I changed the control record to `warn!` and kept only the once-per-second device/inode check as a narrow fallback for an external rename/create or a dropped control record. It is no longer a second scheduled reopen path; the normal path is still just log writing, with the fallback check only when due. The focused logging tests and the Linux-target `cube-hypervisor` check pass at `5e1ab61`.

This seems to be the smallest safe version of the simplification while preserving the existing rotation behavior.

lisongqian:
@ZhengkaiWang Thank you for your thoughts! Here is my view:

> I changed the control record to `warn!`

Agree with you.
>  kept only the once-per-second device/inode check as a narrow fallback for an external rename/create or a dropped control record. It is no longer a second scheduled reopen path; the normal path is still just log writing, with the fallback check only when due. 

I think the fallback should be removed since no additional CPU overhead should be added for unexpected behavior. Reopen configuration should be kept the same as log rotation. A stable operating environment without external operations. The log of shim is the same.

lisongqian (approved):
LGTM

zhuangel (approved):
Almost done, could you please fix it @ZhengkaiWang.

lisongqian (approved):

PR #1293, opened 2026-08-05
fix(examples): improve code-sandbox-quickstart for local dev and create-time envs

AkatQuas:
## Summary
- Integrate optional `e2b-dev-sidecar` in `env_utils.py` when `CUBE_REMOTE_PROXY_BASE` is set, enabling data-plane access from outside the cluster without `*.cube.app` DNS
- Add `apply_create_time_envs()` workaround for cubebox/VNC templates where create-time `envs` do not survive entrypoint startup
- Update `create_with_envs.py` to use `printf` and re-apply envs after create; document sidecar env vars in `.env.example`
- Add `aiohttp`, `httpx`, and `cubesandbox` to quickstart requirements

## Test plan
- [x] `python3 cmd.py` with `.env` (`CUBE_REMOTE_PROXY_BASE=http://<node-ip>`)
- [x] `python3 create_with_envs.py` prints `session is user-session-test`
- [x] `python3 debug_local.py` (local helper, not included in this PR)


lkml-likexu:
@AkatQuas Thanks for your contribution and pls fix the DCO check.

tinklone:
Thanks for the updates. I found three remaining items that should be fixed before/around merge:

1. The DCO check is failing because the PR commits do not include `Signed-off-by` trailers. Please re-sign the commits, for example with `git rebase --signoff origin/master`, then push with `--force-with-lease`.

2. The docs mention `9090` as a common `CUBE_REMOTE_PROXY_BASE` port, but `9090` is the plaintext gRPC ingress, not an HTTPS CubeProxy endpoint. Please remove `9090` from the suggested HTTPS sidecar values in `.env.example`, `README.md`, and `README_zh.md`.

3. `apply_create_time_envs()` reports `failed after 3 attempts` even when it exits early on a non-retryable error. Please track the actual number of attempts made and use that in the warning message.

PR #1303, opened 2026-08-06
docs(architecture): align CubeVS network model naming

zyl1121:
## Motivation

The architecture overview uses `Network Architecture` / `网络架构` for the CubeVS networking deep dive, while the linked pages and README use `CubeVS Network Model` / `CubeVS 网络模型`. This mismatch makes the terminology inconsistent across the English and Chinese documentation.

## What Changed

* Rename the English and Chinese network-document headings to `CubeVS Network Model` / `CubeVS 网络模型`.
* Update both architecture overview references in each language to use the matching names.
* Update the English navigation sidebar entry (`docs/.vitepress/config.mjs`) to the same name.
* Keep the existing technical content and relative links unchanged.

## Validation

* Ran `git diff --check`.
* Reviewed the updated English and Chinese overview links and network-page headings for consistent terminology.

This documentation-only change does not affect runtime behavior.


docs/.vitepress/config.mjs
```diff
@@ -218,7 +218,7 @@ export default withMermaid(defineConfig({
               text: 'System Design',
               items: [
                 { text: 'Architecture Overview', link: '/architecture/overview' },
-                { text: 'Networking (CubeVS)', link: '/architecture/network' }
+                { text: 'CubeVS Network Model', link: '/architecture/network' }
```
chenhengqi, line 221: I think the original one is better. We should probably update the README instead.

zyl1121: > I think the original one is better. We should probably update the README instead.

 The PR is already merged, should we revert it, or will you open another PR to fix that?

PR #1304, opened 2026-08-06
chore(cubeops): remove accidentally committed build binary

zyl1121:
## Motivation

`CubeOps/bin/cubeops` is a 42 MB Linux build artifact that was accidentally committed in #984, even though `CubeOps/bin/` is already covered by `.gitignore`. Committed binaries bloat the repository and inevitably drift from freshly built ones; the binary is produced by the build pipeline at release time.

## What Changed

* Remove the tracked `CubeOps/bin/cubeops` build artifact.
* Keep `CubeOps/bin/` in `.gitignore` so local builds continue to drop artifacts there without polluting git.

## Validation

* `git diff --check` passed
* confirmed that no code or build script references the tracked `CubeOps/bin/cubeops` path, and that the binary is built by the root Makefile (`_output/bin/cubeops`) and packaged by `build-release-bundle.sh` at release time

## Scope

This PR only removes the accidentally committed binary. It does not change the build pipeline, release packaging, or runtime behavior.


lkml-likexu:
@liciazhu  How could something like this happen?

liciazhu:
> @liciazhu How could something like this happen?

My bad—those are compilation test outputs that were mistakenly committed.

zyl1121:
@lkml-likexu I think we should improve the Claude auto-review prompt and add a workflow precheck, so that unexpected files like the `1.md` removed in #1043, or large build artifacts like the one removed in this PR can be caught earlier.

1. **Claude auto-review prompt**

   Claude should look at the full changed-file list, not just the text diff, and call out suspicious files such as unexpected additions, generated artifacts, files in unusual locations, large files, and binary/blob files.

2. **Workflow precheck**

   The workflow should fail if an added or modified file exceeds the configured size threshold, or if it is detected as binary/blob, unless it is explicitly allowlisted. The error message can print the file path and explain how to add an exception.

This way Claude can help notice suspicious cases, while the workflow makes sure the basic file rules are always checked.

lkml-likexu:
@zyl1121 Feel free to fire a PR at any time to update .claude/agents.

PR #1313, opened 2026-08-08
docs(agents): flag unusual PR file changes

zyl1121:
## Motivation

This PR follows the suggestion accepted by @lkml-likexu during the review of #1304: improve the Claude auto-review guidance so unexpected files such as the `1.md` removed in #1043 or the 42 MB `CubeOps/bin/cubeops` build binary committed in #984 can be noticed before merge.

Unusual files can still be valid when intentional, so the reviewer should report the file facts and ask for justification rather than assume they must be removed.

## What Changed

- Require the CubeSandbox project reviewer to inspect the complete changed-file inventory before focusing on the rendered text diff.
- Review added, modified, renamed, and deleted paths for unrelated, misplaced, generated, transient, machine-local, tool-produced, or unexplained opaque files.
- Treat paths, extensions, diff metadata, size, and content-type indicators as supporting evidence rather than definitive proof of type or intent.
- State when the complete inventory or supporting evidence is unavailable instead of asserting that the check passed.
- Ask the author to remove unintended files or explain why intentional files belong in the pull request.

## Validation

- Verified that the auto-review context supplies a complete changed-file inventory and that the guidance fails explicitly when another review context does not.
- `git diff --check` passed.
- Confirmed that the final diff only updates `.claude/agents/cube-project-reviewer.md`.

## Scope

This PR only updates the CubeSandbox project-reviewer guidance. It does not add a deterministic workflow, change repository checks, or block intentional files.

### DIFF
--- a/.claude/agents/cube-project-reviewer.md
+++ b/.claude/agents/cube-project-reviewer.md
@@ -1,6 +1,6 @@
 ---
 name: cube-project-reviewer
-description: Use this agent to review a pull request against CubeSandbox project-specific conventions and release gates that generic reviewers miss — OCI image multi-arch support, bilingual README coverage, feature-change test coverage, unit-test gate, orphaned project wiring, upgradability, terraform/k8s deployment design, Fix regression tests, workflow/CODEOWNERS maintenance, non-English character policy, and commit-message convention.
+description: Use this agent to review a pull request against CubeSandbox project-specific conventions and release gates that generic reviewers miss — OCI image multi-arch support, bilingual README coverage, feature-change test coverage, unit-test gate, orphaned project wiring, upgradability, terraform/k8s deployment design, Fix regression tests, workflow/CODEOWNERS maintenance, non-English character policy, commit-message convention, and unusual file changes.
 tools: Glob, Grep, Read, WebFetch, TodoWrite, WebSearch, BashOutput, KillBash
 model: inherit
 ---
@@ -72,6 +72,17 @@ Flag any violation with the specific rule broken.
 
 **13. Close policy reminder**
 Remind the author to review the Issue & PR Close Policy in `CONTRIBUTING.md` (e.g. linking issues, closing semantics) so the PR follows project process.
+
+**14. Flagged file changes**
+Before focusing on the text diff, review the complete changed-file inventory supplied in the review context. If no complete inventory is available, say that this check could not be completed rather than assuming the rendered diff includes every file.
+
+Review every added, modified, renamed, and deleted path for changes that need explanation:
+- the file is unrelated to the pull request's stated purpose
+- its location or lifecycle conflicts with the repository structure, neighboring files, ignore rules, or established project conventions
+- it appears to be generated, transient, machine-local, or tool-produced output rather than an intentional source, configuration, documentation, or test input
+- its content is opaque or unavailable and the pull request does not explain why the file belongs
+
+Use paths, extensions, missing extensions, diff metadata, file size, and content-type indicators only as supporting evidence; none proves file type or intent by itself. Ask the author to remove an unintended change or explain why an intentional one belongs. State uncertainty when the available evidence is insufficient, and leave the final decision to maintainers.
 
 ## Output
 


lkml-likexu:
@zyl1121 Adding a block of Python code is a bit too much for this part of the task. I would prefer to simply update the MD file to add natural language rules for a (future, smarter) agent. If you insist on adding it, please check the review comments:

## Noteworthy issues

### 1. 530 lines of Python embedded in a YAML heredoc — untestable and unlintable *(most important)*
`pr-file-precheck.yml:194-722`

The evaluator is delivered to `python3` over stdin with no module name and no file on disk, so **there is no import path**: nothing can `import` `detect_signature`, `update_comment`, `valid_approvals`, or the fingerprint logic. In its current form the code is not just untested but structurally untestable in CI, and a syntax error would only surface at runtime on a live `pull_request_target` event. ruff/mypy also can't see it.

The logic that most needs coverage is exactly the branch-dense, security-relevant kind:
- `detect_signature` (14 magic-byte branches with offset math — the PE offset read `prefix[60:64]` / `prefix[offset:offset+4]` and the tar `>= 262` / `[257:262]` slices are classic off-by-one/short-input cases that silently return `None` = a missed binary).
- fingerprint / approval-id stability — a persisted **cross-run contract**; any drift in `finding_record` fields, the sort key, JSON separators, or the `\0` joiner silently invalidates every outstanding approval.
- `update_comment` state machine (pending → acknowledged → resolved, `same_episode` continuity) — getting `same_episode` wrong either loses a valid acknowledgement or (security-relevant) carries an approval across changed findings.
- `valid_approvals` permission gating — the security boundary; a regression lets an unauthorized user approve flagged content.

**Recommendation:** extract to `scripts/file_precheck.py` (or `.github/scripts/`) and invoke it hash-verified exactly like `scripts/gh.sh` already is in `code-review.yml` — the base-only checkout model fully supports this. Then add a focused `pytest` suite (the pure functions above need zero mocking; inject `api`/`permission_for` for the state-machine tests — this is precisely what the PR says it validated manually, so commit it) plus a `py_compile`/lint CI gate on `.github/scripts/**`. The PR's claimed "focused local detection tests" and "mock API comment-state tests" are not committed and are therefore unverifiable and non-reusable. Given the sensitivity (permission gating + persisted approvals), treat extraction + tests as a strong pre-merge ask.

### 2. Duplicated `parse_changed_files` / `-z` diff parsing across both workflows
`code-review.yml:69-101` and `pr-file-precheck.yml:385-405`

The NUL-parsing loop (split, trailing-empty pop, R/C two-field handling, surrogateescape decode) is copy-pasted into both workflows with subtle divergences (`previous = None` vs `""`; `previous_path`/`size_bytes` vs `previous`/`size`). Any git-format edge-case fix must be applied twice and can silently drift. Extracting per #1 would eliminate this.

### 3. `permission_for` detects HTTP 404 by substring-matching an error string
`pr-file-precheck.yml:500-503`

```python
if "failed: 404:" in str(error):
    return "none"
```

Brittle: false-positives if a path/detail body ever contains that literal, and silently stops working if the `api()` error-message format changes. Prefer a custom exception carrying `.code` and check `error.code == 404`. This gates the security boundary, so robustness matters.

### 4. New workflow uses `GITHUB_TOKEN` while the repo's other `pull_request_target` workflow uses a GitHub App token
`pr-file-precheck.yml:176` vs `code-review.yml:46` (`actions/create-github-app-token@v2`)

The choice is internally consistent (`bot_login = "github-actions[bot]"` matches `GITHUB_TOKEN`, and statuses/comments from `GITHUB_TOKEN` intentionally don't re-trigger workflows), but it diverges from the established pattern. Please confirm it's deliberate. Also note: `code-review.yml` carries a security-intent comment block at the top; the new file uses `pull_request_target` too but lacks an equivalent rationale comment — worth adding.

### 5. `.claude/agents/` has no CODEOWNERS entry
`.github/CODEOWNERS` (only `/.github/workflows/ @tinklone` + catch-all `* @fslongjin @ls-ggg @chenhengqi`)

The new workflow file is covered by the workflows rule, but this PR also edits `.claude/agents/cube-project-reviewer.md` — a governance-sensitive file that defines what the automated reviewer enforces — which falls only under the catch-all. Consider adding an explicit `/.claude/` (or `/.claude/agents/`) owner so review-policy changes require a designated owner.

---

## Minor

- **Doc schema wording** (`cube-project-reviewer.md:18`): the `files.txt` description ("status, previous path, and blob size") omits the primary `path` field and doesn't note that `size_bytes` is `null` for deletions/non-blob entries — so the agent shouldn't treat a missing size as an error. Also the "may be truncated" clause refers to `pr.diff`, not `files.txt` (files.txt is never truncated — that's why it exists). Suggest: "including status, path, previous path, and blob size (null for deletions and non-blob entries)."
- **Redundant approval check** (`pr-file-precheck.yml:513`): `command_re.fullmatch(normalized)` is dead after the exact `normalized != expected` equality (approval_id is always 64 hex chars). Harmless; can be simplified.
- **`size > size_limit` fragility** (`pr-file-precheck.yml:440`): correct only because the `object_type != "blob"` guard above ensures `size` is set; a future reorder reintroduces a `None > int` crash. Consider asserting non-None past the blob guard.
- **`render_findings` grouping** (`pr-file-precheck.yml:468-490`) merges only *adjacent* items and depends on the `inspect_findings` sort order; correct but subtle — a one-line comment noting the coupling would help.
- **Concurrency `group` vs job `if`** duplicate a ~10-line predicate (the `if` adds `github.event.issue.pull_request` the group omits). GitHub offers no way to share them, but a cross-reference comment ("keep in sync") reduces drift risk. Also: the bot's own duplicate-comment cleanup (`:662-666`) deletes marker-bearing comments, and a `deleted` event matches the shared `pr-file-precheck-{number}` group with `cancel-in-progress: true` — worth confirming it can't cancel a legitimately concurrent evaluation.

PR #1285, opened 2026-08-05
refactor(network): embed network-agent into Cubelet and simplify lifecycle

FakeLearne:
Replace the standalone network-agent process with an in-process Cubelet network runtime so TAP lifecycle, CubeVS, CubeEgress, port mapping, and startup recovery no longer depend on a separate daemon, gRPC/HTTP, or fdserver transport.

- Add Cubelet/network/runtime as the NetworkController: TapPool (Ready/Active/Cleaning only), StateStore under /dev/shm, PortBinder, CubeVS/CubeEgress adapters, sync TapCleaner, StartupRecover, and a maintenance loop that retries durable cleanup once per tick.
- Wire CubeletNetworkPlugin Create/Destroy/CleanUp to EnsureNetwork / ReleaseNetwork; expose RegisterHTTP for CubeEgress policy dump and tap listing; hand shim TAP fds via GetTapFile one-shot duplicates over cubetap.sock (no plugin-side fd cache).
- Remove network-agent module, pkg/networkagentclient, enable_network_agent config, and related one-click / k8s / docs / CI deploy paths.
- Keep one-shot compatibility for old network-agent state files during upgrade recovery; restrict /v1/network/taps to loopback.
- Retain TAP fds across pool/active handoffs to remove create hot-path burst convoys; collapse PortBinder Reserve/Commit into allocate-as-owned with ReleaseOwnership; drop unused GetNetwork/ListNetworks APIs.
- Move CubeEgress loopback admin off :9090 to :9091 (configurable via `CUBE_EGRESS_ADMIN_PORT`) so it no longer collides with CubeProxy plaintext gRPC on 9090; wire the new default through one-click, Helm, and Cubelet `cube_egress_admin_url`.

PR #1318, opened 2026-08-11
perf/fix(network): cut policy apply latency and harden TAP create under density

FakeLearne:
## Summary
Follow-up to the network-agent → Cubelet embed refactor (#1285). This PR tightens create-path latency when network policy is enabled, and hardens TAP allocation under high-concurrency density load.
- **CubeVS policy hot path:** cache HashOfMaps inner maps and avoid per-call outer `Lookup`s that trigger `synchronize_rcu` (tens of ms). Delete policy maps on TAP destroy, and limit the DNS reaper to live ifindices so create latency with policy stays near the no-policy baseline.
- **TAP create under density:** retry dump-style netlink reads on `ErrDumpInterrupted` / `EINTR`, and stop doing a full `LinkList` on every pool-miss create — `cleanupConflictingTap` only needs a by-name lookup. This stops `EnsureNetwork` failing with "results may be incomplete or inconsistent" under concurrent create pressure.
- **Cleanup:** remove an unused Cubelet helper left over from the above changes.
- **cube-bench:** extend network-policy / warmup coverage so the latency and density paths are easier to exercise in CI/local benches.

### DIFF
--- a/CubeNet/cubevs/dns_reaper.go
+++ b/CubeNet/cubevs/dns_reaper.go
@@ -23,7 +23,13 @@ func reapDNSState() {
 	reapDNSQueryTrack(now)
 }
 
-// reapDNSLearnedPolicies scans allow_out_v3 and removes expired DNS-learned entries.
+// reapDNSLearnedPolicies removes expired DNS-learned allow_out_v3 entries.
+//
+// The fast path iterates ifindex_to_mvmmeta so Ready-pool TAPs (metadata
+// deleted, allow_out flushed) are skipped without walking every HashOfMaps
+// outer key. If metadata cannot be loaded, fall back to iterating allow_out_v3
+// outers so a transient pin/ENOENT failure cannot stall DNS TTL expiry for
+// every subsequent tick.
 func reapDNSLearnedPolicies(now uint64) {
 	allowOut, err := loadPinnedMap(MapNameAllowOutV3)
 	if err != nil {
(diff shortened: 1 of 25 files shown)


CubeNet/cubevs/dns_reaper.go
```diff
@@ -35,13 +41,46 @@ func reapDNSLearnedPolicies(now uint64) {
...
+	defer meta.Close()
+
 	var (
-		ifindex    uint32
-		innerMapID uint32
+		ifindex uint32
+		mvmMeta mvmMetadata
+	)
+	iter := meta.Iterate()
+	for iter.Next(&ifindex, &mvmMeta) {
```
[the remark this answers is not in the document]
FakeLearne: AttachFilter → initNetPolicy unconditionally creates and caches allow_out_v2 before the reaper starts, so these sandboxes hit the cache rather than repeating outer lookups; skipping by DNSPolicyFlags could also leave stale learned entries unreaped.

Cubelet/network/runtime/systemnet/device_test.go
```diff
@@ -0,0 +1,82 @@
...
+func TestWithDumpRetrySucceedsAfterInterrupt(t *testing.T) {
+	calls := 0
+	got, err := WithDumpRetry(func() (int, error) {
+		calls++
+		if calls < 3 {
+			return 0, netlink.ErrDumpInterrupted
+		}
+		return 42, nil
+	})
+	if err != nil {
```
chenhengqi: Use testify packages for assertions.

Cubelet/network/runtime/startup_recover.go
```diff
@@ -71,6 +71,95 @@ func (s *NetworkController) recover() error {
...
+func (s *NetworkController) buildStaleNetPolicyKeepSet() map[uint32]struct{} {
+	keep := make(map[uint32]struct{})
+	if taps, err := s.tapAdapter.List(); err == nil {
+		for _, tap := range taps {
+			if tap != nil && tap.Index > 0 {
+				keep[uint32(tap.Index)] = struct{}{}
+			}
+		}
+	} else {
+		CubeLog.WithContext(context.Background()).Warnf(
```
chenhengqi: If we reach here, all inner maps will be deleted?

CubeNet/cubevs/netpolicy.go
```diff
@@ -135,29 +116,25 @@ func initNetPolicy(ifindex uint32) error {
...
+		return err
 	}
-	defer inner.Close()
+	return flushInnerEntries(inner, value)
+}
 
+func flushInnerEntries(inner *ebpf.Map, value any) error {
 	var key lpmKey
 	iter := inner.Iterate()
 	for iter.Next(&key, value) {
```
chenhengqi: We should use a 2-pass flush here, collect keys and then delete.

Cubelet/network/runtime/systemnet/device_test.go
FakeLearne: Done.

CubeNet/cubevs/netpolicy.go
FakeLearne: Done.

Cubelet/network/runtime/startup_recover.go
FakeLearne: Agreed — if List() fails the keep set is incomplete, so we now skip this GC pass entirely instead of treating live TAP ifindexes as stale.

chenhengqi (approved):

PR #1321, opened 2026-08-11
fix(cubenet): resolve direct on-link neighbors via ARP

### LINKED ISSUE #1209: [Bug Report] Direct mode unconditionally sets destination MAC to gateway MAC
## Summary

In CubeVS direct egress mode, all outbound traffic has its destination MAC hard-set to the node gateway MAC, bypassing the kernel routing/neighbor subsystem entirely. This is correct for off-link (public/cross-subnet) traffic, but breaks when the real destination is on the node's local subnet.

## Problem

The egress path fills the L2 destination address with a load-time gateway MAC constant for **every** destination, without checking whether the destination is on-link. When a sandbox reaches a host in the node's own subnet, the packet is still sent to the gateway, forcing it to forward the packet back out the same interface (hairpinning).

If the upstream gateway does not support hairpinning / same-interface L3 reflection, these flows are dropped.

## Impact

- **Off-link traffic** (public / cross-subnet): works correctly.
- **On-link traffic** (same subnet as node): broken on gateways without hairpinning support — typically traditional L2 networks or bare-metal with standard L3 switches.
- SDN/VPC environments (the intended deployment) are generally unaffected, since same-subnet traffic is handled by the distributed virtual router.

Note: this is a hairpinning dependency, not an ICMP redirect dependency — redirects are ignored by design and do not affect correctness.

## Suggested Direction

Add an on-link destination check in the egress path: for destinations within the node subnet, resolve the real neighbor MAC (or fall back to kernel routing) instead of hard-setting the gateway MAC.

jay3cx:
## Motivation

In direct egress mode, CubeVS sends outbound traffic to the node gateway MAC.

This works for off-link destinations, but on-link traffic then depends on the gateway forwarding the packet back through the same interface. On networks without that hairpin behavior, same-subnet destinations are unreachable.

## What Changed

- Pass the node's primary IPv4 mask to CubeVS to identify on-link destinations in direct mode.
- Add an LRU IP-to-MAC map for direct on-link neighbors.
- Try `bpf_fib_lookup` first. If the kernel has no neighbor, replace the packet with an ARP request and keep a pending map entry.
- Learn ARP replies only for IPs already present in the map.
- Once the MAC is learned, rewrite later packets with that MAC and send them out the node NIC.
- Keep the existing gateway-MAC path for off-link direct egress and route-aware egress.
- Keep the `from_world` dispatch path inlined so Linux 5.4 can load the DNS tail call.

A cache miss consumes the triggering packet (it becomes the ARP request). A later revalidation probe does the same. No userspace resolver or host iptables/sysctl setup is required.

## Validation

- Ran `go test -count=1 ./...` and `go vet ./...` in `CubeNet/cubevs`.
- Ran `go test -count=1 ./network/runtime/systemnet ./network/runtime` and `go vet ./network/runtime/...` in `Cubelet`.
- Tested on Ubuntu 20.04.6 with Linux `5.4.0-216-generic` on arm64:
  - Generated the CubeVS BPF objects with Go 1.24.8 and clang 14.
  - Loaded all generated BPF programs successfully.
  - Attached `from_cube` and `from_world` and confirmed that both were JIT-compiled.
  - Verified the datapath in isolated network namespaces using veth pairs:
    - On a cold neighbor miss, the first ICMP packet was converted into an ARP request.
    - The ARP reply populated the neighbor map, and the second ICMP packet succeeded.
    - On-link UDP and TCP traffic passed.
    - Packet capture showed on-link traffic using the peer MAC.
    - Off-link traffic continued to use the configured gateway MAC.

## Scope

This change only affects IPv4 on-link traffic in direct egress mode.

Off-link direct egress and route-aware egress keep their existing behavior.

Fixes #1209

Signed-off-by: jay3cx <137191587@qq.com>


CubeNet/cubevs/util_test.go
```diff
@@ -0,0 +1,40 @@
...
+			}
+		})
+	}
+}
+
+func TestIPMaskToUint32RejectsNonIPv4Mask(t *testing.T) {
+	if _, err := ipMaskToUint32(net.CIDRMask(64, 128)); err == nil {
+		t.Fatal("ipMaskToUint32 accepted an IPv6 mask")
+	}
+	if _, err := ipMaskToUint32(net.IPMask{255, 0, 255, 0}); err == nil {
```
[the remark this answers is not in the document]
jay3cx: This assertion does pass. net.IPMask.Size does not always return len(mask) * 8: when the mask is not in canonical form (ones followed by zeros), it returns (0, 0).

Therefore net.IPMask{255, 0, 255, 0} is rejected by the existing bits != 32 check. I also re-ran TestIPMaskToUint32RejectsNonIPv4Mask with Go 1.24.8, including on Linux 5.4.0-216-generic, and it passes.

Reference: https://pkg.go.dev/net#IPMask.Size

CubeNet/cubevs/util.go
```diff
@@ -26,6 +27,19 @@ func ipToUint32(ip net.IP) uint32 {
 	return uint32(ip[0]) | uint32(ip[1])<<8 | uint32(ip[2])<<16 | uint32(ip[3])<<24
 }
 
+// ipMaskToUint32 converts an IPv4 mask to the byte layout used by IPv4
+// addresses in BPF packet headers on little-endian hosts.
+func ipMaskToUint32(mask net.IPMask) (uint32, error) {
+	if len(mask) != net.IPv4len {
+		return 0, fmt.Errorf("invalid IPv4 mask length: %d", len(mask))
+	}
```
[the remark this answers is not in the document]
jay3cx: This branch is not dead. net.IPMask.Size returns (0, 0) when a mask is not canonical, rather than returning bits = len(mask) * 8 unconditionally.

For net.IPMask{255, 0, 255, 0}, Size returns (0, 0), so the bits != 32 condition rejects it as intended. The corresponding unit test passes with Go 1.24.8.

Reference: https://pkg.go.dev/net#IPMask.Size

CubeNet/src/mvmtap.bpf.c
```diff
@@ -648,23 +679,28 @@ static __noinline __attribute__((noinline)) __u32 do_udp_nat(struct __sk_buff *s
...
 
 /* Subprog-based version used by dns_finish. */
 static __always_inline int finish_udp_nat(struct __sk_buff *skb, struct mvm_meta *mvm_meta)
 {
+	__u32 daddr;
 	__u32 dst_ifindex = do_udp_nat(skb, mvm_meta);
 
-	if (dst_ifindex)
-		return bpf_redirect(dst_ifindex, egress_redirect_flags);
+	if (dst_ifindex) {
```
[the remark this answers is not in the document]
jay3cx: Validated on Ubuntu 20.04.6 with Linux 5.4.0-216-generic using veristat v0.3. The BPF objects were regenerated from the current patch, and all 8 programs loaded successfully.

Relevant verifier results:
- dns_finish: 41,639 processed instructions, 1,382 states, peak 196
- from_cube: 45,845 processed instructions, 1,032 states, peak 916
- largest result was dns_handle_response_prog at 190,156 processed instructions, also successful

The CubeVS unit-test binary built with Go 1.24.8 also passed on the same 5.4 kernel. No verifier-related code change was required.

Cubelet/network/runtime/systemnet/cube_dev.go
```diff
@@ -27,6 +33,58 @@ type CubeDev struct {
 	Mac   net.HardwareAddr
 }
 
+// ConfigureCubeDevHostRouting prepares cube-dev for the direct-egress on-link
+// fallback. Packets have already been SNATed to the node's local address, so
+// cube-dev must accept local sources, use loose reverse-path validation, and
+// forward packets into the host routing stack.
+func ConfigureCubeDevHostRouting(dev *CubeDev, host *HostDevice) error {
```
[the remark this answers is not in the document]
jay3cx: Confirmed. The supported Cubelet images include iptables and run privileged, so the required sysctls are writable. Failure is intentional because otherwise on-link forwarding would be broken.

No teardown is added here because the egress mode is startup-only and the rule is narrow and idempotent. Cleanup can be added if live mode switching is introduced.

jay3cx:
> ## Review: fix(cubenet): route direct on-link egress via host (PR #1321)
> 评审：修复(cubenet)：通过主机路由直接链路出口（PR #1321 ）
> _AI-generated review — not a human approval.AI 生成的评审——并非人工审批。_
> 
> ### Overview
> The PR fixes #1209: in direct-egress mode CubeVS always rewrote the destination MAC to the node gateway MAC, which breaks same-subnet traffic when the gateway doesn't hairpin. The approach:该 PR 修复了 #1209 ：在直接出口模式下，CubeVS 总是将目的 MAC 重写为节点网关 MAC，当网关不支持 hairpin 时，这会破坏同子网流量。解决方案：
> 
> * Propagates the primary node IPv4 mask into CubeVS (`nodenic_netmask`) and detects on-link destinations in BPF.将节点主 IPv4 掩码传播到 CubeVS（ `nodenic_netmask` ），并在 BPF 中检测链路直连目的地。
> * For on-link traffic, preserves the sandbox-side L2 pair, keeps the SNAT, and redirects into `cube-dev` ingress so the host routing/neighbor tables resolve the real destination MAC.对于直连流量，保留沙箱侧的 L2 地址对，维持 SNAT，并将其重定向至 `cube-dev` 入口，以便主机路由/邻居表解析出真实的目的 MAC 地址。
> * Preserves the gateway-MAC fast path for off-link traffic.对于非直连流量，保留网关 MAC 快速路径。
> * Configures `cube-dev` forwarding / `accept_local` / loose `rp_filter` and a narrow idempotent `FORWARD` iptables rule.配置 `cube-dev` 转发、 `accept_local` 、宽松的 `rp_filter` 以及一条窄范围的幂等 `FORWARD` iptables 规则。
> 
> The design is sound and well-integrated: the "SNAT then inject into cube-dev ingress" mechanism already exists in the codebase (L7-proxy path in `from_cube`, node-IP path), the per-protocol L2 preservation is correctly gated, the `dns_finish` rewrite to a single `goto` is behavior-preserving, and the Go-side defensive copies and test seams are good. The iptables helper (`runIptablesEnsure`) is genuinely idempotent.该设计合理且集成良好：“先 SNAT 再注入 cube-dev 入口”的机制已在代码库中存在（ `from_cube` 中的 L7 代理路径、节点 IP 路径），按协议保留 L2 的操作正确设置了门控， `dns_finish` 重写为单个 `goto` 保持了行为不变，Go 侧的防御性拷贝和测试接缝也做得很好。iptables 辅助函数（ `runIptablesEnsure` ）确实具有幂等性。
> 
> I found one confirmed defect (a failing unit test) plus two areas that should be addressed before merge.我发现了一个已确认的缺陷（一个失败的单元测试），另外还有两个问题需要在合并前处理。
> 
> ### Findings
> #### 1. HIGH — `TestIPMaskToUint32RejectsNonIPv4Mask` fails: non-contiguous masks are not rejected
> 1. 高优先级 — `TestIPMaskToUint32RejectsNonIPv4Mask` 失败：未拒绝非连续掩码
> `CubeNet/cubevs/util.go:35` — `ipMaskToUint32` only validates `len(mask) == 4` and that `mask.Size()` reports 32 bits. `net.IPMask.Size()` always returns `bits = len(mask)*8`, so **any** 4-byte mask passes the second check. The new test `TestIPMaskToUint32RejectsNonIPv4Mask` (`CubeNet/cubevs/util_test.go:37`) asserts that `net.IPMask{255, 0, 255, 0}` is rejected, but the implementation accepts it (returning `0x00ff00ff`) and the test fails with `"ipMaskToUint32 accepted a non-contiguous IPv4 mask"`. `CubeNet/cubevs/util.go:35` — `ipMaskToUint32` 仅验证 `len(mask) == 4` 以及 `mask.Size()` 报告为 32 位。 `net.IPMask.Size()` 始终返回 `bits = len(mask)*8` ，因此任何 4 字节掩码都能通过第二项检查。新测试 `TestIPMaskToUint32RejectsNonIPv4Mask` （ `CubeNet/cubevs/util_test.go:37` ）断言 `net.IPMask{255, 0, 255, 0}` 应被拒绝，但实现却接受了它（返回 `0x00ff00ff` ），导致测试以 `"ipMaskToUint32 accepted a non-contiguous IPv4 mask"` 失败。
> 
> This contradicts the PR's validation claim of `go test -count=1 ./...` in `CubeNet/cubevs`.这与 PR 中在 `CubeNet/cubevs` 中声称的 `go test -count=1 ./...` 验证逻辑相矛盾。
> 
> The gap also matters beyond the test: a non-contiguous mask would silently produce a wrong on-link comparison in `direct_egress_is_onlink` while the error message claims it was validated. Fix: validate contiguity (e.g., once a non-`0xff` byte is seen, every following byte must be `0`), or drop the second test assertion.测试之外，这个差距也很重要：不连续的掩码会在 `direct_egress_is_onlink` 中静默产生错误的直连比较，而错误信息却声称已验证。修复方法：验证连续性（例如，一旦看到非 `0xff` 字节，后续每个字节都必须是 `0` ），或者删除第二个测试断言。
> 
> #### 2. MEDIUM — New hard runtime dependency on `iptables` + `/proc/sys` writes in direct mode
> 2. 中等 — 在直接模式下，对 `iptables` 和 `/proc/sys` 的写入新增了硬性运行时依赖
> `Cubelet/network/runtime/systemnet/cube_dev.go:40` — `ConfigureCubeDevHostRouting` is now called unconditionally in the direct-egress branch of `initCubeVS`. Previously direct mode only ran `CleanupCubeRouter` and made no `iptables`/sysctl calls. Now a missing `iptables` binary or a read-only `/proc/sys/net/ipv4/conf/cube-dev/*` fails `initCubeVS`, which fails `newProductionControllerDeps`, which fails Cubelet startup. `Cubelet/network/runtime/systemnet/cube_dev.go:40` — `ConfigureCubeDevHostRouting` 现在在 `initCubeVS` 的直接出口分支中被无条件调用。此前，直接模式仅运行 `CleanupCubeRouter` ，且不进行任何 `iptables` /sysctl 调用。现在，缺少 `iptables` 二进制文件或只读的 `/proc/sys/net/ipv4/conf/cube-dev/*` 会导致 `initCubeVS` 失败，进而导致 `newProductionControllerDeps` 失败，最终导致 Cubelet 启动失败。
> 
> Also, the `FORWARD` rule and the three sysctls are never cleaned up when the node switches to route-aware mode or shuts down. The rule is narrow (`-i cube-dev -o <eth> -s <nodeIP>/32 -d <onlinkNet>`) so the residue is low-risk, but it should be documented or paired with a teardown.此外，当节点切换到路由感知模式或关闭时， `FORWARD` 规则和三个 sysctl 设置从未被清理。该规则范围较窄（ `-i cube-dev -o <eth> -s <nodeIP>/32 -d <onlinkNet>` ），因此残留风险较低，但应记录在案或配合拆除机制。
> 
> #### 3. MEDIUM — Linux 5.4 verifier budget risk on the `dns_finish` path
> 3. 中 — Linux 5.4 验证器预算在 `dns_finish` 路径上的风险
> `CubeNet/src/mvmtap.bpf.c:699` — the PR adds inlined logic (`direct_egress_is_onlink`, `redirect_egress`, plus a new `bpf_skb_load_bytes`) to `finish_udp_nat`, which is inlined into `dns_finish`. The codebase contains extensive comments explaining that `dns_finish` already sits close to the 5.4 1M-instruction verifier limit (that's exactly why `do_udp_nat` was made `__noinline`). `from_cube` likewise gains three inlined `redirect_egress` call sites and is constrained to have no bpf-to-bpf calls. `CubeNet/src/mvmtap.bpf.c:699` — 该 PR 向 `finish_udp_nat` 添加了内联逻辑（ `direct_egress_is_onlink` 、 `redirect_egress` ，以及新的 `bpf_skb_load_bytes` ），而 `finish_udp_nat` 又被内联到 `dns_finish` 中。代码库中包含大量注释，说明 `dns_finish` 已经接近 5.4 的 100 万条指令验证器限制（这正是 `do_udp_nat` 被设为 `__noinline` 的原因）。 `from_cube` 同样增加了三个内联的 `redirect_egress` 调用点，并且被限制为不允许 bpf 到 bpf 的调用。
> 
> The PR was only validated on Linux 6.6 and explicitly notes 5.4 was not tested. Please run the verifier (or at least `veristat`, per `CubeNet/cubevs/Makefile`) on a 5.4 kernel/compiler target before merge.该 PR 仅在 Linux 6.6 上验证过，并明确说明未在 5.4 上测试。请在合并前，在 5.4 内核/编译器目标上运行验证器（或至少按照 `CubeNet/cubevs/Makefile` 运行 `veristat` ）。
> 
> #### 4. LOW — `NodeIPMask` is now a hard requirement of `cubevs.Init`
> 4. 低 — `NodeIPMask` 现在是 `cubevs.Init` 的硬性要求。
> `CubeNet/cubevs/miscs.go` — `ipMaskToUint32(params.NodeIPMask)` errors are joined into the result **before** the nil-variable guard, and `rewriteConstants` is invoked for `loadLocalgw`/`loadNodenic` too. A nil `NodeIPMask` therefore fails `Init` for _all_ objects, even though `localgw`/`nodenic` never reference `nodenic_netmask`. The only production caller (`controller.go`) sets it, so this is latent — but any other embedder of the cubevs package will now fail at startup with an opaque "invalid IPv4 mask length" error. Consider only failing when the spec actually contains `globalNameNodeNetmask`. `CubeNet/cubevs/miscs.go` — `ipMaskToUint32(params.NodeIPMask)` 的错误会在 nil 变量检查之前合并到结果中，并且 `rewriteConstants` 也会对 `loadLocalgw` / `loadNodenic` 调用。因此，即使 `localgw` / `nodenic` 从未引用 `nodenic_netmask` ，nil 的 `NodeIPMask` 也会导致所有对象在 `Init` 处失败。唯一的生成环境调用方（ `controller.go` ）会设置它，所以这目前是潜在问题——但任何其他嵌入 cubevs 包的使用者现在都会在启动时因不透明的“无效 IPv4 掩码长度”错误而失败。建议仅在规范实际包含 `globalNameNodeNetmask` 时才触发失败。
> 
> #### 5. LOW — On-link detection is subnet-based, not route-based
> 5. 低优先级——在线检测基于子网而非路由
> `direct_egress_is_onlink` approximates on-link as "shares the node IP's subnet." This is a reasonable simplification for the reported bug, but destinations reachable through a directly-connected interface on a different subnet, or multiple subnets on the node, are misclassified as off-link and keep the gateway-MAC fast path (old behavior). Worth a one-line comment documenting the approximation so future readers don't treat it as authoritative. `direct_egress_is_onlink` 将在线近似为“与节点 IP 共享子网”。对于所报告的问题，这是一个合理的简化，但通过不同子网上的直连接口可达的目的地，或节点上存在多个子网时，会被错误分类为离线，并保留网关 MAC 快速路径（旧行为）。值得添加一行注释说明这一近似，以免未来读者将其视为权威定义。
> 
> ### Minor / nits
> * `CubeNet/src/cubevs.h:101` comment says `nodenic_netmask` is "network byte order", while `ipMaskToUint32` and `ipToUint32` produce the host-little-endian layout of the network-ordered bytes. The Go comment in `util.go` is accurate; the `.h` comment is misleading. `CubeNet/src/cubevs.h:101` 的注释称 `nodenic_netmask` 为“网络字节序”，而 `ipMaskToUint32` 和 `ipToUint32` 生成的是网络字节序字节的主机小端布局。 `util.go` 中的 Go 注释是准确的； `.h` 的注释具有误导性。
> * `ConfigureCubeDevHostRouting` validates the mask length/bits but not contiguity — same gap as Finding 1; masks sourced from netlink are always contiguous, so this is cosmetic. `ConfigureCubeDevHostRouting` 验证了掩码长度/位数，但不验证连续性——与发现 1 存在相同的缺口；来自 netlink 的掩码始终是连续的，因此这只是外观上的问题。
> 
> ### Verified correct
> * Byte-order math for `ipMaskToUint32` matches the existing `ipToUint32` convention, and the BPF `(daddr & mask) == (nodenic_ip & mask)` comparison is consistent. `ipMaskToUint32` 的字节序计算与现有的 `ipToUint32` 约定一致，且 BPF `(daddr & mask) == (nodenic_ip & mask)` 比较是一致的。
> * NAT only rewrites saddr, so using the pre-NAT `daddr` in `from_cube` and the `bpf_skb_load_bytes`-loaded daddr after `do_udp_nat` in `dns_finish` is safe.NAT 仅重写 saddr，因此在 `from_cube` 中使用 NAT 前的 `daddr` ，并在 `dns_finish` 中使用 `do_udp_nat` 后加载的 `bpf_skb_load_bytes` daddr 是安全的。
> * The `dns_finish` `goto` refactor is behavior-preserving. `dns_finish` `goto` 的重构保持了行为不变。
> * The FORWARD rule is narrow and idempotent; `accept_local` + `rp_filter=2` on the ingress device are the right knobs for this path.FORWARD 规则范围狭窄且具有幂等性；入口设备上的 `accept_local` + `rp_filter=2` 正是该路径的正确控制点。



> ## Review: fix(cubenet): route direct on-link egress via host (PR #1321)
> 评审：修复(cubenet)：通过主机路由直接链路出口（PR #1321 ）
> _AI-generated review — not a human approval.AI 生成的评审——并非人工审批。_
> 
> ### Overview
> The PR fixes #1209: in direct-egress mode CubeVS always rewrote the destination MAC to the node gateway MAC, which breaks same-subnet traffic when the gateway doesn't hairpin. The approach:该 PR 修复了 #1209 ：在直接出口模式下，CubeVS 总是将目的 MAC 重写为节点网关 MAC，当网关不支持 hairpin 时，这会破坏同子网流量。解决方案：
> 
> * Propagates the primary node IPv4 mask into CubeVS (`nodenic_netmask`) and detects on-link destinations in BPF.将节点主 IPv4 掩码传播到 CubeVS（ `nodenic_netmask` ），并在 BPF 中检测链路直连目的地。
> * For on-link traffic, preserves the sandbox-side L2 pair, keeps the SNAT, and redirects into `cube-dev` ingress so the host routing/neighbor tables resolve the real destination MAC.对于直连流量，保留沙箱侧的 L2 地址对，维持 SNAT，并将其重定向至 `cube-dev` 入口，以便主机路由/邻居表解析出真实的目的 MAC 地址。
> * Preserves the gateway-MAC fast path for off-link traffic.对于非直连流量，保留网关 MAC 快速路径。
> * Configures `cube-dev` forwarding / `accept_local` / loose `rp_filter` and a narrow idempotent `FORWARD` iptables rule.配置 `cube-dev` 转发、 `accept_local` 、宽松的 `rp_filter` 以及一条窄范围的幂等 `FORWARD` iptables 规则。
> 
> The design is sound and well-integrated: the "SNAT then inject into cube-dev ingress" mechanism already exists in the codebase (L7-proxy path in `from_cube`, node-IP path), the per-protocol L2 preservation is correctly gated, the `dns_finish` rewrite to a single `goto` is behavior-preserving, and the Go-side defensive copies and test seams are good. The iptables helper (`runIptablesEnsure`) is genuinely idempotent.该设计合理且集成良好：“先 SNAT 再注入 cube-dev 入口”的机制已在代码库中存在（ `from_cube` 中的 L7 代理路径、节点 IP 路径），按协议保留 L2 的操作正确设置了门控， `dns_finish` 重写为单个 `goto` 保持了行为不变，Go 侧的防御性拷贝和测试接缝也做得很好。iptables 辅助函数（ `runIptablesEnsure` ）确实具有幂等性。
> 
> I found one confirmed defect (a failing unit test) plus two areas that should be addressed before merge.我发现了一个已确认的缺陷（一个失败的单元测试），另外还有两个问题需要在合并前处理。
> 
> ### Findings
> #### 1. HIGH — `TestIPMaskToUint32RejectsNonIPv4Mask` fails: non-contiguous masks are not rejected
> 1. 高优先级 — `TestIPMaskToUint32RejectsNonIPv4Mask` 失败：未拒绝非连续掩码
> `CubeNet/cubevs/util.go:35` — `ipMaskToUint32` only validates `len(mask) == 4` and that `mask.Size()` reports 32 bits. `net.IPMask.Size()` always returns `bits = len(mask)*8`, so **any** 4-byte mask passes the second check. The new test `TestIPMaskToUint32RejectsNonIPv4Mask` (`CubeNet/cubevs/util_test.go:37`) asserts that `net.IPMask{255, 0, 255, 0}` is rejected, but the implementation accepts it (returning `0x00ff00ff`) and the test fails with `"ipMaskToUint32 accepted a non-contiguous IPv4 mask"`. `CubeNet/cubevs/util.go:35` — `ipMaskToUint32` 仅验证 `len(mask) == 4` 以及 `mask.Size()` 报告为 32 位。 `net.IPMask.Size()` 始终返回 `bits = len(mask)*8` ，因此任何 4 字节掩码都能通过第二项检查。新测试 `TestIPMaskToUint32RejectsNonIPv4Mask` （ `CubeNet/cubevs/util_test.go:37` ）断言 `net.IPMask{255, 0, 255, 0}` 应被拒绝，但实现却接受了它（返回 `0x00ff00ff` ），导致测试以 `"ipMaskToUint32 accepted a non-contiguous IPv4 mask"` 失败。
> 
> This contradicts the PR's validation claim of `go test -count=1 ./...` in `CubeNet/cubevs`.这与 PR 中在 `CubeNet/cubevs` 中声称的 `go test -count=1 ./...` 验证逻辑相矛盾。
> 
> The gap also matters beyond the test: a non-contiguous mask would silently produce a wrong on-link comparison in `direct_egress_is_onlink` while the error message claims it was validated. Fix: validate contiguity (e.g., once a non-`0xff` byte is seen, every following byte must be `0`), or drop the second test assertion.测试之外，这个差距也很重要：不连续的掩码会在 `direct_egress_is_onlink` 中静默产生错误的直连比较，而错误信息却声称已验证。修复方法：验证连续性（例如，一旦看到非 `0xff` 字节，后续每个字节都必须是 `0` ），或者删除第二个测试断言。
> 
> #### 2. MEDIUM — New hard runtime dependency on `iptables` + `/proc/sys` writes in direct mode
> 2. 中等 — 在直接模式下，对 `iptables` 和 `/proc/sys` 的写入新增了硬性运行时依赖
> `Cubelet/network/runtime/systemnet/cube_dev.go:40` — `ConfigureCubeDevHostRouting` is now called unconditionally in the direct-egress branch of `initCubeVS`. Previously direct mode only ran `CleanupCubeRouter` and made no `iptables`/sysctl calls. Now a missing `iptables` binary or a read-only `/proc/sys/net/ipv4/conf/cube-dev/*` fails `initCubeVS`, which fails `newProductionControllerDeps`, which fails Cubelet startup. `Cubelet/network/runtime/systemnet/cube_dev.go:40` — `ConfigureCubeDevHostRouting` 现在在 `initCubeVS` 的直接出口分支中被无条件调用。此前，直接模式仅运行 `CleanupCubeRouter` ，且不进行任何 `iptables` /sysctl 调用。现在，缺少 `iptables` 二进制文件或只读的 `/proc/sys/net/ipv4/conf/cube-dev/*` 会导致 `initCubeVS` 失败，进而导致 `newProductionControllerDeps` 失败，最终导致 Cubelet 启动失败。
> 
> Also, the `FORWARD` rule and the three sysctls are never cleaned up when the node switches to route-aware mode or shuts down. The rule is narrow (`-i cube-dev -o <eth> -s <nodeIP>/32 -d <onlinkNet>`) so the residue is low-risk, but it should be documented or paired with a teardown.此外，当节点切换到路由感知模式或关闭时， `FORWARD` 规则和三个 sysctl 设置从未被清理。该规则范围较窄（ `-i cube-dev -o <eth> -s <nodeIP>/32 -d <onlinkNet>` ），因此残留风险较低，但应记录在案或配合拆除机制。
> 
> #### 3. MEDIUM — Linux 5.4 verifier budget risk on the `dns_finish` path
> 3. 中 — Linux 5.4 验证器预算在 `dns_finish` 路径上的风险
> `CubeNet/src/mvmtap.bpf.c:699` — the PR adds inlined logic (`direct_egress_is_onlink`, `redirect_egress`, plus a new `bpf_skb_load_bytes`) to `finish_udp_nat`, which is inlined into `dns_finish`. The codebase contains extensive comments explaining that `dns_finish` already sits close to the 5.4 1M-instruction verifier limit (that's exactly why `do_udp_nat` was made `__noinline`). `from_cube` likewise gains three inlined `redirect_egress` call sites and is constrained to have no bpf-to-bpf calls. `CubeNet/src/mvmtap.bpf.c:699` — 该 PR 向 `finish_udp_nat` 添加了内联逻辑（ `direct_egress_is_onlink` 、 `redirect_egress` ，以及新的 `bpf_skb_load_bytes` ），而 `finish_udp_nat` 又被内联到 `dns_finish` 中。代码库中包含大量注释，说明 `dns_finish` 已经接近 5.4 的 100 万条指令验证器限制（这正是 `do_udp_nat` 被设为 `__noinline` 的原因）。 `from_cube` 同样增加了三个内联的 `redirect_egress` 调用点，并且被限制为不允许 bpf 到 bpf 的调用。
> 
> The PR was only validated on Linux 6.6 and explicitly notes 5.4 was not tested. Please run the verifier (or at least `veristat`, per `CubeNet/cubevs/Makefile`) on a 5.4 kernel/compiler target before merge.该 PR 仅在 Linux 6.6 上验证过，并明确说明未在 5.4 上测试。请在合并前，在 5.4 内核/编译器目标上运行验证器（或至少按照 `CubeNet/cubevs/Makefile` 运行 `veristat` ）。
> 
> #### 4. LOW — `NodeIPMask` is now a hard requirement of `cubevs.Init`
> 4. 低 — `NodeIPMask` 现在是 `cubevs.Init` 的硬性要求。
> `CubeNet/cubevs/miscs.go` — `ipMaskToUint32(params.NodeIPMask)` errors are joined into the result **before** the nil-variable guard, and `rewriteConstants` is invoked for `loadLocalgw`/`loadNodenic` too. A nil `NodeIPMask` therefore fails `Init` for _all_ objects, even though `localgw`/`nodenic` never reference `nodenic_netmask`. The only production caller (`controller.go`) sets it, so this is latent — but any other embedder of the cubevs package will now fail at startup with an opaque "invalid IPv4 mask length" error. Consider only failing when the spec actually contains `globalNameNodeNetmask`. `CubeNet/cubevs/miscs.go` — `ipMaskToUint32(params.NodeIPMask)` 的错误会在 nil 变量检查之前合并到结果中，并且 `rewriteConstants` 也会对 `loadLocalgw` / `loadNodenic` 调用。因此，即使 `localgw` / `nodenic` 从未引用 `nodenic_netmask` ，nil 的 `NodeIPMask` 也会导致所有对象在 `Init` 处失败。唯一的生成环境调用方（ `controller.go` ）会设置它，所以这目前是潜在问题——但任何其他嵌入 cubevs 包的使用者现在都会在启动时因不透明的“无效 IPv4 掩码长度”错误而失败。建议仅在规范实际包含 `globalNameNodeNetmask` 时才触发失败。
> 
> #### 5. LOW — On-link detection is subnet-based, not route-based
> 5. 低优先级——在线检测基于子网而非路由
> `direct_egress_is_onlink` approximates on-link as "shares the node IP's subnet." This is a reasonable simplification for the reported bug, but destinations reachable through a directly-connected interface on a different subnet, or multiple subnets on the node, are misclassified as off-link and keep the gateway-MAC fast path (old behavior). Worth a one-line comment documenting the approximation so future readers don't treat it as authoritative. `direct_egress_is_onlink` 将在线近似为“与节点 IP 共享子网”。对于所报告的问题，这是一个合理的简化，但通过不同子网上的直连接口可达的目的地，或节点上存在多个子网时，会被错误分类为离线，并保留网关 MAC 快速路径（旧行为）。值得添加一行注释说明这一近似，以免未来读者将其视为权威定义。
> 
> ### Minor / nits
> * `CubeNet/src/cubevs.h:101` comment says `nodenic_netmask` is "network byte order", while `ipMaskToUint32` and `ipToUint32` produce the host-little-endian layout of the network-ordered bytes. The Go comment in `util.go` is accurate; the `.h` comment is misleading. `CubeNet/src/cubevs.h:101` 的注释称 `nodenic_netmask` 为“网络字节序”，而 `ipMaskToUint32` 和 `ipToUint32` 生成的是网络字节序字节的主机小端布局。 `util.go` 中的 Go 注释是准确的； `.h` 的注释具有误导性。
> * `ConfigureCubeDevHostRouting` validates the mask length/bits but not contiguity — same gap as Finding 1; masks sourced from netlink are always contiguous, so this is cosmetic. `ConfigureCubeDevHostRouting` 验证了掩码长度/位数，但不验证连续性——与发现 1 存在相同的缺口；来自 netlink 的掩码始终是连续的，因此这只是外观上的问题。
> 
> ### Verified correct
> * Byte-order math for `ipMaskToUint32` matches the existing `ipToUint32` convention, and the BPF `(daddr & mask) == (nodenic_ip & mask)` comparison is consistent. `ipMaskToUint32` 的字节序计算与现有的 `ipToUint32` 约定一致，且 BPF `(daddr & mask) == (nodenic_ip & mask)` 比较是一致的。
> * NAT only rewrites saddr, so using the pre-NAT `daddr` in `from_cube` and the `bpf_skb_load_bytes`-loaded daddr after `do_udp_nat` in `dns_finish` is safe.NAT 仅重写 saddr，因此在 `from_cube` 中使用 NAT 前的 `daddr` ，并在 `dns_finish` 中使用 `do_udp_nat` 后加载的 `bpf_skb_load_bytes` daddr 是安全的。
> * The `dns_finish` `goto` refactor is behavior-preserving. `dns_finish` `goto` 的重构保持了行为不变。
> * The FORWARD rule is narrow and idempotent; `accept_local` + `rp_filter=2` on the ingress device are the right knobs for this path.FORWARD 规则范围狭窄且具有幂等性；入口设备上的 `accept_local` + `rp_filter=2` 正是该路径的正确控制点。

- NodeIPMask is now required only for the CollectionSpec that contains from_cube and references nodenic_netmask. localgw and nodenic no longer fail when NodeIPMask is absent, and a unit test covers both cases.
- direct_egress_is_onlink now explicitly documents that “on-link” means the primary node IPv4 prefix, not every directly connected route on a multi-homed node.
- The nodenic_netmask comment now describes the value as packet-byte layout instead of network byte order.

The CubeVS unit tests pass, and the BPF programs were also validated successfully on Linux 5.4.0-216-generic.

chenhengqi:
Redirect traffic to cube-dev looks wierd. Have you tired `bpf_fib_lookup()`?

jay3cx:
> Redirect traffic to cube-dev looks wierd. Have you tired `bpf_fib_lookup()`?

I looked into `bpf_fib_lookup()`, but on Linux 5.4, which CubeVS supports, it returns `NO_NEIGH` without triggering ARP when the neighbor is not cached. Since `bpf_redirect_neigh()` is unavailable there, we still need the host-stack fallback through `cube-dev`.

chenhengqi:
How about:
* Add a new map storing ip=>mac
* Add a perf event array for notification, if the mac is not found, notify user space to perform an ARP

jay3cx:
> How about:
> 
> * Add a new map storing ip=>mac
> * Add a perf event array for notification, if the mac is not found, notify user space to perform an ARP

I gave this a quick try. It works once the map is populated, but the packet that triggers the miss gets dropped before userspace finishes ARP. The first ping/UDP packet was lost, and TCP connected after one retry (~1s). Is that acceptable?

chenhengqi:
Given we don't have MAC in place, I think the lost is acceptable. A second thought: we can simply turn the lost packet to a ARP packet.

jay3cx:
@chenhengqi Switched to the ARP + map approach as discussed. Also added probe rate-limit and zeroed ARP padding. Squashed to one commit — mind taking another look?

CubeNet/src/nodenic.bpf.c
```diff
@@ -409,6 +456,10 @@ int from_world(struct __sk_buff *skb)
 	struct iphdr *l3;
 	int ret;
 
+	if (skb->protocol == bpf_htons(ETH_P_ARP)) {
+		learn_direct_neighbor(skb);
+		return TC_ACT_OK;
```
chenhengqi, line 440: How does Linux network stack handle this packet?

jay3cx: TC_ACT_OK -> packet goes to the normal Linux ARP path. We only observe and update `direct_neigh`.

The probe was sent by BPF, so the kernel typically has no pending neigh; with arp_accept=0 it also won't create one from that reply. Host ARP handling is unchanged.

Cubelet/network/runtime/systemnet/device.go
```diff
@@ -59,7 +68,8 @@ func GetHostDevice(ifName string) (*HostDevice, error) {
 	return &HostDevice{
 		Index:      link.Attrs().Index,
 		Name:       link.Attrs().Name,
-		IP:         addrs[0].IP,
+		IP:         append(net.IP(nil), ip...),
```
chenhengqi: No `append` here.

Cubelet/network/runtime/systemnet/device.go
```diff
@@ -41,13 +42,21 @@ func GetHostDevice(ifName string) (*HostDevice, error) {
...
 		return nil, fmt.Errorf("ipv4 address on %s is not unique", ifName)
 	}
+	if addrs[0].IPNet == nil {
+		return nil, fmt.Errorf("invalid ipv4 address on %s", ifName)
+	}
+	ip := addrs[0].IP.To4()
+	_, bits := addrs[0].Mask.Size()
+	if ip == nil || bits != 32 {
+		return nil, fmt.Errorf("invalid ipv4 address on %s", ifName)
+	}
```
chenhengqi: I think these checks are redundant.

CubeNet/cubevs/miscs.go
```diff
@@ -57,6 +58,13 @@ func rewriteConstants(vars map[string]*ebpf.VariableSpec, params Params) error {
 		err = errors.Join(err, v.Set(params.EgressRedirectFlags))
 	}
 	err = errors.Join(err, vars[globalNameNodeIP].Set(ipToUint32(params.NodeIP)))
+	if spec.Programs[programNameFromCube] != nil {
```
chenhengqi: This check is unnecessary.

CubeNet/cubevs/miscs.go
```diff
@@ -231,6 +239,9 @@ func Init(params Params) error {
 	_ = os.Remove(pinPath("tungrp_to_tuns")) // NOCC:Path Traversal()
 	// dns_query_track is runtime pending-query state, not persisted policy.
 	_ = os.Remove(pinPath(MapNameDNSQueryTrack)) // NOCC:Path Traversal()
+	// Direct-neighbor state is tied to the current host interface and must not
+	// survive a CubeVS restart with stale MAC addresses.
+	_ = os.Remove(pinPath(MapNameDirectNeighbors)) // NOCC:Path Traversal()
```
chenhengqi: No need to remove this pinned map.

chenhengqi: I think all these 3 `os.Remove()` should be dropped.

CubeNet/cubevs/util.go
```diff
@@ -26,6 +27,19 @@ func ipToUint32(ip net.IP) uint32 {
 	return uint32(ip[0]) | uint32(ip[1])<<8 | uint32(ip[2])<<16 | uint32(ip[3])<<24
 }
 
+// ipMaskToUint32 converts an IPv4 mask to the byte layout used by IPv4
+// addresses in BPF packet headers on little-endian hosts.
+func ipMaskToUint32(mask net.IPMask) (uint32, error) {
```
chenhengqi: Make this as simple as `ipToUint32()`.

CubeNet/src/nodenic.bpf.c
```diff
@@ -350,7 +394,10 @@ static int icmp_nat_session(struct __sk_buff *skb, struct ethhdr *l2, struct iph
 	return bpf_redirect(sess->vm_ifindex, 0);
 }
 
-static int do_icmp_nat(struct __sk_buff *skb)
+/* from_world performs a DNS tail call, so these dispatch helpers must remain
```
chenhengqi: Is there real world issue without this change?

CubeNet/src/nodenic.bpf.c
```diff
@@ -17,6 +17,50 @@
...
+	if (data + sizeof(struct arp_packet) > data_end)
+		return;
+
+	packet = data;
+	if (packet->eth.h_proto != bpf_htons(ETH_P_ARP) ||
+	    packet->arp.ar_hrd != bpf_htons(ARPHRD_ETHER) ||
+	    packet->arp.ar_pro != bpf_htons(ETH_P_IP) ||
+	    packet->arp.ar_hln != ETH_ALEN ||
+	    packet->arp.ar_pln != sizeof(__be32) ||
+	    (packet->arp.ar_op != bpf_htons(ARPOP_REQUEST) &&
```
chenhengqi: We only expect reply here, no?

CubeNet/src/nodenic.bpf.c
```diff
@@ -17,6 +17,50 @@
...
+	     packet->arp.ar_op != bpf_htons(ARPOP_REPLY)))
+		return;
+
+	eth_src = (union macaddr *)packet->eth.h_source;
+	arp_src = (union macaddr *)packet->arp.ar_sha;
+	if (eth_src->p1 != arp_src->p1 || eth_src->p2 != arp_src->p2 ||
+	    (arp_src->addr[0] & 1) || (arp_src->p1 == 0 && arp_src->p2 == 0))
+		return;
```
chenhengqi, line 52: Add some comments here.

CubeNet/src/mvmtap.bpf.c
chenhengqi, line 1: Can we change `set_mac_pair()` instead of touching every `redirect_*` in this file?

CubeNet/src/nodenic.bpf.c
jay3cx: Yeah. On 5.4 from_world fails to load without them:
tail_calls are not allowed in programs with bpf-to-bpf calls

CubeNet/src/mvmtap.bpf.c
jay3cx: Moved it into the old set_mac_pair sites so we don't touch every redirect.

CubeNet/src/nodenic.bpf.c
jay3cx: reply only now.

chenhengqi:
Thanks. Will take another round of review tomorrow.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -106,6 +106,146 @@ static __always_inline bool should_do_nat(const struct iphdr *l3)
...
+	long err;
+
+	if (!direct_egress_is_onlink(daddr)) {
+		set_mac_pair(l2, egress_smacaddr_p1, egress_smacaddr_p2,
+			     egress_dmacaddr_p1, egress_dmacaddr_p2);
+		return EGRESS_MAC_READY;
+	}
+
+	now = bpf_ktime_get_ns();
+	neighbor = bpf_map_lookup_elem(&direct_neigh, &daddr);
```
chenhengqi, line 230: Before doing a map lookup, can we try `bpf_fib_lookup` first?

CubeNet/src/mvmtap.bpf.c
```diff
@@ -444,6 +584,7 @@ static __always_inline __u32 do_icmp_nat(struct __sk_buff *skb, struct mvm_meta
 	__u16 snat_id;
 	__u64 flags;
 	__u64 now;
+	int mac_result;
```
chenhengqi: Sort variables in reverse xmas tree order.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -106,6 +106,146 @@ static __always_inline bool should_do_nat(const struct iphdr *l3)
...
+	packet.arp.ar_hln = ETH_ALEN;
+	packet.arp.ar_pln = sizeof(__be32);
+	packet.arp.ar_op = bpf_htons(ARPOP_REQUEST);
+	macaddr = (union macaddr *)packet.arp.ar_sha;
+	macaddr->p1 = nodenic_macaddr_p1;
+	macaddr->p2 = nodenic_macaddr_p2;
+	packet.arp.ar_sip = nodenic_ip;
+	packet.arp.ar_tip = daddr;
+
+	/* Leave padding after the ARP header so a delayed CHECKSUM_PARTIAL
```
chenhengqi: It's hard to understand what's going on here. Need more comments.

CubeNet/src/mvmtap.bpf.c
jay3cx: fib_lookup first, then ARP if it misses.

jay3cx: Done

CubeNet/src/mvmtap.bpf.c
jay3cx: More comments have been added.

chenhengqi:
@jay3cx I think we can place the MAC address preparation work right after:

https://github.com/TencentCloud/CubeSandbox/blob/ca3adc98506ed0d1f86c4cfe61ed6f9a9eeca895/CubeNet/src/mvmtap.bpf.c#L1021-L1033

So that we don't have to touch protocol-specific code.

jay3cx:
> @jay3cx I think we can place the MAC address preparation work right after:
> 
> https://github.com/TencentCloud/CubeSandbox/blob/ca3adc98506ed0d1f86c4cfe61ed6f9a9eeca895/CubeNet/src/mvmtap.bpf.c#L1021-L1033
> 
> So that we don't have to touch protocol-specific code.

@chenhengqi Moved it there.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -106,6 +106,164 @@ static __always_inline bool should_do_nat(const struct iphdr *l3)
...
+		/* Touch only the deadline so a concurrent learn cannot lose its MAC. */
+		retry_at = now + DIRECT_NEIGH_PROBE_INTERVAL_NS;
+		neighbor->next_probe_at_ns = retry_at;
+		neighbor = bpf_map_lookup_elem(&direct_neigh, &daddr);
+		if (!neighbor)
+			return EGRESS_MAC_DROP;
+		/* Later deadline: another CPU already refreshed this entry. */
+		if (!direct_neighbor_is_zero(neighbor) &&
+		    neighbor->next_probe_at_ns > retry_at)
+			goto set_neighbor;
```
chenhengqi: We can simply always do an ARP probe here, right? I am not sure how frequent this concurrent learning will happen.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -106,6 +106,164 @@ static __always_inline bool should_do_nat(const struct iphdr *l3)
...
+	if (err)
+		return err;
+	return bpf_skb_store_bytes(skb,
+				   sizeof(struct arp_packet) + 2 * DIRECT_ARP_ZERO_CHUNK_LEN,
+				   zeroes,
+				   DIRECT_ARP_FRAME_LEN - sizeof(struct arp_packet) -
+				   2 * DIRECT_ARP_ZERO_CHUNK_LEN, 0);
+}
+
+static __always_inline long direct_egress_arp_request(struct __sk_buff *skb, __u32 daddr)
```
chenhengqi, line 149: See https://github.com/TencentCloud/CubeSandbox/blob/189d6eef1b9aeb6813df3db1880cb6780fa633cc/CubeNet/src/mvmtap.bpf.c#L253

I still have doubt in the implementation here. Please explain the purpose of bpf_skb_change_tail/bpf_skb_change_head/direct_egress_clear_arp_padding here. I haven't consult an AI agent yet.

lkml-likexu:
@jay3cx Thanks for your contribution and rebase is required.

CubeNet/src/mvmtap.bpf.c
chenhengqi: @jay3cx

jay3cx: TAP packets can still have `CHECKSUM_PARTIAL` set. After we turn the packet into an ARP request, that state is still there: `__bpf_skb_change_tail()` trims through `bpf_skb_trim_rcsum()`, which does not clear `ip_summed`, so the original `csum_start` survives all the way to `dev_queue_xmit()`. `skb_checksum_help()` then writes the L4 checksum at the old offset, and it `BUG_ON`s if that write would land past `skb_headlen()`. Getting the length wrong panics the kernel. It does not just emit a bad ARP.

`bpf_skb_change_tail()` is there because the helper will not shrink below `__bpf_skb_min_len()`. For `CHECKSUM_PARTIAL` that floor is `skb_checksum_start_offset() + csum_offset + sizeof(__sum16)`. On a normal TCP packet that is 52 bytes, so `bpf_skb_change_tail(skb, 42, 0)` returns `-EINVAL`. 96 is just a length above that floor. Anything ≥ 52 would load. We cannot shrink straight to a 42-byte ARP frame.

Staying above that floor only keeps the write in bounds. It does not keep it off the ARP header. That is what `bpf_skb_change_head(32)` is for. I checked this on 5.4.0-216-generic with tx-checksumming off so `skb_checksum_help()` actually runs. With only `change_tail(96)`, offset 50 is `eb92`. With `change_tail(96)` plus `change_head(32)`, offset 50 is zero and `ffff` shows up at offset 82, which is 50 + 32.

`direct_egress_clear_arp_padding()` then zeros the rest of the 128-byte buffer so we don't broadcast leftover payload. I dumped the frame. Nothing from the original packet is left in it.

`tcp_reply_reset()` does not need any of this. It still has valid IP and TCP headers, and the checksum field is where the stack expects it.

I'll fix the comment in the code too. "keeps that slot" is the wrong picture.

CubeNet/src/mvmtap.bpf.c
jay3cx: Agreed. Dropped the extra re-lookup and just probe now.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -106,6 +106,164 @@ static __always_inline bool should_do_nat(const struct iphdr *l3)
...
+		    neighbor->next_probe_at_ns > retry_at)
+			goto set_neighbor;
+	} else {
+		pending.next_probe_at_ns = now + DIRECT_NEIGH_PROBE_INTERVAL_NS;
+		err = bpf_map_update_elem(&direct_neigh, &daddr, &pending, BPF_NOEXIST);
+		if (err)
+			return EGRESS_MAC_DROP;
+	}
+
+	if (direct_egress_arp_request(skb, daddr))
```
zhouxianping: A packet is dropped on the initial miss or during subsequent probes; however, one could trigger the kernel to perform ARP learning by notifying a user-space process via `perf_event`. In the event of a miss, a fallback destination MAC address (`egress_dmacaddr`) can be used to encapsulate the packet. Furthermore, since eBPF does not need to send or receive ARP packets itself, the implementation appears simpler.

lkml-likexu:
@jay3cx Thanks for your contribution. Please click "Resolve conversation" to ensure that all comments from cubesandboxbot are either approved or rejected.

jay3cx:
> @jay3cx Thanks for your contribution. Please click "Resolve conversation" to ensure that all comments from cubesandboxbot are either approved or rejected.

done

PR #1322, opened 2026-08-11
feat(pause): cubecow pause/resume with shim exit and volume/host-mount support

ls-ggg:
## Summary

- Unify Pause/Resume on the **cubecow CoW** path: pause snapshots use the same local CoW layout/catalog as user snapshots (no separate pausevm bypass).
- After Pause succeeds, the **shim must exit**; Resume recreates the sandbox from the pause snap (same sandboxID) instead of keeping a live paused process.
- Keep **plugin volumes** and **host-mount** working across Pause/Resume: Detach on Pause / Attach on Resume with aligned Master↔node refcounts; on pause resume reconnect virtiofs devices without remount/`exec.mount` when guest binds already exist; reject Resume if pause-detached volumes were deleted meanwhile.
- Destroy **never** resumes a paused sandbox (CoW pause already exited the shim). Pause post-cleanup uses `cube.pause.keep_tombstone` (replaces `cube.pause.skip_auto_resume`).
- Restore create-time **network policy** on Resume (egress `CubeNetworkConfig` + CubeProxy route refresh) and cover it in `sdk_compat` pause/resume network cases.

## Test plan

- [x] CubeMaster UT: `pausesnap`, `volume/refcount`, `sandbox` resume-pause validation, `sandboxlock`
- [x] Cubelet UT: CoW store helpers / pause cow / Destroy lifecycle (no resume-before-delete)
- [x] Host pause/volume/hostdir/network/cubelet-restart regression on test host
- [x] `sdk_compat` pause/resume + pause/resume network e2e on test host
- [ ] CI checks on this PR (Format Check gofmt fixed in squashed commit)

Assisted-by: Cursor:Composer


docs/zh/dev/redis-key-spec.md
```diff
@@ -64,6 +64,7 @@ cube:{ver}:{scope}:{resource}[:{sub}...]:{id}
 | 沙箱 lifecycle 注册表 | `cube:v1:shared:sandbox:lifecycle:meta` | Hash | shared | CubeMaster | cube-lifecycle-manager | 无（生命周期由 `HDEL` 管理） |
 | 沙箱 lifecycle 事件流 | `cube:v1:shared:sandbox:lifecycle:events` | Stream | shared | CubeMaster | cube-lifecycle-manager | MAXLEN ~ 100000 |
 | 沙箱 lifecycle 状态 | `cube:v1:shared:sandbox:lifecycle:state:{sandboxID}` | String | shared | cube-lifecycle-manager | cube-lifecycle-manager | SET TTL（默认 60s） |
+| 沙箱操作锁（pause/resume/delete） | `master:lock:sb:{sandboxID}` | String | master | CubeMaster | CubeMaster | SET NX EX 10s（解锁 DEL；不续期） |
```
fslongjin: this should follow the naming rules `cube:v1:{scope}:lock:{resource}:{id}` 

To align with existing key naming conventions, please use `sandbox` instead of `sb` in the resource field

ls-ggg: Fixed!

Cubelet/services/cubebox/update.go
```diff
@@ -95,9 +101,18 @@ func (s *service) Update(ctx context.Context, req *cubebox.UpdateCubeSandboxRequ
...
+		if !storage.IsCowBackend() {
+			rsp.Ret.RetMsg = "pause requires cubecow storage backend; legacy pausevm is removed"
+			rsp.Ret.RetCode = errorcode.ErrorCode_InvalidParamFormat
+			return rsp, nil
+		}
+		return s.updateWithPauseCow(ctx, req, sb)
 	case constants.UpdateActionResume:
-		return s.UpdateWithResume(ctx, req, sb)
+		// Resume is Master Create(same sandboxID from pause snap), not Update(resume).
+		rsp.Ret.RetMsg = "pause resume is owned by CubeMaster Create; Update(resume) is not supported"
```
fslongjin, line 113: Resume used to check node CPU/memory **before** waking a paused sandbox (`admitResume` / `paused_resource_release_ratio`). That check lived on `Update(resume)`.

This branch rejects `Update(resume)` and resumes by `Create` instead, but `Create` never runs that check. `admitResume` is now only reached from tests.

What operators will see:
- The documented HTTP 409 `resume rejected by paused_resource_release_ratio policy: need 1024MB > quota 512MB` will not happen anymore.
- If the node is full, Resume looks like a normal create failure (often a 500-style error), not a clear “not enough capacity, retry later”.
- `delete_resume_test.go` still tests in-place `Task.Resume` (`resumeTaskLocked`), which is no longer how production Resume works.

Please run the same capacity check on the Create-from-pause path (or on Master before it calls Create), keep the 409 wording, and stop treating the old Resume helpers as the live path.

CubeMaster/pkg/cubeproxy/admin.go
```diff
@@ -0,0 +1,226 @@
...
+	}
+	url := strings.TrimRight(adminURL, "/") + backendCacheDeletePath
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, url, bytes.NewReader(body))
+	if err != nil {
+		return err
+	}
+	req.Header.Set("Content-Type", "application/json")
+	if cfg := config.GetConfig(); cfg != nil && cfg.CubeProxyConf != nil {
+		if tok := strings.TrimSpace(cfg.CubeProxyConf.AdminToken); tok != "" {
+			req.Header.Set("X-Cube-Admin-Token", tok)
```
fslongjin, line 213: After Resume, the sandbox gets a **new guest IP** (and maybe new host ports). Master already rewrites Redis. This call then asks every CubeProxy to drop its local route cache so requests do not keep using the old IP.

Master only sends `X-Cube-Admin-Token` when `cube_proxy_conf.admin_token` is set. Kubernetes/Terraform already set a token on CubeProxy (`$cube_admin_token`). If Master sends no token, CubeProxy returns **403**, the cache is not cleared, and same-node traffic keeps hitting the old IP → **504**.

I could not find `cube_proxy_conf` in the CubeMaster Helm `conf.yaml` or the TKE terraform addon. Please set Master `admin_token` (and `admin_port`, default 8082) to the same values as CubeProxy, otherwise the cache-clear fix will not work on a real cluster.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
```diff
@@ -0,0 +1,499 @@
...
+	nodeID := strings.TrimSpace(rec.NodeID)
+	if nodeID == "" {
+		if n, ok := localcache.GetNodesByIp(probeIP); ok {
+			nodeID = n.ID()
+		}
+	}
+	log.G(ctx).Warnf("resume: healing timed-out pause binding sandbox=%s snap=%s host=%s (cubelet PAUSED; no ext_info)",
+		req.SandboxID, rec.SnapshotID, probeIP)
+	// No ApplyFromExtInfo: timeout means Master never saw volume events.
+	if err := pausesnap.Complete(ctx, req.SandboxID, rec.SnapshotID, nodeID, probeIP, req.InstanceType, nil); err != nil {
```
fslongjin, line 174: If Pause times out on Master but Cubelet actually finished pausing, Resume tries to recover here. `Complete` is called with `pluginVolumeIDs = nil`.

Later, Resume only refuses to start when that list still contains a volume that was deleted. An empty list means **the check is skipped**.

Concrete sequence:
1. Pause RPC hits the 120s deadline; Master stores FAILED.
2. Cubelet already paused and detached plugin volumes, but Master never saw that event, so it does not update volume use-counts.
3. The user deletes a plugin volume.
4. Resume “recovers” the timed-out Pause and attaches from `sandbox_spec.json` anyway.

Please record the real volume IDs before marking the pause ready (for example from Cubelet / the packed spec), or do not allow Resume after a timed-out Pause (delete only).

Cubelet/services/cubebox/service.go
```diff
@@ -238,6 +240,13 @@ func (s *service) Create(ctx context.Context, req *cubebox.RunCubeSandboxRequest
 		ExtInfo:   map[string][]byte{},
 	}
 
+	// Pause Resume: Master sends a thin Create (ids only); expand containers /
+	// volumes / annotations from sandbox_spec.json packed in the pause snap.
+	if err := expandPauseSnapshotPackage(req); err != nil {
```
fslongjin, line 245: Resume now enters Cubelet through `Create` (this expansion). Pause `Update` and Destroy take `sandboxLifecycleLocks`. **Create does not.**

Pause marks the sandbox PAUSED *before* it finishes tearing down the live VM. Master gives up the Pause RPC at 120s and may start Resume as soon as `List` shows PAUSED.

Then Create can delete the same sandbox record / containerd objects while Pause cleanup is still running on that ID.

Please make Create-from-pause take the same per-sandbox lock as Pause/Destroy, or wait until cleanup has finished — not only until the status is PAUSED.

CubeMaster/pkg/service/sandbox/sandbox_remove.go
```diff
@@ -178,40 +185,50 @@ func dealScfSandbox(ctx context.Context, req *types.DeleteCubeSandboxReq, t *tas
...
-		}()
-
+	return sandboxlock.WithLock(ctx, req.GetSandboxID(), sandboxlock.Options{
+		Value: "delete",
+		TTL:   sandboxlock.DeleteTTL,
+	}, func(ctx context.Context) error {
+		ctx = context.WithoutCancel(ctx)
+		hostIP := strings.Split(callEp, ":")[0]
+		_, nodeOK := localcache.GetNodesByIp(hostIP)
+		handled, err := pausesnap.TryDeletePaused(ctx, req.GetRequestID(), req.GetSandboxID(), hostIP)
```
fslongjin, line 195: This is the new delete-paused path: remove the leftover paused record and the snapshot. It does **not** wake the VM first.

`docs/guide/lifecycle.md` and `docs/zh/guide/lifecycle.md` still tell users the opposite (internal restore, 5s budget, 409 if the node is full, 503 + Retry-After: 5). SDKs and operators will retry on errors this path no longer returns.

Please update both languages to match this function. CubeAPI tests that still expect `failed to resume paused sandbox before delete: shim timeout...` should move to this behavior too.

CubeMaster/pkg/sandboxlock/lock.go
```diff
@@ -0,0 +1,132 @@
...
+	}
+	return withLock(ctx, r, rediskey.LockSandbox(sandboxID), val, opts, fn)
+}
+
+func withLock(ctx context.Context, r doer, key, val string, opts Options, fn func(context.Context) error) error {
+	if err := acquire(ctx, r, key, val, opts); err != nil {
+		return err
+	}
+	defer func() {
+		if _, err := r.Do("DEL", key); err != nil {
```
fslongjin: Unlock is a plain Redis `DEL`. It does not check that this caller still owns the key.

If Pause runs longer than the lock TTL (180s here), another pause/resume/delete can take the same key. When the first call returns, this `defer DEL` removes **the second caller’s** lock. Both can then run at once on the same sandbox.

Normal Pause RPC is 120s, so this is uncommon, but a stuck DB write or clock skew is enough. Please delete the key only if the value still matches the token this call wrote (a small Lua `GET` then `DEL`).

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -0,0 +1,465 @@
...
+
+func pauseShimLikelyExited(err error) bool {
+	if err == nil {
+		return false
+	}
+	msg := strings.ToLower(err.Error())
+	return strings.Contains(msg, "ttrpc: closed") ||
+		strings.Contains(msg, "connection refused") ||
+		strings.Contains(msg, "no such process") ||
+		strings.Contains(msg, "not found")
```
fslongjin: `ttrpc: closed` / `connection refused` after a successful snapshot are expected: the shim is supposed to exit.

Treating any error that contains `not found` as success is broader. That string also appears when the task was never there, or the snapshot never ran. Cubelet would still write catalog files and tear down the live sandbox.

Please keep the “connection already gone” cases, and do not treat a generic `not found` as “snapshot worked”.

docs/dev/redis-key-spec.md
```diff
@@ -64,6 +64,7 @@ The following are the standard keys currently registered in the system (`v1`). N
 | Sandbox lifecycle registry | `cube:v1:shared:sandbox:lifecycle:meta` | Hash | shared | CubeMaster | cube-lifecycle-manager | none (lifecycle via `HDEL`) |
 | Sandbox lifecycle events | `cube:v1:shared:sandbox:lifecycle:events` | Stream | shared | CubeMaster | cube-lifecycle-manager | MAXLEN ~ 100000 |
 | Sandbox lifecycle state | `cube:v1:shared:sandbox:lifecycle:state:{sandboxID}` | String | shared | cube-lifecycle-manager | cube-lifecycle-manager | SET TTL (default 60s) |
+| Sandbox op lock (pause/resume/delete) | `cube:v1:master:lock:sandbox:{sandboxID}` | String | master | CubeMaster | CubeMaster | SET NX EX 60s (DEL on unlock; no renew) |
```
fslongjin: This row says the lock expires in **60s**. In code (`sandboxlock/lock.go`) Pause uses **180s**, and Resume/Delete use **120s**.

Please write the real TTLs here (and in `docs/zh/dev/redis-key-spec.md`, which currently also says 60s).

docs/zh/dev/redis-key-spec.md
```diff
@@ -64,6 +64,7 @@ cube:{ver}:{scope}:{resource}[:{sub}...]:{id}
 | 沙箱 lifecycle 注册表 | `cube:v1:shared:sandbox:lifecycle:meta` | Hash | shared | CubeMaster | cube-lifecycle-manager | 无（生命周期由 `HDEL` 管理） |
 | 沙箱 lifecycle 事件流 | `cube:v1:shared:sandbox:lifecycle:events` | Stream | shared | CubeMaster | cube-lifecycle-manager | MAXLEN ~ 100000 |
 | 沙箱 lifecycle 状态 | `cube:v1:shared:sandbox:lifecycle:state:{sandboxID}` | String | shared | cube-lifecycle-manager | cube-lifecycle-manager | SET TTL（默认 60s） |
+| 沙箱操作锁（pause/resume/delete） | `cube:v1:master:lock:sandbox:{sandboxID}` | String | master | CubeMaster | CubeMaster | SET NX EX 60s（解锁 DEL；不续期） |
```
fslongjin: Same mismatch as the English spec: this says `EX 60s`, but Pause locks for 180s and Resume/Delete for 120s. Please keep the two pages in sync with `sandboxlock`.

.gitignore
```diff
@@ -82,3 +82,16 @@ __pycache__
 !/tests/perf/cubebench.sh
 
 web/tsconfig.tsbuildinfo
+
+# Local toolchain downloads (e.g. portable protoc)
+.tools/
+
+# Local pause/volume stress helpers (kept outside the repo)
+scripts/pause_scenario_stress.py
+scripts/volume_pause_matrix.py
```
fslongjin: The PR test plan marks `scripts/volume_pause_matrix.py` as done (49/49), but that file is gitignored and is not in the repo. Later reviewers cannot rerun it.

Listing personal helper names (`_dbg_agent_in_guest.py`, `_smoke_pause_volume.py`, `.cursor/`) in the project `.gitignore` also does not belong here.

Please either add a real, reviewed test under `tests/` or drop the claim. Keep machine-local ignores out of this file.

(`Cubelet/pkg/utils/localstorage_test.go` seeding `TestSetBs_SingleDb` is unrelated to Pause/Resume — easier to follow if that lands in its own commit/PR.)

CubeProxy/tests/backend_cache_test.lua
```diff
@@ -0,0 +1,54 @@
+package.path = "./lua/?.lua;" .. package.path
```
fslongjin, line 1: Good to have a unit test for dropping the per-sandbox route cache.

`CubeProxy/Makefile` `test` still only runs `test_start.sh` / admin access log / Lua syntax / sentinel. CI `cube-proxy-test` will not execute this file.

Please add something like `lua tests/backend_cache_test.lua` to that target (same “run if lua exists” pattern as the sentinel tests).

Cubelet/services/cubebox/cube_container_create.go
```diff
@@ -110,6 +141,33 @@ func (l *local) Create(ctx context.Context, opts *workflow.CreateContext) error
 			return ret.Err(errorcode.ErrorCode_PreConditionFailed, "already exists")
 		}
 	}
+	// Resume-from-pause reuses the same sandboxID. A PAUSED tombstone may still
+	// be in the store for List; replace it. Reject only non-paused collisions.
+	if desired := strings.TrimSpace(realReq.GetAnnotations()[constants.MasterAnnotationDesiredSandboxID]); desired != "" {
+		if sb, err := l.cubeboxManger.Get(ctx, desired); err == nil && sb != nil && sb.SandboxID == desired {
+			if sb.GetStatus() != nil && sb.GetStatus().IsPaused() {
```
fslongjin: `IsPaused()` is true for both **pausing** (snapshot still running, VM still there) and **paused** (snapshot done, leftover record only).

Resume Create uses that to decide “this ID is safe to delete and recreate”. If Create runs while Pause is still in the pausing state (no Cubelet lock on Create; see `service.go` Create), this will delete a sandbox that is still running.

Please only replace the leftover record when the state is `PAUSED`, not `PAUSING`.

fslongjin (commented):
Thanks for the CoW Pause/Resume work. I went through the change with a focus on what users and operators will actually hit after merge.

A few items below are on the code. Two more that are not on a changed line:

**User docs are still describing the old delete path.** `docs/guide/lifecycle.md` and `docs/zh/guide/lifecycle.md` still say that deleting a paused sandbox first restores it, can return 409 on capacity, and may wait 5 seconds for that restore. This PR deletes the paused leftover record and the snapshot directly — it never wakes the VM. Please update both English and Chinese pages so clients do not retry on errors that no longer exist.

**CubeAPI still tests the old delete-before-wake errors.** `CubeAPI/src/routes.rs` still maps `failed to resume paused sandbox before delete: shim timeout...` and the old capacity-on-resume message. Master no longer does that restore. Those tests will keep the old API contract alive. Please replace them with the new delete-paused behavior (or drop them).

Inline notes are on the related files.

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -0,0 +1,465 @@
...
+		SnapshotPath:    snapshotPath,
+		MetaDir:         snapshotPath,
+		RootfsVol:       rootfsObject.Name,
+		RootfsKind:      rootfsObject.Kind,
+		MemoryVol:       memoryObject.Name,
+		MemoryKind:      memoryObject.Kind,
+		RootfsSizeBytes: rootfsObject.SizeBytes,
+		Kind:            storage.CatalogKindPauseSnapshot,
+	}); err != nil {
+		stepLog.Warnf("failed to persist pause snapshot catalog for %s: %v", snapID, err)
```
luzhixing12345: catalog.json is important, should not just report warning

CubeShim/shim/src/sandbox/sb.rs
luzhixing12345, line 1346: Currently, the Shim sets its internal state to `Paused` just before executing the actual pause operation, whereas the hypervisor's `pause → snapshot → delete` sequence is performed in stages. If any stage fails(disconnect agent/vm pause to snapshot), the caller cannot accurately determine whether the VM is still running, whether the snapshot has been fully committed to disk, or whether the VM has already been deleted.

It is recommended to transition directly to the `ERROR` state upon failure.

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -0,0 +1,465 @@
...
+	}
+	tmpSnapshotPath := snapshotPath + ".tmp"
+	if _, err := pathutil.ValidatePathUnderBase(DefaultSnapshotDir, tmpSnapshotPath); err != nil {
+		rsp.Ret.RetCode = errorcode.ErrorCode_InvalidParamFormat
+		rsp.Ret.RetMsg = fmt.Sprintf("invalid pause tmp snapshot path: %v", err)
+		return rsp, nil
+	}
+
+	memorySizeBytes := snapshotMemorySizeBytes(resourceSpec.Memory)
+	sourceRootfs, err := storage.GetSandboxRootfs(ctx, req.SandboxID, rootVolumeName)
```
luzhixing12345, line 143: after calculate memorySizeBytes, we should first check the snapshot metadata fs capacity (with statfs) and CubeCow backend capacity with backend metrics before write data to disk

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -0,0 +1,465 @@
...
+	if err != nil {
+		rsp.Ret.RetCode = errorcode.ErrorCode_PreConditionFailed
+		rsp.Ret.RetMsg = fmt.Sprintf("failed to resolve sandbox rootfs: %v", err)
+		return rsp, nil
+	}
+	rootfsObject, err := storage.CommitRootfs(ctx, sourceRootfs, snapID)
```
luzhixing12345: Cubelet first executes `CommitRootfs` on the running guest rootfs before calling the Shim to pause the VM and save its memory. Since the guest can still write to the disk during the interval between these two operations, an inconsistent state—characterized by "new memory" combined with "old disk data"—may arise upon restoration.

Cubelet/services/cubebox/service.go
ls-ggg: Fixed in 7793522.

Create-from-pause now takes the same per-sandbox `sandboxLifecycleLocks` as Pause/Destroy (`resumeFromPauseSandboxID` + `LockContext`). Tombstone replace is restricted to `CONTAINER_PAUSED` only (not `PAUSING`).

Cubelet/services/cubebox/cube_container_create.go
ls-ggg: Fixed in 7793522.

Tombstone replace now requires `CONTAINER_PAUSED` specifically; `PAUSING` is rejected as already exists.

Cubelet/services/cubebox/update.go
ls-ggg: Won't fix in this PR.

With the new Resume path, Resume is treated the same as a normal Create: whether the host has enough capacity should be decided primarily by CubeMaster scheduling. A Cubelet-level fallback check is still worthwhile, but that belongs to a broader Create-path improvement, not a Resume-specific fix here.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
ls-ggg: Won't fix in this PR.

On Pause RPC timeout we never receive `ext_info`, so Master cannot adjust volume refcounts. Heal-to-READY with an empty `pluginVolumeIDs` list intentionally skips that adjustment and still allows Resume to remount from the packed pause `sandbox_spec`. The main downside is a possibly inflated refcount that can block volume deletion when nothing still references the volume. Under the current architecture this is the best trade-off for this extreme timeout path.

CubeMaster/pkg/service/sandbox/sandbox_remove.go
ls-ggg: Fixed (docs) in 7793522.

Updated EN/ZH `lifecycle.md` (and the sdk_compat framework note): deleting a paused sandbox no longer wakes/resumes the MicroVM; it removes the tombstone + pause snapshot directly.

CubeMaster/pkg/sandboxlock/lock.go
ls-ggg: Fixed in 7793522.

`sandboxlock` now stores a per-acquire random token (optional debug prefix) and unlocks with a Lua script that `DEL`s only when `GET` still matches the holder token.

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Fixed in 7793522.

PauseToSnapshot no longer self-exits after Update OK; Cubelet reaps the shim via keep_tombstone `localTask.Delete`. The Update-path `pauseShimLikelyExited` heuristic (including the broad `not found` match) was removed — Update errors are treated as Pause failure.

docs/dev/redis-key-spec.md
ls-ggg: Fixed (docs) in 7793522.

`docs/dev/redis-key-spec.md` now documents op-specific TTLs (pause 180s / resume·delete 120s / default 60s) and token-safe Lua unlock.

docs/zh/dev/redis-key-spec.md
ls-ggg: Fixed (docs) in 7793522.

`docs/zh/dev/redis-key-spec.md` updated to match: pause 180s / resume·delete 120s / default 60s, token-safe Lua unlock.

CubeMaster/pkg/cubeproxy/admin.go
ls-ggg: Fixed (docs) in 7793522.

Documented in EN/ZH lifecycle: CubeMaster `cubeproxy.admin_token` must match CubeProxy `$cube_admin_token`, otherwise backend-cache purge returns 403 and stale routes can cause same-node 504 after Resume.

CubeProxy/tests/backend_cache_test.lua
ls-ggg: Fixed in 7793522.

`CubeProxy/Makefile` `test` now runs `backend_cache_test.lua` alongside the sentinel Lua test when `lua`/`luajit` is available.

.gitignore
ls-ggg: Fixed in 7793522.

Removed personal local-script / `.cursor` entries from `.gitignore`, and updated the PR test plan so it no longer claims an in-repo 49/49 matrix script.

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Fixed in 7793522.

`WriteSnapshotCatalog` failure now fails Pause (`failPause`) and cleans up the renamed snapshot directory; we no longer mark PAUSED without a usable catalog.

CubeShim/shim/src/sandbox/sb.rs
ls-ggg: Fixed in 7793522.

On mid-pause failure after entering the pause path, shim-local state is now set to **`Exited`** (not left as misleading `Paused`). We did not add a new `SandBoxState::Error` variant — `Exited` is the existing terminal state used for this case.

On success, PauseToSnapshot no longer self-exits after Update OK; Cubelet reaps the shim via keep_tombstone Delete. Customer-visible sandbox state remains Cubelet/Master.

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Follow-up / out of scope for this PR.

Agreed that pre-checking metadata FS / CubeCow capacity before Pause writes is valuable. Tracking as a follow-up.

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Fixed in 7793522.

Order is now PauseToSnapshot (freeze memory) first, then `CommitRootfs` on the still-present live rootfs volume before keep_tombstone Destroy, so the disk snapshot matches the frozen guest memory.

CubeShim/shim/src/sandbox/sb.rs
ls-ggg: (superseded) See the corrected Fixed reply above — mid-pause failure now sets shim state to `Exited` in 7793522.

CubeProxy/nginx.conf
```diff
@@ -476,11 +476,12 @@ http {
     # this server.
     #
     # Routes:
-    #   POST /admin/meta/upsert     body=JSON {sandbox_id, ...}    → cube_sandbox_meta
-    #   POST /admin/meta/delete     body=JSON {sandbox_id}         → drops meta+state+last_active
-    #   POST /admin/state           body=JSON {sandbox_id, state}  → cube_sandbox_state
-    #   GET  /admin/last_active                                     → JSON snapshot of dict
-    #   GET  /admin/last_active?since=<ms>                          → only entries newer than `since`
+    #   POST /admin/meta/upsert           body=JSON {sandbox_id, ...}    → cube_sandbox_meta
+    #   POST /admin/meta/delete           body=JSON {sandbox_id}         → drops meta+state+last_active+backend cache
```
chenhengqi, line 480: There is a onging work in #1252. I think we can reuse the `/admin/meta/delete` endpoint.

luzhixing12345:
we should add `Cubelet/api/services/cubebox/v1/cubebox_grpc.pb.go` and `Cubelet/doc/cubelet-api.md` to this git commit, after run ./deploy/one-click/build-release-bundle-builder.sh these files will be updated.

CubeProxy/nginx.conf
ls-ggg: Good point — thanks for the pointer to #1252.

We looked at reusing that path. The distinction is:

- **What Resume needs:** invalidate CubeProxy's in-process `local_cache` route entries only (so the next request misses and reloads the rewritten Redis `SandboxProxyMap`). Lifecycle dicts (`cube_sandbox_meta` / `state` / `last_active`) and Redis must stay intact for a paused sandbox.
- **What `/admin/meta/delete` in #1252 does:** clear the fixed route-cache keys *and then* delete META/STATE/LAST — that full "sandbox is gone" semantics is correct for Destroy, but calling it on Resume would drop lifecycle bookkeeping and break auto-resume / CLM coordination.
- So we can reuse the **cache-clear helper** from #1252 (delete `meta_cached` + fixed suffixes; Redis miss → reload), but **not** the whole `/admin/meta/delete` endpoint as-is.

For this PR we keep `POST /admin/backend_cache/delete` to avoid coupling on #1252 landing. After #1252 merges, I'll open a follow-up PR to extract/share that local-cache invalidation and consolidate Resume onto it (without deleting lifecycle meta).

ls-ggg:
@luzhixing12345 Thanks — good catch on the generated docs.

- `Cubelet/doc/cubelet-api.md` was stale vs the Pause-related proto fields (`UpdateCubeSandboxResponse.ext_info`, `CleanupTemplateResponse.plugin_volume_ids`). Regenerated via the builder image (`make proto` in Cubelet) and included in the latest commit.
- `Cubelet/api/services/cubebox/v1/cubebox_grpc.pb.go` was already part of this PR (message-field-only proto change; no substantive gRPC stub delta beyond the existing generator version line).

luzhixing12345:
try the test code, it will continuously create and pause sandbox util no space left on device

```python
from datetime import datetime
import time
n = 0
while True:
    n += 1
    sb = Sandbox.create()
    print(sb.sandbox_id)
    print(f"create {n} sandbox at {datetime.now().isoformat(timespec='seconds')}")
    sb.pause()
    print(f"pause {n} sandbox at {datetime.now().isoformat(timespec='seconds')}")
    time.sleep(1)
```

when no space left on device, resume will be rejected.

Clipboard_Screenshot_1786607926

and sandbox status goto unknown

```bash
sandbox_id                          status     host_id           create_at              pause_at
be862840dc3245599334e139fb89c379    paused     21.130.250.164    2026-08-13 15:50:11    2026-08-13 15:50:12
858d72ba3cfd42d5ad6a9091284f8517    paused     21.130.250.164    2026-08-13 15:50:13    2026-08-13 15:50:13
30286440b5d3402d885a02d94183a615    paused     21.130.250.164    2026-08-13 15:50:15    2026-08-13 15:50:56
f67eaed75e194dce80465cca5107a7cd    paused     21.130.250.164    2026-08-13 15:50:57    2026-08-13 15:50:58
a54ff9971d974e9bb49ccfe5374b4788    paused     21.130.250.164    2026-08-13 15:50:59    2026-08-13 15:51:00
84f1b145032a4eb19e9bdd8a4cfc0175    paused     21.130.250.164    2026-08-13 15:51:01    2026-08-13 15:51:41
344c1e2100b34b558a6c5180c6dc50b5    paused     21.130.250.164    2026-08-13 15:51:42    2026-08-13 15:51:43
d492ac24ca3e40f0b979d6425d936a96    paused     21.130.250.164    2026-08-13 15:51:44    2026-08-13 15:51:45
23fecaa8ac2f4bec98808f1cbbccdcbf    paused     21.130.250.164    2026-08-13 15:51:46    2026-08-13 15:51:47
26ba3bb133ba42b086a85f34825864fe    paused     21.130.250.164    2026-08-13 15:51:48    2026-08-13 15:52:40
cdee1cf553424826aee330f5097473b1    unknown    21.130.250.164    2026-08-13 15:58:36    -
```

however, destroy the cubebox by webui/cli doesn't cleanup the disk to free more space

```bash
~/CubeSandbox# df -h /data/cubelet
Filesystem      Size  Used Avail Use% Mounted on
/dev/loop0       25G   25G  314M  99% /data/cubelet

~/CubeSandbox# cubemastercli cubebox destroy cdee1cf553424826aee330f5097473b1
2026/08/13 16:01:02 doDestroySandbox RequestId:f7b442e3-7060-4538-a97b-ce26a26518f5,sandBoxId:cdee1cf553424826aee330f5097473b1, code:200, message:Success,cost:25
2026/08/13 16:01:02 destroyed: cdee1cf553424826aee330f5097473b1


~/CubeSandbox# df -h /data/cubelet
Filesystem      Size  Used Avail Use% Mounted on
/dev/loop0       25G   25G  314M  99% /data/cubelet
```

now all the sandbox could not be resumed.

there is something wrong with the cleanup code. check Cubelet/services/cubebox/service.go:704

ls-ggg:
@luzhixing12345 fixed — please retry the disk-full → unknown → delete scenario.

Cause: when Pause failed halfway the shim was still alive, but delete took the tombstone path (CubeBox row only), so the half-finished `tpl-snap-*-memory` objects stayed on disk. Delete of unknown / FAILED now does a full Destroy to kill the live shim, then CleanupTemplate to drop the partial snapshot.

Assisted-by: Cursor:Composer

ls-ggg:
Testing for this PR has largely passed. Cross-node pause/resume will land in a follow-up PR with further related changes, so let's merge this one first.

Assisted-by: Cursor:Composer

PR #1323, opened 2026-08-11
feat(examples): add derived sandbox-code image with run_code env inheritance

### LINKED ISSUE #565: [Bug Report] Create-time envs are not fully propagated to sandbox runtime execution
## Summary

Create-time envs passed through `Sandbox.create(envs=...)` are not fully propagated to sandbox runtime execution.

On current `origin/master`, both `commands.run` and `run_code` miss those create-time envs. After applying the related CubeAPI-side fix, `commands.run` works, but `run_code` still does not.

This issue is a follow-up to #427. That earlier issue captured the broader propagation gap, and maintainer feedback there clarified that generic environment refresh at sandbox startup is not currently supported.

## Environment

- CubeSandbox version / commit: current `origin/master`
- Host OS and kernel version: Rocky Linux 9.6, `5.14.0-570.17.1.el9_6.x86_64`
- KVM info (`modinfo kvm`): host supports KVM; the issue is in the control/runtime integration path rather than KVM itself
- Deployment mode: cluster
- Relevant component: CubeAPI / sandbox runtime image / bundled lightweight code interpreter

## Steps to Reproduce

1. Create a sandbox through the E2B SDK with create-time envs on current `origin/master`:

```python
from e2b_code_interpreter import Sandbox

with Sandbox.create(
    template="tpl-xxxx",
    envs={"CUBE_TEST_CREATE": "from-e2b-create"},
) as sandbox:
    print("commands.run:")
    print(sandbox.commands.run("echo $CUBE_TEST_CREATE").stdout)

    print("run_code:")
    result = sandbox.run_code(
        "import os; print(os.environ.get('CUBE_TEST_CREATE', '<NOT SET>'))"
    )
    print(result.logs.stdout)
```

2. Observe the behavior of `commands.run` and `run_code`.

## Actual Behavior

On current `origin/master`:

```text
commands.run:


run_code:
['<NOT SET>\n']
```

Both execution paths fail to see the create-time env var.

## Related PR / Partial Verification

A separate CubeAPI-side PR is prepared for the CubeAPI -> envd propagation layer.

Related CubeAPI-side PR: #566

With that PR applied, the same script produces:

```text
commands.run:
from-e2b-create

run_code:
['<NOT SET>\n']
```

So the related PR fixes the CubeAPI -> envd propagation layer for envd-backed command execution, such as `commands.run`.

The remaining gap is that `run_code` still does not inherit sandbox-level envs, because the bundled lightweight code interpreter in the sandbox image does not currently read envd `/envs`.

### Verified split of the problem

The issue currently breaks down into two layers:

1. CubeAPI create-path propagation
2. runtime/image-side `run_code` propagation

The related CubeAPI-side fix addresses the first layer by:

- accepting E2B SDK `envs` as an alias of `envVars`
- posting create-time envs to envd after sandbox creation succeeds

That is enough to restore `commands.run`.

This keeps the follow-up narrowly scoped relative to #427: it does not claim that every runtime path now supports startup-time env refresh.

### Why `run_code` still fails

`run_code` does not go through the same envd-backed command path as `commands.run`.

The bundled lightweight code interpreter in the sandbox image does not currently read sandbox-level envs from envd `/envs`, so even after envd has the correct values, `run_code` still prints `<NOT SET>`.

### Verified behavior with the related CubeAPI-side fix

Cluster validation confirmed:

- E2B SDK `envs` path works for `commands.run`
- CubeSandbox SDK native `envVars` path also works for `commands.run`
- per-call command env overrides still work
- after a per-call override, the next `commands.run` falls back to the sandbox-level env again
- `pause` / `resume` preserves the same `commands.run` env visibility in the current lifecycle model

So for the currently verified lifecycle:

- a single create-time envd `/init` call is sufficient for `commands.run`
- a separate resume-time re-init path does not appear necessary

### Suggested follow-up direction

For the remaining `run_code` gap, the most direct follow-up appears to be updating the bundled lightweight code interpreter to read sandbox-level envs from envd `/envs` before applying per-call env overrides.

Preferred precedence:

```text
sandbox-level envs from envd < per-call envs
```

### Maintainer feedback requested

Would updating the bundled lightweight code interpreter to read envd `/envs` be the preferred follow-up direction for making `run_code` consistent with `commands.run`?

A separate CubeAPI-side fix is prepared for the first layer: propagating create-time envs into envd so that `commands.run` works correctly.

zyl1121:
## Motivation

PR #566 made sandbox create-time environment variables available to envd-backed command execution, so `commands.run` now observes create-time environment variables correctly.

However, `run_code` still behaves differently. The bundled lightweight code interpreter does not load create-time values from envd, so code executed through `run_code` cannot see environment variables that were provided when the sandbox was created.

This means `commands.run` and `run_code` can observe different environments in the same sandbox. That is surprising for users who configure API keys, feature flags, or runtime settings at sandbox creation time.

This PR adds a derived `sandbox-code` example image that lets the lightweight code interpreter inherit sandbox create-time environment variables, while keeping per-call `run_code` environment overrides temporary and scoped to a single execution.

The per-call lifecycle follows [e2b-dev/code-interpreter#141](https://github.com/e2b-dev/code-interpreter/pull/141): inject environment values into the user execution, clean them up in a background execution, and make the next request wait for cleanup.

Fixes #565.
Follow-up to #566.

## What Changed

* Added `examples/run-code-env-inheritance/`, a derived image based on the official `sandbox-code` image.
* Replaced only `/opt/lightweight-code-interpreter/server.py` in the derived image.
* Loaded create-time environment variables from envd once per Jupyter kernel and applied environment setup in a separate background execution before unchanged user code.
* Marked sandbox-level values as applied only after environment setup succeeds, so a compilation failure in the first user cell does not suppress later initialization.
* Preserved per-call override behavior by applying `env` and `env_vars` after create-time values.
* Snapshotted prior kernel values for per-call keys, restored existing values after execution, removed only previously absent keys, and made the next execution wait for cleanup.
* Rolled back the per-call snapshot immediately when background environment setup fails, while leaving sandbox initialization retryable on the next execution.
* Disabled Jupyter history storage for silent setup and cleanup executions.
* Made the envd fetch timeout configurable through `ENVD_TIMEOUT`, while preserving fail-fast HTTP 502 behavior for unavailable or invalid `/envs` responses.
* Added focused interpreter tests for envd loading, caching, precedence, prior-value restoration, first-cell compilation failures, cleanup, retry behavior, and invalid envd responses.
* Extended shared SDK compatibility adapters with optional per-call `run_code` environment support.
* Added shared E2E cases for both E2B and CubeSandbox backends, covering temporary per-call restoration and first-cell syntax failures. They are disabled by default and skip before sandbox creation unless `SDK_E2E_RUN_CODE_ENV_INHERITANCE=true` is explicitly set for a compatible template.
* Updated English and Chinese SDK compatibility documentation and example instructions, including registry-backed and one-click local-image template creation paths.

## Validation

* Derived-image interpreter tests: 14 passed.
* Default SDK compatibility behavior without the opt-in flag: 4 skipped before sandbox creation.
* Opt-in SDK compatibility E2E through E2B and CubeSandbox: 4 passed.
* Built the derived image, created a template, and verified:

  * create-time environment inheritance
  * temporary per-call override precedence
  * restoration of sandbox-level and kernel-defined values after per-call overrides
  * cleanup that tolerates user code deleting a per-call key
  * sandbox environment initialization before a failing first user cell
  * rollback and retry after background environment setup failure

## Scope

This PR adds an opt-in derived `sandbox-code` example image and shared compatibility coverage.

The replacement server is based on the lightweight code interpreter already bundled in the official `sandbox-code` image. Existing execution, context, restart, and timeout behavior is intentionally preserved; this example changes only sandbox environment loading and the per-call environment overlay lifecycle.

It does not modify the default `sandbox-code` image or any existing templates. Users who want this behavior need to build the derived image and create a template from it.

Create-time environment variables are read once per Jupyter kernel. This PR does not add generic environment refresh behavior. Per-call `run_code` values remain temporary overlays for a single execution.

The derived interpreter requires envd's `/envs` endpoint and fails with HTTP 502 when that dependency is unavailable or returns an invalid payload; it does not fall back to the stock interpreter behavior.

The shared environment-inheritance E2E case is disabled by default and skips before sandbox creation unless `SDK_E2E_RUN_CODE_ENV_INHERITANCE=true` is explicitly set for a compatible template.

### DIFF
--- a/examples/run-code-env-inheritance/Dockerfile
+++ b/examples/run-code-env-inheritance/Dockerfile
@@ -1,0 +1,6 @@
+# syntax=docker/dockerfile:1.7
+
+ARG SANDBOX_CODE_IMAGE=cube-sandbox-cn.tencentcloudcr.com/cube-sandbox/sandbox-code:latest
+FROM ${SANDBOX_CODE_IMAGE}
+
+COPY lightweight-code-interpreter/server.py /opt/lightweight-code-interpreter/server.py
--- a/examples/run-code-env-inheritance/README.md
+++ b/examples/run-code-env-inheritance/README.md
@@ -1,0 +1,118 @@
+# Derived `sandbox-code` image: make `run_code` inherit sandbox-level environment variables
+
+[中文](README_zh.md)
+
+This example provides a derived image based on the official `sandbox-code` image. It replaces only the lightweight code interpreter so `run_code` can read sandbox-level environment variables injected through `Sandbox.create(envs=...)`, while preserving temporary per-call environment overrides.
+
+## Background
+
+Environment variables passed to `Sandbox.create(envs=...)` are available to `commands.run`, but `run_code` in the official `sandbox-code` image cannot read them because its lightweight code interpreter does not inject the sandbox environment from envd into the Jupyter kernel.
+
+This derived image aligns the environments inherited by `run_code` and `commands.run` without modifying the default `sandbox-code` image or any existing templates.
+
+## Environment semantics
+
+Before the first user execution in each Jupyter kernel, the lightweight code interpreter reads and caches the sandbox-level environment variables from `http://127.0.0.1:49983/envs`, then injects them into the kernel. Each `run_code` call applies its `env` or `env_vars` values afterward, with the following precedence:
+
+```text
+Sandbox.create(envs=...) < per-call run_code environment variables
+```
+
+Sandbox-level values and per-call overrides are applied in a separate background kernel execution before the user code, so a compilation failure in the first user cell does not prevent environment initialization. Before applying per-call values, the interpreter snapshots the previous kernel value of each affected key. Afterward, a background cleanup restores the environment:
+
+- Keys that existed before the call are restored to their previous kernel values.
+- Keys that did not exist before the call are removed.
+
+The next execution waits for pending cleanup so per-call values cannot leak into later executions. This lifecycle follows E2B code-interpreter. Sandbox-level environment variables are loaded only once per Jupyter kernel; this example does not provide a general runtime refresh mechanism.
+
+This derived interpreter requires envd's `/envs` endpoint. If envd is unavailable, the endpoint is unsupported, the request exceeds `ENVD_TIMEOUT` (2 seconds by default), or the payload is invalid, `run_code` fails with HTTP 502 instead of falling back to the stock interpreter behavior. Set `ENVD_TIMEOUT` in the image environment to adjust the fetch timeout.
+
+## Build and test the image
+
+From the repository root:
+
+```bash
+docker build \
+  -t sandbox-code-env-inheritance:latest \
+  examples/run-code-env-inheritance
+```
+
+Override `SANDBOX_CODE_IMAGE` when using the international registry or a pinned base image:
+
+```bash
+docker build \
+  --build-arg SANDBOX_CODE_IMAGE=cube-sandbox-int.tencentcloudcr.com/cube-sandbox/sandbox-code:latest \
+  -t sandbox-code-env-inheritance:latest \
+  examples/run-code-env-inheritance
+```
+
+After building the image, run the focused unit tests inside it so they use the same Python dependencies as the runtime:
+
+```bash
+docker run --rm \
+  --entrypoint python \
+  -v "$PWD/examples/run-code-env-inheritance:/work:ro" \
+  -e PYTHONPATH=/work/lightweight-code-interpreter \
+  sandbox-code-env-inheritance:latest \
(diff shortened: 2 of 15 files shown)


lkml-likexu:
@zyl1121 Would you mind to rebase this one ?

zyl1121:
> Would you mind to rebase this one ?

Rebased and resolved the conflicts. All E2E tests have passed.

luzhixing12345:
LGTM. I tested it according to the instructions in the readme and didn't encounter any blocking issues.

PR #1345, opened 2026-08-13
fix(cubeapi): map CubeMaster 130400 to HTTP 400 on sandbox paths

dushulin:
## Summary

CubeMaster returns `MasterParamsError` (130400) when the request itself is invalid — e.g. a `host-mount` `hostPath` outside `allowed_host_mount_prefixes`, or a malformed field. On the sandbox paths cube-api funnelled that into `AppError::Internal` → **HTTP 500**.

`templates::map_err` already classifies 130400 as 400 via `CubeMasterError::is_params_error()`, whose doc comment states *"should surface as HTTP 400, not 500"*. The sandbox paths were never wired to it.

Two consequences: a client gets 500 for its own bad input and retries a request that can never succeed; and the server-side success-rate SLI derived from `cubeapi_business_request_total` records a client mistake as a server fault.

## Where the fix has to go

`parse_response` converts every non-success `ret_code` into `CubeMasterError::Api` **before** the caller can inspect the envelope, so the `ret.into_result()` / `ensure_*` checks never observe a business code — the pre-existing comment above `map_update_cubemaster_err` already noted this for the pause/resume path. The classification therefore belongs on the transport error:

| site | role |
| --- | --- |
| new `params_error_or_internal` | applied at the client-facing call sites: create, list, logs (v1/v2), timeout, refresh, detail |
| `sandbox_not_found_or_internal` | delegates its fallthrough to it, so the delete and update mappers inherit the same behaviour |

`is_invalid_path_parameter()` is included alongside `is_params_error()`, matching `templates::map_err` — a malformed sandbox id is also client input.

`set_timeout` / `refresh` additionally gain the 404 mapping their own envelope check already performed, so both halves of those paths now agree.

**Only 130400 is remapped.** `130408 CubeletUnHealthy` reads like a client error but is a backend fault, so remapping the whole `1304xx` prefix would hide real faults from the SLI.

## Test plan

- [x] `cargo test --bin cube-api` — 113 passed, 0 failed
- [x] `cargo fmt --check` clean, `cargo clippy` no new warnings
- [x] Tests drive the **real service methods** against a mocked CubeMaster returning a 130400 envelope (`create_sandbox`, `set_timeout`, `refresh`), asserting `AppError::BadRequest` **and** `StatusCode::BAD_REQUEST`
- [x] Each test was confirmed to fail when the mapping is reverted to `internal_error`, so it gates the real request path rather than a helper in isolation
- [x] `create_sandbox_keeps_backend_faults_internal` is the negative control: 130408 and 130593 must stay `Internal` / 500

### DIFF
--- a/CubeAPI/src/services/sandboxes.rs
+++ b/CubeAPI/src/services/sandboxes.rs
@@ -101,7 +101,7 @@ impl SandboxService {
             .cubemaster
             .list_sandboxes(&req)
             .await
-            .map_err(internal_error)?;
+            .map_err(params_error_or_internal)?;
 
         ensure_create_result(resp.ret.ret_code, resp.ret.ret_msg)?;
 
@@ -274,7 +274,7 @@ impl SandboxService {
             .cubemaster
             .create_sandbox(&req)
             .await
-            .map_err(internal_error)?;
+            .map_err(params_error_or_internal)?;
 
         resp.ret.into_result().map_err(internal_error)?;
 
@@ -431,7 +431,7 @@ impl SandboxService {
             Err(e) if e.is_not_found() => {
                 Err(AppError::NotFound(format!("sandbox {} not found", sandbox_id)))
             }
-            Err(e) => Err(internal_error(e)),
+            Err(e) => Err(params_error_or_internal(e)),
         }
     }
 
@@ -468,7 +468,7 @@ impl SandboxService {
                 "sandbox {} not found",
                 sandbox_id
             ))),
-            Err(e) => Err(internal_error(e)),
+            Err(e) => Err(params_error_or_internal(e)),
         }
     }
 
@@ -484,7 +484,7 @@ impl SandboxService {
             .cubemaster
             .set_sandbox_timeout(&req)
             .await
-            .map_err(internal_error)?;
+            .map_err(|e| sandbox_not_found_or_internal(e, sandbox_id))?;
 
         resp.ret
             .into_result()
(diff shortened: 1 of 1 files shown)


dushulin:
Thanks — all three points are correct, and the diff has been reworked accordingly.

Confirming the analysis: `parse_response` returns `Err(CubeMasterError::Api { .. })` for every `ret_code` that is neither 0 nor 200, so `ensure_create_result` / `ensure_update_result` only ever observe a success code in practice. `ensure_create_result` has a single call site (inside `list`) and its transport error was `map_err(internal_error)`, so the guard I added there was unreachable; more importantly `create_sandbox` mapped its transport error to `internal_error` before `resp.ret.into_result()` ran, so the create path — the one the report came from — still returned 500. The pre-existing comment above `map_update_cubemaster_err` had already documented this for pause/resume; I failed to generalise it.

Changes in the force-push:

- New `params_error_or_internal` classifies the **transport** error and is applied at the client-facing call sites: `create_sandbox`, `list`, `get_logs`, `get_logs_v2`, `set_timeout`, `refresh`, `fetch_sandbox_detail`.
- `sandbox_not_found_or_internal` delegates its fallthrough to it, so the delete and update mappers inherit the behaviour without a second copy of the predicate.
- The unreachable `RET_CODE_BAD_REQUEST` branches in `ensure_create_result` / `ensure_update_result` are removed, along with the now-unused constant.
- Tests now drive the real service methods against a mocked CubeMaster returning a 130400 envelope (`create_sandbox`, `set_timeout`, `refresh`). Each one was verified to fail when the mapping is reverted to `internal_error`, so they gate the actual request path rather than a helper in isolation. `create_sandbox_keeps_backend_faults_internal` is the negative control over 130408 / 130593.

`set_timeout` / `refresh` additionally gained the 404 mapping their own envelope check already performed, so both halves of those paths now agree.

luzhixing12345:
good catch, LGTM, also fix timeout and refresh

PR #1347, opened 2026-08-14
feat(cubenet): support arbitrary custom L7 port + scheme in egress rules

zhouxianping:
## Summary
  L7 egress rules can now pin the TCP port CubeEgress intercepts on, via
  `Match.port` + `Match.scheme`. This extends the existing host/path/SNI
  matching so a rule applies to a specific `(host, port, scheme)` tuple instead
  of always the default `{80/http, 443/https}` set.

  ## Changes by layer
  - **CubeNet dataplane** (`mvmtap`/`session`/`tcp`): port+scheme → `skb->mark`
    on egress SYN; spec tests updated.
  - **CubeEgress**: Lua policy + iptables mark validation for custom
    `(host, port, scheme)`.
  - **Cubelet**: embedded network runtime + policy builder handle port/scheme.
  - **CubeAPI / CubeMaster**: validate egress rule port/scheme at API and master.
  - **SDKs** (go / node / python): port/scheme validation in egress rule match.
  - **deploy/one-click**: bind L7 skb-mark config into the install flow.
  - **docs** (`security-proxy.md` EN + ZH): document custom L7 ports, the four
    port/scheme quadrants, and constraints.
  - **examples**: L7 custom-port echo demos.

  ## SDK compatibility e2e (this session)
  - `cases/network/test_l7_custom_port.py`: 5 cases — custom HTTP port injection,
    scheme-only default-set interception, HTTPS custom-port interception,
    port-without-scheme rejection, subnet-host rejection.
  - `framework/capabilities.py`: `NETWORK_L7_CUSTOM_PORT`.
  - `CubeNet/cubevs/Makefile`: `test: gen` so stale BPF objects are regenerated
    before `make test` (prevents verifier failures from outdated `.o`).

  ## Validation
  - `go test ./...` under `CubeNet/cubevs` passes (after `make gen`).
  - SDK compatibility e2e: `test_l7_custom_port.py` 5 passed against a template
    that trusts the L7 interception CA.


fslongjin:
## Review of PR #1347 — arbitrary custom L7 port + scheme in egress rules

Overall this is a well-engineered change: the unified `classify_egress_flow()` verdict, the `/48` port-in-key LPM trie, the `key_prefixlen` guard for exact-key merge, the fail-closed unknown-scheme handling, and the static asserts mirroring BPF structs are all high quality. The findings below are what I'd want addressed before merge, in severity order.

### Correctness

**1. [High] Migration rollback ordering can permanently lose `allow_out` policy**
`CubeNet/cubevs/migration.go:98-118` (with `Init` rollback at `miscs.go:299-305`)

`migratePersistentPolicyMaps()` deletes the legacy `allow_out`/`allow_out_v2` pin (line 109) immediately after `migrateAllowOutMap` succeeds, **before** `migrateDNSAllowMap` (line 114). If the DNS migration then fails, `Init`'s deferred rollback removes the freshly-populated `allow_out_v3`/`dns_allow_v2` pins; on the next restart `allowOutMigrationSource()` returns `""` (legacy pin is gone) and the allow_out policy is never re-migrated. The "leave the legacy pin in place so the migration can be retried" comment only covers an allow-map failure, not a later DNS failure. Suggestion: migrate both maps first and only unlink the legacy pins after both succeed. The existing rollback tests (`TestMigrateDNSAllowMapFailureKeepsLegacyPin`, `TestMigrateAllowOutMapFailureKeepsLegacyPin`) only exercise single-map failure and miss this combined case.

**2. [High] Port-less deny rules are silently narrowed to `{80/http, 443/https}`; a custom-port allow can bypass a broader deny**
`CubeEgress/lua/access_phase.lua:144-152` via `port_scheme.matches`

Pre-PR, a deny rule with a host + scheme matched **all** intercepted traffic to that host. Post-PR, `port_scheme.matches(nil, nil, dst_port, scheme)` expands to only `{80/http, 443/https}`, so a deny rule with no port matches only the default set. Combined with a custom-port allow rule (e.g. `deny host="*.example.com"` + `allow host="api.example.com", port=8443, scheme=https`), `api.example.com:8443` traffic is allowed because the deny rule never matches port 8443 and `decide()` is first-match-wins. For **deny** rules the narrowing is fail-open (for allow rules it's fail-closed). Consider making deny apply port-agnostically within the host, or at minimum documenting the narrowing prominently and auditing existing deny rules on upgrade.

**3. [Medium] Non-atomic iptables chain install can leave a fail-open interception state**
`CubeEgress/scripts/cube-proxy-iptables-init.sh:124-144`

The chain is built by `iptables -F` then two sequential `-A` TPROXY rules with no rollback and no post-condition check. If the second (HTTPS) rule fails after the first succeeds, the script aborts before `install_routing()`, so mark-based `ip rule` routing is never installed and HTTPS-marked packets fall through to the default route — a silent bypass of all L7 interception/deny for the HTTPS scheme. Consider building into a scratch chain + atomic swap (or verifying both rules are present before proceeding).

**4. [Medium] Cubelet `cloneEgressRules` shallow-copies the new `Match.Port` pointer**
`Cubelet/network/runtime/policy_builder.go:47-51`

`match := *r.Match` shares `Port *int` (and the pre-existing `Host`/`Path`/`SNI`/`Scheme` pointers) with the caller's request, despite the function contract claiming a deep copy. `Method` is correctly deep-copied. CubeMaster fixed this with `cloneIntPtr` (`types.go:227`); Cubelet should mirror it.

### Test coverage / CI

**5. [High] The new `CubeNet/cubevs` unit tests (~2000 lines, 8 files) are not executed by CI or `run.sh`**
`.github/workflows/unit-test-check.yml:185-189` maps `CubeNet/*` → `cubelet` → `cubelet-pkg-test` (`go test -short ./pkg/...` under Cubelet). `CubeNet/cubevs` is a separate Go module, so `custom_port_test.go`, `dns_learn_test.go`, `egress_policy_test.go`, `l7_mark_test.go`, `migration_test.go`, `tcp_state_test.go`, `netpolicy_test.go`, `dump_test.go` are never compiled/run. Add a `go test ./...` for the cubevs module to the workflow (and `tests/unittest/run.sh`).

**6. [High] The `FLOW_HTTP` verdict — the primary new plaintext path — is never asserted**
`CubeNet/cubevs/egress_policy_test.go:12-17`

The verdict constants define `reject=0, SNAT=1, HTTPS=3` but no `HTTP=2`, and the seed cases only cover HTTPS. Add `flowVerdictHTTP = 2` and a case seeding a `/48` with `Scheme: L7SchemeHTTP`.

**7. [High] iptables TPROXY rule generation is stubbed out in tests**
`CubeEgress/tests/cube-proxy-iptables-init_test.sh:98` overrides `install_chain()` with `echo`, so the actual two mark→TPROXY rules (`--on-port 8080/8443`, mark/mask match) and the gateway-from-CIDR arithmetic are never asserted. A regression in `--on-port` or the mark match would pass CI.

**8. [Medium] Migration success path and dump port rendering are untested**
- `CubeNet/cubevs/migration_test.go` has full bpffs success tests for DNS migration but only rollback tests for `allow_out_v2`→`allow_out_v3`; the 16-byte L7+expires → `/48` expansion upgrade path is unproven.
- `CubeNet/cubevs/dump_test.go` never exercises `l7PortEntriesToDump` / `dumpPolicyInnerMap` NBO→host port rendering; a byte-order regression would show `0x20EB` instead of `8443`.

**9. [Medium] HTTPS custom-port e2e asserts nothing that distinguishes interception from passthrough**
`tests/e2e/sdk_compat/cases/network/test_l7_custom_port.py` asserts only `code=200` + non-empty body against a public endpoint. Plain SNAT would satisfy the same invariant, and the case is gated behind `SDK_E2E_L7_HTTPS_CAP_TRUSTED=1` (silently skipped by default). Consider an endpoint that echoes a request-observable signal, or document that the leg only proves reachability.

### Docs / conventions

**10. [Medium] Non-English character in a Go comment**
`CubeNet/cubevs/dump.go:184` — `Scheme string `json:"scheme"` // "http" | "https" | ""(未知)` violates the project's non-English-character policy. Change to `"" (unknown)`.

**11. [Medium] "Enforced by the SDK client" overstates SDK coverage**
`docs/guide/security-proxy.md:148` (and ZH `:135`): the SDK clients only validate the `port` range + scheme + port-requires-scheme; the subnet-CIDR rejection, the `(host, port)` scheme-conflict rejection, and the 8-tuple cap are server-side only. Reword to "validated server-side; the SDK client also pre-validates the port/scheme pair".

### Minor

**12. [Low] Scheme-only rules silently narrow datapath interception to a single port**
`Cubelet/network/runtime/policy_builder.go:218` expands scheme-only `https` to port 443 only. Pre-PR the host stayed intercepted on the full `{80,443}` set and the scheme was a CubeEgress filter; with a CubeEgress default-deny sandbox, plaintext HTTP to that host now bypasses the proxy entirely. Docs were updated to the new semantics, but the Python SDK `Match.__post_init__` docstring (`sdk/python/cubesandbox/_policy.py:87-92`) and the Go SDK test comment still describe the old behavior — callers will be misled.

**13. [Low] Dead validation helpers with a contradictory counting model**
`CubeNet/cubevs/netpolicy.go:759-805` — `validateNetPolicyEntryCounts`/`countUniqueLPMEntries`/`countUniqueDNSAllowEntries` are only referenced from tests and count one entry per LPM key, contradicting `expandedAllowOutEntryCount` (per-port). If re-wired into production they'd reintroduce the E2BIG overflow the PR's own comment describes. Recommend deleting them.

**14. [Low] `from_envoy` reply path now does 2 dependent hash lookups on every TCP reply packet**
`CubeNet/src/localgw.bpf.c:19-46` — `update_l7_session_from_cubedev` runs on every L7-proxy reply packet (pull headers + `lookup_session` + `bpf_ktime_get_ns`), where the previous `from_envoy` did zero map lookups. Gate on control-flag packets (`syn||fin||rst`) to remove the steady-state data/ACK cost; verify against the tcp-state tests before landing.

**15. [Low] Cross-layer validation parity: CubeAPI scheme check doesn't trim whitespace**
`CubeAPI/src/services/sandboxes.rs:1107` uses `eq_ignore_ascii_case` without `trim()`, so `" https "` is rejected by CubeAPI but accepted by CubeMaster and all SDKs. Also: `persistentPolicyGenerationExists` (`miscs.go:274-291`) hard-fails with no self-healing when exactly one of `allow_out_v3`/`dns_allow_v2` is pinned.

---

Thanks for the thorough spec-test harness and the e2e coverage — these are above average for the repo. The correctness items (#1, #2) and the CI gap (#5) are the ones I'd prioritize.

CubeNet/cubevs/netpolicy.go
```diff
@@ -600,37 +983,63 @@ func populateAllowOutInnerMap(outerMap *ebpf.Map, ifindex uint32, entries []allo
...
+					// never with a shorter covering entry (whose
+					// flags would otherwise leak into this /48).
+					// Flags only: the static (zero) expiry wins over
+					// a learned same-key entry, so the entry becomes
+					// permanent rather than ageing out at the old TTL.
+					if oldVal.KeyPrefixlen == uint8(key.Prefixlen) {
+						val.Flags |= oldVal.Flags
+					}
+				case !errors.Is(lerr, ebpf.ErrKeyNotExist):
+					return fmt.Errorf("inner map lookup failed: %w, cidr: %s", lerr, entry.source)
```
FakeLearne: ## allow_out gets silently narrowed to the L7 port set when a rule targets the same host

When the same target appears both in allow_out and as an L7 rule host, mergeAllowOutWithL7 (and mergeDNSAllowRules for domains) folds the two into a single entry with flags |= L7_REQUIRED plus the L7 port set. Downstream, that flag makes both writers emit only exact (ip, port)/48 keys:

populateAllowOutInnerMap takes the L7 branch and continues — the plain /32 write below is never reached.
dns_learn_response_ip picks the /48 loop; its /32 branch is mutually exclusive with it.
So the port-agnostic allow the user explicitly asked for in allow_out is dropped.

For example:
allow_out: ["10.1.2.3"]
rules:     [ Match(host="10.1.2.3", port=8443, scheme="https") ]

tcp/8443 would still be intercepted, but every other port to 10.1.2.3 should now miss allow_out_v3, fall through to the 10.0.0.0/8 deny, and get an RST. Domains look the same: allow_out: ["github.com"] + Match(host="github.com", scheme="https") under allow_internet_access=false would take out git@github.com (tcp/22). On master both cases wrote a /32 and every port stayed open, so this reads as a silent behavioural regression — no error, no log.

Suggested fix: track whether the host also came from a plain allow_out (e.g. a hasPlainAllow field on allowOutPolicyEntry and a spare bit in dns_allow_value) and emit both shapes: the /32 (or subnet) any-port entry and the /48 L7 entries. The LPM fallback then does the right thing — /48 for the intercepted ports, /32 for everything else.

lkml-likexu:
@FakeLearne  Please confirm whether the following test cases are already present in `tests/e2e/sdk_compat/` (pls delete for yes). If not, please move them there.

```
tests/test_l7_custom_port_e2e.py::test_l7_custom_http_inject_and_https_deny SKIPPED (use --run-e2e or set CUBE_E2E=1)
tests/test_l7_custom_port_e2e.py::test_l7_custom_https_allow_reaches_real_upstream SKIPPED (use --run-e2e or set CUBE_E2E=1)
tests/test_l7_custom_port_validation_e2e.py::test_l7_custom_scheme_conflict_rejected SKIPPED (use --run-e2e or set CUBE_E2E=1)
tests/test_l7_custom_port_validation_e2e.py::test_l7_custom_port_budget_exceeded_rejected SKIPPED (use --run-e2e or set CUBE_E2E=1)
```

Issue #1349, opened 2026-08-14
CubeMaster go.mod: transitive dep CubeNet/cubevs not resolvable — replace doesn't propagate from Cubelet, go list -m all fails


chenhuiluo:
CubeMaster/go.mod (community main, unmodified) references Cubelet, which requires CubeNet/cubevs. However, CubeMaster itself neither requires nor replaces cubevs, so resolving the module graph fails:

$ cd CubeMaster && go list -m all
   go: github.com/tencentcloud/CubeSandbox/CubeNet/cubevs@v0.0.0: invalid version: unknown revision CubeNet/cubevs/v0.0.0

The same error surfaces in GoLand's dependency sync (and any tool that walks the full module graph). Notably, go build ./... does NOT fail because cubevs is never imported by CubeMaster's code (the compiler only resolves the import graph, not the full require graph) — which makes the issue easy to miss but still real.

Root cause (per Go official docs)

Go's replace directive only applies in the main module's go.mod and is ignored by other modules
(https://go.dev/ref/mod#when-are-replace-and-exclude-statements-used). Concretely:

- Cubelet/go.mod requires cubevs v0.0.0 (line 195) AND replaces it to the local path ../CubeNet/cubevs (line 217). Self-contained.
- CubeMaster/go.mod replaces Cubelet to ../Cubelet (line 189), but CubeMaster does NOT repeat the cubevs require+replace.
- When CubeMaster is the main module, Cubelet's replace cubevs => ../CubeNet/cubevs does not propagate, so   
MVS still tries to fetch cubevs@v0.0.0 from the remote, which has no such tag → unknown revision.

Proposed fix

Add to CubeMaster/go.mod (consistent with the existing Cubelet/CubeDB replace style):

require github.com/tencentcloud/CubeSandbox/CubeNet/cubevs v0.0.0
replace github.com/tencentcloud/CubeSandbox/CubeNet/cubevs => ../CubeNet/cubevs

Alternatives considered

- A repo-level go.work: solves it for developers locally, but GoLand and go list -m all may behave differently across environments, and
go.work is typically gitignored / not the canonical way to express intra-repo module wiring.
- Just the replace (no require): per the same Go doc, "a replace directive alone does not add a module to the module graph; a require ... is also needed" — so require + replace together is the correct fix.

Verification

After adding both lines in CubeMaster/go.mod:
$ cd CubeMaster && go list -m all 2>&1 | grep cubevs
[...]

PR #1351, opened 2026-08-14
fix(cubemaster): add missing cubevs replace so module graph resolves

chenhuiluo:
CubeMaster consumes Cubelet, which requires CubeNet/cubevs. Per the Go module spec, a replace directive does not propagate to consumers — only the main module's go.mod replace applies. CubeMaster's go.mod had a replace for Cubelet but not for its transitive dep cubevs, so resolving the module graph (go list -m all, go mod tidy, IDE sync) tried to fetch cubevs@v0.0.0 from the remote and failed with "unknown revision CubeNet/cubevs/v0.0.0".

Add the replace to point cubevs at the local ../CubeNet/cubevs, mirroring the existing cubelog/CubeDB/Cubelet replace pattern. No require is needed because cubevs already enters the graph transitively via Cubelet.

Verified: go list -m all and go mod tidy now succeed on CubeMaster.


chenhuiluo:
Fixes #1349

PR #1352, opened 2026-08-14
fix(lifecycle): apply explicit connect timeouts

zyl1121:
## Motivation

Connect accepts an optional idle timeout, but explicit values were ignored when the sandbox was already running.

Paused Connect also needs a clear boundary for `timeout=0`: applying it immediately after Resume can re-trigger the lifecycle action while connecting, while E2B-compatible Connect timeout values are expected to be positive or `-1`.

Running Connect should be safe for retries and recovery flows: a shorter requested timeout must not shorten a longer existing deadline.

The Python SDK did not expose the E2B-compatible Connect timeout argument.

## What Changed

- Use the existing set-timeout path when Connect targets an already-running sandbox, but only when the requested deadline is later than the current one; use `set_timeout(...)` for intentional shortening. Pass a paused-sandbox timeout through Resume only when it extends the recorded deadline, and preserve known deadlines during transient pre-running states without inferring a timeout from missing lifecycle metadata.
- Preserve the current timeout when omitted, use `-1` for no expiry, ensure positive values leave at least that much lifetime, and reject `0` or values below `-1` with HTTP 400.
- Keep the deprecated Resume contract unchanged, including its legacy `timeout=0` preserve behavior; `set_timeout(0)` remains available for immediate expiry.
- Add the optional `timeout` argument to Python `Sandbox.connect` and preserve the SDK compatibility adapter's existing wait-budget contract.
- Add focused API, SDK, and compatibility coverage, document the timeout semantics and lifecycle-transition conflicts, and synchronize the OpenAPI 400 and 409 responses.

## Validation

- CubeAPI test suite: 152 passed, including paused/running no-shortening and transient-state timeout regressions.
- Python SDK test suite: 263 passed, 4 skipped.
- SDK compatibility offline suite: 122 passed, 173 live tests skipped as expected.
- Existing runtime E2E continues to cover positive timeout refresh through paused Connect and explicit Resume, verified the exact Redis deadline refresh and actual pause, and cleaned the test sandbox.
- SDK compatibility coverage waits for a running sandbox before explicit Connect calls, gates CubeSandbox-only no-shortening semantics from E2B, and verifies that the longer deadline is preserved without measuring from fixture creation time.

## Scope

This PR only makes explicit Connect timeout values effective, defines their validation boundary, and adds the corresponding Python SDK compatibility surface.

### DIFF
--- a/CubeAPI/src/cubemaster/mod.rs
+++ b/CubeAPI/src/cubemaster/mod.rs
@@ -1397,7 +1397,8 @@ pub struct SandboxUpdateRequest {
     /// "pause" | "resume"
     #[serde(rename = "action")]
     pub action: String,
-    /// TTL in seconds (for resume; 0 = keep original). Optional for pause.
+    /// Idle timeout in seconds for resume (-1 = never, 0 = immediate).
+    /// Omitted for pause or when the caller keeps the current timeout.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub timeout: Option<i32>,
 }
--- a/CubeAPI/src/handlers/sandboxes.rs
+++ b/CubeAPI/src/handlers/sandboxes.rs
@@ -300,6 +300,7 @@ pub async fn pause_sandbox(
     request_body = ResumedSandbox,
     responses(
         (status = 201, description = "Sandbox resumed", body = Sandbox),
+        (status = 400, description = "Invalid timeout value", body = ApiError),
         (status = 404, description = "Sandbox not found", body = ApiError),
         (status = 409, description = "Sandbox is already running", body = ApiError),
         (status = 500, description = "Unexpected backend error", body = ApiError)
@@ -310,6 +311,9 @@ pub async fn resume_sandbox(
     Path(sandbox_id): Path<String>,
     Json(body): Json<ResumedSandbox>,
 ) -> AppResult<impl IntoResponse> {
+    body.validate()
+        .map_err(|e| AppError::BadRequest(e.to_string()))?;
+
     state
         .logger
         .log(
@@ -346,6 +350,7 @@ pub async fn resume_sandbox(
     request_body = ConnectSandbox,
     responses(
         (status = 200, description = "Sandbox connection info", body = Sandbox),
+        (status = 400, description = "Invalid timeout value", body = ApiError),
         (status = 404, description = "Sandbox not found", body = ApiError),
         (status = 500, description = "Unexpected backend error", body = ApiError)
     )
@@ -355,6 +360,9 @@ pub async fn connect_sandbox(
     Path(sandbox_id): Path<String>,
     Json(body): Json<ConnectSandbox>,
 ) -> AppResult<impl IntoResponse> {
+    body.validate()
+        .map_err(|e| AppError::BadRequest(e.to_string()))?;
+
     state
         .logger
         .log(
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -359,11 +359,12 @@ pub struct SandboxDetail {
 // ─── Sandbox — pause/resume/connect/snapshot ──────────────────────────────
 
 /// Request body for POST /sandboxes/{id}/resume (deprecated).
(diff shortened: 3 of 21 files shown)


lkml-likexu:
@zyl1121 Any new idea for this issue or rebase it ?

zyl1121:
> @zyl1121 Any new idea for this issue or rebase it ?

Rebased and tested again. The fix still works as expected.

luzhixing12345:
@zyl1121 Thanks for your contribution. After reviewing and testing it, I found no major issues with the overall implementation; however, I noticed a difference between the current PR's `connect(timeout=N)` design and that of E2B.

For a running sandbox, E2B's `connect(timeout=N)` only extends the deadline if the new timeout is longer [see documentation](https://github.com/e2b-dev/E2B/blob/ccaf9fc0ffe6ac39c7ec786af7608ab1de19467b/packages/python-sdk/e2b/sandbox_sync/main.py#L260-L284).

For instance, if a sandbox has 200 seconds of remaining life, calling `Sandbox.connect(id, timeout=30)` leaves that remaining time unchanged. In contrast, the current PR resets the timeout to 30 seconds starting from the current moment (by internally calling `set_timeout` directly).

While I understand that Cube does not need to replicate E2B's entire lifecycle mechanism, the core behavior of public APIs sharing the same name—especially `connect(timeout=N)`—should align. I prefer E2B's design here because the primary responsibility of `connect` is to "obtain a usable sandbox connection," and it may be frequently invoked by the SDK, retry logic, or recovery mechanisms.

Consider a scenario where a sandbox has one hour of life remaining: if `Sandbox.connect(id, timeout=60)` is called using the current PR's approach, the lifecycle would be abruptly shortened to 60 seconds. This is risky, as the caller might simply want to ensure at least 60 seconds of availability rather than intentionally discarding a longer existing duration; E2B's semantics are safer in this regard. If a user genuinely intends to shorten the lifecycle, they should use the explicitly named `sandbox.set_timeout(60)`.

Would you consider changing the running-sandbox path so that Connect only extends the current deadline, equivalent to:

```python
new_end_at = max(current_end_at, now + timeout)
```

What's your option?

zyl1121:
> For a running sandbox, E2B's connect(timeout=N) only extends the deadline if the new timeout is longer [see documentation](https://github.com/e2b-dev/E2B/blob/ccaf9fc0ffe6ac39c7ec786af7608ab1de19467b/packages/python-sdk/e2b/sandbox_sync/main.py#L260-L284).

> While I understand that Cube does not need to replicate E2B's entire lifecycle mechanism, the core behavior of public APIs sharing the same name—especially connect(timeout=N)—should align. I prefer E2B's design here because the primary responsibility of connect is to "obtain a usable sandbox connection," and it may be frequently invoked by the SDK, retry logic, or recovery mechanisms.

That makes sense. I agree that `connect(timeout=N)` should only ensure at least that much remaining lifetime, rather than shortening an already longer deadline.

I'll update the running-sandbox path to only extend the deadline when needed, and leave `set_timeout` as the explicit API for shortening it.

zyl1121:
@luzhixing12345 Updated. `connect(timeout=N)` now preserves a longer existing deadline and only extends it when needed.

Also updated the timeout docs to match the new semantics and added regression coverage for the shorter-timeout case.

luzhixing12345:
bot comment https://github.com/TencentCloud/CubeSandbox/pull/1352#discussion_r4056971919 is correct, please modify here

```diff
- None => timeout > 0,
+ None => false,
```

By the way, please squash into one commit, it's about to merge.

zyl1121:
> bot comment [#1352 (comment)](https://github.com/TencentCloud/CubeSandbox/pull/1352#discussion_r4056971919) is correct, please modify here
> 
> ```diff
> - None => timeout > 0,
> + None => false,
> ```
> 
> By the way, please squash into one commit, it's about to merge.

Updated. I applied the requested `None` handling and also tightened the related edge cases: `Stopped` now remains read-only, the paused Connect test no longer counts resume latency against the timeout check, and the Resume timeout schema documentation is now aligned.

I've also squashed the PR into a single commit.

PR #1363, opened 2026-08-17
feat(clm): add active-standby HA with leader election and reconciler (#1211)

### LINKED ISSUE #1211: [Feature Request] Add Active-Standby support for CLM
## Problem / Motivation

Currently, we only allow CLM to be deployed on controll node. This can be fragile on production and may block the auto-pause/auto-resume workflow.

## Proposed Solution

Introduce Active-Standby support to improve availability. With this in place, we will also need reconcilation to achieve eventual consistency for sandbox status.

shsaihdsaiudh:
Fixes #1211

## What

Adds active-standby HA to `cube-lifecycle-manager` so a crashed replica no longer blocks the auto-pause/auto-resume workflow:

- **Leader election via Redis lease** (`cube:v1:shared:lock:lifecycle_manager:leader`, SETNX + token-matched renew/release Lua, default TTL 15s). Only the leader runs the stateful loops: stream consumer, stale-pending claim, last-active poller, idle sweeper, and the periodic reconciler.
- **Standby behavior**: keeps the HTTP server up and reports 503 on `/readyz` so the Service routes `/internal/resume` only to the leader; a standby can still serve resume via the meta-hash fallback on a registry miss.
- **Failover** (bounded by one leader TTL): the new leader bootstraps from the meta hash, replays the snapshot to every CubeProxy, claims the dead consumer's pending stream entries via `XAUTOCLAIM` (min-idle = reconcile interval), and the reconciler converges remaining drift between the meta hash, the in-memory registry, and proxy meta dicts.
- **Hot-loop guard**: after 3 consecutive fast-failing leader stints (each shorter than the leader TTL) the process exits so the pod supervisor restarts it; clean stints and stints surviving at least one TTL reset the counter.
- **Config safety**: in HA mode `CUBE_LCM_RECONCILE_INTERVAL` must be >= `CUBE_LCM_LEADER_TTL` (validated at startup) — the interval doubles as the `XAUTOCLAIM` min-idle, and a smaller value could steal pending stream entries from a merely partitioned — still alive — old leader.

HA is opt-in via `CUBE_LCM_HA_ENABLED` (default off keeps single-replica behavior); the Helm chart enables it by default with `replicas: 2`. One-click and terraform deployments remain single-replica.

## Upgradability note

HA deployments that set `reconcileInterval < leaderTTL` will now fail fast at startup — raise `reconcileInterval` before upgrading. Helm defaults (60s/15s) satisfy the constraint.

## Testing

- Unit tests: `cube-lifecycle-manager` `go test ./...` all pass (leaderelect, reconciler, resumer, httpapi, config, redisstream, and the new leader supervisor).
- HA failover exercised end-to-end against a live Redis + 2-replica deployment (leader kill / partition / restart scenarios).


chenhengqi:
@shsaihdsaiudh Could you please rebase against latest master?

shsaihdsaiudh:
Rebased onto latest master as requested, and pushed `acee4693` addressing the review findings:

- **reconciler: stale-meta refresh never resets `LastActiveMs`** — now mirrors the stream `OpUpdate` path with `ResetLastActive`, so a timeout extension that only survives in the meta hash can't be undone by the sweeper. Covered by `TestReconcileRefreshesStaleMeta`.
- **`ClaimPending` cursor + trimmed entries** — the cursor is now threaded across passes, so one call drains the whole failover backlog instead of 100 entries per reconcile interval; value-less (trimmed) entries are now acked explicitly, closing the Redis 6.2 PEL leak (no-op on ≥ 7).
- **claim loop stealing the leader's own in-flight entries** — `claimStalePending` now drains the predecessor's backlog after promotion and then parks for the rest of the stint. XAUTOCLAIM can't exclude the calling consumer, and a slow 100-event `consumeStream` batch can exceed minIdle; a live leader never accumulates stale entries of its own, so anything left behind by a wedged/demoted stint belongs to the *next* leader's drain.
- **`ReconcileInterval` validation** — `> 0` is now validated unconditionally (the reconciler/claim tickers run in single-replica mode too; a zero interval panicked `time.NewTicker` at startup). The `>= leaderTTL` rule stays HA-only.
- **`OnLost` racing the demoted stint** — the elector now waits (bounded, 30s) for the demoted stint's goroutines to drain before invoking `OnLost`, so the registry reset can't race an in-flight `handleEvent`, and a re-elected stint never overlaps the old one. Covered by `TestElectorDrainsStintBeforeOnLost`.
- **`CUBE_LCM_HA_ENABLED` truthy set** — switched to `strconv.ParseBool` (same as `CUBE_LCM_EVENTBUS_ENABLED`): mixed-case bools work, and look-truthy-but-unknown values (`yes`, `on`, ...) now fail loudly at startup instead of silently disabling HA.
- **standby resume fallback docs** — corrected in README and the EN/ZH lifecycle guides: during the failover window the Service has no ready endpoints and resume requests fail (503 + `Retry-After`, clients retry); the meta-hash fallback serves the freshly-promoted leader during bootstrap, plus direct pod-IP access.
- **chart `replicas`/`ha.enabled` coupling** — `validate.yaml` now fails rendering for `lifecycleManager.replicas > 1` with `ha.enabled: false`, and values.yaml carries a prominent warning.

Verified locally: `go build`, `go vet`, `gofmt`, full `go test ./...` plus `-race` on the touched packages, `helm lint` + `helm template` (default values, the new fail rule, and `replicas: 1` + HA off), and the chart guard script suite.

PR #1374, opened 2026-08-18
cubeops: reject an empty JWT signing secret

### LINKED ISSUE #1373: [Bug Report] CubeOps can boot with an empty JWT signing secret, allowing forged admin tokens
`BootstrapJWTSecret` can return an empty string with a `nil` error, so CubeOps starts with a
zero-length HMAC key and any client that guesses this can forge a valid `admin` access token.

## Environment

- Deployment mode: single-node / cluster (both; the default one-click path is affected)
- Relevant component: CubeOps

## Steps to Reproduce

The reachable trigger is a `jwt_secret` row that exists but holds an empty value, which the
one-click deployment can produce because it deliberately leaves `JWT_SECRET` unset
(`deploy/one-click/env.example:257-258`, `deploy/one-click/scripts/systemd/cubeops-start.sh:27`).

1. Deploy CubeOps without setting `JWT_SECRET`, so the secret is auto-generated into
   `t_system_setting`.
2. Blank the stored value, simulating a truncated restore or a partially-applied migration:
   `UPDATE t_system_setting SET setting_value = '' WHERE setting_key = 'jwt_secret';`
3. Restart CubeOps. It starts normally and logs
   `JWT secret loaded from database (t_system_setting)`.
4. Mint a token off-box with an empty HMAC key:

   ```go
   claims := AccessClaims{
       RegisteredClaims: jwt.RegisteredClaims{
           ExpiresAt: jwt.NewNumericDate(time.Now().Add(time.Hour)),
           Subject:   "admin",
           Audience:  jwt.ClaimStrings{"cubeops:access"},
       },
       Username: "admin", Role: "admin", Typ: "access",
   }
   tok, _ := jwt.NewWithClaims(jwt.SigningMethodHS256, claims).SignedString([]byte(""))
   ```

5. `curl -H "Authorization: Bearer $tok" http://<cubeops>/api/v1/auth/session`

The same end state is also reachable from a transient DB read error, because
`GetSystemSetting` discards the error whenever the scanned value is empty (see #5 — same root
cause, tracked separately).

## Expected Behavior

CubeOps refuses to start when it cannot resolve a non-empty JWT signing secret, and the JWT
manager refuses to sign or verify with a zero-length key.

## Actual Behavior

Startup succeeds and the forged token is accepted. Reproduced against the shipped claim shapes:

```
=== RUN   TestEmptySecretSignsAndVerifies
    jwt_test.go:39: server signed a token with an EMPTY secret: eyJhbGciOiJIUzI1NiIsInR5cCI6IkpXVCJ9.eyJ...
    jwt_test.go:58: FORGED admin token ACCEPTED: username="admin" role="admin" typ="access"
--- PASS
```

`golang-jwt/v5`'s `SigningMethodHMAC.Sign` accepts a `[]byte{}` key without complaint, so an
empty secret is valid HMAC — every `/api/v1` route behind `auth.Middleware` is reachable.

## Additional Context

Call chain:

- `CubeOps/internal/store/setting.go:19-28` — `GetSystemSetting` returns `("", nil)` for a row
  that is missing, a value that is empty, **and** any DB error.
- `CubeOps/internal/store/setting.go:33-41` — `GetOrCreateSystemSetting` reads back through it.
- `CubeOps/internal/store/db.go:139-148` — `BootstrapJWTSecret` returns `winner` unchecked.
- `CubeOps/cmd/cubeops/main.go:53-58` — only `err` is checked.
- `CubeOps/internal/auth/jwt.go:54-60` — `NewJWTManager` stores `[]byte("")` unvalidated.

`bootstrapMasterKey` (`db.go:81-121`) shares the first two links but fails safe, because
`crypto.InstallMasterKey("")` rejects a zero-length key (`crypto/aes_gcm.go:60-62`). The JWT path
has no equivalent guard.

Related: #5 (the store layer swallowing DB errors) is the upstream cause and is worth fixing on
its own; this issue is the missing guard that turns it into an auth bypass.

dwin-gharibi:
Closes #1373.

## Motivation

`BootstrapJWTSecret` could hand back an empty string with a `nil` error, after which
`NewJWTManager` stored a zero-length HMAC key. `golang-jwt/v5` treats `[]byte{}` as a perfectly
valid HMAC key, so anyone who guessed the secret was empty could mint a `sub=admin`,
`aud=cubeops:access`, `typ=access` token that passed `VerifyAccessToken` unchanged — full access to
every `/api/v1` route behind `auth.Middleware`.

The default one-click deployment is the affected configuration: it deliberately leaves `JWT_SECRET`
unset so the secret is auto-generated into `t_system_setting`, which is the code path that can
return `""`.

## What this changes

Two independent guards, so neither alone has to be perfect:

1. **`CubeOps/internal/store/db.go`** — `BootstrapJWTSecret` now fails when the resolved secret is
   empty instead of returning it as a success. `main.go` already exits on that error, so startup
   aborts with a clear message rather than coming up insecure.

2. **`CubeOps/internal/auth/jwt.go`** — `JWTManager` refuses a zero-length key in all four token
   operations (`GenerateAccessToken`, `GenerateRefreshToken`, `VerifyAccessToken`,
   `VerifyRefreshToken`) via a shared `checkSecret()` helper. This keeps the exported
   `NewJWTManager` signature unchanged while making forged-token acceptance impossible even if an
   empty secret reaches the manager some other way.

Deliberately **not** changed here:

- `GetSystemSetting` / `GetSetting` / `GetUserPassword` still swallow DB errors. That is the
  upstream cause and is a behaviour change with a wider blast radius, so it is a separate fix
  (see the sibling branch for report finding #5). This PR is safe to land on its own and does not
  depend on it.
- No comment changes; no unrelated cleanup.

## Testing

New: `CubeOps/internal/auth/jwt_empty_secret_test.go`

- `TestEmptySecretRejectedOnGenerate` — both generators refuse an empty key.
- `TestEmptySecretRejectedOnVerify` — both verifiers refuse a token when configured with an empty
  key, including a token that was legitimately signed with a different, non-empty key.
- `TestNonEmptySecretStillWorks` — regression guard that the normal sign/verify round trip is
  untouched.

```
$ cd CubeOps && go test ./...
ok  github.com/tencentcloud/CubeSandbox/CubeOps/internal/auth      3.025s
ok  github.com/tencentcloud/CubeSandbox/CubeOps/internal/store     57.508s
ok  github.com/tencentcloud/CubeSandbox/CubeOps/internal/service   8.063s
... 11 packages ok, 0 failures
```

CI gates checked locally:

- `gofmt -l ./internal ./cmd` — clean (`fmt-check`).
- `go build ./...` — clean.
- `go test ./...` — 11 ok, 0 failures (`unit-test-check` runs `make cubeops-test`, which is
  `go test ./... -v -count=1`).

## Risk / rollout

Low risk, but it **is** a fail-closed change: an operator whose `jwt_secret` row is currently empty
will see CubeOps refuse to start instead of silently accepting forged tokens. That is the intended
behaviour. The remedy is to clear the row so it is regenerated, or to set `JWT_SECRET` explicitly:

```sql
DELETE FROM t_system_setting WHERE setting_key = 'jwt_secret';
```

Existing valid sessions are unaffected — the secret itself is not rotated by this change.

### DIFF
--- a/CubeOps/e2e/doc.go
+++ b/CubeOps/e2e/doc.go
@@ -1,0 +1,12 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package e2e holds end-to-end tests that drive the real cubeops binary against
+// a real database over HTTP.
+//
+// Every test file in this package carries the e2e build tag, so the default
+// unit gate (go test ./...) compiles this file and reports "no test files"
+// instead of failing on a package with no buildable sources. Run the suite with:
+//
+//	go test -tags e2e ./e2e/...
+package e2e
--- a/CubeOps/e2e/empty_jwt_secret_test.go
+++ b/CubeOps/e2e/empty_jwt_secret_test.go
@@ -1,0 +1,131 @@
+//go:build e2e
+
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"crypto/hmac"
+	"crypto/sha256"
+	"encoding/base64"
+	"encoding/json"
+	"fmt"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+)
+
+// forgeAccessToken mints the token an attacker would build once they guess the
+// signing secret is empty: sub=admin, aud=cubeops:access, typ=access, signed
+// with a zero-length HMAC key. It deliberately does not go through JWTManager,
+// which now refuses to sign with an empty key.
+func forgeAccessToken(secret string) string {
+	seg := func(v any) string {
+		raw, _ := json.Marshal(v)
+		return base64.RawURLEncoding.EncodeToString(raw)
+	}
+	now := time.Now().Unix()
+	header := seg(map[string]any{"alg": "HS256", "typ": "JWT"})
+	claims := seg(map[string]any{
+		"sub": "admin", "username": "admin", "role": "admin", "scopes": []string{},
+		"typ": "access", "aud": []string{"cubeops:access"},
+		"iat": now, "exp": now + 900,
+	})
+	signingInput := header + "." + claims
+	mac := hmac.New(sha256.New, []byte(secret))
+	mac.Write([]byte(signingInput))
+	return signingInput + "." + base64.RawURLEncoding.EncodeToString(mac.Sum(nil))
+}
+
+func readJWTSecret(t *testing.T, e *env) string {
+	t.Helper()
+	db := e.db(t)
+	defer db.Close()
+	var v *string
+	err := db.QueryRow(
+		"SELECT setting_value FROM t_system_setting WHERE setting_key = 'jwt_secret'").Scan(&v)
+	if err != nil {
+		t.Fatalf("read jwt_secret: %v", err)
+	}
+	if v == nil {
+		return "<null>"
+	}
+	return *v
+}
+
+func setJWTSecret(t *testing.T, e *env, value any) {
+	t.Helper()
+	db := e.db(t)
+	defer db.Close()
+	if _, err := db.Exec(
+		"REPLACE INTO t_system_setting (setting_key, setting_value) VALUES ('jwt_secret', ?)",
+		value); err != nil {
+		t.Fatalf("seed jwt_secret: %v", err)
+	}
+}
+
+// TestEmptyJWTSecretEndToEnd is the wire-level guard for #1373. It runs the real
+// binary against a real database and presents a forged admin token to a real
+// authenticated route.
+func TestEmptyJWTSecretEndToEnd(t *testing.T) {
+	e := newEnv(t)
+	defer e.teardown()
+
+	// First boot runs migrations and seeds admin/admin.
+	first := e.start(t)
+	first.stop()
+
+	for _, stored := range []any{"", " ", "\t", nil} {
+		label := fmt.Sprintf("%#v", stored)
+		t.Run("degenerate stored secret "+label, func(t *testing.T) {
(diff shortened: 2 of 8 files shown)


liciazhu:
Thanks for the fix. Two issues in the store-layer repair path still need attention before this is ready to merge.

1. Repair misses whitespace-only stored values

The repair SQL only matches exact empty strings (setting_value = ''), but the code detects emptiness with TrimSpace(winner) == "". If the DB stores ' ', the UPDATE matches zero rows and the repair does nothing; the re-read still returns ' ', so the function errors out and the service fails to start. The auth layer can block whitespace keys, but the store layer can't repair them — the two guards disagree on what "empty" means. Aligning them (e.g. WHERE NULLIF(TRIM(setting_value), '') IS NULL) would fix this.

2. A whitespace-only JWT_SECRET env var leaves the service half-broken

if envSecret != "" passes JWT_SECRET=" " straight through, so the store repair path is skipped. The service boots and looks healthy, but login/refresh permanently fail (checkSecret() rejects them) with no startup warning. Guarding the env branch with TrimSpace and falling back to the generated path would avoid this confusing state.

Both come from the same root cause: the definition of "empty" isn't applied consistently across detection, repair, and the env branch. Aligning on TrimSpace everywhere (plus a store test for a whitespace-only value) should resolve both.

Please revise to address these two points before merging. Thanks!

dwin-gharibi:

liciazhu:
LGTM.
Re-reviewed: empty/whitespace secret rejection is now enforced at both JWT signing/verification and bootstrap layer. Ran full CubeOps test suite + race detector locally — all pass.

fslongjin:
This issue is not reachable on production paths (one-click install / Kubernetes), so I've decided to close this PR.

PR #1376, opened 2026-08-18
cubemaster: fix the data races in localcache

### LINKED ISSUE #1375: [Bug Report] Five data races in CubeMaster localcache, including a torn interface{} read that can corrupt memory
`pkg/base/localcache` mutates cache entries and counters without synchronization. The worst case is
`put()` overwriting a `*util.CacheValue` in place while `Get()` readers are dereferencing it, which
can tear the two-word `interface{}` and hand a caller a value whose type word and data word come from
different objects.

## Environment

- Relevant component: CubeMaster

## Steps to Reproduce

```
cd CubeMaster && go test -race -count=1 ./...
```

9 data races and 9 failing tests, five of the races in production code.

## Expected Behavior

Race-clean under the detector.

## Actual Behavior

```
race 2: Write at localcache.go:201  | Previous write   (consecutiveFailNum)
race 4: Write at localcache.go:166  | Previous read at localcache.go:209   (CacheValue.Value)
race 5: Read  at localcache.go:122  | Previous write at localcache.go:167  (CacheValue.Expired/LastAccess)
race 7: Write at localcache.go:111  | Previous read at localcache.go:299   (chCacheExit)
```

## Additional Context

Five distinct defects:

**a. `CacheValue` mutated in place (worst).** `put()` at `:164-169` overwrites `itm.Value`,
`itm.Expired` and `itm.LastAccess` on an entry that `Get()` is concurrently reading at `:122`,
`:139-140` and `:148`. The `Lock()`/`Unlock()` pairs at `:145-147` and `:161-163` protect only the
`valueList` LRU reordering, not the field access. `util.CacheValue` (`util/util.go:14-19`) has plain
fields.

Concurrency is by design: `asyncRefresh` (`:219-233`) spawns a goroutine into `loadAndRefresh` →
`put` while request goroutines call `Get`. `sharedCalls.Do` dedupes concurrent *loads* for one key but
does not exclude concurrent *readers*.

`Value` is an `interface{}` — two machine words. A torn read pairs one value's type word with
another's data word, and the caller's type assertion then dereferences a mismatched pointer.

**b. `consecutiveFailNum` mixes atomic and plain access.** `atomic.AddInt64` at `:197` versus a plain
store at `:201`. `errStrategy()` reads it atomically at `:305`. The counter drives the cache's
degradation strategy.

**c. `curCacheSize` read non-atomically** at `:183`, written with `atomic.AddInt64` at `:165`, `:169`,
`:176`.

**d. `Destroy()` nils a channel a live goroutine selects on** — `chCacheExit = nil` at `:111` versus
`case <-localCache.chCacheExit:` in `errStrategy()` at `:299`.

**e. `errStrategy()` mutates the shared config struct** — `localCacheConfig.ExpiredUse = true` at
`:308` while `Get()` reads `.ExpiredUse` at `:124`.

The race detector is not wired into any test target, which is why these survived — see the separate
issue for adding `-race` to CI.

dwin-gharibi:
Closes #1375.

## Motivation

`go test -race ./...` in CubeMaster reported 9 races, five of them in `pkg/base/localcache`
production code. The most serious is `put()` mutating a `*util.CacheValue` in place while `Get()`
readers dereference it: `Value` is a two-word `interface{}`, so a torn read can pair one value's type
word with another's data word and the caller's type assertion then dereferences a mismatched pointer.

## What this changes

All in `CubeMaster/pkg/base/localcache/localcache.go`:

**1. Cache entries are now immutable once published.** `put()` no longer overwrites the live
`CacheValue`; it builds a replacement and swaps `element.Value` under the existing mutex:

```go
localCache.Lock()
prev := element.Value.(*util.CacheValue)
next := &util.CacheValue{Key: prev.Key, Value: val, LastAccess: ..., Expired: expired}
element.Value = next
localCache.valueList.MoveToBack(element)
localCache.Unlock()
```

`Get()` correspondingly reads the `*CacheValue` pointer **and** does its `MoveToBack` inside one
critical section, then works from that snapshot. Because nothing mutates a published entry any more,
the subsequent field reads need no lock.

This moves `Get`'s `MoveToBack` from after the refresh decision to before it. The LRU effect is the
same — exactly one `MoveToBack` per hit, on access — and it removes the duplicated
lock/`MoveToBack` block that the `DemotionExpiredUse` path used to need.

**2. `consecutiveFailNum`** — the plain store at `:201` becomes `atomic.StoreInt64`, matching the
`atomic.AddInt64` at `:197` and the `atomic.LoadInt64` in `errStrategy`.

**3. `curCacheSize`** — the read in `put()` becomes `atomic.LoadInt64`, matching its atomic writes.

**4. `Destroy()`** — no longer assigns `chCacheExit = nil` (which raced with `errStrategy`'s select).
Idempotency now comes from a `sync.Once`, so repeated `Destroy()` calls still cannot double-close.

**5. `ExpiredUse`** — `errStrategy` no longer writes the shared `*LocalCacheConfig` that `Get` reads.
`LocalCache` gains an `expiredUse atomic.Bool`, seeded from config in `NewCache` and driven by
`errStrategy`; `Get` reads it atomically. The yaml-tagged `LocalCacheConfig.ExpiredUse` field is
unchanged, so config parsing and every existing test that sets it still work.

Also fixed, because it is the last thing standing between this package and a green `-race` run:
**`localcache_test.go:48`** reassigned the shared `ctx` in a loop while spawned goroutines read it. The
context is now derived per iteration and passed as a parameter, which also removes an unintended
2700-deep context chain.

No comment changes.

## Testing

New: `CubeMaster/pkg/base/localcache/localcache_race_test.go`

- `TestConcurrentGetAndRefreshOnSameKey` — 32 readers × 300 iterations on one hot key with a 1 ms
  expiry, so `Get` and the async-refresh `put` interleave continuously. It type-asserts every hit,
  which is what would catch a torn `interface{}`.
- `TestDestroyIsIdempotent` — repeated `Destroy()` does not panic.
- `TestConcurrentDestroyDoesNotRaceWithBackgroundLoops` — four concurrent `Destroy()` calls.

Red/green verified — with `localcache.go` reverted to master, the new tests alone report **4 data
races**. With the fix:

```
$ go test -race -count=1 ./pkg/base/localcache/...
ok  github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/localcache       10.739s
ok  github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/localcache/util   2.118s
```

Module-wide, `go test -race ./...`:

| | master | this branch |
|---|---|---|
| data races | 9 | **2** |
| failing tests incl. `TestLocalCache`, `TestNoValue`, `TestSaveFile` | 9 | 6 |

CI gates checked locally:

- `gofmt -l ./pkg/base/localcache` — clean (`fmt-check`).
- `GOOS=linux go build ./pkg/base/localcache/` — clean.

## Still red after this change (pre-existing, not touched here)

The 2 remaining races and the 6 remaining failures are all pre-existing on master and belong to other
branches:

- `pkg/base/queueworker/queue_test.go:65,67,75` — test-side counter race.
- `pkg/base/recov/runtime_test.go` vs `runtime.go:30` — test-side read of an atomically-written
  counter.
- `pkg/base/bufferqueue/bufferqueue_test.go:131,138` — same class (this one is not yet tracked; worth
  its own issue).
- `TestDestroySandboxMissingSandboxReturnsNotFound`, `TestHostChangeGoodBodyReturnsSuccessEnvelope`,
  `TestTemplateLocalityFilterSelect`, `Test_local_appendNodeByCluster` — fail on master too
  (`gomonkey` cannot patch binaries on darwin; these need the Linux builder).

## Risk / rollout

Moderate — this touches the hot read path of a shared cache. The behavioural contract is unchanged:
same LRU semantics, same expiry decisions, same return values. The notable internal change is that a
`Get` hit now returns a value snapshot taken at lock time rather than whatever the entry happened to
hold mid-refresh, which is the point.

One extra allocation per `put()` on an existing key (the replacement `CacheValue`, ~48 bytes). Given
that `put` already runs a loader call and a JSON-ish payload, this is not a meaningful cost.

### DIFF
--- a/CubeMaster/pkg/base/localcache/localcache.go
+++ b/CubeMaster/pkg/base/localcache/localcache.go
@@ -79,7 +79,7 @@ func NewCache(name string, loader LoaderFunc, localCacheConfig *LocalCacheConfig
 		localCache.loadFile(localCacheConfig.LoadFileName)
 	}
 	localCache.localCacheConfig = localCache.SetupConfig(localCacheConfig)
-	localCache.expiredUse.Store(localCache.localCacheConfig.ExpiredUse)
+	localCache.expiredUse.Store(localCache.localCacheConfig != nil && localCache.localCacheConfig.ExpiredUse)
 
 	localCache.chShrinkCache = make(chan bool, 1)
 	localCache.chCacheExit = make(chan bool)
@@ -126,7 +126,6 @@ func (localCache *LocalCache) Get(ctx context.Context, key string) (interface{},
 
 		localCache.Lock()
 		itm := element.Value.(*util.CacheValue)
-		localCache.valueList.MoveToBack(element)
 		localCache.Unlock()
 
 		if time.Now().Add(-itm.Expired).After(time.Unix(itm.LastAccess, 0)) {
@@ -135,6 +134,10 @@ func (localCache *LocalCache) Get(ctx context.Context, key string) (interface{},
 				r, f, err := localCache.loadAndRefresh(ctx, key)
 
 				if err != nil && localCache.localCacheConfig.DemotionExpiredUse {
+
+					localCache.Lock()
+					localCache.valueList.MoveToBack(element)
+					localCache.Unlock()
 					return itm.Value, true, nil
 				}
 
@@ -148,6 +151,9 @@ func (localCache *LocalCache) Get(ctx context.Context, key string) (interface{},
 			localCache.asyncRefresh(ctx, key)
 		}
 
+		localCache.Lock()
+		localCache.valueList.MoveToBack(element)
+		localCache.Unlock()
 		return itm.Value, true, nil
 	}
 
--- a/CubeMaster/pkg/base/localcache/localcache.go
+++ b/CubeMaster/pkg/base/localcache/localcache.go
@@ -246,10 +246,10 @@ func (localCache *LocalCache) shrinkCache() {
 		select {
 		case <-localCache.chShrinkCache:
 			curTime := time.Now()
-			curCacheSize := localCache.curCacheSize
+			curCacheSize := atomic.LoadInt64(&localCache.curCacheSize)
 			var shrinkNum, shrinkSize, size int64
 			for {
-				if localCache.curCacheSize > localCache.localCacheConfig.LowCacheSize {
+				if atomic.LoadInt64(&localCache.curCacheSize) > localCache.localCacheConfig.LowCacheSize {
 					localCache.Lock()
 					element := localCache.valueList.Front()
 					if element == nil {
@@ -330,10 +330,13 @@ func (localCache *LocalCache) saveFile(file string) {
 		var itm *util.CacheValue
 		switch value := item.Object.(type) {
 		case *list.Element:
+			localCache.Lock()
+			stored := value.Value
+			localCache.Unlock()
 			var ok bool
-			itm, ok = value.Value.(*util.CacheValue)
+			itm, ok = stored.(*util.CacheValue)
 			if !ok {
-				CubeLog.Errorf("Cache(%s) cannot persist key %s: list element contains %T", localCache.name, key, value.Value)
+				CubeLog.Errorf("Cache(%s) cannot persist key %s: list element contains %T", localCache.name, key, stored)
 				continue
 			}
 		case *util.CacheValue:
--- a/CubeMaster/pkg/base/localcache/localcache.go
+++ b/CubeMaster/pkg/base/localcache/localcache.go
@@ -105,11 +105,11 @@ func (localCache *LocalCache) Destroy() {
 	if localCache == nil {
 		return
 	}
+	if localCache.chCacheExit == nil {
+		return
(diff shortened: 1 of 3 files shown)


fslongjin (approved):
LGTM

Verified the races on origin/master with `go test -race`, and this branch is clean. The immutable `CacheValue` swap under the mutex is the right fix for the torn `interface{}` read; the atomic / `sync.Once` / `expiredUse` changes are correct.

fslongjin:
image
Hello, please re-organize your commits, add a line `Signed-off-by` to them. @dwin-gharibi

fslongjin (approved):
LGTM

Re-reviewed after the rebase onto current master (`35cba98c`). Previously reviewed localcache content is unchanged; the extra commits only move the `Destroy` nil-guard outside `Once` and make the shrink signal non-blocking, which is correct. `go test -race ./pkg/base/localcache/` is still green.

PR #1378, opened 2026-08-18
cubeops: key the login rate limiter on the real peer IP, and bound the failure map

### LINKED ISSUE #1377: [Bug Report] CubeOps login rate limiter is bypassable with a spoofed X-Forwarded-For, and its failure map is unbounded
The login rate limiter keys its buckets on the raw client-supplied `X-Forwarded-For` header, with no
trusted-proxy check and no validation. Sending a different value per request gives an attacker a fresh
bucket every time, so the 5-failures-per-minute limit never triggers. The same header also makes the
failure map grow without bound.

## Environment

- Relevant component: CubeOps

## Steps to Reproduce

```
for i in $(seq 1 50); do
  curl -s -o /dev/null -w '%{http_code}\n' \
    -X POST http://<cubeops>/api/v1/auth/login \
    -H "X-Forwarded-For: 10.0.0.$i" \
    -d '{"username":"admin","password":"guess-'"$i"'"}'
done
```

## Expected Behavior

After 5 failures from the same client, further attempts return `429`.

## Actual Behavior

All 50 return `401` — the limiter never fires. Reproduced against the real handler:

```
no XFF, same peer IP        : [401 401 401 401 401 429 429 429 429 429]   <- baseline works
rotating XFF, same peer IP  : 50 attempts -> 50 x 401, 0 x 429
whitespace-padded same IP   : 20 attempts -> 0 x 429
```

The third case matters on its own: `"  198.51.100.7"` and `"198.51.100.7"` land in different buckets, so
even a single address yields unlimited buckets by varying leading whitespace.

## Additional Context

`CubeOps/internal/auth/ratelimit.go:87-98`:

```go
func clientIP(c *gin.Context) string {
	if xff := c.GetHeader("X-Forwarded-For"); xff != "" {
		for i := 0; i < len(xff); i++ {
			if xff[i] == ',' { return xff[:i] }
		}
		return xff          // caller-controlled, unvalidated, untrimmed
	}
	return c.ClientIP()
}
```

This reads the header directly, bypassing gin's `SetTrustedProxies` / `RemoteIPHeaders` machinery
entirely. The module comment states the limiter "protects the weak default credentials (admin/admin) from
brute-force attacks" — it does not.

**Second defect, same file.** The comments at `:36-38` and `:61-62` claim pruning prevents "unbounded
memory growth from IPs that never return after their window expires". It does not: `recordFailure` and
`isBlocked` prune **only the key currently being touched**, and there is no background sweeper. N distinct
header values create N permanent map entries, each holding a `[]time.Time` — remote memory exhaustion on
an unauthenticated endpoint.

dwin-gharibi:
Closes #1377.

## Motivation

`clientIP` read `X-Forwarded-For` directly, so the limiter's bucket key was fully attacker-controlled:
rotating the header gave a fresh full-quota bucket per request and the 5-per-minute limit never fired.
Varying only leading whitespace on a single address had the same effect. Reading the header by hand also
bypassed gin's `SetTrustedProxies` machinery, so there was no notion of *which* proxies are allowed to
assert a client IP.

Separately, the failure map was unbounded: pruning only ever touched the key being accessed, and there was
no sweeper, so N distinct header values meant N permanent entries.

## What this changes

**1. `internal/auth/ratelimit.go` — `clientIP` delegates to `c.ClientIP()`.** gin resolves the client IP
against the configured trusted-proxy set, so a forwarded header is honoured only when it comes from a
proxy the operator has declared. Untrusted callers are keyed on their real TCP peer address.

**2. `internal/config/config.go` — new `TrustedProxies []string`** (`trusted_proxies` in YAML,
`CUBE_OPS_TRUSTED_PROXIES` as a comma-separated env var), defaulting to **empty**.

**3. `internal/server/server.go` — `r.SetTrustedProxies(s.cfg.TrustedProxies)`** during router
construction. An invalid entry is logged and the router falls back to trusting nothing, which is the safe
direction.

**4. `internal/auth/ratelimit.go` — bounded map.** When the map reaches `sweepThreshold` (4096) entries,
`recordFailure` sweeps every key and drops those whose window has expired. Combined with (1), the key
space is now the set of real peer IPs seen within the window, which traffic bounds naturally.

No comment changes.

## Deployment impact — please read before merging

**This changes behaviour for anyone running CubeOps behind a reverse proxy.** gin's default is to trust
all proxies; the new default trusts none. Concretely:

- **Direct exposure (no proxy):** improvement, no action needed. Spoofing no longer works.
- **Behind nginx/LB, `trusted_proxies` unset:** every request is attributed to the proxy's IP, so all
  clients share one bucket and 5 failures from anyone briefly blocks logins for everyone. Operators
  **must** set `CUBE_OPS_TRUSTED_PROXIES` (or `trusted_proxies`) to the proxy address/CIDR.

I chose the fail-closed default deliberately — the alternative (default to trusting all proxies) keeps the
bypass. The Helm chart and one-click templates should set `trusted_proxies` to the ingress/CubeProxy CIDR
as a follow-up; I did not touch deployment manifests here to keep this PR to one component.

## Testing

New: `CubeOps/internal/auth/ratelimit_xff_test.go`, driving the real `LoginRateLimit()` middleware and
`markLoginFailure` through a gin router.

- `TestRotatingForwardedForCannotBypassTheLimiter` — 40 attempts from one peer with 40 different header
  values; the limiter must still block.
- `TestWhitespacePaddedForwardedForCannotBypassTheLimiter` — the padding variant.
- `TestLimiterStillFiresWithoutForwardedFor` — baseline, so a broken test cannot pass vacuously.
- `TestDistinctPeerIPsGetIndependentBuckets` — one peer being blocked must not block another.
- `TestTrustedProxyForwardedForIsHonoured` — with the proxy in `SetTrustedProxies`, two clients behind it
  are still limited independently. This is the case that would break if the fix had simply ignored the
  header.
- `TestSweepBoundsTheFailureMap` — the map stays at or below `sweepThreshold`.

Red/green — with `clientIP` reverted to the header-reading version (keeping the rest so the package still
compiles):

```
--- FAIL: TestRotatingForwardedForCannotBypassTheLimiter
    ratelimit_xff_test.go:68: rotating X-Forwarded-For bypassed the limiter entirely
--- FAIL: TestWhitespacePaddedForwardedForCannotBypassTheLimiter
    ratelimit_xff_test.go:89: whitespace-padded X-Forwarded-For bypassed the limiter
```

with the fix:

```
$ go test ./internal/auth/
ok  github.com/tencentcloud/CubeSandbox/CubeOps/internal/auth  2.336s

$ go test ./...
11 packages ok, 0 failures
```

CI gates checked locally:

- `gofmt -l ./internal` — clean (`fmt-check`).
- `go build ./...` — clean.
- `go test ./...` — 11 ok (`unit-test-check` → `make cubeops-test`).

### DIFF
--- a/CubeOps/config.example.yaml
+++ b/CubeOps/config.example.yaml
@@ -11,6 +11,13 @@ log_file_num: 10              # number of rotated log files to retain
 log_file_size: 100            # max size in MB per log file before rotation
 jwt_secret: ""             # leave empty to auto-generate on first start
 
+# Reverse proxies allowed to assert the client IP via X-Forwarded-For.
+# Empty (the default) trusts none: the login rate limiter keys on the real TCP
+# peer, which is correct for direct exposure. Behind nginx/an LB you MUST list
+# the proxy here as an IP or CIDR, otherwise every client shares one bucket.
+# Env override: CUBE_OPS_TRUSTED_PROXIES (comma-separated).
+trusted_proxies: []
+
 # --- Database (MySQL) ---
 # IMPORTANT: Replace the placeholder values below with your real credentials
 # before deploying. The defaults shown here are EXAMPLES ONLY and will not
--- a/CubeOps/config.example.yaml
+++ b/CubeOps/config.example.yaml
@@ -17,7 +17,13 @@ jwt_secret: ""             # leave empty to auto-generate on first start
 # proxy here as an IP or CIDR, otherwise every client shares one bucket.
 # Keep this as narrow as possible -- any peer whose source address matches is
 # trusted to assert any client IP, so a broad range such as 10.0.0.0/8 hands
-# that power to every workload on the network. 0.0.0.0/0 and ::/0 are rejected.
+# that power to every workload on the network. 0.0.0.0/0 and ::/0 are rejected,
+# and ranges wider than /16 (v4) or /64 (v6) are accepted but warned about at
+# startup. An empty list is also warned about, since it means every client
+# behind a proxy shares one login bucket.
+# Note: gin expands a bare loopback entry such as 127.0.0.1 to the whole
+# 127.0.0.0/8, and in containerised deployments the proxy usually connects from
+# a bridge address rather than loopback, so 127.0.0.1 is often not what you want.
 # Env override: CUBE_OPS_TRUSTED_PROXIES (comma-separated).
 trusted_proxies: []
 
--- a/CubeOps/config.example.yaml
+++ b/CubeOps/config.example.yaml
@@ -13,8 +13,11 @@ jwt_secret: ""             # leave empty to auto-generate on first start
 
 # Reverse proxies allowed to assert the client IP via X-Forwarded-For.
 # Empty (the default) trusts none: the login rate limiter keys on the real TCP
-# peer, which is correct for direct exposure. Behind nginx/an LB you MUST list
-# the proxy here as an IP or CIDR, otherwise every client shares one bucket.
+# peer, which is correct for direct exposure. Behind nginx/an LB, list the
+# proxy here as an IP or CIDR, otherwise every client shares one bucket.
+# Keep this as narrow as possible -- any peer whose source address matches is
+# trusted to assert any client IP, so a broad range such as 10.0.0.0/8 hands
+# that power to every workload on the network. 0.0.0.0/0 and ::/0 are rejected.
 # Env override: CUBE_OPS_TRUSTED_PROXIES (comma-separated).
 trusted_proxies: []
 
--- a/CubeOps/e2e/doc.go
+++ b/CubeOps/e2e/doc.go
@@ -1,0 +1,12 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+// Package e2e holds end-to-end tests that drive the real cubeops binary against
+// a real database over HTTP.
+//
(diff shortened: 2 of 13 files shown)


liciazhu:
Thanks for the PR. 

Points to consider
1. Broad CIDRs still accepted. /0 variants are rejected (verified), but 10.0.0.0/8  and 0.0.0.0/1  pass validation. A misconfiguration here lets everything on that network spoof XFF and bypass the limiter — partially re-opening the hole. Docs warn against it, but enforcement (or at least a regression test) would help.
2. Deployment regression risk. With empty default, all clients behind nginx/LB share one bucket. After upgrade, 5 failures from anyone throttle logins for the whole proxy; forgetting to set trusted_proxies breaks or amplifies the limiter. Please make this prominent in release notes and consider a startup warning.
3. TestEvictionKeepsTheMostRecentAttackers never exercises the eviction path. 

Overall
I'd like the two configuration guardrails above tightened (or clearly documented) before merge, and the eviction test should be adjusted to actually reach the threshold. Thanks for the effort!

dwin-gharibi:

liciazhu:
LGTM. 
Re-reviewed: both prior findings fixed (eviction test now reaches the threshold; validateTrustedProxies trims entries). Ran full CubeOps suite + race detector locally — all pass.

lkml-likexu:
@dwin-gharibi Would you help rebase it ?

PR #1380, opened 2026-08-18
cubeapi: key the rate limiter on the validated identity, not an unvalidated header

### LINKED ISSUE #1379: [Bug Report] CubeAPI rate limiter is bypassable with an unvalidated X-API-Key, and lumps all Bearer clients into one bucket
The rate limiter buckets on the raw `X-API-Key` header. When a client authenticates with
`Authorization: Bearer`, that header is never validated by the auth middleware but is still used as the
limiter key — so rotating it gives a fresh full-quota bucket per request. Clients that send no `X-API-Key`
all share one `"anonymous"` bucket, and the keyed state store is never reclaimed.

## Environment

- Relevant component: CubeAPI

## Steps to Reproduce

Run CubeAPI with `CUBE_API_KEY=supersecret --rate-limit-per-sec 3`, then send 30 requests each way:

```
curl -H "Authorization: Bearer supersecret" http://127.0.0.1:3000/sandboxes
curl -H "Authorization: Bearer supersecret" -H "X-API-Key: rotating-$i" http://127.0.0.1:3000/sandboxes
```

## Expected Behavior

Both are limited after 3 requests/second.

## Actual Behavior

```
A) Bearer only (shared 'anonymous' bucket, limit=3/s)   : 30 requests -> 429=26  non429=4
B) Bearer + rotating unvalidated X-API-Key              : 30 requests -> 429=0   non429=30
```

## Additional Context

`CubeAPI/src/middleware/rate_limit.rs:16-35`:

```rust
let key = request.headers()
    .get("X-API-Key")
    .and_then(|v| v.to_str().ok())
    .unwrap_or("anonymous")      // there is no IP fallback, despite the doc comment
    .to_string();
```

`extract_credential` (`middleware/auth.rs:25-45`) **prefers Bearer**, so with a Bearer token the
`X-API-Key` header is never validated — yet it still decides the bucket.

Three consequences:

1. **Bypass** — a valid Bearer token plus a random `X-API-Key` per request means `governor` allocates a
   fresh full-quota bucket every time.
2. **Cross-tenant DoS** — every Bearer client with no `X-API-Key` shares `"anonymous"`. In callback mode,
   which is the multi-tenant mode, the limiter degenerates into one global bucket.
3. **Unbounded memory** — `RateLimiter::keyed` uses a DashMap and nothing calls `retain_recent`
   (`grep -rn 'retain_recent|shrink_to_fit' CubeAPI/src/` is empty), so the bypass in (1) is also a
   memory-exhaustion vector.

The doc comment claims a fallback to IP that does not exist, so a reviewer would believe per-IP limiting
is in place.

dwin-gharibi:
Closes #1379.

## Motivation

The limiter bucketed on the raw `X-API-Key` header. Because `extract_credential` prefers
`Authorization: Bearer`, that header is never validated when a Bearer token is present — but it still
chose the bucket. Rotating it therefore gave a fresh full-quota bucket per request and the limit never
applied. Clients sending no `X-API-Key` all shared a single `"anonymous"` bucket, so in callback
(multi-tenant) mode any one tenant could starve the rest. And the keyed state store was never reclaimed.

## What this changes

**1. `middleware/auth.rs` — the auth middleware publishes the identity it validated.** A new
`RateLimitIdentity(String)` is inserted into the request extensions once the credential has been accepted,
in both modes:

- simple-key mode: after the key comparison succeeds
- callback mode: after the callback returns 200

The identity is prefixed (`bearer:` / `apikey:`) so the two credential kinds cannot collide.

**2. `middleware/rate_limit.rs` — the limiter reads that extension** instead of a header, falling back to
a single `"unauthenticated"` bucket if it is somehow absent. In practice it never is: `rate_limit` is only
ever layered together with `unified_auth`, and `unified_auth` runs first.

**3. `middleware/auth.rs` — the API key comparison is now constant-time.** `provided != expected_key`
short-circuits at the first differing byte, which leaks the key one byte at a time to a patient attacker.
Replaced with a length-check plus an XOR-accumulate loop. CubeOps already does this
(`subtle.ConstantTimeCompare`), so the two services now agree.

**4. `state.rs` — a background task calls `rate_limiter.retain_recent()` every 60s**, so the DashMap no
longer grows without bound.

No comment changes.

## Testing

Re-ran the probe that originally demonstrated the bypass, against the real binary with
`CUBE_API_KEY=supersecret --rate-limit-per-sec 3`, 30 requests per case:

| case | master | this branch |
|---|---|---|
| A) Bearer only | 26 × 429 | 27 × 429 |
| B) Bearer + rotating unvalidated `X-API-Key` | **0 × 429** | **26 × 429** |
| C) valid `X-API-Key` only | — | 27 × 429 |

Case B is the fix. Case C confirms API-key clients are still limited, and case A that Bearer clients are
too — all three shapes now behave the same.

CubeAPI's own suite:

```
$ cargo test
test result: ok. 109 passed; 0 failed; 0 ignored
```

CI gates checked locally:

- `cargo fmt --check` — clean (`fmt-check`).
- `cargo build` — clean.
- `cargo clippy --all-targets` — no new warnings. The two `field_reassign_with_default` hits reported in
  `auth.rs` are the pre-existing ones in that file's test module; their line numbers moved because this
  change inserts code above them.

## Behaviour change worth noting

Bearer clients are now limited **per token** rather than sharing one global bucket. That is the intended
behaviour, but it does mean a deployment that was unknowingly relying on the shared-bucket accounting will
see different 429 patterns. Concretely: a single Bearer client that previously consumed the shared quota
now gets its own, so aggregate throughput across many Bearer clients goes *up*, while an individual abusive
client is now actually constrained.

## Not fixed here

`extract_credential` matches the `Bearer ` scheme case-sensitively, so a spec-compliant
`authorization: bearer <token>` is rejected (RFC 7235 §2.1 makes the scheme case-insensitive). I confirmed
it still reproduces on this branch (5/5 requests → 401) and left it alone: it is a separate defect with its
own issue, and mixing an interop fix into a rate-limiting fix would muddy both.

### DIFF
--- a/CubeAPI/Cargo.lock
+++ b/CubeAPI/Cargo.lock
@@ -536,6 +536,7 @@ dependencies = [
  "reqwest",
  "serde",
  "serde_json",
+ "sha2",
  "thiserror 1.0.69",
  "tokio",
  "tower 0.4.13",
--- a/CubeAPI/Cargo.toml
+++ b/CubeAPI/Cargo.toml
@@ -59,6 +59,7 @@ uuid = { version = "1", features = ["v4", "serde"] }
 # ── High-concurrency in-memory state ──────────────────────────────────────
 # Lock-free concurrent HashMap: O(1) reads without global lock
 dashmap = "5"
+sha2 = "0.10"
 # Atomic counters / flags
 # (std::sync::atomic is sufficient for simple cases)
 
--- a/CubeAPI/src/config/mod.rs
+++ b/CubeAPI/src/config/mod.rs
@@ -18,7 +18,7 @@ pub struct ServerConfig {
     #[serde(default = "default_worker_threads")]
     pub worker_threads: usize,
 
-    /// Rate limit: max requests per second per API key
+    /// Rate limit: max requests per second per validated identity
     #[serde(default = "default_rate_limit")]
     pub rate_limit_per_sec: u32,
 
@@ -111,6 +111,13 @@ fn default_log_prefix() -> String {
 }
 
 impl ServerConfig {
+    pub fn auth_is_configured(&self) -> bool {
+        self.auth_callback_url
+            .as_deref()
+            .is_some_and(|u| !u.is_empty())
+            || self.cube_api_key.as_deref().is_some_and(|k| !k.is_empty())
+    }
+
     pub fn from_env() -> anyhow::Result<Self> {
         let _ = dotenvy::dotenv();
         let cfg = config::Config::builder()
--- a/CubeAPI/src/main.rs
+++ b/CubeAPI/src/main.rs
@@ -98,7 +98,7 @@ struct Cli {
     #[arg(long, value_name = "PREFIX")]
     log_prefix: Option<String>,
 
-    /// Rate limit: max requests per second per API key (default: 100).
+    /// Rate limit: max requests per second per validated identity (default: 100).
     ///
     /// Overrides the RATE_LIMIT_PER_SEC environment variable.
     #[arg(long, value_name = "N")]
--- a/CubeAPI/src/middleware/auth.rs
+++ b/CubeAPI/src/middleware/auth.rs
@@ -134,8 +134,10 @@ pub async fn unified_auth(
                         "Invalid API key or token".to_string(),
                     ));
                 }
-                let identity = identity_of(&credential);
-                request.extensions_mut().insert(RateLimitIdentity(identity));
+                request.extensions_mut().insert(RateLimitIdentity(format!(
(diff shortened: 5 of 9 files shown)


liciazhu:
LGTM. I've verified the fix empirically with governor 0.6.3 — the rotating-header bypass is closed (keying on the validated identity), and the GC doesn't reset active buckets. The remaining points are non-blocking; I'd just suggest adding a GC regression test when you have a chance. Thanks!

dwin-gharibi:

PR #1382, opened 2026-08-18
cubeops: propagate database errors from the settings and user getters

### LINKED ISSUE #1381: [Bug Report] CubeOps store layer swallows every database error, so outages surface as "invalid credentials"
Three CubeOps store getters turn *any* database error into `("", nil)` whenever the scanned value is
empty, so a database outage is reported to the operator as `401 invalid credentials` instead of a
`500`.

## Environment

- Relevant component: CubeOps

## Steps to Reproduce

1. Bring up CubeOps against MySQL/Postgres and confirm you can log in.
2. Stop the database (or block the port) so reads fail:
   `docker stop <mysql-container>`
3. `curl -si -X POST http://<cubeops>/api/v1/auth/login -d '{"username":"admin","password":"admin"}'`

## Expected Behavior

`500` with the underlying database error, so the operator can see the real fault. The comment in
`AuthService.Login` states exactly this intent: *"Genuine infrastructure errors (DB down, etc.) are
still returned verbatim so the operator can diagnose them."*

## Actual Behavior

`401 {"error":"invalid credentials"}`. The operator chases a phantom credentials problem while the
real fault is the database.

Reproduced with verbatim copies of both functions:

```
DB unreachable  -> pwd="" err=<nil>  (the connection error is DISCARDED)
user not found  -> pwd="" err=<nil>
ctx canceled    -> pwd="" err=<nil>  (discarded)

Login: err==nil, so BOTH branches of `if err != nil` are skipped -> DEAD CODE
Login: -> ErrInvalidCredentials (401) even though the DATABASE IS DOWN
```

## Additional Context

The pattern appears three times:

- `CubeOps/internal/store/setting.go:24` — `GetSystemSetting`
- `CubeOps/internal/store/setting.go:59` — `GetSetting`
- `CubeOps/internal/store/setting.go:111` — `GetUserPassword`

```go
if errors.Is(err, sql.ErrNoRows) || val == "" {
    return "", nil          // row missing, value empty, AND any DB error
}
return val, err
```

`val` is the zero value whenever the query fails, so `val == ""` is true for a genuine error too.

Because of this, both branches of `if err != nil` in `AuthService.Login`
(`internal/service/auth.go:82-91`) are **unreachable** — control falls straight through to
`VerifyPassword("", password)`. `ChangePassword` (`:179-185`) has the same shape.

There is also a contract mismatch that let this survive: the service-layer `UserStore` fake in
`internal/service/auth_test.go:23-29` returns `("", errors.New("user not found"))` for an unknown
user, i.e. it models not-found as an *error*, while the real store returns `("", nil)`. The tests
therefore passed against a contract production never implemented.

`staticcheck` does not flag this (the assignment is used), and `go vet` does not either — it needs a
reader to notice the shadow-free-but-dead branch.

Related: #1 (empty JWT signing secret) is the same root cause reaching a different consumer —
`GetOrCreateSystemSetting` reads back through `GetSystemSetting`, so a transient read error yields an
empty JWT secret with no error.

dwin-gharibi:
Closes #1381.

## Motivation

`GetSystemSetting`, `GetSetting` and `GetUserPassword` collapsed three different outcomes — row
missing, value empty, and *any* database error — into `("", nil)`. Because the scanned value is the
zero value whenever the query fails, `val == ""` was true for a real error too, so the error was
discarded.

The visible symptom is that a database outage is reported as `401 invalid credentials` rather than a
`500`, which is the opposite of what `AuthService.Login`'s own comment promises. It also made both
branches of `if err != nil` in `Login` unreachable dead code.

## What this changes

**`CubeOps/internal/store/setting.go`** — all three getters now separate the two cases:

```go
if err != nil {
    if errors.Is(err, sql.ErrNoRows) {
        return "", nil
    }
    return "", err
}
return val, nil
```

The resulting contract is: **`("", nil)` means absent; a non-nil error means the read failed.** An
empty stored value is still reported as absent, which is what every caller already wanted.

**`CubeOps/internal/service/auth.go`** — `Login` and `ChangePassword` updated for that contract:

- `Login` returns `500` for a read failure and `ErrInvalidCredentials` when the user is absent
  (`stored == ""`) or the password does not match. User enumeration is still not possible: absent
  and wrong-password are indistinguishable to the caller.
- `ChangePassword` likewise returns the wrapped error for a read failure and
  `ErrInvalidOldPassword` when the user is absent or the old password is wrong.

**`CubeOps/internal/service/auth_test.go`** — the `fakeUserStore` returned
`("", errors.New("user not found"))` for an unknown user, modelling not-found as an *error*, while
the real store returns `("", nil)`. That mismatch is why the tests passed against a contract
production never implemented. The fake now returns `("", nil)`; its assertion (unknown user →
`ErrInvalidCredentials`) is unchanged and still passes.

Callers that already discard the error (`internal/service/openclaw.go:685-701`,
`internal/service/agenthub.go:457`) are untouched — they still get `""` and still ignore the error, so
their behaviour is unchanged.

`bootstrapMasterKey` (`internal/store/db.go:83,93`) and `BootstrapJWTSecret` (`:139`) already had
`if err != nil` guards that were previously dead; they now actually fire, which is the intended
fail-closed behaviour.

## Testing

New: `CubeOps/internal/service/auth_db_error_test.go`

- `TestLoginSurfacesInfrastructureError` — a DB error is wrapped and returned, and is **not**
  `ErrInvalidCredentials`.
- `TestLoginUnknownUserStillReportsInvalidCredentials` — no enumeration regression.
- `TestChangePasswordSurfacesInfrastructureError` — same for the password-change path.
- `TestChangePasswordUnknownUserReportsBadOldPassword` — absent user still reports a bad old
  password.

```
$ cd CubeOps && go test ./...
ok  .../internal/auth        ok  .../internal/config     ok  .../internal/crypto
ok  .../internal/cubemaster  ok  .../internal/handler    ok  .../internal/httputil
ok  .../internal/logging     ok  .../internal/redact     ok  .../internal/server
ok  .../internal/service     ok  .../internal/store
12 packages, 0 failures
```

CI gates checked locally:

- `gofmt -l ./internal ./cmd` — clean (`fmt-check`).
- `go build ./...` — clean.
- `go test ./...` — 0 failures (`unit-test-check` → `make cubeops-test`).

## Risk / rollout

**This is a behaviour change on the login path** and is the reason it is split out from #1: during a
database outage, `/api/v1/auth/login` now returns `500` instead of `401`. Any monitoring or client
retry logic keyed on `401` for that case will see `500` instead — which is the correct signal, but
worth calling out.

The `UserStore` contract change is internal to CubeOps (`internal/`), so there is no external API
impact.

### DIFF
--- a/CubeOps/e2e/db_outage_test.go
+++ b/CubeOps/e2e/db_outage_test.go
@@ -1,0 +1,91 @@
+//go:build e2e
+
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package e2e
+
+import (
+	"net/http"
+	"strings"
+	"testing"
+)
+
+// dbDetailMarkers are fragments of a driver error that must never reach a
+// client. The host/port come from the container the harness started.
+func dbDetailMarkers(e *env) []string {
+	return []string{
+		"dial tcp", "connection refused", "invalid connection",
+		"sql:", "mysql:", e.mysqlURL.port,
+	}
+}
+
+func assertNoDBDetail(t *testing.T, e *env, where, body string) {
+	t.Helper()
+	for _, marker := range dbDetailMarkers(e) {
+		if strings.Contains(strings.ToLower(body), strings.ToLower(marker)) {
+			t.Errorf("%s response leaks %q to the caller: %s", where, marker, body)
+		}
+	}
+}
+
+// TestDatabaseOutageEndToEnd is the wire-level guard for #1381. It logs in
+// against a healthy database, then kills the database underneath the running
+// server and asserts that:
+//
+//   - login reports 500 (infrastructure), not 401 (invalid credentials)
+//   - none of the three public/authenticated 500 paths echo driver detail
+func TestDatabaseOutageEndToEnd(t *testing.T) {
+	e := newEnv(t)
+	defer e.teardown()
+
+	inst := e.start(t)
+	defer inst.stop()
+
+	// Healthy baseline: a real login and a real refresh token to use later.
+	token := login(t, inst.baseURL, "admin", "admin")
+	code, body := do(t, http.MethodGet, inst.baseURL+"/api/v1/auth/session", token, "")
+	if code != http.StatusOK {
+		t.Fatalf("baseline session check failed: %d %s", code, body)
+	}
+
+	code, refreshBody := do(t, http.MethodPost, inst.baseURL+"/api/v1/auth/login", "",
+		`{"username":"admin","password":"admin"}`)
+	if code != http.StatusOK {
+		t.Fatalf("baseline login failed: %d %s", code, refreshBody)
+	}
+	refreshToken := extractRefreshToken(t, refreshBody)
+
+	// Pull the database out from under the running server.
+	e.teardown()
+
+	t.Run("login reports an outage rather than bad credentials", func(t *testing.T) {
+		code, body := do(t, http.MethodPost, inst.baseURL+"/api/v1/auth/login", "",
+			`{"username":"admin","password":"admin"}`)
+		if code == http.StatusUnauthorized {
+			t.Fatalf("a database outage was reported as invalid credentials: %d %s", code, body)
+		}
+		if code != http.StatusInternalServerError {
+			t.Fatalf("login during an outage returned %d, want 500: %s", code, body)
+		}
+		assertNoDBDetail(t, e, "login", body)
+	})
+
+	t.Run("change-password does not leak driver detail", func(t *testing.T) {
+		code, body := do(t, http.MethodPost, inst.baseURL+"/api/v1/auth/change-password", token,
+			`{"old_password":"admin","new_password":"another-password"}`)
+		if code != http.StatusInternalServerError {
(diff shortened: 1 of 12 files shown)


liciazhu:
Thanks for this PR.
One issue to address before merge:
The same  err.Error() leak pattern is still present in the  Refresh handler ( handler.go ~line 140):
httputil.WriteError(c, http.StatusInternalServerError, err.Error())
This is the same class of vulnerability the PR addresses for  Login and  ChangePassword, but the third entry point was missed.
Suggested fix:
 1. Replace  err.Error() with  "internal server error" on the  Refresh 500 path, consistent with the other two handlers.
 2. Add a  TestRefreshOutageDoesNotLeakDatabaseDetailsToTheCaller test case to cover it.
Everything else looks good — the service layer changes, store contract updates, and test coverage are solid. Happy to approve once the Refresh path is brought in line.

dwin-gharibi:

liciazhu:
LGTM. 
Re-reviewed: both prior findings fixed (Refresh + agenthub 500 paths now also scrubbed; store getters distinguish sql.ErrNoRows from real driver errors). Ran full CubeOps suite + race detector locally — all pass.

PR #1383, opened 2026-08-18
cubemaster: fix integration tests panicking on unopened dao

### LINKED ISSUE #1372: [Bug Report] CubeMaster integration test package panics: dao: Open has not been called yet
## Summary

  CubeMaster's integration test package panics at TestMain on current master: pkg/base/db.Init now
  requires the shared dao.Open handle to be established first, but the test still initializes mocks
  before the app starts.

  ## Environment

  - CubeSandbox version / commit: master @ 0daf59aa (2026-08-18)
  - Host OS and kernel version: N/A (test-only; reproduces on any host with Go >= 1.25)
  - KVM info (modinfo kvm): N/A
  - Deployment mode: N/A (does not affect deployments; only the integration test package)
  - Relevant component: CubeMaster (integration test package) / CubeDB (dao)

  ## Steps to Reproduce

  1. Check out master at 0daf59aa.
  2. Save the following config as repro.yaml (the log section is required by config validation):

  common:
    http_port: 18089
    http_bind: "127.0.0.1"
  log:
    path: "/tmp/cm-log"
    file_size: 100
    file_num: 10
    level: "info"
  instance_db_config:
    addr: "127.0.0.1:3306"
    user: "root"
    pwd: "root"
    db_name: "test"
  redis:
    nodes: "127.0.0.1:6379"

  3. Run (no MySQL/Redis needed to trigger):

  cd CubeMaster
  mkdir -p /tmp/cm-log
  CUBE_MASTER_CONFIG_PATH=$PWD/repro.yaml go test ./integration/ -run TestDemo -count=1

  ## Expected Behavior

  TestMain should open the shared database handle (as app.Run does via dao.Open) before any package
  touches it, so the integration tests can start (with a reachable MySQL/Redis) instead of panicking.

  ## Actual Behavior

  panic: dao: Open has not been called yet

  goroutine 1 [running]:
  github.com/tencentcloud/CubeSandbox/CubeDB/dao.Default()
        CubeDB/dao/dao.go:143 +0xc7
  github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/db.Init(...)
        CubeMaster/pkg/base/db/db.go:21
  github.com/tencentcloud/CubeSandbox/CubeMaster/integration.mock_db(...)
        CubeMaster/integration/mock_init.go:657
  github.com/tencentcloud/CubeSandbox/CubeMaster/integration.MockInit()
        CubeMaster/integration/mock_init.go:111 +0x2c
  github.com/tencentcloud/CubeSandbox/CubeMaster/integration.TestMain(...)
        CubeMaster/integration/main_test.go:33 +0x85
  main.main()
        _testmain.go:66 +0xa5
  FAIL  github.com/tencentcloud/CubeSandbox/CubeMaster/integration

  ## Additional Context

  - Root cause: PR #984 ("Decoupling of CubeAPI Operations Logic and Architecture Refactoring")
    changed pkg/base/db.Init into a thin wrapper that returns dao.Default() — the connection is
    established by dao.Open at app startup — but integration.TestMain still calls MockInit() before
    app.Run() (where dao.Open runs), so dao.Default() is reached before the handle exists.

  - The package is excluded from the regular unit-test CI job (make test greps out /integration), and
    the Makefile integration-test target references a test/conf.yaml that is not present in the repo,
    so CI does not catch this.

  - Suggested fix: in integration/main_test.go, call dao.Open(ctx, ...) (from config.GetDbConfig())
    before MockInit(); dao.Open is idempotent, so the later app.Run() open is a no-op. Optionally
    follow with dao.Migrate(ctx).

  - Reproducing requires a log section in the config; a minimal config without it fails config
    validation first (validate config fail: log config is nil).

aptacc2421:
Fixes #1372

## Summary

CubeMaster's integration test package panics at TestMain on current master: pkg/base/db.Init is now a thin wrapper over dao.Default(), which panics with "dao: Open has not been called yet" until the shared handle is established by dao.Open at app startup. Both the integration test bootstrap (TestMain) and mock-debug mode (cmd/cubemaster/main.go) call MockInit() before app.Run(), so mock_db() reached dao.Default() before any Open.

## What Changed

Open the shared dao handle in mock_db() from the loaded DB config first. dao.Open is idempotent for the same identity, so the later open in app.Run()'s initDatabaseSchema is a no-op and schema migration still runs there.

## Validation

- TestDemo passes against a reachable MySQL/Redis (PASS, ~120s).
- Without a database, the package now fails with a clean connection error instead of panicking.
- Single commit, single file (CubeMaster/integration/mock_init.go); gofmt clean.

Assisted-by: DeepSeek Harness:deepseek-v4-flash

### DIFF
--- a/CubeMaster/integration/mock_init.go
+++ b/CubeMaster/integration/mock_init.go
@@ -22,6 +22,7 @@ import (
 	"github.com/alicebob/miniredis/v2"
 	"github.com/gomodule/redigo/redis"
 	"github.com/google/uuid"
+	"github.com/tencentcloud/CubeSandbox/CubeDB/dao"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/config"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/constants"
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/base/db"
@@ -654,10 +655,26 @@ func metricNow() []byte {
 }
 
 func mock_db() {
-	mocktest_OssDb = db.Init(config.GetDbConfig())
+	// db.Init is a thin wrapper that returns dao.Default(), which panics
+	// with "dao: Open has not been called yet" until the shared handle is
+	// established by dao.Open. Open it here from the loaded DB config;
+	// dao.Open is idempotent for the same identity, so the later
+	// dao.Open in app.Run's startup path (initDatabaseSchema) is a no-op.
 	// Schema (including t_cube_host_type) is owned by the dao.Migrate
-	// path that the integration test bootstrap runs before tests.
+	// path that the app bootstrap runs before tests. The dao config must
+	// be built from the same snapshot as initDatabaseSchema: config
+	// hotswap between MockInit and app.Run could change the identity and
+	// make the second dao.Open fail with "dao: already opened with ...".
+	daoCfg, err := db.ConfigFromDBConfig(config.GetDbConfig())
+	if err != nil {
+		stdlog.Fatalf("integration: dao config fail: %v", err)
+	}
+	if _, err := dao.Open(mocktest_Ctx, daoCfg); err != nil {
+		stdlog.Fatalf("dao open fail:%v", err)
+	}
+	mocktest_OssDb = db.Init(config.GetDbConfig())
 }
+
 func mock_getstr() string {
 	return fmt.Sprintf("%d.%d.%d.%d", rand.Int31n(254), rand.Int31n(254), rand.Int31n(254), rand.Int31n(254))
 }
--- a/CubeMaster/integration/mock_init.go
+++ b/CubeMaster/integration/mock_init.go
@@ -655,16 +655,12 @@ func metricNow() []byte {
 }
 
 func mock_db() {
-	// db.Init is a thin wrapper that returns dao.Default(), which panics
-	// with "dao: Open has not been called yet" until the shared handle is
-	// established by dao.Open. Open it here from the loaded DB config;
-	// dao.Open is idempotent for the same identity, so the later
-	// dao.Open in app.Run's startup path (initDatabaseSchema) is a no-op.
-	// Schema (including t_cube_host_type) is owned by the dao.Migrate
-	// path that the app bootstrap runs before tests. The dao config must
-	// be built from the same snapshot as initDatabaseSchema: config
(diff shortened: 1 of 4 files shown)


aptacc2421:
Addressed the review feedback: the DBConfig → dao.Config mapping is now extracted into db.ConfigFromDBConfig in pkg/base/db and used by both app startup (initDatabaseSchema) and the integration/mock-debug bootstrap (mock_db), so dao.Open's identity-dedup cannot drift between the two call sites. Added a focused unit test for the mapping; TestDemo still passes against a reachable MySQL/Redis.

aptacc2421:
Addressed both new non-blocking notes:

1. ConfigFromDBConfig's nil error now reads "db config is nil: instance_db_config is not set" — keeps the neutral description while retaining the actionable YAML key for mock-debug mode (the app path is already guarded by coreInit).
2. Added comments in both mock_db and initDatabaseSchema noting the two dao.Open calls must be built from a single config snapshot, since a config hotswap between the two opens would trip dao.Open's identity check.

Build and the ConfigFromDBConfig unit test pass.

CubeMaster/integration/mock_init.go
```diff
@@ -654,10 +655,26 @@ func metricNow() []byte {
...
+	// with "dao: Open has not been called yet" until the shared handle is
+	// established by dao.Open. Open it here from the loaded DB config;
+	// dao.Open is idempotent for the same identity, so the later
+	// dao.Open in app.Run's startup path (initDatabaseSchema) is a no-op.
 	// Schema (including t_cube_host_type) is owned by the dao.Migrate
-	// path that the integration test bootstrap runs before tests.
+	// path that the app bootstrap runs before tests. The dao config must
+	// be built from the same snapshot as initDatabaseSchema: config
+	// hotswap between MockInit and app.Run could change the identity and
+	// make the second dao.Open fail with "dao: already opened with ...".
```
fslongjin: Thanks for the contribution! The overall behavioral change looks good to me. However, I think we can clean up the comments here a bit. We should focus on describing what the code does, rather than explaining the rationale behind this panic. Otherwise, it might confuse future readers and comes across as noise.

fslongjin (changes requested):

CubeMaster/integration/mock_init.go
aptacc2421: Thanks for review! Already changed.

fslongjin (approved):

PR #1384, opened 2026-08-18
feat(CubeOps): migrate node management to CubeOps

### LINKED ISSUE #1204: [Feature Request] Move node management & ops logic from CubeMaster to CubeOps for better HA
## Problem / Motivation

CubeMaster currently carries both the **sandbox control-plane responsibilities** (scheduling, lifecycle coordination, etc.) and a large amount of **ops / node management logic**. These two responsibilities are coupled together, which makes CubeMaster harder to understand and evolve, and also harder to deploy in a robust high-availability (HA) fashion: ops processes and node management state/behavior sit on the same critical path as the sandbox control plane.

CubeOps already exists as a dedicated ops / management backend (WebUI ops APIs, cluster-side management plane, etc.). The next step is to continue this split: **move ops- and node-management-related responsibilities from CubeMaster to CubeOps**, and redesign the relevant architecture so that CubeMaster can focus on control-plane work that scales cleanly under HA.

If the split is not done, the following issues remain:

1. **CubeMaster HA is constrained**: ops logic embedded in CubeMaster amplifies the failure blast radius and coordination complexity of multi-replica deployments.
2. **Blurred responsibility boundaries**: node management and day-to-day ops processes are still mixed with sandbox scheduling / lifecycle, making ownership and evolution unclear.
3. **Incomplete ops → control-plane interaction**: CubeOps needs clearer capability to **proactively notify CubeMaster** of ops-side events, rather than continuing to have CubeMaster carry these ops responsibilities itself.

---

## Goals / Requirements

1. **Migrate node management logic to CubeOps**: node management responsibilities currently in CubeMaster should ultimately be owned by CubeOps.
2. **CubeOps must be able to proactively notify CubeMaster**: CubeOps should be capable of actively telling CubeMaster about relevant ops changes / events so that the control plane can perceive them and act accordingly.
3. **The relevant architecture must be redesigned**: the boundaries, ownership, and interaction between CubeOps and CubeMaster around node management / ops processes should be redesigned, not simply relocated.
4. **Support better CubeMaster HA**: the direct goal of this work is to extract ops logic out of CubeMaster so that CubeMaster can be deployed and run in a high-availability manner more reliably.

### Expected Benefits

- Clearer boundary between the ops plane (CubeOps) and the control plane (CubeMaster).
- Reduced ops burden on CubeMaster, improving readiness for multi-replica HA deployment.
- More explicit and predictable CubeOps → CubeMaster notification path in day-to-day operations.

---

## Additional Context

Previously, WebUI / ops-facing API capabilities were already split off into CubeOps to make CubeAPI lighter. This request extends the same direction to **CubeMaster**: extract the remaining ops and node management logic into CubeOps, redesign the CubeOps ↔ CubeMaster relationship (including proactive ops-side notifications), and treat CubeMaster's high-availability deployment as an explicit goal.

The implementation path and detailed architecture will be discussed separately after the requirements are aligned.

liciazhu:
Migrate node management and operations from CubeMaster/DB to the CubeOps service, and expose them through a new CLI and the WebUI.
1. CubeOps node management
   - Cubelet reports registration/heartbeat/metrics to CubeOps instead of CubeMaster/DB; CubeMaster syncs the node view from CubeOps via an external node loader (polls every 5s, default changed 30s->5s).
2. cubeopscli
   - New CLI with node subcommands: list/ls, isolate, unisolate,delete/rm to query node status against CubeOps.
3. Cubelet / CubeMaster <-> CubeOps interaction
   - Cubelet now targets CubeOps; CubeMaster consumes CubeOps through the external node loader and the new cubemastercli node query.
4. Deployment
   - Wire cube-ops and cubeopscli into all-in-one / k8s / terraform deployments, including REDIS_URL for cube-ops.
5. WebUI node ops
   - Add node isolate / unisolate actions and operation history to the node detail page.
6. Update Docs

Resolves #1204


Cubelet/dynamicconf/conf.yaml
fslongjin, line 1: I think the configuration here is incorrect; there's no need to modify the dynamic conf.

CubeMaster/cmd/cubemastercli/commands/cubebox/testutil_test.go
```diff
@@ -0,0 +1,40 @@
+// Copyright (c) 2024 Tencent Inc.
```
fslongjin: please change it to 2026

CubeMaster/pkg/localcache/template_locality.go
```diff
@@ -21,16 +23,21 @@ func SyncNodeTemplates(nodeID string, templateIDs []string) {
 		previous = discoverNodeTemplateSet(nodeID)
 	}
 
+	log.G(context.Background()).Infof("SyncNodeTemplates nodeID=%s current=%v previous=%v previousCached=%v", nodeID, current, previous, ok)
```
fslongjin: please use a context with RequestTrace instead of `context.Background()`

Cubelet/pkg/masterclient/client.go
```diff
@@ -152,15 +152,15 @@ func parseEndpoints(raw string) []string {
...
 }
 
 func (c *Client) RegisterNode(ctx context.Context, req *RegisterNodeRequest) error {
-	return c.post(ctx, "/internal/meta/nodes/register", req)
+	return c.post(ctx, "/internal/node-agent/v1/nodes/register", req)
 }
 
 func (c *Client) UpdateNodeStatus(ctx context.Context, nodeID string, req *UpdateNodeStatusRequest) error {
-	return c.post(ctx, "/internal/meta/nodes/"+nodeID+"/status", req)
+	return c.post(ctx, "/internal/node-agent/v1/nodes/"+nodeID+"/status", req)
```
fslongjin: why not `/v1/node-agent/`?

CubeMaster/pkg/localcache/node_cache.go
```diff
@@ -327,6 +334,14 @@ func (l *local) updateNodeFromMetaData(n *node.Node) error {
...
 		old.InvalidateLabelsCache()
 		old.SetSchedulingDisabled(n.SchedulingDisabled())
+		// On guest-image / cube-agent version change, trigger templatecenter's
+		// incremental compat scan.
+		if onNodeVersionsChanged != nil {
+			if compatVersionsChanged(old.Versions, n.Versions) {
+				go onNodeVersionsChanged(old.ID())
+			}
+		}
+		old.Versions = n.Versions
```
fslongjin: This looks like a port of master's `nodemeta.persistVersions` "version changed → compat scan" hook, but the meaning of that path changed after multi-version inventory.

Node upgrades no longer overwrite old inventory. Templates/sandboxes pin the version string from create time; restore/recreate resolve from inventory and fail if the version is missing — they do not fall back to the live toolbox. `ScanNodeCompat` already documents that it only updates the matrix; `READY+STALE` stays schedulable.

Comparing `old.Versions` vs `n.Versions` (the live current values) therefore does not answer "can this template still run on this node?". The matrix API also re-runs `evaluateCompat` on read, so the incremental scan is not required for correctness.

Please keep `old.Versions = n.Versions` — CubeOps → localcache still needs that data. I'd drop the change callback above, so we don't refresh `compat_status` from an outdated model on every ~5s full sync.

There is also a concrete regression vs master: empty versions were skipped there, and `inventory_incomplete` was merged with the previous snapshot so a missing field was not treated as a version change. Here, a CubeOps payload that omits `Versions` becomes `"" != "v1"`, false-triggers a scan, and can persist `UNKNOWN` onto replicas.

CubeMaster/pkg/localcache/node_cache.go
```diff
@@ -363,3 +378,28 @@ func (l *local) updateNodeMetric(n *node.Node) error {
...
+func compatRelevantVersions(versions []node.ComponentVersion) map[string]string {
+	out := map[string]string{"guest-image": "", "cube-agent": ""}
+	for _, v := range versions {
+		switch v.Component {
+		case "guest-image", "cube-agent":
+			out[v.Component] = v.Version
+		}
+	}
+	return out
+}
```
fslongjin: These helpers keep one version per component name (last write wins) and detect whether the live toolbox `guest-image` / `cube-agent` current value moved.

Cubelet's collector also does not scan `component_versions/` — heartbeats report the live current set only. After multi-version, a node can hold v1 in inventory (still used by old templates) and v2 in the live directory at the same time. This code treats a live v1→v2 bump as "incompatible", which is the opposite of the inventory model.

If we drop the callback above, please delete these helpers too so they are not later reused as the source of truth for compatibility.

CubeMaster/pkg/nodemeta/service.go
```diff
@@ -6,767 +6,73 @@ package nodemeta
...
-				}),
-			}).Create(&rows).Error; err != nil {
-				return err
-			}
+	// Wire version-change detection: localcache fires this when a node's
+	// guest-image / cube-agent version changes between sync cycles, and
+	// templatecenter's compat scan hook picks it up for an incremental rescan.
+	localcache.RegisterNodeVersionsChangedCallback(func(nodeID string) {
+		if OnGuestAgentVersionChanged != nil {
+			OnGuestAgentVersionChanged(nodeID)
```
fslongjin: This is the wiring for the incremental scan above. If `node_cache.go` stops detecting live version changes, please also remove `RegisterNodeVersionsChangedCallback` and the localcache callback registration, so we don't leave a path that only rewrites DB state when the live current value moves.

`scheduleInitialCompatScan` and the on-read matrix evaluation can stay as a product hint ("live toolbox has moved; redo if you want new guest metrics"). That is visibility, not a correctness gate.

CubeMaster/pkg/base/node/node.go
```diff
@@ -102,7 +112,13 @@ type Node struct {
...
 	NicQueues      int64 `json:"nic_queues,omitempty"`
 
-	NodeLabels map[string]string `json:"NodeLabels,omitempty"`
+	NodeLabels     map[string]string `json:"NodeLabels,omitempty"`
+	LocalTemplates []string          `json:"LocalTemplates,omitempty"`
+
+	// Versions carries the real version of each component installed on the
+	// node. Populated by CubeOps /internal/v1/nodes; consumed by templatecenter
+	// compat scan via localcache.GetNode.
+	Versions []ComponentVersion `json:"Versions,omitempty"`
```
fslongjin, line 121: Please double-check the JSON tag. Cubelet reports `json:"versions"` (lowercase); this field is `"Versions"`. If CubeOps `/internal/v1/nodes` emits `node.Node` PascalCase, this is fine. If it forwards the cubelet heartbeat as-is, this will always unmarshal as an empty slice, and both the incremental detector and the matrix "current" versions will be dead.

Separate from "should we keep the compat check", but the check cannot work until this data path is aligned.

fslongjin (changes requested):

deploy/one-click/build-release-bundle.sh
```diff
@@ -350,6 +351,14 @@ components["cubeops"] = {
...
 }
 
+# ── cubeopscli from CORE_BIN_DIR ──
+components["cubeopscli"] = {
+    "version": cube_version,
+    "commit": cube_commit,
+    "build_time": cube_build_time,
+    "digest_sha256": required_sha256(os.path.join(core_bin_dir, "cubeopscli")),
+}
+
```
fslongjin: CubeMastercli, Cube Ops, and cubeopscli don't need to report their versions — they are not compute node components.

CubeOps/internal/cubemaster/client.go
```diff
@@ -456,3 +432,63 @@ func setTraceHeaders(req *http.Request, rid string) {
...
+	if err := json.Unmarshal(raw, &env); err != nil {
+		return 0, fmt.Errorf("count node sandboxes: unmarshal envelope: %w", err)
+	}
+	if env.Ret.RetCode != 0 && env.Ret.RetCode != 200 {
+		return 0, &CMError{RetCode: env.Ret.RetCode, RetMsg: env.Ret.RetMsg}
+	}
+	// Array shape; JSON null decodes to a nil slice (count 0) with err == nil.
+	var items []json.RawMessage
+	if err := json.Unmarshal(env.Data, &items); err == nil {
+		return len(items), nil
```
fslongjin: **[High] Inventory check cannot tell "empty node" from "cubelet down" — ordinary delete of an empty node breaks, and `--force` becomes the default**

`CountNodeSandboxes` now goes through CubeMaster `/cube/sandbox/list?host_id=...` instead of asking Cubelet directly (what `CubeMaster/pkg/service/sandbox/node_deletion.go` did on master).

Two problems stacked:

1. CubeMaster `ListSandbox` **swallows** `cubelet.List` errors (`sandbox_list.go`) and still returns `ret_code: 0`. `ListCubeSandboxRes.Data` is `json:"data,omitempty"` — both a truly empty node and an unreachable cubelet **omit** the `data` field. `encoding/json` and jsoniter both omit a nil slice (verified).
2. This function still treats JSON `null` as count **0** (`json.Unmarshal` into a slice succeeds). The new `"null data"` case in `client_test.go` locks that in. The comment above claims `data=null` is fail-closed; the array branch does the opposite.

Net effect vs master:

| Situation | Expected | Actual |
|---|---|---|
| Isolated, empty, cubelet healthy | delete allowed | ordinary delete **fails** (missing `data` → unexpected shape) |
| Isolated, still has sandboxes, cubelet down | refuse | ordinary delete also fails (accidentally OK) |
| Operators then always `--force` | inventory still checked | `--force` **skips** the check — a down cubelet can lose a node that still hosts sandboxes |

`--force` still requires isolation, but on master an empty node deleted cleanly without it. Please either talk to Cubelet directly again, or make CubeMaster return a hard error when `cubelet.List` fails — and treat omitted/`null` data as `ErrSandboxCheckFailed`, never 0. Flip the `"null data"` test to expect an error.

deploy/one-click/scripts/one-click/common.sh
```diff
@@ -209,6 +209,35 @@ resolve_control_plane_cubemaster_addr() {
   die "ONE_CLICK_CONTROL_PLANE_IP or ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR is required for compute role"
 }
 
+resolve_control_plane_cubeops_addr() {
```
fslongjin, line 212: **[High] Runtime now requires CubeOps, but install preflight / env still only require CubeMaster**

This new resolver is the address Cubelet actually registers against (`:3010`). Compute-role **install** still does not match it:

- `check_compute_control_plane_preflight` (`deploy/one-click/lib/common.sh`) still only accepts `ONE_CLICK_CONTROL_PLANE_IP` / `ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR`.
- `env.example` still documents CubeMaster as the meta endpoint and never mentions `ONE_CLICK_CONTROL_PLANE_CUBEOPS_ADDR` / `CUBEOPS_ADDR` (the file in this PR only adds a blank line).
- `install.sh` persists `ONE_CLICK_CONTROL_PLANE_CUBEMASTER_ADDR` into the runtime env, but **not** the CubeOps address.

Terraform happens to write `ONE_CLICK_CONTROL_PLANE_CUBEOPS_ADDR`; a manual one-click / custom CubeOps port will pass preflight and then fail to register. Please require (or derive) the CubeOps addr in preflight, `env.example`, and the `install.sh` upsert together.

CubeMaster/conf.yaml
```diff
@@ -3,7 +3,10 @@ common:
   http_readtimeout: 120
   http_writetimeout: 360
   http_idletimeout: 360
-  sync_meta_data_interval: 30s
+  # CubeOps node-management base URL. 
+  cube_ops_addr: "http://127.0.0.1:3010"
```
fslongjin, line 8: **[High] Shipped default silently forces CubeOps-loader mode; upgrade order is unspecified**

A non-empty `cube_ops_addr` makes `nodemeta.Init` register the CubeOps loader and **stop writing** the local DB node tables (`pkg/nodemeta/service.go`). The comment here / in `service.go` says empty falls back to the DB loader — with this default that fallback is unreachable.

The Helm chart gates this correctly (`{{- if or .Values.cubeOps.enabled .Values.externalControlPlane.enabled }}`). This bare `conf.yaml` and one-click (`install.sh` defaults to `http://127.0.0.1:3010`) do not.

Together with deleting `/internal/meta` and pointing Cubelet at `/internal/node-agent/v1`:

- CubeMaster before CubeOps → ~31s boot retry, then start failure or an empty schedule
- New Cubelet against old control plane, or old Cubelet against new Master → registration fails
- Leaving `cube_ops_addr` empty as a "fallback" yields an empty/stale node view because nothing writes those tables anymore

Please ship `cube_ops_addr: ""` (or a clearly marked example), and document the breaking upgrade order: **CubeOps (node-agent) → Cubelet → then enable `cube_ops_addr`**. Mixed Cubelet generations are not supported. There is also no node-table data-migration step.

deploy/kubernetes/chart/templates/ops.yaml
```diff
@@ -68,6 +68,15 @@ spec:
             # do not fail when Master already applied the same migrations.
             - name: CUBEMASTER_MIGRATION_SKIP_FINGERPRINT_CHECK
               value: "true"
+            - name: REDIS_HOST
```
fslongjin, line 71: **[Medium-High] CubeOps metric fan-out has no Sentinel wiring; Redis init failure is warn-only**

Cubelet heartbeats used to refresh CubeMaster in-memory `MetricUpdate` **and** Redis. Now CubeOps writes `cube:v1:master:node:metric:{id}` and CubeMaster only reads it. Prefilter drops nodes whose metric is older than `MetricUpdateTimeout` (default 1h).

This block only sets `REDIS_HOST` / `PORT` / `PASSWORD`. CubeProxy and LCM in the **same chart** already pass `REDIS_MASTER_NAME` / `SENTINEL_NODES`. On Sentinel, `cube.redisHost` is not the master — these writes fail.

Worse, `CubeOps/internal/server/server.go` treats `nodemetric.Init` failure as a **warning** and keeps serving. Heartbeats then never refresh Redis; after ~1h every node is excluded and creates look mysteriously stuck.

Please mirror the proxy Sentinel env here, and fail CubeOps startup (or readiness) if Redis cannot be reached. Subsequent CubeMaster metadata syncs also do **not** copy `MetricUpdate` from the CubeOps node view — Redis is the only refresh path.

CubeMaster/pkg/localcache/export.go
```diff
@@ -224,39 +241,6 @@ func NotifyEvent(e *Event) error {
 	}
 }
 
-// EvictNode synchronously removes every scheduler-local view owned by nodeID.
-// It is used after a node deletion commits and by metadata reloads on replicas
-// that observe the deleted registration later.
-func EvictNode(nodeID string) {
```
fslongjin, line 230: **[Medium] Removing `EvictNode` leaves stale template-locality after CubeOps deletes a node**

On master, node deletion called `EvictNode`, which did `SyncNodeTemplates(nodeID, nil)` and dropped `templateNodeCache`. Deletion now lives in CubeOps; CubeMaster only notices the id missing on the next loader sync and goes through `checkDirty` → `delNodeCache`, which does **not** clear template locality.

`db_cache.go` now calls `SyncNodeTemplates` for nodes **still** in the CubeOps list, but deleted ids never get an empty-list reconcile. The scheduler can keep treating a ghost node as having a warm template replica (most visible when template affinity is on).

Please restore this cleanup (or call `SyncNodeTemplates(id, nil)` + delete `templateNodeCache` from `delNodeCache` / the dirty path).

CubeMaster/pkg/localcache/db_cache.go
```diff
@@ -32,24 +32,26 @@ func (l *local) DB() *gorm.DB {
...
 
 	if externalNodeLoader != nil {
-		nodes, err := externalNodeLoader(context.Background())
+		nodes, err := externalNodeLoader(ctx)
 		if err != nil {
 			retCode = 500
 			return err
 		}
 
+		log.G(ctx).Infof("syncAllFromDB: externalNodeLoader returned %d nodes", len(nodes))
```
fslongjin, line 53: **[Medium] Successful empty CubeOps list never evicts — deleting the last node(s) leaves ghosts**

`checkDirty` still returns immediately when `allFromDb` is empty (pre-existing guard against a bad empty load). That used to be rare. Node delete is now a first-class CubeOps API: deleting the last node(s) yields a **successful** `[]` from `/internal/v1/nodes`. This log will print `nodes=0`, `checkDirty` no-ops, and CubeMaster keeps scheduling onto the removed nodes until restart.

Please distinguish loader **error** (keep cache) from loader **success with 0 nodes** (evict). Do not use "empty list" as the fuse now that emptying the cluster is an intentional operator action.

Same path: `SyncNodeTemplates` below only runs for ids still in the CubeOps list — combined with removing `EvictNode`, deleted nodes also leak template-locality (see comment on `export.go`).

CubeOps/internal/nodemanagement/service/service.go
```diff
@@ -0,0 +1,664 @@
...
+	Sets    map[string]map[string]struct{}
+}
+
+type NodeService struct {
+	store               store.NodeStore
+	declaredVersions    map[string]string
+	declaredVersionSets map[string]map[string]struct{}
+
+	mu    sync.RWMutex
+	nodes map[string]*model.NodeSnapshot
```
fslongjin: **[High] In-process node map makes CubeOps a single point of failure — live state should be in Redis so this control plane can actually HA**

This `nodes map[string]*model.NodeSnapshot` is the live source of truth for heartbeats, health, isolation, delete, and the `/internal/v1/nodes` view CubeMaster polls. MySQL is only the durable registration/status ledger; `ListNodes` / `GetNode` serve this in-memory map. The README already admits *"in-process map, multi-replica concurrent writes are not supported yet"*.

That is the opposite of why node management moved out of CubeMaster (#1204):

- CubeOps crash / rolling restart → CubeMaster sees an empty or stale schedule until cubelets re-heartbeat or DB reload finishes.
- `replicas: 2` splits heartbeats across processes; isolate/delete and the scheduler list diverge. Helm `cubeOps.replicas` has no upper bound; Terraform `cube_ops_replicas` only checks `>= 1`.
- Resource metrics already fan out to Redis (`nodemetric`); the node roster does not.

**Capping replicas at 1 is a stopgap, not the design.** If CubeOps owns node lifecycle, the hot snapshot (heartbeat time, Healthy, cordon, labels, scheduler view) belongs in **Redis** (or equivalent shared store). Process memory can be a short-TTL cache. DB can stay the durable ledger; it should not be the hot path, and a single-process map must not be the HA story.

Until that shared state exists: force `replicas == 1` in Helm/Terraform and document CubeOps as a SPOF. Do not treat single-replica as the end state of this migration — that just moved the SPOF from CubeMaster to CubeOps.

docs/guide/kubernetes/install.md
```diff
@@ -212,7 +212,7 @@ kubectl get pods -n cube-system -o wide
 
 # 2) Have compute nodes registered with CubeMaster?
 kubectl exec -n cube-system deploy/cube-cubemastercli -- \
-  sh -lc 'cubemastercli --address "$CUBEMASTERCLI_ADDRESS" --port "$CUBEMASTERCLI_PORT" node list'
+  sh -lc 'cubeopscli --address "$CUBEOPSCLI_ADDRESS" --port "$CUBEOPSCLI_PORT" node list'
```
fslongjin, line 215: **[Docs] Command was switched to `cubeopscli`, but this comment (and several other pages) still say Cubelet registers with CubeMaster**

This helm-test snippet now talks to CubeOps, yet the line above still asks *"Have compute nodes registered with CubeMaster?"*. Operators following leftover copy will keep pointing Cubelet at `:8089` / `CUBEMASTER_ADDR`.

Same stale mapping is still in pages this PR already touched, and in pages it missed:

- `docs/guide/cubemaster-scheduler-config.md` / `docs/zh/guide/cubemaster-scheduler-config.md` §91 was updated to `/internal/node-agent/v1`, but the table at line 113 still says `ONE_CLICK_CONTROL_PLANE_IP` / `CUBEMASTER_ADDR` is *"the CubeMaster endpoint used by Cubelet for registration and reports"*. Line 235 still tells people to check CubeMaster reachability for compute nodes.
- `docs/guide/self-build-deploy.md` and the Chinese counterpart were **not changed at all** — they still document `CUBEMASTER_ADDR` as the compute-node registration address.
- `deploy/one-click/env.example` still labels CubeMaster as the compute-role **meta** endpoint and never mentions `ONE_CLICK_CONTROL_PLANE_CUBEOPS_ADDR` / `CUBEOPS_ADDR` (this PR only adds a blank line there).

Please sweep EN+ZH together: Cubelet registers with **CubeOps `:3010`**; CubeMaster `:8089` is HTTP/scheduling/artifacts only. That is the same wiring gap as the one-click preflight comment on `scripts/one-click/common.sh`.

CubeMaster/pkg/base/node/node.go
liciazhu: I have already checked this issue.End-to-end verification passed:
Cubelet masterclient/client.go:69 sends versions (lowercase);
CubeOps UpdateNodeStatusRequest.Versions (model/node.go:105) receives it as versions;
CubeOps /internal/v1/nodes converts it via ToSchedulerNode (internal.go:74), outputting using the JSON tag Versions (uppercase, model/node.go:266) on SchedulerNode.Versions;
CubeMaster node.Node.Versions (node.go:121) receives it as Versions (uppercase).
Both ends align correctly, and the data path is functioning as expected. I ran the temporary test TestVerifyVersionsTagAlignmentAcrossPipeline (not in workspace); both deserialization stages PASS.

CubeOps/internal/cubemaster/client.go
```diff
@@ -456,3 +432,70 @@ func setTraceHeaders(req *http.Request, rid string) {
...
+	if env.Ret.RetCode != 0 && env.Ret.RetCode != 200 {
+		return 0, &CMError{RetCode: env.Ret.RetCode, RetMsg: env.Ret.RetMsg}
+	}
+	// data:null means CubeMaster could not reach the cubelet; fail closed.
+	if bytes.Equal(bytes.TrimSpace(env.Data), []byte("null")) {
+		return 0, errors.New("count node sandboxes: data=null (cubelet may be unreachable)")
+	}
+	// data key omitted (CubeMaster omits an empty list via omitempty) = no sandboxes.
+	if len(env.Data) == 0 {
+		return 0, nil
```
fslongjin: The sandbox inventory check for node deletion goes through the wrong layer.

CubeOps asks CubeMaster `/cube/sandbox/list?host_id=...`, and CubeMaster in turn asks Cubelet. That indirection inherits two pre-existing problems this PR did not fix: on Cubelet list failure, `ListSandbox` swallows the error and still returns success; and an empty list vs. a failed list are indistinguishable in the JSON (both omit `data` due to `omitempty`). The net effect is that when a node is still heartbeating (so it looks Healthy) but the CubeMaster→Cubelet list RPC fails (timeout, firewall, lock), `CountNodeSandboxes` treats the omitted `data` as 0 and `DeleteNode` proceeds — deleting a node that may still host sandboxes.

This is not fixable by parsing CubeMaster's envelope more carefully. The design is wrong: ask the component that owns the data. The sandbox list lives on Cubelet, so CubeOps should query Cubelet directly (or via a CubeMaster passthrough that actually surfaces the error) and refuse deletion on failure — matching what master did before this migration.

fslongjin: The sandbox inventory check for node deletion goes through the wrong layer.

CubeOps asks CubeMaster `/cube/sandbox/list?host_id=...`, and CubeMaster in turn asks Cubelet. That indirection inherits two pre-existing problems this PR did not fix: on Cubelet list failure, `ListSandbox` swallows the error and still returns success; and an empty list vs. a failed list are indistinguishable in the JSON (both omit `data` due to `omitempty`). The net effect is that when a node is still heartbeating (so it looks Healthy) but the CubeMaster→Cubelet list RPC fails (timeout, firewall, lock), `CountNodeSandboxes` treats the omitted `data` as 0 and `DeleteNode` proceeds — deleting a node that may still host sandboxes.

This is not fixable by parsing CubeMaster's envelope more carefully. The design is wrong: ask the component that owns the data. The sandbox list lives on Cubelet, so CubeOps should query Cubelet directly (or via a CubeMaster passthrough that actually surfaces the error) and refuse deletion on failure — matching what master did before this migration.

deploy/one-click/scripts/systemd/cubeops-start.sh
```diff
@@ -46,16 +46,10 @@ if [[ -n "${CUBEMASTER_MIGRATION_SKIP_FINGERPRINT_CHECK:-}" ]]; then
...
-  redis_pass="${CUBE_SANDBOX_REDIS_PASSWORD:-}"
-  redis_auth=""
-  if [[ -n "${redis_pass}" ]]; then
-    redis_auth=":${redis_pass}@"
-  fi
-  export REDIS_URL="redis://${redis_auth}${CUBE_SANDBOX_REDIS_HOST}:${CUBE_SANDBOX_REDIS_PORT:-6379}"
-fi
+# Redis (optional). Split HOST/PORT/PASSWORD avoids embedding the password
+# in REDIS_URL; CubeOps assembles the connection internally.
+export REDIS_HOST="${CUBE_SANDBOX_REDIS_HOST:-127.0.0.1}"
```
fslongjin: One-click deploy wires CubeOps to Redis differently from Helm, and the one-click path breaks when an external Redis/Sentinel is configured.

`cubeops-start.sh` only exports `REDIS_HOST/PORT/PASSWORD` from the bundled-local-Redis variables (`CUBE_SANDBOX_REDIS_*`, default `127.0.0.1:6379`). It never reads `CUBE_EXTERNAL_REDIS_HOST` and never exports `REDIS_MASTER_NAME` / `REDIS_SENTINEL_*`. Meanwhile `install.sh` masks the local Redis service when an external Redis or Sentinel is configured. With this PR's fail-fast `nodemetric.Init`, the result is: one-click + external Redis → local Redis is masked → CubeOps still dials `127.0.0.1` → ping fails → CubeOps won't start. Helm + Sentinel is already wired correctly (`ops.yaml` exports the Sentinel env vars), so the one-click launcher just needs the same derivation CubeProxy/LCM already use.

Secondary: the `/health` probe always returns ok, so a runtime Redis outage stays invisible to k8s; and `test-redis-sentinel-guard.sh` asserts Sentinel wiring for CubeProxy/CLM but not CubeOps.

CubeOps/internal/nodemanagement/service/service.go
```diff
@@ -0,0 +1,665 @@
...
```
fslongjin: Banning multi-replica via `replicas > 1` fail is the wrong fix for the HA gap. The node live state (heartbeat, health, isolation, scheduling view) should be stored in Redis so CubeOps itself can run multiple replicas — not pinned to one pod with an in-process map.

This matters more than a generic HA nicety because CubeMaster now depends on CubeOps at boot: `nodemeta.Init` fails fast if CubeOps is unreachable, so a CubeOps outage or rolling-restart gap propagates directly to CubeMaster (no scheduling candidates). Moving the single point of failure from CubeMaster to CubeOps does not satisfy #1204.

Redis is already in the stack for node metrics; the node snapshot should live alongside it (memory as a short-TTL cache). The single-replica guard can stay as a temporary safety belt, but it should not be presented as the completed HA story.

CubeOps/internal/cubemaster/client_test.go
```diff
@@ -462,3 +463,99 @@ func TestClient_FullRetCodeRoundTrip(t *testing.T) {
...
+		body    string
+		want    int
+		wantErr string // non-empty substring expected when an error is wanted
+		isCM    bool   // error must be *CMError
+	}{
+		// Array shape.
+		{"array empty", `{"ret":{"ret_code":0},"data":[]}`, 0, "", false},
+		{"array one", `{"ret":{"ret_code":0},"data":[{"id":"sb-1"}]}`, 1, "", false},
+		// data key omitted (CubeMaster omits an empty list via omitempty) = no sandboxes.
+		{"data key omitted", `{"ret":{"ret_code":200},"size":1,"total":2}`, 0, "", false},
```
fslongjin: The tests lock in the wrong contract for the sandbox inventory check, which will block the design fix proposed above.

`client_test.go` case `"data key omitted"` asserts that a missing `data` field counts as 0 sandboxes. But that is exactly the shape CubeMaster returns when the Cubelet list RPC fails (error swallowed, `data` omitted via `omitempty`) — indistinguishable from a genuine empty list. If you change `CountNodeSandboxes` to fail-closed on a missing `data` (the design direction above), this test will break; right now it is codifying the unsafe behavior. The test suite has no case constructed from CubeMaster's real `ListCubeSandboxRes` encoder covering "Cubelet list error → success envelope + omitted data".

`TestDeleteNode_EmptyNode_Succeeds` runs with `checker == nil`, so it exercises "no checker wired" rather than "empty node + real list". The production path always wires `cubemaster.Client` in `server.go`, and the path that matters — healthy heartbeat + `Count` returns 0 via omitted data — has no test.

Separate, smaller: `make cubeops` and CI `expected_bins` only build `cubeops`, not `cubeopscli`. The release bundle relies on the one-click builder to produce it separately, so the CI gate never catches a broken `cubeopscli` build.

docs/guide/cubemaster-scheduler-config.md
```diff
@@ -88,7 +88,7 @@ scheduler:
 
 ## How node metadata affects scheduling
 
-Cubelet registers nodes and continuously reports status through CubeMaster's `/internal/meta` API. CubeMaster stores this metadata and keeps local cache snapshots for scheduling.
+Cubelet registers nodes and continuously reports status through CubeOps's `/internal/v1/node-agent` API. CubeOps persists this metadata to MySQL/Redis; CubeMaster syncs the node view from CubeOps every few seconds and keeps local cache snapshots for scheduling.
```
fslongjin, line 91: Several pages still say Cubelet registers/reports to CubeMaster after this migration. Operators following the old copy will keep pointing Cubelet at `:8089` / `CUBEMASTER_ADDR` and wonder why the node never appears.

Residual spots I found:

- `docs/guide/cubemaster-scheduler-config.md:18` — "Each Cubelet reports them to CubeMaster"
- `docs/zh/guide/cubemaster-scheduler-config.md:112` — `CUBE_SANDBOX_NODE_IP` "注册到 CubeMaster"
- `docs/guide/kubernetes/upgrade.md:84` and `docs/zh/guide/kubernetes/upgrade.md:84` — "the node has re-registered with CubeMaster" / "节点已重新注册到 CubeMaster"
- `docs/guide/tencentcloud-terraform-deploy.md:313` — "Cubelet node status/resource reporting interval to CubeMaster"
- `deploy/one-click/env.example:39` — comment still says "CubeMaster meta endpoint" for the compute-role control-plane address
- `docs/zh/guide/kubernetes/install.md:211` — section header "计算节点是否已注册到 CubeMaster" (the kubectl target is correct, only the wording is stale)

These should all say CubeOps (`:3010`). It would help to do one pass with the registration target as the single source of truth so the pages stop drifting apart.

CubeMaster/pkg/localcache/node_cache.go
```diff
@@ -191,6 +194,11 @@ func (l *local) delNodeCache(n *node.Node) {
 		CubeLog.WithContext(context.Background()).Warnf("node is nil")
 		return
 	}
+	// Clean template locality: empty list deregisters all replicas.
+	SyncNodeTemplates(context.Background(), n.ID(), nil)
```
fslongjin: `SyncNodeTemplates` now accepts a `ctx` (good), but this call site still passes `context.Background()`, so the request trace is lost. Please thread the caller's context through `delNodeCache` so the template-locality cleanup shows up in traces.

CubeOps/internal/nodemanagement/nodemetric/redis_metric.go
```diff
@@ -0,0 +1,284 @@
...
+	if cfg == nil {
+		return nil
+	}
+	sentinel := cfg.RedisMasterName != ""
+	url := cfg.RedisURL
+	if !sentinel && url == "" && cfg.RedisHost != "" {
+		url = buildRedisURL(cfg.RedisHost, cfg.RedisPort, cfg.RedisPassword)
+	}
+	if !sentinel && url == "" {
+		return nil
```
fslongjin: This PR made `Init` fail-fast when Redis is configured but unreachable (good), but the "Redis not configured at all" branch still returns nil and lets CubeOps start. With no Redis, node metrics are never written, CubeMaster reads stale data, and after `MetricUpdateTimeout` (default 1h) every node is evicted from scheduling — a silent degradation that is hard to attribute.

Since CubeOps is now the authoritative node-management plane and CubeMaster depends on it at boot, an unconfigured Redis should not be a silently-healthy state. Either require Redis at startup (fail fast here too), or at least surface it in the `/health` probe so k8s marks the pod NotReady.

CubeOps/internal/cubemaster/client.go
```diff
@@ -456,3 +432,70 @@ func setTraceHeaders(req *http.Request, rid string) {
...
+	if err := json.Unmarshal(env.Data, &items); err == nil {
+		return len(items), nil
+	}
+	// { "list": [...] } / { "sandboxes": [...] } shape. A raw map lets an
+	// object without either key fall through to the fail-closed branch.
+	var obj map[string]json.RawMessage
+	if err := json.Unmarshal(env.Data, &obj); err == nil {
+		var total int
+		seen := false
+		for _, key := range []string{"list", "sandboxes"} {
```
fslongjin: This parser over-compatibilities shapes CubeMaster never returns, while letting the one shape it actually returns on failure pass as 0.

`ListCubeSandboxRes.Data` is `[]*SandboxBriefData` with `json:"data,omitempty"`. The real envelope is either a JSON array (`[...]`) or the field omitted entirely (empty list *or* Cubelet list error). It never emits `{"list":[...]}` or `{"sandboxes":[...]}`. So the `{"list","sandboxes"}` branches here are dead weight that only widens the attack surface for misinterpretation.

Meanwhile the omitted-`data` case (line 468-471) — the one shape CubeMaster *does* produce on a Cubelet list failure — is treated as 0 sandboxes and allowed to proceed. The fail-closed branch at the bottom only catches shapes that do not happen in practice.

Net: the parser is backwards. Tighten it to the actual contract (array = count, omitted = indeterminate → fail), and drop the speculative object shapes.

CubeMaster/pkg/localcache/node_cache.go
```diff
@@ -160,9 +167,9 @@ func (l *local) dealEvent(ctx context.Context) {
 	}
 }
 
-func (l *local) checkDirty(allFromDb map[string]struct{}) {
+func (l *local) checkDirty(ctx context.Context, allFromDb map[string]struct{}) {
 	if len(allFromDb) == 0 {
-		CubeLog.WithContext(context.Background()).Warnf("checkDirty allFromDb is empty")
+		CubeLog.WithContext(ctx).Warnf("checkDirty allFromDb is empty, skip cleanup")
```
fslongjin: The empty-list short-circuit was put back, and it again leaves ghost nodes in the scheduler cache after the last node is deleted.

CubeOps `ListNodes` returns a **successful empty list** when MySQL has zero hosts — it does not error. After a successful load, `syncAllFromDB` calls `checkDirty`. This early return then skips eviction, so cached nodes stay schedulable.

A CubeOps / loader **failure** already returns from `syncAllFromDB` and never reaches `checkDirty`. So this guard does not protect against a failed sync; it only blocks the legitimate “cluster is now empty” path.

Please drop the `len(allFromDb) == 0` skip. Failed load → error; successful empty load → evict.

docs/guide/kubernetes/install.md
```diff
@@ -210,9 +210,9 @@ helm upgrade --install cube ./deploy/kubernetes/chart \
 # 1) Are Pods Ready?
 kubectl get pods -n cube-system -o wide
 
-# 2) Have compute nodes registered with CubeMaster?
-kubectl exec -n cube-system deploy/cube-cubemastercli -- \
-  sh -lc 'cubemastercli --address "$CUBEMASTERCLI_ADDRESS" --port "$CUBEMASTERCLI_PORT" node list'
+# 2) Have compute nodes registered with CubeOps?
+kubectl exec -n cube-system deploy/cube-cubeopscli -- \
```
fslongjin: There is no `cube-cubeopscli` Deployment in the chart. `cubeopscli` ships in the `cubemastercli` Pod.

Following this snippet fails immediately with NotFound. The Chinese install page, FAQ, and `NOTES.txt` already use `deploy/cube-cubemastercli`.

Please change this to:

```
kubectl exec -n cube-system deploy/cube-cubemastercli -- \
  sh -lc 'cubeopscli --address "$CUBEOPSCLI_ADDRESS" --port "$CUBEOPSCLI_PORT" node list'
```

CubeOps/internal/server/server.go
```diff
@@ -115,6 +140,14 @@ func (s *Server) buildRouter() *gin.Engine {
 	storeH.Register(authed)
 	agenthubH.Register(authed)
 
+	// Internal routes — no auth. These endpoints must not be exposed through
+	// nginx or a public Bind address. Callers: CubeMaster, cubeopscli.
+	internalH.Register(r.Group("/internal/v1"))
```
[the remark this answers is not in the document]
liciazhu: This is a WebUI deployment issue and will require adjustments to the WebUI deployment configuration.
Currently, the WebUI runs inside a Docker container and accesses CubeOps via host.docker.internal:3010. The hostname host.docker.internal resolves to the Docker bridge interface docker0 (typically 172.17.0.1). For the container to reach the CubeOps service running on the host, CubeOps must bind to all network interfaces (0.0.0.0) rather than exclusively to 127.0.0.1

CubeMaster/pkg/localcache/export.go
```diff
@@ -179,19 +178,13 @@ func GetNode(id string) (*node.Node, bool) {
...
 }
 
-func metadataHealthTimeout() time.Duration {
-	return nodehealth.MetadataTimeout(config.GetConfig().Common.SyncMetaDataInterval)
-}
-
-func cloneNodeWithCurrentHealth(n *node.Node, now time.Time) *node.Node {
+// cloneNodeWithCurrentHealth returns a defensive copy of n; CubeMaster trusts
+// the CubeOps Healthy verdict without overriding it on sync staleness.
+func cloneNodeWithCurrentHealth(n *node.Node, _ time.Time) *node.Node {
```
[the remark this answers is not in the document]
liciazhu: Remove CubeMaster's fallback logic that forcibly marks nodes unhealthy based on MetaDataUpdateAt timeout. CubeMaster should fully trust CubeOps' health determination, keeping node lifecycle ownership strictly within CubeOps. If the scheduler targets a node reported as healthy but actually unhealthy, it will retry on another node.

docs/guide/kubernetes/upgrade.md
```diff
@@ -81,7 +81,7 @@ Expect:
 
 - Control-plane Pods finished rolling per Deployment strategy (`cube-master` uses Recreate; other CP Deployments use RollingUpdate)
 - Compute: matching DaemonSet Pods run the new images and are Ready
-- If you bumped Big Pod runtime: existing sandboxes on that node were interrupted; new sandboxes can be created; the node has re-registered with CubeMaster
+- If you bumped Big Pod runtime: existing sandboxes on that node were interrupted; new sandboxes can be created; the node has re-registered with CubeOps
```
fslongjin, line 84: The re-register target is updated, but the upgrade / mixed-fleet procedure is still missing.

After this migration an old Cubelet still reports to CubeMaster `:8089`, and a new Cubelet reports to CubeOps `:3010`. The docs never say:

- bring CubeOps up (and reachable) before Cubelets whose `meta_server_endpoint` points at it
- do not mix old and new Cubelets in one cluster during the cutover
- CubeMaster now fails to start if `cube_ops_addr` is set and CubeOps is down

A short “upgrade order” paragraph on this page (and the Chinese counterpart) would close it.

CubeOps/internal/nodemanagement/service/health.go
```diff
@@ -0,0 +1,43 @@
...
+		return HealthStatus{Healthy: false, UnhealthyReason: model.ReasonHeartbeatExpired}
+	}
+	if !reportedReady {
+		return HealthStatus{Healthy: false, UnhealthyReason: model.ReasonReportedNotReady}
+	}
+	return HealthStatus{Healthy: true}
+}
+
+// MetadataTimeout returns the health-check timeout (syncInterval + 10s slack).
+// Default 5s aligns with CubeMaster's SyncMetaDataInterval → 15s window.
```
fslongjin: The `syncInterval` argument is unused on the production path, and this comment is now wrong.

`applyCurrentHealth` always calls `MetadataTimeout(0)`, which hard-codes a 15s heartbeat window (`5s + 10s` slack). CubeMaster `sync_meta_data_interval` in this PR is **1s**, not 5s, so “Default 5s aligns with CubeMaster's SyncMetaDataInterval → 15s window” is stale.

Either thread the real interval through, or drop the unused parameter and document a fixed 15s health window that is not tied to CubeMaster's sync period.

fslongjin (approved):

PR #1391, opened 2026-08-19
feat(templatecenter): standalone template build service

wbzdssm:
Add CubeTemplateCenter (TC), a standalone process that takes over the data-plane half of template building -- pull the image, run the build in an envd sandbox, produce the rootfs ext4, compute its fingerprint, and report the result back to CubeMaster. CubeMaster keeps the control plane: it persists jobs, serves the public /cube/template* API, drives cross-node distribution, and serves artifact downloads to Cubelet.

Wiring is deliberately minimal and identical on every deploy path:

  - one boolean switch, templatecenter_enabled, in CubeMaster's conf.yaml. Unset/false builds in-process exactly as before; true hands builds to TC. No build/route mode strings.
  - two target-named address env vars, shared by whoever needs them: CUBE_TEMPLATE_CENTER_ADDR (CubeMaster -> TC) and CUBE_MASTER_ADDR (TC -> CubeMaster, with the retired spellings as fallback). Addresses are deployment facts, so they live in the environment, not in conf.

Core code:

  - CubeTemplateCenter/: the new module. Reuses CubeMaster/pkg/templatecenter for the build logic; adds tcconfig (env reading), build (execution + status reporter), reconcile, httpservice (template routes only), and the unauthenticated internal build endpoint POST /tc/api/v1/build.
  - CubeMaster/pkg/templatecenter/: remote build resume, artifact presence healing (a READY row whose ext4 file is gone is detected at reuse, distribution and download time, and the row demoted so the next create rebuilds it), force delete for templates stuck on in-flight work, and distribution readiness guards.
  - CubeDB/migrate/fingerprint.go: an applied migration version that is absent from this binary's tree is reported, not fatal -- a database ahead of the binary is the safe direction; real content drift stays fatal.

Deploy:

  - Helm: controlPlane.templateCenter.enabled=true wires both halves (conf switch + both address envs) and rejects the invalid combinations (multi-replica, un-colocated on a non-shared volume, etc.). TC is one replica, co-located with CubeMaster over a shared artifact PVC.
  - one-click: configs/single-node/templatecenter.yaml, the systemd unit and start/postcheck scripts, install.sh placeholder resolution, and the address envs.
  - images: cube-templatecenter in release-docker-images.yml and build-cube-images.sh.


luzhixing12345:
delay to the next version

luzhixing12345:
move to #1659

PR #1393, opened 2026-08-19
perf(clm): Reduce cross-replica resume wakeup latency

### LINKED ISSUE #1047: [Feature Request] zero-latency optimization for cross-sidecar wakeup via Redis Pub/Sub
## Problem / Motivation

In the current auto-resume flow, when concurrent cross-node resumes occur, late-arriving requests enter `waitForRunning()` and hard-poll Redis at a fixed **200ms** interval. 

Since the actual sandbox resume typically takes only **60ms**, this coarse-grained fixed polling is highly suboptimal. It introduces an average of 100ms (and up to 200ms) of extra detection latency, forcing user requests to wait pointlessly even after the sandbox has already resumed. Additionally, under high concurrency (the "thundering herd" problem), this triggers a massive Redis `GET` read amplification (read storm).

**Goals:**
1. Completely eliminate the fixed 200ms extra latency on the cross-node wait path.
2. Eliminate the Redis `GET` read storm under high concurrency.
3. Gracefully avoid the classic "missed message race window" associated with on-demand Pub/Sub subscriptions.

## Proposed Solution

We propose deprecating "on-demand polling" and upgrading to a **"Global Singleton Pub/Sub Subscription"** architecture with a **"1-Second Local Cache"**. 

When a sidecar updates a state to `running` or `paused`, it will simultaneously publish a wakeup event to a sharded Pub/Sub channel (`cube:v1:shared:events:<shard>`).

**1. Global Singleton Mechanism:** 
Upon startup, the sidecar uses just 1 connection to establish a permanent global subscription. Since the subscription is always online, it fundamentally eradicates the classic missed-message race condition where "the event is published right before the channel is established", and it also prevents concurrent waiters from exhausting the Redis connection pool.

**2. The 1-Second Local Cache (Edge & Race Scenarios Response):** 
Although the global subscription is always online, a microscopic scheduling time difference can still exist when the business goroutine executes `waitForRunning()` (e.g., the goroutine is scheduled 1 millisecond late and barely misses the live event). To completely close this race window, the global subscriber will cache received events in a local dictionary with a 1-second TTL. 

Guarded by this memory barrier, request flows automatically fall into one of three scenarios:

-  **Scenario 1: Normal Queueing — Instant Wakeup, Zero-Overhead Pass (~98%)**
  - The request misses the local cache, registers a listener with the local EventBus, and performs a single fallback `GET` (to close any extreme early mismatch). Confirming it's still `resuming`, it suspends safely. Tens of milliseconds later, the global subscriber receives the live Pub/Sub event and instantly wakes up the request. **Once awakened, it trusts the event payload and proceeds immediately without issuing another Redis `GET` confirmation.** (Zero-Latency, Zero-GET).

-  **Scenario 2: Edge Race Condition — Cache Hit, Pass Post-GET (~1%)**
  - The request arrives just milliseconds after the `running` event was broadcast, hitting the 1-second TTL cache in local memory. To prevent the cache from causing split-brain, the request immediately issues **1 `GET`** to Redis to verify the final true state. Once confirmed, it passes instantly. (Bypasses the 200ms hard-wait spike).

-  **Scenario 3: Abnormal Fallback — Polling for Truth (~1%)**
  - In extreme network anomalies (e.g., Pub/Sub disconnect drops messages), the request triggers a timeout protection and automatically degrades to **exponential backoff polling** (`20ms → 40ms → 80ms → 160ms → 300ms`). It actively `GET`s Redis on each jittered tick until it reaches a terminal state.

## Alternatives Considered

**1. Reusing Existing Redis Streams**
The system currently uses Streams for `create/delete` events. However, state transitions (`running/paused`) are ephemeral "fire-and-forget" wakeup signals. Pub/Sub does not require message persistence or consumer group cursor maintenance, making it much more lightweight than Streams and perfectly suited for low-latency wakeups.

**2. Scalability Concerns with Pub/Sub (100+ Subscribers)**
Industry benchmarks and underlying implementation principles provide a definitive answer regarding whether Pub/Sub will bottleneck Redis: **Zero pressure**.
- **Connections & Complexity**: The time complexity of `PUBLISH` is **O(N+M)**. Even if we scale to 300+ Sidecar nodes in the future, compared to the default Redis `maxclients` limit of 10,000, maintaining a few hundred idle long connections consumes negligible resources.
- **Throughput**: When subscribers number in the thousands and payloads are minimal (e.g., a short string of a dozen bytes), Pub/Sub distribution latency is stably **`< 1ms`**. Genuine bottlenecks only appear with tens of thousands of concurrent subscribers causing Output Buffer bloat—two orders of magnitude beyond our hundred-node scale.

## Additional Context

- **Extensibility**: The sharded Pub/Sub channels will be abstracted into a generic event bus. Future iterations can reuse this for Lifecycle Metadata Updates and Proxy Fleet configuration broadcasts.

chenhengqi:
Use Redis Pub/Sub as a best-effort wakeup hint while keeping Redis state as the source of truth and retaining the polling fallback.

Closes #1047

### DIFF
--- a/cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
+++ b/cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
@@ -20,6 +20,7 @@ import (
 	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/config"
 	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/cubemasterclient"
 	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/discovery"
+	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/eventbus"
 	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/httpapi"
 	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/lifecycle"
 	"github.com/tencentcloud/CubeSandbox/cube-lifecycle-manager/internal/proxypush"
@@ -67,6 +68,18 @@ func run() error {
 	stream := redisstream.New(rdb, logger.Named("redis"))
 	masterClient := cubemasterclient.New(cfg.CubeMasterURL, cfg.HTTPTimeout)
 	reg := registry.New()
+
+	// The eventbus carries best-effort cross-replica wakeup hints. Redis
+	// remains the source of truth and the wait path retains polling fallback.
+	var bus *eventbus.Bus
+	if cfg.EventBusEnabled {
+		bus = eventbus.New()
+		stream.SetNotifyEnabled(true)
+		stream.SetLocalBus(bus)
+		logger.Info("eventbus enabled")
+	} else {
+		logger.Info("eventbus disabled (waitForRunning will poll)")
+	}
 
 	rootCtx, cancel := signalContext()
 	defer cancel()
@@ -147,6 +160,7 @@ func run() error {
 		ProxyPush:    pushClient,
 		StateLockTTL: cfg.StateLockTTL,
 		Log:          logger.Named("resumer"),
+		EventBus:     bus,
 	})
 
 	sweep := sweeper.New(sweeper.Options{
@@ -170,6 +184,9 @@ func run() error {
 	if discSvc != nil {
 		loopCount++
 	}
+	if cfg.EventBusEnabled {
+		loopCount++
+	}
 	stateSyncDeps := statesync.Deps{
 		Registry:  reg,
 		Redis:     stream,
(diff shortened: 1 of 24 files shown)


chenhengqi:
cc @novahe

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -68,6 +69,18 @@ func run() error {
 	masterClient := cubemasterclient.New(cfg.CubeMasterURL, cfg.HTTPTimeout)
 	reg := registry.New()
 
+	// The eventbus carries best-effort cross-replica wakeup hints. Redis
+	// remains the source of truth and the wait path retains polling fallback.
+	var bus *eventbus.Bus
```
FakeLearne, line 74: CUBE_LCM_EVENTBUS_ENABLED=false (the documented kill switch) does not actually disable the eventbus wait path, and can panic.

bus is a *eventbus.Bus. When the flag is off it stays a nil pointer, but assigning it to Options.EventBus (waitBus interface) creates a typed-nil interface: {type: *eventbus.Bus, value: nil}. That is not == nil.

waitForRunning then takes the eventbus branch and calls EventBus.Wait(), which dereferences the nil *Bus (b.mu.Lock()).

This is reachable on a single replica: if the sweeper already holds pausing and a /internal/resume arrives, we enter waitForRunning. Default-on is fine; the kill switch is not.

Fix: keep a true nil interface unless the feature is on:

```golang
var bus waitBus
if cfg.EventBusEnabled {
    b := eventbus.New()
    stream.SetNotifyEnabled(true)
    stream.SetLocalBus(b)
    bus = b
}
```

(waitBus lives in resumer, so either export a small constructor there or pass EventBus only inside the if and leave the field unset.) A unit test with EventBusEnabled=false plus a pausing/resuming peer lock would catch this.

PR #1399, opened 2026-08-20
feat(network): update sandbox egress policy in place, re-evaluating live flows

FakeLearne:
Add PUT /sandboxes/{sandboxID}/network so a running sandbox's egress policy can be replaced without recreating it, and make the update reach traffic that already exists rather than only future connections.

The route, the status codes (204/400/404/409) and the full-replacement semantics follow E2B, so an omitted field clears rather than preserves. One divergence is deliberate: E2B's nftables chain accepts ESTABLISHED,RELATED first, so tightening a policy there has no effect on open connections and a long-lived connection can keep using a revoked destination indefinitely. We re-evaluate instead.

## Datapath

Each sandbox carries a policy generation (mvm_meta.policy_version), bumped only after both planes hold the new policy. Every session caches the generation it was admitted under, so the next packet on an established flow is re-judged exactly once per update:

- same verdict: restamped and left alone, costing one policy lookup per flow;
- no longer allowed, or the verdict changed: the session pair is deleted. TCP is answered with an RST, matching how every other unreachable TCP packet is handled here, so the guest fails fast instead of stalling on retransmits; UDP and ICMP have nothing to reset and are dropped.

A verdict *change* retires the flow rather than migrating it. The SNAT and L7 paths disagree about both the reply tuple and which side terminates the TCP connection, so migration is not expressible; the client reconnects and is judged as any new flow.

Deleting rather than flagging keeps the retirement self-enforcing, and is what lets this stay small. A later non-SYN packet finds no session and is reset, so a revoked flow cannot resume even if a subsequent update re-allows the destination, while a SYN legitimately opens a fresh connection under the current policy. Both directions go at once, so reply-path callers need no change either, and neither does the reaper.

That property holds only once the legacy 80/443 connection drain in do_tcp_nat is removed, which this does. While that branch is present a deleted session is indistinguishable from a lost one, and the drain silently undoes the revocation. It had also lost its own justification independently: written (1b5d2c35) when L7 flows created no session and re-ran the policy lookup per packet, so a DNS-learned entry expiring broke live connections, it was made redundant by #1347, after which L7 flows cache their verdict in nat_session and an established flow never consults the policy maps again. Two of the three causes its comment cites do not hold either -- both session maps are pinned and no startup path flushes them, and they are BPF_MAP_TYPE_HASH rather than LRU, so nothing is evicted. The remaining case, a flow idle past the reaper timeout, contradicts the reaper it depends on: deleteSessions removes both directions, and the drain then revives the egress direction alone, without recreating a session, on the sole evidence that the proxy still holds a socket -- scoped by which flows happen to leave a host socket rather than by policy, which is why it only ever applied to 80/443. Consequently an L7 keepalive connection left idle past the reaper timeout is now reset rather than revived, and the application reconnects.

policy_version fits in existing padding, so mvm_meta and nat_session keep their sizes and the pinned-map layout is unchanged. The kernel verifier accepts del_session() while the lookup pointer is still held, checked against mvmtap, nodenic and localgw.

## Control plane

cubevs gains UpdateTAPDevicePolicy, a third apply mode beside applyNetPolicy (additive) and replaceNetPolicy (flush then refill). Neither fits a live sandbox: flushing blanks the policy for as long as the refill takes, and swapping the inner map defeats the HashOfMaps inner cache and pays a synchronize_rcu per update. So it diffs against the installed maps and writes only what changed. DNS-learned rows are left untouched, and deny_out's desired set always includes the always-denied private and link-local ranges.

Ordering is part of the contract: CubeEgress, then CubeVS (which bumps the generation), then durable state. A failure before the bump leaves flows on their cached verdict instead of judging them against a half-applied map, and a failure before the state write replays the previous policy on restart. Nothing is rolled back; the diff is computed from the live maps, so replaying converges.

For clone, snapshot and pause/resume the new policy is written to all three places that outlive the update: the network runtime state file (restart recovery), the Cubelet cubebox store (pause packaging) and Master's sandboxspec (what clone replays). The sandboxspec write happens synchronously before the response, so update then snapshot then clone is read-your-writes; a failure there is logged and does not fail the call, following the same best-effort convention the create path already applies to that store.

## DNS resolver allowance

The create path folds the sandbox's resolver addresses into allow_out so domain rules can be resolved at all, but only when the policy names a domain. An update carries just the caller's targets, so it would silently revoke DNS and black-hole every domain rule it had just installed. The resolver list is now recorded on every create -- unconditionally, because a later update may introduce the first domain -- and folded back in under the same "policy still names a domain" condition, so dropping every domain also withdraws the implicit DNS access.

Recording is separate from installing: an all-IP policy still gets no resolver access, since allow_out cannot be narrowed to port 53. The gate therefore asks whether a target is *installed* as a domain, not whether it looks like a DNS name. The latter accepts "10.0.0.1" -- digits are valid label characters and nothing requires a non-numeric top label -- and would fold the resolver into every policy written with bare IPv4 literals, granting one extra /32 on all ports that the caller never asked for. cubevs exports that install decision as IsAllowOutDomainTarget, mirroring splitAllowOutTargets, so the update gate cannot diverge from the create gate, which has always been IP-first.

An update that neither installs nor clears L7 rules does not contact CubeEgress at all. That is decided from the old and new rule sets, not from whether the proxy happens to be configured, so an L3-only sandbox never depends on a component it does not use.

## Known limitation

Addresses already learned for a domain outlive the rule that produced them until their DNS TTL expires, so revoking a domain promptly needs a short resolver TTL. Provenance tracking for learned entries is deferred.

## Testing

Go unit tests cover the map diff (revocation, expanded L7 port sets, deny_out convergence, DNS port-set replacement), the generation bump and its survival across metadata rewrites, and the controller's ordering, resolver refold, L3-only skip and "a failed update keeps the old policy" guarantees. The resolver gate is exercised over bare, masked, subnet and L7-host target forms, because only the bare literals distinguish the install decision from a name-shape check. A new BPF case drives session_policy_revoked through the existing egress-policy harness across seven verdict/generation combinations.

Ten end-to-end cases were added under a new network_dynamic_update capability. Two of them are what actually distinguish this feature: a revoked connection must be torn down, and a still-permitted one must be left alone -- the second is what stops an implementation that simply kills every session from passing. Their guest-side holder uses TCP keepalive rather than application bytes, because a peer that hangs up on unexpected data is indistinguishable from a policy reset, and it classifies RESET, ALIVE, EOF and DATA separately so only the unambiguous outcomes are asserted.

Three more cover the compatibility requirement: a clone inherits the updated policy rather than the create-time one (which also covers read-your-writes, since clone snapshots immediately after the update returns), a clone of a narrowed policy is not more permissive, and the policy survives pause/resume. These have to be end-to-end -- clone and snapshot are not network-aware, they just replay a stored create request, so nothing under templatecenter changed here and review alone cannot tell whether the spec write-back landed.

Verified on a live single-node deployment: all ten cases pass on repeated runs, and examples/network-policy/network_dynamic_update.py walks an IP allow list, a connection carried across a revoking update (observed as RESET), a domain allow list, and switching on L7 interception mid-run.

### DIFF
--- a/CubeNet/cubevs/netpolicy.go
+++ b/CubeNet/cubevs/netpolicy.go
@@ -1009,7 +1009,7 @@ func splitAllowOutTargets(targets []string) ([]string, []string, error) {
 		if net.ParseIP(target) != nil || isDottedDecimalLikeTarget(target) {
 			return nil, nil, fmt.Errorf("unsupported allow_out IP target: %s", target) //nolint:err113
 		}
-		if !isDNSAllowTarget(target) {
+		if !IsAllowOutDomainTarget(target) {
 			return nil, nil, fmt.Errorf("invalid allow_out domain target: %s", target) //nolint:err113
 		}
 		domains = append(domains, target)
@@ -1042,6 +1042,25 @@ func isDottedDecimalLikeTarget(target string) bool {
 		}
 	}
 	return true
+}
+
+// IsAllowOutDomainTarget reports whether an allow_out target is installed as a
+// dns_allow_v2 domain rule rather than an allow_out_v3 CIDR.
+//
+// IP and CIDR literals take precedence, exactly as splitAllowOutTargets decides
+// it: isDNSAllowTarget() alone accepts an all-numeric name like "10.0.0.1"
+// because digits are valid DNS label characters, so asking it directly reports a
+// bare IPv4 literal as a domain. Callers outside this package need the install
+// decision, not the name-shape check, so this is the one they get.
+func IsAllowOutDomainTarget(target string) bool {
+	target = strings.TrimSpace(target)
+	if target == "" || isIPv4Target(target) || strings.Contains(target, "/") {
+		return false
+	}
+	if net.ParseIP(target) != nil || isDottedDecimalLikeTarget(target) {
+		return false
+	}
+	return isDNSAllowTarget(target)
 }
 
 func isDNSAllowTarget(target string) bool {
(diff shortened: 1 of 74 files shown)


CubeNet/src/session.h
```diff
@@ -170,6 +170,69 @@ static __always_inline __u8 classify_egress_flow(__u32 ifindex, __u32 daddr,
...
+{
+	__u8 verdict;
+
+	if (sess->policy_version == mvm_meta->policy_version)
+		return false;
+
+	verdict = classify_egress_flow(ifindex, daddr, dport);
+	if (verdict == FLOW_REJECT || verdict != session_verdict(sess))
+		return true;
+	sess->policy_version = mvm_meta->policy_version;
```
zhouxianping: During `classify_egress_flow`, `mvm_meta->policy_version` was modified synchronously in user space, resulting in an advanced version number being erroneously updated to `mvm_meta->policy_version`. It is recommended to first cache the value of `mvm_meta->policy_version` in a local variable and then use that local variable's value for policy checks.

CubeNet/src/session.h
```diff
@@ -227,6 +291,13 @@ static __always_inline bool create_nat_session(struct __sk_buff *skb,
 	sess.state = initial_state;
 	sess.packet_class = packet_class;
 	sess.l7_scheme = l7_scheme;
+	/* Stamp the generation this verdict was taken under. A missing meta
+	 * leaves 0, which reads as stale and costs one re-check on the next
+	 * packet -- the safe direction.
+	 */
+	meta = bpf_map_lookup_elem(&ifindex_to_mvmmeta, &vm_ifindex);
+	if (meta)
+		sess.policy_version = meta->policy_version;
```
zhouxianping: The version should be cached before calling `classify_egress_flow`, and this cached version should be recorded in the session; this prevents a scenario where a concurrent modification of `meta->policy_version` in user space causes the session to incorrectly record a version that is ahead of the actual state.

CubeNet/src/session.h
```diff
@@ -227,6 +291,13 @@ static __always_inline bool create_nat_session(struct __sk_buff *skb,
 	sess.state = initial_state;
 	sess.packet_class = packet_class;
 	sess.l7_scheme = l7_scheme;
+	/* Stamp the generation this verdict was taken under. A missing meta
+	 * leaves 0, which reads as stale and costs one re-check on the next
+	 * packet -- the safe direction.
+	 */
+	meta = bpf_map_lookup_elem(&ifindex_to_mvmmeta, &vm_ifindex);
```
zhouxianping: The `ifindex_to_mvmmeta` lookup has already been performed in the preceding steps; simply pass the `version` via the parameter.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -783,47 +799,29 @@ static __always_inline __u64 do_tcp_nat(struct __sk_buff *skb, struct mvm_meta *
...
-			 * No session is re-created, so each packet on the drained flow
-			 * re-enters this path (per-packet socket lookup — acceptable for
-			 * draining connections that will eventually close).
+			/* No session: the flow was never authorized, or it was retired
+			 * (reaped, or revoked by a policy update). Either way there is no
+			 * record that this connection is allowed, so answer like every
+			 * other unreachable TCP packet here instead of trusting that a
+			 * live proxy socket implies a past authorization.
 			 */
-			if (l4->dest == bpf_htons(80) || l4->dest == bpf_htons(443)) {
```
zhouxianping: Deleting this code will cause L7 connections established prior to the upgrade to stop working after the upgrade.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -783,47 +799,29 @@ static __always_inline __u64 do_tcp_nat(struct __sk_buff *skb, struct mvm_meta *
...
+do_recheck:
+	/* A policy update revoked this flow: retire the session pair so the tuple is
+	 * free for an immediate reconnect, and answer with an RST like every other
+	 * unreachable TCP packet here, so the guest learns now instead of stalling
+	 * until its retransmit timer gives up. Never RST an RST, or two peers that
+	 * both consider the flow dead would trade resets forever.
+	 */
+	if (session_policy_revoked(sess, mvm_meta, skb->ingress_ifindex,
+				   key.dst_ip, key.dst_port)) {
+		del_session(&key, sess);
```
zhouxianping: Why delete the conversation?

CubeNet/src/session.h
FakeLearne: All three are the same race. Fixed by reading it once at the top of each `do_*_nat()`, before any classification, and passing it by value to `session_policy_revoked()` and `create_nat_session()`. That drops the redundant `ifindex_to_mvmmeta` lookup too.

CubeNet/src/cubevs.h
```diff
@@ -145,12 +145,24 @@ const volatile __u16 nodenic_macaddr_p2 = 0x16dd;
...
+ * so bumping it would orphan every live session for this TAP.
+ */
 struct mvm_meta {
 	__u32 version;
 	__u32 ip;
 	__u8 uuid[64];
 	__u8 dns_policy_flags;
-	__u8 reserved[55];
+	__u8 reserved0[3];	/* aligns policy_version; reserved starts at an odd offset */
+	__u32 policy_version;
```
chenhengqi, line 175: This can be a `__u8`.

CubeNet/cubevs/tap.go
```diff
@@ -83,6 +83,9 @@ func UpsertTAPDeviceMetadata(ifindex uint32, ip net.IP, id string, version uint3
 		IP:      mvmIP,
 		UUID:    stringToByteArray(id),
 		Version: version,
+		// First generation for a brand-new TAP. 0 is reserved for "unset", so
+		// pre-upgrade sessions read as stale and get re-checked exactly once.
+		PolicyVersion: 1,
```
chenhengqi: This may invalidate some connections because of expired learned DNS entries.

CubeNet/cubevs/netpolicy.go
```diff
@@ -1174,6 +1212,183 @@ func netPolicyValueV3Expired(value netPolicyValueV3, now uint64) bool {
...
+		return err
+	}
+	defer m.Close()
+
+	var meta mvmMetadata
+	if err := m.Lookup(&ifindex, &meta); err != nil {
+		return fmt.Errorf("map.Lookup failed: %w, name: %s", err, MapNameIfindexToMVMMetadata)
+	}
+	meta.PolicyVersion++
+	if meta.PolicyVersion == 0 {
```
chenhengqi: This seems wrong.

FakeLearne: Done

CubeNet/cubevs/tap.go
FakeLearne: Done

chenhengqi (approved):

PR #1401, opened 2026-08-20
fix(cubemaster): return zero EndAt for NeverTimeout sandbox

### LINKED ISSUE #1400: [Bug Report] SetTimeout with NeverTimeout (-1) calculates EndAt in the past, causing premature expiration
## Summary

When setting sandbox timeout to `NeverTimeout` (`-1`), `CubeMaster` returns an `EndAt` timestamp roughly 1 second in the past (`now - 1000ms`) instead of `0` (never expires). This causes `CubeAPI` / CLM / SDKs to treat the newly updated sandbox as already expired immediately.

## Environment

- CubeSandbox commit: `927fd6f` (master)
- Relevant component: CubeMaster / CubeAPI

## Steps to Reproduce

1. Create a sandbox and invoke `POST /cube/sandbox/timeout` with payload:
   ```json
   {
     "sandboxID": "<valid-sandbox-id>",
     "timeout": -1
   }
   ```
2. Inspect the response payload `end_at`.
3. In `CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go`, assert `rsp.EndAt == 0` for `timeout = -1`:
   ```text
   sandbox_timeout_test.go:116: timeout=-1 (NeverTimeout) must return EndAt=0 (never expires), got 1787230892317
   --- FAIL: TestSetTimeoutValidationAllowsNeverTimeout (0.00s)
   ```

## Expected Behavior

`SetTimeout` with `timeout: -1` (`types.NeverTimeout`) should return `EndAt = 0` (omitted in JSON), indicating the sandbox has no expiration deadline.

## Actual Behavior

`refreshTimeoutMeta` in `CubeMaster/pkg/service/sandbox/sandbox_timeout.go` delegates to the timeout provider, which correctly returns `0`. However, the handler uses:

```go
else if endAt > 0 {
    return endAt
}
```

Since `endAt == 0`, it falls through to the fallback duration arithmetic:

```go
return time.Now().UnixMilli() + int64(timeoutSeconds)*1000 // now - 1000ms
```

This sets `EndAt` to an elapsed timestamp (`now - 1000ms`). `CubeAPI` receives a non-zero timestamp, wraps it as `Some(DateTime)` in the past, and clients/dashboards/CLM treat the sandbox as already expired.

## Additional Context

- Related code: `CubeMaster/pkg/service/sandbox/sandbox_timeout.go` (around lines 124-135)

MeiSiristhebest:
Closes #1400.

## Motivation

When `SetTimeout` is invoked with `timeout = -1` (`types.NeverTimeout`), `refreshTimeoutMeta` passes `-1` to the underlying lifecycle timeout provider, which correctly calculates `EndAt = 0` (unlimited). However, because `refreshTimeoutMeta` guarded the provider result with `else if endAt > 0`, the zero value fell through to the fallback calculation:

```go
return time.Now().UnixMilli() + int64(timeoutSeconds)*1000 // now - 1000ms
```

This returned a non-zero timestamp in the past (`now - 1s`), causing CubeAPI, client SDKs, and CLM to interpret the sandbox as already expired immediately upon setting `NeverTimeout`.

## What this changes

1. **`CubeMaster/pkg/service/sandbox/sandbox_timeout.go`** — Consolidated into a single `getTimeoutProvider()` call that accepts `endAt` directly when `endAt > 0 || timeoutSeconds < 0`, followed by a negative-aware fallback returning `0`.
2. **`CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go`** — Added comprehensive regression test coverage:
   - `TestSetTimeoutValidationAllowsNeverTimeout`: exercises the no-provider path.
   - `TestSetTimeoutWithInstalledProviderAllowsNeverTimeout`: verifies the provider round-trip with `timeoutSeconds == -1` and `EndAt == 0` propagation.
   - `TestSetTimeoutWithInstalledProviderErrorAllowsNeverTimeout`: verifies that the fallback safely yields `0` even when the provider returns an error.

## Testing

- Unit test regression verification:
  `go test -v -count=5 -run "TestSetTimeout" ./pkg/service/sandbox` (passed)
- Full package test suite:
  `go test ./pkg/service/sandbox` (passed)
- Code formatted with `gofmt`.

### DIFF
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout.go
@@ -126,9 +126,14 @@ func refreshTimeoutMeta(ctx context.Context, sandboxID string, timeoutSeconds in
 		endAt, err := p.RefreshTimeout(ctx, sandboxID, timeoutSeconds)
 		if err != nil {
 			log.G(ctx).Warnf("lifecycle: RefreshTimeout sandbox=%s failed: %v", sandboxID, err)
+		} else if timeoutSeconds < 0 {
+			return 0
 		} else if endAt > 0 {
 			return endAt
 		}
+	}
+	if timeoutSeconds < 0 {
+		return 0
 	}
 	return time.Now().UnixMilli() + int64(timeoutSeconds)*1000
 }
--- a/CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_timeout_test.go
@@ -6,6 +6,7 @@ package sandbox
 
 import (
 	"context"
+	"errors"
 	"testing"
 
 	"github.com/tencentcloud/CubeSandbox/CubeMaster/pkg/errorcode"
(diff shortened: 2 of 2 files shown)


luzhixing12345:
LGTM, this fix is ​​concise and accurate.

PR #1403, opened 2026-08-20
refactor(grpcpool): harden ref-count concurrency invariants and simplify GracefulStop

### LINKED ISSUE #1402: grpc-middleware/pool: harden ref-count concurrency invariants and simplify GracefulStop
## Summary

In both `CubeMaster/pkg/base/grpc-middleware/pool` and `Cubelet/pkg/grpc-middleware/pool`, the gRPC connection pool implementation has concurrency invariant and lifecycle edge cases that require hardening:

1. **Atomic reference count underflow**: `decrRef()` decrements `p.ref` using `atomic.AddInt32(&p.ref, -1)`. When `newRef < 0`, it only clamped the local variable `newRef = 0` and never updated `p.ref` in memory. Under redundant closes or edge timings, this leaves `p.ref` negative, corrupting reference tracking and breaking the `ref == 0` invariant required for cache eviction sweeps (`workerGrpcConnPool.checkWorkerConn`).
2. **Missing ref balance in `Get()`**: On error return paths (`current == 0` when pool is closed, dial failures on saturated pools, and dial failures during pool expansion), `Get()` incremented `p.ref` via `incrRef()` but returned without calling `decrRef()`, leading to unbalanced reference counts.
3. **Double-`Close()` race and unnecessary goroutine in `GracefulStop`**: `GracefulStop` spawned an asynchronous background polling goroutine alongside `time.After`. On timeout, both the main caller and the background routine could invoke `Close()` concurrently.
4. **Data race in `GetActiveTimeAndRef`**: Reading `p.active` without `p.RLock()` and reading `p.ref` directly without atomic load posed data-race risks with concurrent callers.

### 1. Ref Count Underflow in `decrRef`

```go
func (p *pool) decrRef() {
	if p == nil {
		return
	}

	newRef := atomic.AddInt32(&p.ref, -1)
	if newRef < 0 {
		newRef = 0 // Only mutates local variable; p.ref in memory remains negative (< 0)
	}
    ...
}
```

If `decrRef` is called when `p.ref == 0`, `p.ref` in memory drops to `-1`. A subsequent `checkWorkerConn` sweep checking `ref == 0` would fail to evict the idle pool from cache.

### 2. Double-Close Race in `GracefulStop`

```go
func (p *pool) GracefulStop(maxWaitTime time.Duration) {
	done := make(chan struct{}, 1)
	go func() {
		for {
			if atomic.LoadInt32(&p.ref) <= 0 {
				p.Close()
				done <- struct{}{}
				break
			}
			time.Sleep(100 * time.Millisecond)
		}
	}()
	select {
	case <-done:
		return
	case <-time.After(maxWaitTime):
		p.Close()
		return
	}
}
```

## Proposed Fix

1. Use atomic CAS loops in `decrRef` and `incrRef` to ensure `p.ref` in memory never underflows below `0` or overflows past `MaxInt32`.
2. Balance `p.ref` by calling `p.decrRef()` on all `Get()` error paths (`ErrClosed`, dial failure).
3. Refactor `GracefulStop` into a clean synchronous timer loop with explicit `defer timer.Stop()` and `defer ticker.Stop()`, eliminating background goroutines and the double-`Close()` race condition.
4. Protect `GetActiveTimeAndRef` with `p.RLock()` and `atomic.LoadInt32(&p.ref)`.
5. Guard `Get()` against concurrent `Close()` after acquiring lock and before index modulo.
6. Add comprehensive unit tests and keep `CubeMaster` and `Cubelet` packages aligned.

MeiSiristhebest:
## Summary

Fixes #1402

This PR hardens the reference counting and lifecycle invariants of the internal `grpcpool` implementation across `CubeMaster` and `Cubelet`:

1. **CAS-based clamp on `decrRef`**: Replaces the local-clamping decrement with an atomic CAS loop, guaranteeing that `p.ref` in memory never drops below `0` even under redundant closes or edge-case timing, preserving the `ref == 0` invariant required for cache eviction sweeps (`checkWorkerConn`).
2. **Balanced ref tracking on `Get()` error paths**:
   - Releases incremented ref when the pool is closed (`current == 0`).
   - Releases incremented ref when `Dial` fails (both in the `current >= MaxActive` non-reuse path and during pool expansion).
   - Re-checks and returns `ErrClosed` (releasing ref) if `CheckConnStatus` encounters a nil connection from a concurrent close race.
3. **Data race elimination**: Protects `GetActiveTimeAndRef` with read lock (`p.RLock()`) and `atomic.LoadInt32(&p.ref)`.
4. **Simplified, synchronous `GracefulStop`**: Replaces the background goroutine, channel signaling, and `time.After` with a synchronous wait loop using explicit `Timer` and `Ticker` cleanup. This eliminates the concurrent double-`Close()` race condition and removes unnecessary goroutine allocations (*Note: `GracefulStop` currently has no active callers in the codebase*).
5. **Synchronized `Cubelet` mirror**: Applies identical hardening and unit tests to `Cubelet/pkg/grpc-middleware/pool` to keep the package implementations aligned (*Note: currently unreferenced across the repo*).

## Verification

Added and verified comprehensive unit tests covering:
- `TestDecrRefUnderflow`: asserts CAS clamp keeps `ref >= 0` across redundant decrements.
- `TestConcurrentIncrDecrRef`: samples `ref >= 0` under high concurrency and asserts final `ref == 0`.
- `TestGetAfterCloseNoRefLeak`: verifies ref balance when pool is closed.
- `TestGetDialFailureNoRefLeak`: verifies ref balance when `Dial()` returns an error.
- `TestGetNilConnAfterCloseRace`: verifies safe `ErrClosed` return when connection is nil.
- `TestGracefulStopImmediateWhenZeroRef` & `TestGracefulStopTimeout`: validates synchronous timeout and immediate termination semantics.
- `TestIncrRefOverflowProtection`: verifies max integer boundary clamp.
- `TestDecrRefShrinkPath`: validates idle connection eviction on ref drop.

All tests pass cleanly (`go test -v -count=10`).

### DIFF
--- a/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
+++ b/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
@@ -180,6 +180,7 @@ func (p *pool) Get() (Conn, error) {
 	nextRef := p.incrRef()
 	current := atomic.LoadInt32(&p.current)
 	if current == 0 {
+		p.decrRef()
 		return nil, ErrClosed
 	}
 
--- a/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
+++ b/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
@@ -206,7 +206,11 @@ func (p *pool) Get() (Conn, error) {
 		}
 
 		c, err := p.opt.Dial(p.ua, p.address)
-		return p.wrapConn(c, true), err
+		if err != nil {
+			p.decrRef()
+			return nil, err
+		}
+		return p.wrapConn(c, true), nil
 	}
 
 	p.Lock()
@@ -232,6 +236,7 @@ func (p *pool) Get() (Conn, error) {
 		atomic.StoreInt32(&p.current, current)
 		if err != nil {
 			p.Unlock()
+			p.decrRef()
 			return nil, err
 		}
 	}
--- a/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
+++ b/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
@@ -51,7 +51,10 @@ type pool struct {
 }
 
 func (p *pool) GetActiveTimeAndRef() (time.Time, int32) {
-	return p.active, p.ref
+	p.RLock()
+	active := p.active
+	p.RUnlock()
+	return active, atomic.LoadInt32(&p.ref)
 }
 
 func New(ua, address string, option Options) (Pool, error) {
@@ -215,6 +218,11 @@ func (p *pool) Get() (Conn, error) {
 
 	p.Lock()
 	current = atomic.LoadInt32(&p.current)
+	if current == 0 {
+		p.Unlock()
+		p.decrRef()
+		return nil, ErrClosed
+	}
 	if current < int32(p.opt.MaxActive) && nextRef > current*int32(p.opt.MaxConcurrentStreams) {
 
 		increment := current
@@ -241,6 +249,10 @@ func (p *pool) Get() (Conn, error) {
 		}
 	}
 	p.Unlock()
+	if current == 0 {
+		p.decrRef()
+		return nil, ErrClosed
+	}
 	next := atomic.AddUint32(&p.index, 1) % uint32(current)
 	return p.CheckConnStatus(p.conns[next])
 }
--- a/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
+++ b/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
@@ -287,7 +287,10 @@ func (p *pool) GracefulStop(maxWaitTime time.Duration) {
 }
 
 func (p *pool) CheckConnStatus(conn *conn) (*conn, error) {
-
+	if conn == nil {
+		p.decrRef()
+		return nil, ErrClosed
+	}
 	return conn, nil
 }
 
--- a/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
+++ b/CubeMaster/pkg/base/grpc-middleware/pool/pool.go
@@ -111,11 +111,15 @@ func New(ua, address string, option Options) (Pool, error) {
 }
 
 func (p *pool) incrRef() int32 {
-	newRef := atomic.AddInt32(&p.ref, 1)
-	if newRef >= math.MaxInt32 {
-		newRef = math.MaxInt32
+	for {
+		old := atomic.LoadInt32(&p.ref)
+		if old >= math.MaxInt32 {
+			return math.MaxInt32
+		}
+		if atomic.CompareAndSwapInt32(&p.ref, old, old+1) {
+			return old + 1
+		}
 	}
-	return newRef
 }
 
 func (p *pool) decrRef() {
@@ -123,9 +127,17 @@ func (p *pool) decrRef() {
 		return
 	}
 
-	newRef := atomic.AddInt32(&p.ref, -1)
-	if newRef < 0 {
-		newRef = 0
+	var newRef int32
+	for {
+		old := atomic.LoadInt32(&p.ref)
+		if old <= 0 {
+			newRef = 0
+			break
+		}
+		if atomic.CompareAndSwapInt32(&p.ref, old, old-1) {
+			newRef = old - 1
+			break
+		}
 	}
 
 	if newRef == 0 && atomic.LoadInt32(&p.current) > int32(p.opt.MaxIdle) {
@@ -136,7 +148,6 @@ func (p *pool) decrRef() {
 		}
 		p.Unlock()
 	}
-
 }
 
 func (p *pool) reset(index int) {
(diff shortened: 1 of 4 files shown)


fslongjin:
Thanks for the PR — the code-level changes are correct as far as they go, and the fail-first test legitimately demonstrates the clamp bug. However, after walking the actual call chain, I don't think the PR description (and issue #1402) substantiates a production leak. Summarizing what I verified against `origin/master`:

## What's real

- **`decrRef` clamp bug**: yes, the old code only clamped the local variable and left `p.ref` negative in memory. The CAS loop is the correct fix.
- **`Get()` on a closed pool leaking a ref**: yes, the missing `decrRef()` before `return nil, ErrClosed` was a real defect, and adding it is correct.

## What the description claims vs. what the call chain shows

### 1. "Negative `ref` makes `checkWorkerConn` never reclaim the pool" — the causal chain breaks in practice

The only production consumer of this pool is `workerGrpcConnPool` (`CubeMaster/pkg/cubelet/grpcconn/worker_conn.go`), and every call site (`actions.go`, ~20 sites) follows a strict `GetWorkerConn()` → single `defer conn.Close()` pattern. The only interleaving that drives `ref` below zero is:

```
GetWorkerConn: cache.Load(p) ──► p.Get(): ref 0→1, returns conn
checkWorkerConn: reads ref==0 ──► cache.Delete(key) ──► (routine pool) p.Close(): ref=0, current=0
caller: defer conn.Close() ──► decrRef(): ref 0→-1
```

Note the pool has **already been removed from the cache** by the time the underflow lands on it. Nothing ever reads its `ref` again — the object is garbage-collected. For the described leak ("pool stuck in `sync.Map` because `ref != 0`") you'd need a closed pool to remain reachable in the cache with a nonzero `ref`, and no current caller produces that state. The fail-first test (`TestDecrRefUnderflow`) validates the clamp is ineffective, not that worker cleanup ever hits this path.

### 2. "`GracefulStop` background goroutine spins forever after timeout" — this is factually incorrect

On the timeout path, `GracefulStop` itself calls `p.Close()`, and `Close()` does `atomic.StoreInt32(&p.ref, 0)`. The background goroutine observes `ref <= 0` on its next tick (≤100 ms later), sends to the buffered channel, and exits. A permanent spin would additionally require `ref` to stay > 0 after `Close()` — please re-check this part of the issue write-up. Also worth noting: `GracefulStop` currently has **zero callers** in the repo (the historical callers in `Cubelet/network/plugin_eni.go` and `Cubelet/storage/local.go` were removed with the ENI/EKS/CBS refactors), so this is at best a latent defect. That said, the synchronous rewrite is a genuine simplification (one goroutine + channel + `time.After` replaced by a plain loop, and it removes the unsynchronized double-`Close()` race between the two old callers of `Close()`), so I'm fine keeping it — as cleanup, not as a bugfix.

### 3. Cubelet-side copy: zero references

`Cubelet/pkg/grpc-middleware/pool` has no importers anywhere in the repo. The mirrored fix lands entirely on dead code. Not harmful, but it shouldn't be counted toward "fixing a production issue in two components."

## The adjacent bug this PR doesn't touch

While tracing this, I found what looks like the *actual* "connections to removed nodes never close" defect: `GetWorkerConn` stores pools under `ua + "+" + addr` (`worker_conn.go:66`), but `CloseWorkerConn` looks up with the bare `addr` (`worker_conn.go:84`). The lookup never matches, so the `CloseWorkerConn` calls in `node_cache.go:220/240` (node delete / node down) are no-ops; stale pools only get reclaimed later by the expiry sweep. This is reachable today on a plain node-removal path and seems more impactful than either issue in #1402 — would you be interested in filing/following up on that one?

## Ask

To merge this, I'd need one of:

1. **A reproduction along a real call chain** — e.g. a test that interleaves `checkWorkerConn`'s Delete→Close with `GetWorkerConn`/`conn.Close()` and shows either (a) a pool remaining in the cache with `ref != 0` indefinitely, or (b) the `GracefulStop` goroutine outliving the call by more than ~100 ms; **or**
2. **Retitle/downgrade**: drop the "goroutine spins forever" and "checkWorkerConn leak" claims, describe this as concurrency hardening of the ref-count invariants (CAS clamp + `ErrClosed` ref release + synchronous `GracefulStop`), and note explicitly that `GracefulStop` has no current callers and the Cubelet copy is unreferenced.

The changes themselves are fine to keep either way — my objection is to the framing, not the diff.

MeiSiristhebest:
Thanks @fslongjin for the thorough walkthrough and insightful code analysis!

I completely agree with your assessment regarding the production call chains and severity framing. I have updated the PR accordingly following your suggestions:

1. **Retitled & Reframed**: Updated the title and description to focus on **concurrency hardening of ref-count invariants and cleanup** (CAS clamp to prevent negative in-memory `ref` under defensive/redundant close scenarios, `Get()` error ref balancing, and synchronous `GracefulStop` cleanup).
2. **Clarified Scope**: Explicitly documented in the PR description that `GracefulStop` currently has no callers and `Cubelet`'s package is unreferenced across the repo (kept for parity).
3. **Hardened `Get()` Dial Failure Paths**: As noted by the static analysis review, also added `p.decrRef()` when `Dial()` returns an error during connection creation/pool expansion, and added unit tests (`TestGetDialFailureNoRefLeak`) to cover these paths.
4. **Follow-up on `CloseWorkerConn`**: That's a great catch on the `ua + "+" + addr` vs `addr` key mismatch in `worker_conn.go`! I'd be very happy to file and follow up on a separate PR to fix that defect.

Please take another look when you have a moment. Thanks again for the detailed review!

fslongjin (changes requested):
please rebase your code onto the branch `origin/master`

MeiSiristhebest:
Done! Rebased cleanly onto the latest `origin/master`. Thanks @fslongjin!

lkml-likexu:
@fslongjin Any review comments for this one ?

PR #1409, opened 2026-08-21
cubecow: introduce s3 backend to support cross-machine pause/resume

lisongqian:


ls-ggg:
I've tested this PR via cherry-pick. LGTM!

PR #1411, opened 2026-08-21
cubeegress: require an explicit opt-in before injecting credentials over plaintext HTTP

### LINKED ISSUE #1410: [Bug Report] Injected credentials can be exfiltrated to any destination IP on the plaintext HTTP path
On the `:8080` plaintext path CubeEgress injects the operator's credential based solely on a sandbox-supplied
`Host` header, then proxies to the sandbox-chosen destination IP. Nothing authenticates that destination, so
a sandbox can send the credential to a host it controls.

## Environment

- Deployment mode: single-node / cluster (default network posture)

## Steps to Reproduce

Given an operator rule that injects a credential and matches on `host`, with no `scheme` constraint — which
matches both schemes, since a missing `match` field matches anything (`rule_matches:146-148`):

```json
{"id":"r-allow-internal-api",
 "match":{"host":"api.internal.corp"},
 "action":{"allow":true,"inject":[{"header":"Authorization","secret":"<real token>"}]}}
```

1. From the sandbox, open a plain TCP connection to an **attacker-controlled IP** on port 80. TPROXY captures
   it and the `:8080` block handles it.
2. Send `GET / HTTP/1.1` with `Host: api.internal.corp`.
3. `rule_matches` matches on `host`; `inject_gates` sees `scheme == "http"` and a non-empty `Host` and passes.
4. `apply_injects` sets `Authorization: <real token>`.
5. `proxy_pass http://$server_addr:80` delivers the request, with the credential, to the attacker.

## Expected Behavior

The credential is not injected toward a destination that has not been authenticated.

## Actual Behavior

It is. Driving the real `access_phase.lua`:

```
  HTTPS, Host==SNI (legit)              allow=true  injected=sk-live-REAL-OPERATOR-CREDENTIAL
  HTTP, spoofed Host -> attacker IP     allow=true  injected=sk-live-REAL-OPERATOR-CREDENTIAL
```

The HTTPS case is safe for a different reason (below); the HTTP case is the bug.

## Additional Context

The module's own gate documentation claims this is covered:

```
-- access_phase.lua:26-27
--   G5 dst IP authentic — implicit via G6 (proxy_ssl_verify)
--   G6 upstream cert    — Pδ; nginx proxy_ssl_verify on
```

That holds for `:8443`, which does `proxy_pass https://$server_addr:$server_port` with `proxy_ssl_verify on`
(`nginx.conf:95`) and `proxy_ssl_name $cube_original_sni` (`:192`) — an attacker IP cannot present a valid
certificate for someone else's SNI, so the handshake fails. Confirmed: with `Host != SNI` and an attacker
destination, G4 already blocks the inject.

`:8080` does `proxy_pass http://$server_addr:$server_port` — plaintext, no certificate, so **G5/G6 have no
enforcement at all there**. The only HTTP-path check is that `Host` is non-empty
(`access_phase.lua:215-219`). `ctx.dst_ip` is recorded for audit (`:296`) but never compared against
anything.

This is precisely the "talk to host A but route as if Host=B" shape the HTTPS branch of G4 was written to
block (`:196-201`), left unblocked on HTTP.

**Reachability.** The destination is only constrained if cubevs restricts it, and `AllowInternetAccess`
defaults to **true** when unset (`Cubelet/network/runtime/policy_builder.go:99,107`, also
`plugin_policy.go:112,116`, `CubeOps/internal/service/openclaw.go:945`;
`CubeAPI/src/models/mod.rs:818` is `Option<bool>` defaulting to `None`). So in the default posture the
destination is unconstrained.

dwin-gharibi:
Closes #1410.

On `:8443` the destination is authenticated: `proxy_ssl_verify on` plus `proxy_ssl_name $cube_original_sni`
means an attacker-controlled IP cannot present a valid certificate for someone else's SNI, so G4's
`Host == SNI` check plus the upstream handshake genuinely bind the credential to the intended host.

On `:8080` there is no certificate, so none of that applies. The only HTTP-path check was that `Host` is
non-empty — and `Host` is supplied by the sandbox, which also chooses the destination IP via TPROXY. A
sandbox could therefore connect to its own server, send `Host: api.internal.corp`, and receive the
operator's credential.

## What this changes

Plaintext injection becomes **opt-in per rule** instead of implicit.

**`CubeEgress/lua/access_phase.lua`**

- `inject_gates` gains a check on the HTTP branch: unless the matched rule sets
  `action.allow_plaintext_inject = true`, injection is refused with the new reason
  `g5_plaintext_inject_not_permitted`, flagged as a `security_event`.
- `decide()` carries `r.action.allow_plaintext_inject == true` onto the decision, and the `inject_gates`
  call site passes it through.

**`CubeEgress/lua/policy.lua`** — `validate_policy` rejects a non-boolean `allow_plaintext_inject`, so a
typo like `"true"` fails at install time rather than silently reading as falsy.

The HTTPS path is untouched. The request itself is still **allowed** when the gate fails — consistent with
the existing contract that a gate failure drops the inject rather than blocking traffic — and the existing
pre-clear of inject headers still runs, so a sandbox cannot smuggle its own forged `Authorization` through in
place of the real one.

No comment changes. Note the G5/G6 comment at `access_phase.lua:26-27` is now doubly inaccurate for `:8080`;
I left it alone under the no-comment-changes constraint, but it should be corrected when someone touches that
block.

## Testing

Driving the real `access_phase.lua` end to end:

| case | master | this branch |
|---|---|---|
| HTTPS, `Host == SNI` (legit) | credential injected | credential injected |
| HTTP, spoofed `Host` → attacker IP, default rule | **credential injected** | **`<none>`**, `g5_plaintext_inject_not_permitted` |
| HTTP, rule sets `allow_plaintext_inject = true` | credential injected | credential injected |
| HTTP, rule sets `allow_plaintext_inject = false` | **credential injected** | `<none>`, `g5_plaintext_inject_not_permitted` |

The first row is the regression guard — the fix must not disturb the HTTPS path, which is the normal case.
The last row is worth noting: on master the field is ignored entirely, so even an operator who explicitly
disabled it still leaked.

Field validation:

```
  allow_plaintext_inject=omitted          accepted=true   want=true   OK
  allow_plaintext_inject=true             accepted=true   want=true   OK
  allow_plaintext_inject=false            accepted=true   want=true   OK
  allow_plaintext_inject="yes" (string)   accepted=false  want=false  OK
  allow_plaintext_inject=1 (number)       accepted=false  want=false  OK
```

`loadfile` on both modules → SYNTAX OK.

CI gates: the Lua files are not covered by `fmt-check` or `unit-test-check`.

## Why opt-in rather than an outright ban or a dst-IP check

- **Outright ban** would break the case `inject_gates`' own comment calls out — "operators may legitimately
  need to inject credentials into plaintext traffic on trusted networks (e.g. intra-cluster 80)".
- **Resolving the Host and requiring `ctx.dst_ip` to be in the resolved set** is the stronger fix, but it puts
  a DNS dependency on the request path and races DNS changes; it also needs a decision about caching and
  failure modes. Worth doing, and it would let the opt-in be removed again.

Opt-in is the smallest change that makes the default safe and makes the risky configuration explicit and
greppable.

## Deployment impact — read before merging

**Any existing rule that injects over plaintext HTTP will stop injecting** until
`allow_plaintext_inject: true` is added to it. Operators relying on intra-cluster `:80` injection must update
those rules. The failure is visible rather than silent: the decision carries
`inject_dropped = g5_plaintext_inject_not_permitted` and raises a `security_event`, so it shows up in the
audit stream immediately.

### DIFF
--- a/CubeEgress/lua/access_phase.lua
+++ b/CubeEgress/lua/access_phase.lua
@@ -216,6 +216,9 @@ local function inject_gates(ctx)
         if not ctx.host or ctx.host == "" then
             return false, "g4_missing_host", true
         end
+        if ctx.allow_plaintext_inject ~= true then
+            return false, "g5_plaintext_inject_not_permitted", true
+        end
     end
     return true
 end
@@ -354,9 +357,10 @@ local function allow(decision)
             end
         end
         local ok, reason, sec = inject_gates({
-            scheme = decision.scheme,
-            sni    = decision.sni,
-            host   = decision.host,
+            scheme                 = decision.scheme,
+            sni                    = decision.sni,
+            host                   = decision.host,
+            allow_plaintext_inject = decision.allow_plaintext_inject,
         })
         if not ok then
             decision.inject_dropped = reason
@@ -471,6 +475,7 @@ function _M.decide()
             decision.rule_id     = r.id
             decision.audit_level = (r.action and r.action.audit) or "metadata"
             decision.inject      = r.action and r.action.inject  -- consumed in Pγ
+            decision.allow_plaintext_inject = r.action and r.action.allow_plaintext_inject == true
             if r.action and r.action.allow == true then
                 return allow(decision)
             else
--- a/CubeEgress/lua/policy.lua
+++ b/CubeEgress/lua/policy.lua
@@ -89,6 +89,11 @@ local function validate_policy(p)
             return false, "rules[" .. i .. "].action.allow required (boolean)"
         end
         -- inject is optional; if present must be array of {header, secret, format}.
+        if r.action.allow_plaintext_inject ~= nil
+           and type(r.action.allow_plaintext_inject) ~= "boolean" then
+            return false, string.format(
+                "rules[%d].action.allow_plaintext_inject must be a boolean", i)
+        end
         if r.action.inject ~= nil then
             if type(r.action.inject) ~= "table" then
                 return false, "rules[" .. i .. "].action.inject must be array"


chenhengqi (changes requested):
See my comments on #1410. Let's gather some consensus that there is a real issue before the fix.

PR #1417, opened 2026-08-21
sdk: recognise non-ASCII domains so the allow_out deny-all guard fires

### LINKED ISSUE #1416: [Bug Report] Go and Node SDKs skip the allow_out deny-all guard for non-ASCII domains
## Summary

`validateAllowOutDomainsRequireDenyAll` returns early when no `allowOut` entry is recognised as a domain.
The Go and Node SDKs validate domain labels ASCII-only, so a non-ASCII or IDN domain is not recognised and
the "you must also deny all other traffic" guard **never fires**. Python recognises it and does fire.

## Environment

- Relevant component: sdk/go, sdk/node

## Steps to Reproduce

Go:

```go
_, err := client.Create(ctx, &CreateOptions{
    Network: &NetworkOptions{AllowOut: []string{"café.example.com"}},
})
// err == nil — no guard
```

Python, same input:

```
ApiError: When specifying allowed domains in allow_out, you must disable public
outbound traffic or include '0.0.0.0/0' in deny_out to block all other traffic.
```

## Expected Behavior

All SDKs raise the guard, so the user is told their egress is not actually locked down.

## Actual Behavior

| `allowOut` entry | Python raises | Go errors | Node throws |
|---|---|---|---|
| `example.com` | yes | yes | yes |
| `exаmple.com` (Cyrillic `а`) | yes | **no** | **no** |
| `bücher.example` | yes | **no** | **no** |
| `例子.com` | yes | **no** | **no** |
| `café.example.com` | yes | **no** | **no** |
| `xn--bcher-kva.example` (punycode) | yes | yes | yes |

A Go or Node user who writes `allowOut: ["café.example.com"]` with public egress still enabled gets **no
error** and believes egress is restricted to one domain. Homoglyph entries such as `exаmple.com` are the
worst case — visually identical to the ASCII form, silently unguarded.

## Additional Context

- `sdk/go/policy.go:130-145` — `isValidDNSDomainName`, `ch >= 'a' && ch <= 'z' || ...` (ASCII only).
- `sdk/node/src/policy.ts:237-249` — `/^[a-zA-Z0-9-]+$/` (ASCII only).
- `sdk/python/cubesandbox/_policy.py:376-386` — `ch.isalnum()`, which is Unicode-aware.

Both non-Python versions claim parity in their comments — Go: "mirrors the current server contract";
Node: "Mirrors the Python SDK".

dwin-gharibi:
Closes #1416.

## Motivation

`validateAllowOutDomainsRequireDenyAll` bails out early when nothing in `allowOut` looks like a domain.
Go and Node validated labels with an ASCII-only character class, so any non-ASCII or IDN domain was not
classified as a domain and the guard silently did not run — the user was told nothing and kept public
egress enabled while believing it was restricted. Python, which uses `str.isalnum()`, does fire.

Homoglyphs are the sharp edge: `exаmple.com` with a Cyrillic `а` is visually identical to the ASCII form
and was completely unguarded in two of the three SDKs.

## What this changes

- `sdk/go/policy.go` — `isValidDNSDomainName` accepts a label character if it is `-`, ASCII
  alphanumeric, or `unicode.IsLetter` / `IsDigit` / `IsNumber`, which is the Go equivalent of Python's
  `str.isalnum()`.
- `sdk/node/src/policy.ts` — the label regex becomes `/^[\p{L}\p{N}0-9-]+$/u`.

Python is already correct and is unchanged. Everything else about the classifier is untouched: IPs,
CIDRs, dotted-decimal lookalikes, `*` handling, length limits and the leading/trailing `-` rules all
behave exactly as before.

No comment changes.

## Why widen the classifier rather than narrow Python

The classifier answers "is this a domain rather than an IP or CIDR?", and a non-ASCII string is
definitively not an IP or a CIDR. Widening makes the guard **fire**, which is the fail-safe direction: the
user gets a clear error telling them to add deny-all. Narrowing Python would have made all three SDKs
consistently *skip* the guard, i.e. consistent but silently unsafe.

I did not add IDNA/punycode encoding. That would be the fuller fix, but it means a new dependency in
`sdk/go` (`golang.org/x/net/idna`) for a module that deliberately has almost none, and it is a separate
decision. Classification alone is enough to close the guard hole.

## Recommended follow-up (not in this PR)

`Cubelet/network/runtime/policy_builder.go:220-246` (`isValidL7DomainName`) is also ASCII-only, so a
non-ASCII SNI/Host match is silently dropped from the cubevs L7 allow list rather than rejected. Now that
the SDKs pass such entries through instead of ignoring them, that server-side path deserves its own issue
— either punycode-normalise or reject explicitly.

## Testing

New:

- `sdk/go/policy_domain_guard_test.go` — `TestAllowOutDomainGuardFiresForUnicodeDomains` over seven
  domains (ASCII, Cyrillic homoglyph, umlaut, CJK, accented, punycode, wildcard+accented). For each it
  asserts the entry is classified as a domain, that the guard fires with public egress on, and that it
  does **not** fire when `deny_out` contains `0.0.0.0/0` or `defaultDenyAll` is set.
  `TestNonDomainTargetsStillNotTreatedAsDomains` covers IPs, CIDRs, IPv6, dotted-decimal lookalikes,
  empty, whitespace, underscore and double-star — so widening the charset did not turn non-domains into
  domains.
- `sdk/node/test/policy-domain-guard.test.ts` — the same matrix, 20 cases.

Red/green:

```
Go, with policy.go reverted:
  --- FAIL: TestAllowOutDomainGuardFiresForUnicodeDomains
      isDomainAllowOutTarget("exаmple.com") = false, want true
      isDomainAllowOutTarget("bücher.example") = false, want true
      isDomainAllowOutTarget("例子.com") = false, want true

Node, with policy.ts reverted:
  × fires for exаmple.com / bücher.example / 例子.com / café.example.com / *.café.example.com
  Failed Tests 5
```

All three SDK suites with the fix:

```
sdk/go     ok  github.com/tencentcloud/CubeSandbox/sdk/go  0.383s
sdk/node   Test Files 16 passed | 1 skipped     Tests 206 passed | 1 skipped
sdk/python 225 passed in 0.39s
```

CI gates checked locally:

- `gofmt -l .` in `sdk/go` — clean (`fmt-check`).
- `npx tsc --noEmit` in `sdk/node` — clean.
- All three suites pass (`sdk-test-check`).

## Risk / rollout

**This turns a silent no-op into an error.** Anyone who currently passes a non-ASCII domain in `allowOut`
without deny-all gets an `ApiError` where they previously got none. That is the point of the guard, and
Python users already saw this error — but it is a visible behaviour change for Go and Node callers on
upgrade. The remedy is the one the error message states: set `allowInternetAccess: false` or add
`0.0.0.0/0` to `denyOut`.

### DIFF
--- a/sdk/go/policy.go
+++ b/sdk/go/policy.go
@@ -6,6 +6,7 @@ package cubesandbox
 import (
 	"net"
 	"strings"
+	"unicode"
 )
 
 // L7 egress policy types — host/path/SNI matching, audit, credential
@@ -136,9 +137,13 @@ func isValidDNSDomainName(domain string) bool {
 			return false
 		}
 		for _, ch := range label {
-			if !(ch >= 'a' && ch <= 'z' || ch >= '0' && ch <= '9' || ch == '-') {
-				return false
+			if ch == '-' || ch >= 'a' && ch <= 'z' || ch >= '0' && ch <= '9' {
+				continue
 			}
+			if unicode.IsLetter(ch) || unicode.IsDigit(ch) || unicode.IsNumber(ch) {
+				continue
+			}
+			return false
 		}
 	}
 	return true
--- a/sdk/go/policy_domain_guard_test.go
+++ b/sdk/go/policy_domain_guard_test.go
@@ -1,0 +1,55 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package cubesandbox
+
+import "testing"
+
+func TestAllowOutDomainGuardFiresForUnicodeDomains(t *testing.T) {
+	domains := []string{
+		"example.com",
+		"exаmple.com",
+		"bücher.example",
+		"例子.com",
+		"café.example.com",
+		"xn--bcher-kva.example",
+		"*.café.example.com",
+	}
+
+	for _, d := range domains {
+		t.Run(d, func(t *testing.T) {
+			if !isDomainAllowOutTarget(d) {
+				t.Fatalf("isDomainAllowOutTarget(%q) = false, want true", d)
+			}
+			if err := validateAllowOutDomainsRequireDenyAll([]string{d}, nil, false); err == nil {
+				t.Fatalf("guard did not fire for %q with public egress enabled", d)
+			}
+			if err := validateAllowOutDomainsRequireDenyAll([]string{d}, []string{"0.0.0.0/0"}, false); err != nil {
+				t.Fatalf("guard fired for %q despite deny-all: %v", d, err)
+			}
+			if err := validateAllowOutDomainsRequireDenyAll([]string{d}, nil, true); err != nil {
+				t.Fatalf("guard fired for %q despite defaultDenyAll: %v", d, err)
+			}
+		})
+	}
+}
+
+func TestNonDomainTargetsStillNotTreatedAsDomains(t *testing.T) {
+	nonDomains := []string{
+		"10.0.0.1",
+		"10.0.0.0/8",
+		"::1",
+		"999.1.2.3",
+		"1.2.3.4",
+		"",
+		"has space",
+		"under_score.example.com",
+		"double**star.example.com",
+	}
+
+	for _, target := range nonDomains {
+		if isDomainAllowOutTarget(target) {
+			t.Errorf("isDomainAllowOutTarget(%q) = true, want false", target)
+		}
+	}
+}
--- a/sdk/node/src/policy.ts
+++ b/sdk/node/src/policy.ts
@@ -244,7 +244,7 @@ function isValidDnsDomainName(domain: string): boolean {
       label.length <= 63 &&
       !label.startsWith("-") &&
       !label.endsWith("-") &&
-      /^[a-zA-Z0-9-]+$/.test(label),
+      /^[\p{L}\p{N}0-9-]+$/u.test(label),
   );
 }
 
--- a/sdk/node/test/policy-domain-guard.test.ts
+++ b/sdk/node/test/policy-domain-guard.test.ts
@@ -1,0 +1,33 @@
+import { describe, expect, it } from "vitest";
+import { validateAllowOutDomainsRequireDenyAll } from "../src/policy";
+
+const UNICODE_DOMAINS = [
+  "example.com",
+  "exаmple.com",
+  "bücher.example",
+  "例子.com",
+  "café.example.com",
+  "xn--bcher-kva.example",
+  "*.café.example.com",
+];
+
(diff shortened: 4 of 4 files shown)


fslongjin:
**Half of the fix is client-side only — the authoritative server guard still has the identical ASCII-only bug.**

This PR correctly widens the *SDK* domain classifiers (it touches only `sdk/go/policy.go` + `sdk/node/src/policy.ts`; Python already used `isalnum`, so it needs no change). Merging it makes the SDKs raise a clear error for a non-ASCII / IDN / homograph `allow_out` entry instead of silently classifying it as a non-domain and skipping the deny-all guard — good, fail-closed, and no ASCII-input regression.

But the classifier this PR edits is only the SDK's local mirror. The authoritative enforcement of the allow_out deny-all contract lives in CubeAPI and still carries the same ASCII-only logic:

- `CubeAPI/src/services/mod.rs:69-81` — `is_valid_dns_domain_name` still accepts only `ch.is_ascii_lowercase() || ch.is_ascii_digit() || ch == '-'`. A non-ASCII target is still not classified as a domain, so `validate_allow_out_domains_require_deny_all` (`:20-41`) does not require a deny-all → **public egress silently stays open for direct API callers**.
- This is the *default endpoint the SDKs point at* (`defaultAPIURL = "http://127.0.0.1:3000"`, `sdk/go/config.go:14`; `"http://127.0.0.1:3000"`, `sdk/node/src/config.ts:87` — CubeAPI's bind port in `main.rs:42`), and the guard runs on sandbox create / network-update / template-create paths (`services/sandboxes.rs:196`, `:513`; `services/templates.rs:643`).
- Net effect after merge: a client using the fixed SDK gets a 400 for a Unicode allow_out entry, but the *same* request sent straight to CubeAPI (curl, an older SDK, or any non-SDK client) is still silently accepted without a deny-all. SDK and server now disagree about the same configuration.

Same-class ASCII-only copies that remain unfixed elsewhere:
- CubeMaster templatecenter `CubeMaster/pkg/templatecenter/request_validation.go:216-232` (template network validation).
- CubeNet data plane `CubeNet/cubevs/netpolicy.go:1055` (`IsAllowOutDomainTarget`) — and `:1012`, which *rejects* a Unicode allow_out domain outright. So even when a caller adds the deny-all the SDK instructs, the entry still cannot take effect end-to-end (fail-closed / error rather than a security hole, but "IDN supported end-to-end" doesn't hold).
- Cubelet L7 `Cubelet/network/runtime/policy_builder.go:381-407` (silently drops a non-ASCII SNI/Host) — the only server-side copy the PR description mentions.

Not blocking this client-side fix, but please register the server-side copies — at minimum CubeAPI `services/mod.rs:69-81`, the one the SDKs default to — as a known limitation / follow-up issue in the PR description before merging. As written, the description reads as if the guard hole is closed, which is only true on the SDK path.

PR #1477, opened 2026-08-21
sdk/python: use typing.List in classes that define a `list` method

### LINKED ISSUE #1476: [Bug Report] Classes with a `list` method shadow the builtin in their own annotations, so mypy cannot check those parameters
## Summary

Four classes in the Python SDK define a method named `list`. Inside a class body the name `list` resolves
to that method, so annotations written as `list[...]` in the same class refer to the method rather than the
builtin. mypy reports them as invalid types and silently stops checking those parameters.

## Environment

- Deployment mode: n/a — SDK
- Relevant component: sdk/python

## Steps to Reproduce

```
cd sdk/python && mypy --python-version 3.11 --ignore-missing-imports cubesandbox
```

## Expected Behavior

No `valid-type` errors; `list[...]` annotations resolve to the builtin and are type-checked.

## Actual Behavior

10 errors:

```
cubesandbox/_template.py:220: error: Function "cubesandbox._template.Template.list" is not valid as a type  [valid-type]
cubesandbox/_template.py:228: error: ... (7 total in _template.py)
cubesandbox/sandbox.py:374:   error: Function "cubesandbox.sandbox.Sandbox.list" is not valid as a type  [valid-type]
cubesandbox/sandbox.py:647:   error: ... (3 total in sandbox.py)
```

Each error means mypy could not resolve that annotation, so the parameter or return value is effectively
untyped — genuine type errors there go unreported.

## Scope

Classes defining a `list` method:

- `cubesandbox/_template.py:149` — `Template.list`
- `cubesandbox/sandbox.py:357` — `Sandbox.list`
- `cubesandbox/_volume.py:379` — `Volume.list`
- `cubesandbox/_filesystem.py:143` — `Filesystem.list`

The first two are the ones mypy currently flags; the other two have `list[...]` annotations in the same
class body and are the same latent problem.

## Note: runtime is NOT affected

Worth stating explicitly to keep the severity honest. All four modules use
`from __future__ import annotations`, and `typing.get_type_hints` evaluates annotations in **module**
globals, not the class namespace — so at runtime `list[...]` resolves to the builtin as intended:

```
get_type_hints(Template.build) -> OK
get_type_hints(Template.list)  -> OK
get_type_hints(Sandbox.list)   -> OK
```

This is a static-analysis defect, not a runtime one.

dwin-gharibi:
Closes #1476.

## Motivation

`Template`, `Sandbox`, `Volume` and `Filesystem` each define a method named `list`. Inside a class body the
name `list` binds to that method, so an annotation written `list[TemplateInfo]` in the same class refers to
the method, not the builtin. mypy reports `Function "...list" is not valid as a type` and, more importantly,
**stops type-checking those parameters** — 10 annotations across two files were effectively `Any`.

## What this changes

Replaces `list[...]` with `typing.List[...]` in the four affected modules and adds `List` to each
`from typing import ...` line:

| file | annotations rewritten |
|---|---|
| `cubesandbox/_template.py` | 11 |
| `cubesandbox/sandbox.py` | 6 |
| `cubesandbox/_volume.py` | 4 |
| `cubesandbox/_filesystem.py` | 2 |

`typing.List` is not shadowed by anything, so the annotations resolve for static analysers as well as at
runtime. The public API — including the `list` method names — is unchanged.

I fixed all four rather than only the two mypy currently flags: `_volume.py` and `_filesystem.py` have the
same construct and would start producing the same errors as soon as anything about their annotations
changed.

No comment changes.

## Correcting the original report

The issue I filed first claimed this broke `typing.get_type_hints` at runtime. **That was wrong**, and I
verified it rather than leaving it in:

```
MASTER get_type_hints(Template.build) -> OK
MASTER get_type_hints(Template.list)  -> OK
MASTER get_type_hints(Sandbox.list)   -> OK
```

All four modules use `from __future__ import annotations`, and `get_type_hints` evaluates against **module**
globals rather than the class namespace, so `list` resolves to the builtin at runtime. The defect is
static-analysis only, and the issue text has been corrected to say so. That also lowers the severity from
what I first suggested.

## Testing

The evidence is the mypy count, since the defect is a type-resolution one:

```
MASTER 'not valid as a type' count: 10
BRANCH 'not valid as a type' count:  0
```

No behavioural change, so no new tests. Existing suite and lint are unaffected:

```
$ python3 -m pytest tests -q
225 passed in 0.38s

$ ruff check --select F,E9 cubesandbox
Found 2 errors.        # both pre-existing F401 on master, identical count
```

CI gates checked locally:

- `pytest` — 225 passed (`sdk-test-check`).
- `ruff` — no new findings (the two `F401` unused-import errors are present on master too and are not
  touched here).

### DIFF
--- a/sdk/python/cubesandbox/_filesystem.py
+++ b/sdk/python/cubesandbox/_filesystem.py
@@ -6,7 +6,7 @@ from __future__ import annotations
 import json
 import logging
 import struct
-from typing import TYPE_CHECKING, Any, Iterator
+from typing import Any, Iterator, List, TYPE_CHECKING
 
 logger = logging.getLogger(__name__)
 
@@ -122,7 +122,7 @@ class Filesystem:
 
     def write_files(
         self,
-        files: list[tuple[str, str | bytes]],
+        files: List[tuple[str, str | bytes]],
         *,
         user: str | None = None,
     ) -> int:
@@ -140,7 +140,7 @@ class Filesystem:
                 ) from e
         return len(files)
 
-    def list(self, path: str) -> list[dict[str, Any]]:
+    def list(self, path: str) -> List[dict[str, Any]]:
         """List entries in a directory."""
         result = self._filesystem_rpc("ListDir", {"path": path})
         return result.get("entries", [])
--- a/sdk/python/cubesandbox/_template.py
+++ b/sdk/python/cubesandbox/_template.py
@@ -4,7 +4,7 @@
 from __future__ import annotations
 
 from dataclasses import dataclass, field
-from typing import Any, Dict
+from typing import Any, Dict, List
 
 import requests
 
@@ -41,7 +41,7 @@ class TemplateBuild:
     message: str = ""
     created_at: str = ""
     finished_at: str = ""
-    logs: list[str] = field(default_factory=list)
+    logs: List[str] = field(default_factory=list)
 
     @classmethod
     def from_dict(cls, data: dict) -> "TemplateBuild":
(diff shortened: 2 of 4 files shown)


wbzdssm:
Thank you for your PR. Could you please provide more details about the test results?

PR #1483, opened 2026-08-22
fix(cubeapi): accept E2B host-keyed network rules

### LINKED ISSUE #1482: [Bug Report] CubeAPI rejects the official E2B SDK's `network.rules` map shape, breaking drop-in SDK compatibility
## Summary

The README advertises "Native E2B SDK compatibility — swap one URL env var, zero business code
changes", and the capability table marks E2B SDK compatibility as "Drop-in". That holds for
everything I exercised except one thing: `SandboxOpts.network.rules`.

The official `e2b` SDK types `rules` as a **map keyed by host**; CubeAPI deserializes it as an
**array** of `EgressRule`. So any application that passes the SDK's documented `network` option
gets an HTTP 422 at sandbox creation and cannot proceed through the SDK at all.

## Environment

- CubeSandbox version / commit: `abe4a4f` (master)
- Host OS and kernel version: n/a
- Deployment mode: n/a
- Base image tag: `2026.16`
- Client: official `e2b` npm SDK `2.45.0`
- Relevant component: CubeAPI

## Steps to Reproduce

Create a sandbox through the official SDK using the `network` option exactly as documented at
https://docs.e2b.dev/sandbox/internet-access :

```js
network: {
  allowOut: ['api.example.com'],
  rules: {
    'api.example.com': [
      { transform: { headers: { 'X-Header': 'Content' } } }
    ],
  },
}
```

## Expected Behavior

The sandbox is created, since this is the shape the official SDK produces and the documented way to
express per-host egress rules.

## Actual Behavior

`POST /sandboxes` rejects the body:

```
HTTP 422
Failed to deserialize the JSON body into the target type: network.rules: invalid type: map
```

The field is declared as a list in `CubeAPI/src/models/mod.rs`:

```rust
/// L7 egress rules, evaluated first-match-wins in list order.
pub rules: Option<Vec<EgressRule>>,
```

with `EgressRule` being `{ name: String, match: EgressRuleMatch, action: EgressRuleAction }`.

## Impact

`network.rules` with `transform.headers` is E2B's documented mechanism for injecting credentials into
outbound HTTP(S) requests, so that the secret never has to be placed inside the sandbox as an
environment variable. With the shapes diverging, a caller using the official SDK's `network` option
cannot create a sandbox — the only workaround is to bypass the SDK and hand-build the request body in
CubeSandbox's own shape, which is exactly the "zero business code changes" property the README
promises.

## Scope — this is the only divergence I hit

To keep the severity honest: everything else I exercised through the official SDK worked. Sandbox
create, list and kill, plus the whole envd surface — filesystem read/write/list/rename/remove/exists,
`commands.run`, `pty.create` — 15 checks passing. `network.rules` is the single failing one.

## Suggested Fix (just a suggestion)

Accept both shapes, e.g. a `#[serde(untagged)]` enum over `Vec<EgressRule>` and
`HashMap<String, Vec<StandardRule>>`, expanding the map form into the existing internal
`Vec<EgressRule>`: each host key becomes `match: { host: <key> }` and
`action: { allow: true, inject: <from transform.headers> }`.

This would be purely additive — callers already using the array form are unaffected. CubeSandbox's
array form is strictly more expressive anyway (it can match on method/path/scheme, can deny, and
carries audit levels), so the map form is a subset that lowers cleanly into it.

zyl1121:
## Motivation

The E2B Python and npm SDKs send `network.rules` as a host-keyed map for per-host request transforms.

CubeAPI currently models this field as an ordered array, so an otherwise valid `Sandbox.create` request is rejected with:

```text
422: network.rules: invalid type: map, expected a sequence
```

This breaks E2B-compatible sandbox creation when SDK users configure network transform rules.

Fixes #1482.

## What Changed

* Accept E2B's host-keyed `network.rules` map in `SandboxNetworkConfig`.
* Convert each host key to `match.host`.
* Convert each non-empty `transform.headers` entry to `action.inject` in the existing internal egress-rule model.
* Ignore E2B no-op entries with no transform or no headers instead of creating an unintended allow-only rule.
* Reject unsupported E2B transform fields instead of silently discarding network intent CubeAPI cannot express.
* Preserve the existing array-shaped `network.rules` format for backward compatibility.
* Document both accepted wire shapes in OpenAPI and add regression coverage for map, array, no-op, schema, and complete `NewSandbox` requests.

## Validation

* Python `e2b==2.34.0` / `e2b-code-interpreter==2.8.1` created and deleted a sandbox with host-keyed `network.rules`.
* npm `e2b==2.45.0` created and deleted a sandbox with host-keyed `network.rules`.
* CubeMaster and Cubelet request payloads preserved the converted host match and injected header in the final sandbox creation request.
* Existing array-shaped rules created a sandbox successfully and remained present in the downstream request.
* Targeted CubeAPI regression tests passed, including E2B no-op handling and OpenAPI schema coverage.
* Full CubeAPI test suite passed with 126 tests.
* CubeAPI release build and generated OpenAPI export passed.

## Scope

This PR only adds CubeAPI request-shape compatibility for E2B-style `network.rules`.

It does not change network policy semantics, the internal egress-rule model, CubeMaster/Cubelet behavior, or SDK code.

### DIFF
--- a/CubeAPI/src/models/mod.rs
+++ b/CubeAPI/src/models/mod.rs
@@ -60,9 +60,109 @@ pub struct SandboxNetworkConfig {
     /// expanded to the requested sandbox port; envd traffic is exempt.
     #[serde(rename = "maskRequestHost", skip_serializing_if = "Option::is_none")]
     pub mask_request_host: Option<String>,
-    /// L7 egress rules, evaluated first-match-wins in list order.
-    #[serde(skip_serializing_if = "Option::is_none")]
+    /// L7 egress rules. Accepts CubeSandbox's ordered rule array and E2B's
+    /// host-keyed transform map for compatibility. Both are normalized to the
+    /// ordered internal rule model.
+    #[serde(
+        default,
+        deserialize_with = "deserialize_egress_rules",
+        skip_serializing_if = "Option::is_none"
+    )]
+    #[schema(value_type = Option<SandboxNetworkRulesInput>)]
     pub rules: Option<Vec<EgressRule>>,
+}
+
+/// Deserialize CubeSandbox's ordered rule list and E2B's host-keyed rule map
+/// into the same internal representation. E2B uses the map form for
+/// per-host request transforms, while CubeEgress evaluates an ordered list.
+fn deserialize_egress_rules<'de, D>(deserializer: D) -> Result<Option<Vec<EgressRule>>, D::Error>
+where
+    D: serde::Deserializer<'de>,
+{
+    let Some(input) = Option::<SandboxNetworkRulesInput>::deserialize(deserializer)? else {
+        return Ok(None);
+    };
+
+    let hosts = match input {
+        SandboxNetworkRulesInput::CubeSandbox(rules) => return Ok(Some(rules)),
+        SandboxNetworkRulesInput::E2B(hosts) => hosts,
+    };
+    let mut rules = Vec::new();
+
+    for (host, entries) in hosts {
+        if host.is_empty() {
+            return Err(serde::de::Error::custom(
+                "network.rules host keys must be non-empty strings",
+            ));
+        }
+
+        let injects: Vec<Vec<EgressRuleInject>> = entries
+            .into_iter()
+            .filter_map(|entry| {
+                entry.transform.and_then(|transform| {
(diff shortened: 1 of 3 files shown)


luzhixing12345:
LGTM. Test passed. A clean and simple fix.

PR #1485, opened 2026-08-22
fix(sdk): do not send a non-positive Connect-Timeout-Ms for a command

### LINKED ISSUE #1484: [Bug Report] commands.run forwards a non-positive timeoutMs to envd, hanging every command (Node & Python)
## Summary

`commands.run` forwards `timeoutMs` to envd as `Connect-Timeout-Ms` without checking that it is
positive. A non-positive value therefore reaches envd as an already-expired deadline, envd never
answers, and the call fails with a transport-level headers timeout rather than with anything that
says what happened.

Two values reach it that way in ordinary use:

- **`0`**, which is how the official `e2b` SDK spells "no deadline". Code ported under the
  "drop-in E2B SDK compatibility" claim passes it.
- **`NEVER_TIMEOUT`**, which this SDK exports itself. Its value is `-1`, and it is forwarded
  verbatim.

The same SDKs already get this right for a pty, and the Go SDK gets it right for both — so this
is one guard missing in one place, in two of the three clients.

| SDK | `commands` guard | `pty` guard |
| --- | --- | --- |
| Node | `if (idleMs !== undefined)` — `src/commands.ts:241` | `if (options.timeoutMs !== undefined && options.timeoutMs > 0)` — `src/pty.ts:315` |
| Python | `if timeout is not None:` — `cubesandbox/_commands.py:149` | `if timeout is not None and timeout > 0:` — `cubesandbox/_pty.py:232` |
| Go | `if timeout <= 0 { return }` — `envd.go:240` | same helper |

## Environment

- CubeSandbox commit: `981ccda4` (master)
- SDK: `@cubesandbox/sdk` 0.3.0, built from `sdk/node`
- Node.js 24
- Relevant component: SDK (Node, Python)

## Steps to Reproduce

Against any running sandbox:

```js
import { Sandbox, NEVER_TIMEOUT } from "@cubesandbox/sdk"

const sb = await Sandbox.connect(sandboxId)

await sb.commands.run("echo hi", { user: "root", timeoutMs: 0 })              // hangs
await sb.commands.run("echo hi", { user: "root", timeoutMs: NEVER_TIMEOUT })  // hangs
await sb.commands.run("echo hi", { user: "root", timeoutMs: 30000 })          // fine
await sb.commands.run("echo hi", { user: "root" })                            // fine
```

## Expected Behavior

A non-positive `timeoutMs` means "no deadline" and no `Connect-Timeout-Ms` header is sent — which
is what `pty` in the same SDK does, and what the Go SDK does for both.

`NEVER_TIMEOUT` in particular should never produce a hang: it is the constant this SDK exports for
saying that a thing does not time out.

## Actual Behavior

```
timeoutMs: 0              → fetch failed / UND_ERR_HEADERS_TIMEOUT
timeoutMs: NEVER_TIMEOUT  → fetch failed / UND_ERR_HEADERS_TIMEOUT
timeoutMs: 30000          → { stdout: "hi\n", stderr: "", exitCode: 0 }   ~22 ms
omitted                   → { stdout: "hi\n", stderr: "", exitCode: 0 }   ~21 ms
```

The failure surfaces as a transport error from the HTTP client, so nothing in it points at the
header that caused it.

## Impact

Every command in every sandbox hangs for the lifetime of the client's headers timeout. For an
application that runs a command per request, that is the whole application — and because the error
is a generic `fetch failed`, it reads as a network or proxy fault rather than as an argument.

`0` is the value E2B's SDK documents for "no timeout", so this is reachable by exactly the port the
README invites.

## Suggested Fix

Match `pty` in the same file's neighbour:

```diff
-    if (idleMs !== undefined) {
+    if (idleMs !== undefined && idleMs > 0) {
       headers["Connect-Timeout-Ms"] = String(Math.trunc(idleMs));
     }
```

and the equivalent in `cubesandbox/_commands.py:149`.

HuChundong:
Fixes #1484.

## What

`commands.run` forwarded `timeoutMs` to envd as `Connect-Timeout-Ms` without checking that it was
positive. envd reads that header as a hard wall-clock deadline, so a zero or negative one has
already passed: the request is never answered, and the caller waits until its HTTP client gives up
on headers and reports a transport error that says nothing about the cause.

Both non-positive values arrive in ordinary use:

- **`0`** is how the `e2b` SDK spells "no deadline", so it comes with exactly the port the README's
  drop-in compatibility invites.
- **`NEVER_TIMEOUT`** is how this SDK spells it, and its value is `-1` — a constant meaning "does
  not time out" whose effect was that nothing ever returned.

## Why this is a slip rather than a design choice

The same guard is already correct everywhere else it appears:

| | `commands` | `pty` |
| --- | --- | --- |
| Node | `if (idleMs !== undefined)` ← fixed here | `if (options.timeoutMs !== undefined && options.timeoutMs > 0)` |
| Python | `if timeout is not None:` ← fixed here | `if timeout is not None and timeout > 0:` |
| Go | `if timeout <= 0 { return }` | same helper |

## What is not changed

`Sandbox.create`'s own `timeout` field. `0` and `NEVER_TIMEOUT` are meaningful to CubeAPI there,
and `test_create_sends_explicit_zero_timeout` / `test_create_sends_never_timeout` still pass
unchanged.

## Tests

Node and Python both gain two cases: no `Connect-Timeout-Ms` header for `0` or `NEVER_TIMEOUT`, and
a header that is still sent for a positive value. Both new cases fail against the previous
behaviour — reverting the one-line guard turns them red and leaves the rest green.

```
node    216 passed | 2 skipped   tsc --noEmit clean
python  257 passed | 4 skipped
```

## Measured

Against a running sandbox, before the fix:

```
timeoutMs: 0              → fetch failed / UND_ERR_HEADERS_TIMEOUT
timeoutMs: NEVER_TIMEOUT  → fetch failed / UND_ERR_HEADERS_TIMEOUT
timeoutMs: 30000          → { stdout: "hi\n", exitCode: 0 }   ~22 ms
omitted                   → { stdout: "hi\n", exitCode: 0 }   ~21 ms
```

### DIFF
--- a/sdk/python/cubesandbox/_commands.py
+++ b/sdk/python/cubesandbox/_commands.py
@@ -146,8 +146,27 @@ class Commands:
             "Connect-Protocol-Version": CONNECT_PROTOCOL_VERSION,
             "Connect-Content-Encoding": "identity",
         }
-        if timeout is not None:
-            headers["Connect-Timeout-Ms"] = str(int(timeout * 1000))
+        # The timeout has two readers that spell "no deadline" differently, so
+        # it is normalised once here.
+        #
+        # envd reads Connect-Timeout-Ms as a hard wall-clock deadline, so a zero
+        # or negative one has already passed and the request never gets an
+        # answer -- it hangs until the HTTP client gives up on headers, and
+        # reports a transport error that says nothing about the cause. The
+        # header is therefore omitted rather than sent as zero.
+        #
+        # httpx reads the same number as a client-side deadline, where only None
+        # means "no deadline": 0 makes every socket operation time out at once,
+        # and -1 is rejected outright. Forwarding the raw value would trade the
+        # hang for an immediate failure, which is not what either value asks
+        # for.
+        #
+        # Both non-positive values arrive in ordinary use: 0 is how the e2b SDK
+        # spells "no deadline", and NEVER_TIMEOUT is how this one does. _pty.py
+        # and the Go SDK already guard this way.
+        effective_timeout = timeout if timeout is not None and timeout > 0 else None
+        if effective_timeout is not None:
+            headers["Connect-Timeout-Ms"] = str(int(effective_timeout * 1000))
         access_token = self._sandbox._data.get("envdAccessToken")
         if access_token:
             headers["X-Access-Token"] = access_token
@@ -159,7 +178,7 @@ class Commands:
             url,
             content=_encode_connect_envelope(json.dumps(payload).encode("utf-8")),
             headers=headers,
-            timeout=timeout,
+            timeout=effective_timeout,
         ) as resp:
             if resp.status_code >= 400:
                 detail = _http_error_detail(resp)
--- a/sdk/node/src/commands.ts
+++ b/sdk/node/src/commands.ts
@@ -238,7 +238,15 @@ export class Commands {
       ...this.sandbox.trafficTokenHeaders(),
       ...userHeaders(user),
     };
-    if (idleMs !== undefined) {
+    // Only a positive deadline is sent. envd reads Connect-Timeout-Ms as a hard
+    // wall-clock deadline, so a zero or negative one has already passed and the
+    // request never gets an answer — it hangs until the HTTP client gives up on
+    // headers, and reports a transport error that says nothing about the cause.
+    //
+    // Both non-positive values arrive in ordinary use: `0` is how the e2b SDK
+    // spells "no deadline", and NEVER_TIMEOUT is how this one does. `pty.ts`
+    // and the Go SDK already guard this way.
+    if (idleMs !== undefined && idleMs > 0) {
       headers["Connect-Timeout-Ms"] = String(Math.trunc(idleMs));
     }
     const accessToken = this.sandbox.envdAccessToken;
--- a/sdk/node/test/commands.test.ts
+++ b/sdk/node/test/commands.test.ts
@@ -7,7 +7,7 @@ import type { AddressInfo } from "node:net";
 import { afterAll, afterEach, beforeAll, describe, expect, it } from "vitest";
 
 import { Config } from "../src/config.js";
-import { Sandbox } from "../src/index.js";
+import { NEVER_TIMEOUT, Sandbox } from "../src/index.js";
 
 const SANDBOX_ID = "sb-test-001";
 const DOMAIN = "cube.app";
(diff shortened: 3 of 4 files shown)


sdk/python/cubesandbox/_commands.py
luzhixing12345: @HuChundong Hi, the Node.js changes LGTM. However, the Python fix appears incomplete: non-positive timeouts are no longer sent to envd, but they are still passed to httpx. timeout=0 fails immediately, while timeout=-1 raises a ValueError.

I think non-positive values should be normalized to None, for example:
```python
effective_timeout = timeout if timeout is not None and timeout > 0 else None
```
Then effective_timeout should be passed to httpx instead of the original timeout.

HuChundong: Thanks — you're right, and thanks for catching it. Fixed in ca207bcd, using the variable name you suggested.

`effective_timeout` is now computed once and both readers of it take the normalised value: the header is omitted, and httpx is given `None` instead of the raw `0`/`-1`. Without this the fix only moved the failure — from a hang to an immediate socket timeout (or a `ValueError` on httpx versions that reject a negative) — so neither value actually meant "no deadline" on that path.

The tests are updated too, because the existing ones could not have caught it: `httpx.MockTransport` answers synchronously and enforces no client-side timeout, so the value passed has no observable effect. They now record the timeout the client was handed (`_recording_client`) and assert `None` for `0`/`NEVER_TIMEOUT` and `30` for a positive one. Reverting the one-line source change turns both parametrized cases red, so they hold the behaviour rather than describing it.

Two places I deliberately did not touch:

- `_pty.py` — its httpx calls take a separate `request_timeout`, so the raw value never reaches a client there.
- `_run_with_e2b_connect` — `0` is the e2b client's own spelling of "no deadline", so forwarding it is correct on that path.

On the bot's other note (a positive timeout below one millisecond truncating to `Connect-Timeout-Ms: 0`): I looked for a way to reach it and could not find one, so I have left it out of this PR rather than widen it. Nothing in the SDK computes a timeout and passes it down — `commands.run` takes the caller's value directly, and the only `deadline = monotonic() + timeout` in the tree is `pause(wait=True)`'s polling loop, which never feeds a command. The parameter's unit also excludes it (`timeoutMs` is documented in milliseconds; the Python one is seconds), and where this SDK does reject absurd timeouts — `-2`, `NaN`, `Infinity` on `setTimeout` — it rejects them at the entry point before any request, which seems the right layer for that class of value rather than the header builder. Happy to send it as a separate PR if you think it is worth closing anyway.

Local runs: `sdk/python` 257 passed / 4 skipped, `sdk/node` 216 passed / 2 skipped (untouched, as a regression baseline). The pre-existing `ruff` and `prettier` complaints on these files are the same on the base commit, so I have not reformatted them here.

HuChundong: Two corrections to my previous comment, and an unrelated finding on the `e2b` path.

**The lockfile is gone.** `sdk/python/uv.lock` was staged by accident (a `git add -A` in my working tree) and was never meant to be part of this PR. Force-pushed without it; the PR is four files again, two of them from the first commit.

**On the `e2b` path, I was half right and should have checked before saying it.** I said `0` is that client's own spelling of "no deadline" and left the path alone. That much holds, but not for the reason I implied — it holds by falsiness, not by intent: `connectrpc`'s sync client does `timeout_ms=timeout_ms or self._timeout_ms` before `if timeout_ms is not None: headers["connect-timeout-ms"] = str(timeout_ms)`, so `0` falls through to the client default and no header is sent.

`NEVER_TIMEOUT` does not. `-1` is truthy, so it reaches the header as `connect-timeout-ms: -1`, which is the same class of value this PR exists to stop sending. So the bot's note is right about that half.

I have not fixed it here, because of a second thing I ran into while checking, which looks more significant than the timeout question:

**`_run_with_e2b_connect` does not match the current `e2b` signature.** It calls

```python
rpc.start(request, headers=..., timeout=timeout, request_timeout=...)
```

but in `e2b` 2.45.1 (today's release) `ProcessClient.start` is `start(self, request, *, headers=None, timeout_ms=None)`. There is no `timeout` and no `request_timeout`, so that call raises `TypeError` — and `run()` only catches `ImportError`, so it does not fall back; it propagates. `e2b` is not a declared dependency of this SDK (not in `dependencies`, not in the `dev` extra), so the path is exercised only on a machine that happens to have it installed, which is presumably why CI does not see this.

If that reading is right, then fixing the `-1` header on that path means first deciding what the call should be — and whether the path should keep a `timeout` parameter at all, given the units differ (`timeout` in seconds here, `timeout_ms` there). That is a bigger change than this PR, and it is your call rather than mine. Happy to open a separate issue with the above, or a PR if you tell me which shape you want.

For this PR the scope is unchanged: the fallback path, which is the one that actually runs by default.

luzhixing12345:
Thanks for your contribution, LGTM

PR #1487, opened 2026-08-23
fix(cubeproxy): correctly invalidate per-port route cache via route e…

zhouxianping:
…poch

The route cache in ngx.shared.local_cache invalidated only the fixed metadata keys, leaving dynamic per-port backend entries ({sid}:{port}:backend_*) to their TTL. After Resume rewrites a sandbox's backend, the shared meta_cached flag re-lit by one port's reload let the remaining ports keep serving stale per-port entries, so a multi-port sandbox routed un-reloaded ports to the old backend.

Embed a per-sandbox route epoch in the per-port keys ({sid}:{epoch}:{port}: backend_*) and bump {sid}:route_epoch in invalidate_sandbox before deleting the fixed keys, sealing every per-port entry without scanning the dict. The epoch key uses a 24h TTL refreshed on each bump so it always outlives the minute-TTL stale entries it seals, and is never deleted (deleting it would reset the epoch while stale entries could still be alive).

Also require every dependent field (backend ip/port, AllowPublicTraffic, TrafficAccessToken, MaskRequestHost) to be present before serving a cached hit. With the previous backend+flag-only gate, an LRU eviction or TTL skew that dropped the auth fields served a hit that skipped the traffic-token check for a restricted sandbox. And make meta_cached a real completion signal: written last on fill/refresh, read first on hit.

Tests: epoch bump/refresh/never-deleted and sealed-key unreachability; the multi-port stale-route regression (fails without the epoch); flag-without- data is a miss; evicting the token field forces a reload.

### DIFF
--- a/CubeProxy/lua/sandbox_backend.lua
+++ b/CubeProxy/lua/sandbox_backend.lua
@@ -121,6 +121,71 @@ local function load_sandbox_proxy_metadata(ins_id)
     return nil, nil
 end
 
+-- Read path. Returns host_ip, host_port, mask_request_host on a valid hit,
+-- nil on a miss. The completeness gate is the presence of EVERY served field
+-- (false is a valid negative-cache sentinel; only nil means missing): a
+-- partial fill or a field lost to TTL skew / LRU eviction leaves a nil and
+-- forces a reload instead of serving an incomplete entry or silently skipping
+-- the traffic-token check. There is no separate meta_cached flag — checking
+-- the fields directly makes one redundant.
+--
+-- The serving backend is computed from the raw route fields the fill mirrors
+-- for the whole sandbox (HostIP / SandboxIP / the per-port host-port mapping),
+-- resolved for the caller's host. A single fill writes every port's mapping at
+-- once, so after an invalidate all ports converge on fresh data from one
+-- reload — no per-port reload, no route epoch. On a hit the TTLs are refreshed.
+local function read_cached_backend(cache, ins_id, container_port, caller_host_ip, timeout)
+    -- The presence of every served field is the completeness gate; there is no
+    -- separate meta_cached flag. A partial fill or a field lost to TTL skew /
+    -- LRU eviction leaves a nil here and forces a reload instead of serving an
+    -- incomplete entry (or skipping the traffic-token check).
+    local target_host_ip = cache:get(ins_id .. ":HostIP")
+    local target_sandbox_ip = cache:get(ins_id .. ":SandboxIP")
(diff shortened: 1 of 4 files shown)


CubeProxy/lua/sandbox_backend.lua
```diff
@@ -121,6 +121,57 @@ local function load_sandbox_proxy_metadata(ins_id)
     return nil, nil
 end
 
+-- Per-port cache keys embed the route epoch so invalidate_sandbox can seal
+-- every dynamic per-port entry at once by bumping {sid}:route_epoch.
+local function backend_cache_keys(ins_id, epoch, container_port)
+    return string.format("%s:%s:%s:backend_ip", ins_id, epoch, container_port),
```
chenhengqi: We can simply invalidate the `backend_ip` (should be `HostIP` instead) key.

CubeProxy/lua/sandbox_backend.lua
```diff
@@ -121,6 +121,73 @@ local function load_sandbox_proxy_metadata(ins_id)
...
+    local target_sandbox_ip = cache:get(ins_id .. ":SandboxIP")
+    local mapped_port = cache:get(ins_id .. ":" .. container_port)
+    local allow_public = cache:get(ins_id .. ":AllowPublicTraffic")
+    local traffic_token = cache:get(ins_id .. ":TrafficAccessToken")
+    local mask_request_host = cache:get(ins_id .. ":MaskRequestHost")
+
+    if not (target_host_ip and target_sandbox_ip
+            and allow_public ~= nil and traffic_token ~= nil
+            and mask_request_host ~= nil) then
+        return nil
```
chenhengqi, line 152: Should be `return nil, nil, nil`?

CubeProxy/lua/sandbox_backend.lua
```diff
@@ -121,6 +121,73 @@ local function load_sandbox_proxy_metadata(ins_id)
...
+-- reload — no per-port reload, no route epoch. On a hit the TTLs are refreshed
+-- data-first, flag-last (same order as the fill path).
+local function read_cached_backend(cache, ins_id, container_port, caller_host_ip, timeout)
+    if not cache:get(ins_id .. ":meta_cached") then
+        return nil
+    end
+
+    local target_host_ip = cache:get(ins_id .. ":HostIP")
+    local target_sandbox_ip = cache:get(ins_id .. ":SandboxIP")
+    local mapped_port = cache:get(ins_id .. ":" .. container_port)
```
chenhengqi, line 144: Should return early if `mapped_port` is nil.

chenhengqi (approved):
LGTM.

PR #1488, opened 2026-08-23
fix(cubenet): scan-driven direct on-link neighbor resolution

zhouxianping:
The #1321 direct-mode on-link resolution built ARP requests in BPF and dropped the triggering packet (first-packet loss) plus a periodic sacrifice packet after each 5-minute revalidate. Replace it with a fib-driven + userspace-scanner design that never drops a packet and never fabricates ARP in the datapath.

The datapath (prepare_egress_l2) now only reports state and always forwards: on-link traffic uses a TTL-bounded cache of the last bpf_fib_lookup result (valid_until is written only on a fib success and never renewed on a hit, so a dead neighbour cannot freeze the cache); a fib failure invalidates the cache, uses a short negative TTL, and falls back to the gateway MAC so the packet is always forwarded. The in-BPF ARP construction, the from_world ARP learning, and the probe/DROP branches are removed, along with the now-unused probe/revalidate constants and struct arp_packet.

A userspace scanner (cubevs/direct_neigh_scanner.go) owns all scheduling: it walks direct_neigh on an interval, GCs idle entries, fires learning triggers with exponential backoff for unresolved destinations, and fires 16s keepalive triggers for learned ones (driving kernel NUD re-validation), rate-limited by a global token bucket. Triggers are plain UDP sends from the node NIC source IP; the learning event is the kernel's own ARP exchange. Init drops any stale direct_neigh pin so the resized value reloads cleanly, and the scanner starts only in direct mode.

### DIFF
--- a/CubeNet/docs/direct-neighbor-resolution.md
+++ b/CubeNet/docs/direct-neighbor-resolution.md
@@ -1,0 +1,101 @@
+# Direct-mode on-link neighbor resolution (scan-driven)
+
+In **direct mode** (no cube-router, `EgressRedirectFlags == 0`), a sandbox may
+egress to a destination that is on the node NIC's own L2 link (on-link). The
+datapath needs the destination's MAC. This document describes how that is
+resolved after the rework that followed PR #1321.
+
+## Design rule
+
+**The kernel neighbor table is the only source of truth for the MAC.** Neither
+the datapath nor userspace fabricates ARP packets, mirrors the neighbor table,
+or subscribes to netlink events. The split of responsibility is:
+
+- **Datapath** (`mvmtap.bpf.c` `prepare_egress_l2`): only *reports* state — it
+  runs `bpf_fib_lookup`, records the result, and *always forwards the packet*.
+- **Userspace scanner** (`cubevs/direct_neigh_scanner.go`): owns all
+  *scheduling* — when to trigger kernel neighbor learning, when to keepalive,
+  when to garbage-collect.
+
+## Datapath behavior (`prepare_egress_l2`)
+
+For an on-link destination, in order:
+
+1. **Positive cache hit** — `direct_neigh[daddr]` has a MAC and `now <
+   valid_until_ns`: use the cached MAC, no fib lookup. `valid_until_ns` is
+   *never* renewed on a hit (iron rule: a cache entry lives at most
+   `CACHE_TTL` before being re-checked against the kernel).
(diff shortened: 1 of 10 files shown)


CubeNet/src/mvmtap.bpf.c
```diff
@@ -120,80 +120,22 @@ static __always_inline bool direct_neighbor_is_zero(const struct direct_neighbor
...
+ *   - otherwise bpf_fib_lookup() refreshes the cache, and on failure the cache
+ *     is invalidated and the gateway MAC is used as a fallback.
+ *
+ * There is no packet drop or packet-into-ARP conversion: an unresolved on-link
+ * destination is forwarded to the gateway (zero loss on hairpin networks; a
+ * bounded blackhole elsewhere until the scanner's trigger makes the kernel ARP
+ * resolve and the next fib refresh caches the MAC). Always returns
+ * EGRESS_MAC_READY.
+ */
 static __always_inline int prepare_egress_l2(struct __sk_buff *skb,
```
chenhengqi: The return type can be changed to `void` now.

CubeNet/src/mvmtap.bpf.c
```diff
@@ -217,41 +157,65 @@ static __always_inline int prepare_egress_l2(struct __sk_buff *skb,
...
-		smac = (const union macaddr *)fib.smac;
-		dmac = (const union macaddr *)fib.dmac;
-		set_mac_pair(l2, smac->p1, smac->p2, dmac->p1, dmac->p2);
+	now = bpf_ktime_get_ns();
+	neighbor = bpf_map_lookup_elem(&direct_neigh, &daddr);
+	if (!neighbor) {
+		/* Untracked destination: only register for the scanner, never
+		 * disturb its scheduling fields. */
+		bpf_map_update_elem(&direct_neigh, &daddr, &pending, BPF_NOEXIST);
+		neighbor = bpf_map_lookup_elem(&direct_neigh, &daddr);
```
chenhengqi, line 168: This `bpf_map_lookup_elem()` seems unnecessary and we can merge these two `if (!neighbor)` branches.

CubeNet/docs/direct-neighbor-resolution.md
```diff
@@ -0,0 +1,101 @@
+# Direct-mode on-link neighbor resolution (scan-driven)
```
chenhengqi, line 1: I think this file should be dropped from this PR.

PR #1489, opened 2026-08-24
feat(s3): cross-node pause/snapshot restore with sandbox-owned volumes

ls-ggg:
## Summary
- Depends on #1409 (now on master): cubecow S3 engine / `export_snapshot` / `import_lvol`. This PR no longer carries that cubecow commit.
- Add Cubelet dual-handle storage (`kind=reflink` + `kind=s3`): S3 Pause / Commit / AppSnapshot upload package objects (rootfs / memory / metadata), Create fetches them on the target node, and metadata lives on a node-local cubecow disk rather than host paths.
- Persist `backend`, `remote_status`, and export uuids on Master. Cross-node Resume / FromSnap is allowed only when `backend=s3` and `remote_status=ready`. XFS stays local. Sandbox volumes are deleted with the sandbox; package snapshots are deleted with the template / pause package.

## Test plan
- [x] `make cubemaster` / `make cubelet`
- [x] `make fmt`
- [x] CI-equivalent unit tests: `cubedb-test`, `cubecow-test-native`, `cubelet-pkg-test`, `cube-api-test`, `shim-test`
- [x] Focused packages: `restoreplace`, `pausesnap`, `remotestatus`, Cubelet `storage` / `cubebox` S3 + rollback + remote tests
- [x] Two-node cluster combo: same-node pause/resume/snapshot/from-snap, plus cross-node Pause/Resume and FromSnap when `remote_status=ready`
- [ ] GitHub Actions fmt-check / unit-test-check on this PR

Assisted-by: Cursor:Composer


CubeDB/migrate/migrations/mysql/20260814143000_create_snapshot_tables.sql
```diff
@@ -0,0 +1,91 @@
...
+-- Snapshot + pause-snapshot tables (xfs|s3 backend, remote_status, export
+-- uuids) and CoW backend on the running-sandbox spec row. Template
+-- definition / replica tables are left unchanged.
+
+-- +goose NO TRANSACTION
+-- +goose Up
+
+CALL cubemaster_acquire_migration_lock('cubemaster_migration_20260814143000_create_snapshot_tables', 60);
+
+CREATE TABLE IF NOT EXISTS `t_cube_snapshot` (
```
fslongjin, line 13: These new tables replace the previous `TemplateDefinition`-based storage, but no backfill is performed from legacy rows (`kind=pause_snapshot` / `kind=snapshot` in `t_cube_template_definition`). After a rolling upgrade, already-paused sandboxes become invisible/unresumable (pause bindings are read only from `t_cube_pause_snapshot`) and pre-existing snapshots become invisible, undeletable and unrollbackable (all snapshot ops read only `t_cube_snapshot`). The Down migration drops these tables, so a downgrade also loses new bindings. Please backfill in the migration (or document a mandatory drain before upgrade).

CubeMaster/pkg/restoreplace/placement.go
```diff
@@ -0,0 +1,335 @@
...
+	rows, err := queryHostFactCandidatesFn(ctx, origin.CPUIDHash, origin.HostKernelRelease, false)
+	if err != nil {
+		return nil, err
+	}
+	out := make([]compatibleCandidate, 0, len(rows))
+	for _, row := range rows {
+		if row == nil || row.HostFacts == nil {
+			continue
+		}
+		out = append(out, compatibleCandidate{NodeID: row.NodeID, NodeIP: row.HostIP})
```
fslongjin, line 287: `defaultListCompatible` appends every candidate with `HostFacts != nil` but never checks `HostFacts.KVMModuleTaint`; `nodemeta.QueryHostFactCandidates` explicitly defers that absolute gate to the caller. This contradicts `templatecenter/restore_compat.go`, which treats a tainted target as a hard blocker. A cross-node pause-resume / from-snapshot create can silently land on a host running a forced/out-of-tree/staging `kvm.ko`. Suggest skipping candidates with a non-empty `KVMModuleTaint`.

CubeMaster/pkg/templatecenter/snapshot_reconciler.go
```diff
@@ -251,27 +270,27 @@ func reconcileSnapshotReplicaPresence(ctx context.Context) error {
...
-					_ = UpsertReplica(ctx, def.TemplateID, model.InstanceType, failedReplicaStatus(model, msg))
-					_ = updateDefinitionFields(ctx, def.TemplateID, map[string]any{
+					msg := fmt.Sprintf("snapshot %s missing from local catalog on node %s", rec.SnapshotID, firstNonEmpty(replica.NodeID, hostIP))
+					_ = UpsertReplica(ctx, rec.SnapshotID, model.InstanceType, failedReplicaStatus(model, msg))
+					_ = updateSnapshotFields(ctx, rec.SnapshotID, map[string]any{
 						"status":     StatusFailed,
 						"last_error": msg,
 					})
 				}
 			case cubeleterrorcode.ErrorCode_PreConditionFailed:
```
fslongjin, line 287: `reconcileSnapshotReplicaPresence` applies XFS-local presence semantics to shared-S3 snapshots: on a single `PreConditionFailed` (catalog miss or transient `requireS3Cow` error) from one node's probe it runs `CleanupTemplate(s3)` and stamps `status=FAILED`. The Cubelet's S3 cleanup falls back to deterministic object refs and deletes the cluster-shared cubecow objects, which other nodes can still restore from. One node's transient S3 hiccup can therefore destroy shared snapshot data, and the FAILED stamp blocks `RollbackSandboxToSnapshot` (requires `status=Ready`). Consider skipping S3 snapshots here (presence is already tracked by `remote_status` + the `remotestatus` poller).

CubeMaster/pkg/service/sandbox/id_resolve.go
```diff
@@ -44,7 +45,7 @@ func ResolveSandboxID(ctx context.Context, input string) (string, error) {
 		return "", err
 	}
 
-	entries := collectClusterSandboxIDs(ctx)
+	entries := append(collectClusterSandboxIDs(ctx), collectPausedSandboxIDs(ctx)...)
```
fslongjin, line 48: `entries` concatenates cluster-scan IDs then paused IDs, so `cacheCollectedSandboxEntries` lets a pause binding overwrite the fresher node-scan location. After a cross-node resume whose binding-delete failed, the stale binding (still pointing at the origin) wins on a cold cache (e.g. Master restart), and a later Destroy/snapshot by ID targets the origin, leaking the live sandbox. Consider skipping the paused entry when the cluster scan already reported the same SandboxID.

CubeMaster/pkg/restoreplace/placement.go
```diff
@@ -0,0 +1,335 @@
...
+			seen[key] = struct{}{}
+			scope = append(scope, key)
+		}
+	}
+	if len(scope) == 0 {
+		return nil, fmt.Errorf("no compatible node for cross-node restore of %s (cpuid_hash/kernel match)",
+			strings.TrimSpace(in.SnapshotID))
+	}
+
+	selected, err := selectNodeFn(ctx, in.InstanceType, cloneReqResForCrossNode(in.ReqRes, scope), scope)
```
fslongjin, line 161: `Decide` builds the candidate `scope` from cpuid/kernel-compatible nodes but never re-checks that `selectNodeFn`'s pick is inside it — `defaultSelectNode` even drops the `scope` argument and `scheduler.Select`'s `BackoffSelect` path ignores `TemplateNodeScope`. When the compatible candidates are all rejected by resource filters, or the locality filter is disabled, an arbitrary incompatible host can be selected for cross-node restore with no error surfaced. Verify the selected node is in `scope` and error otherwise.

Cubelet/services/cubebox/appsnapshot.go
```diff
@@ -230,36 +240,38 @@ func (s *service) AppSnapshot(ctx context.Context, req *cubebox.AppSnapshotReque
...
+	layout.resetTmpDir()
+	if err := layout.prepareWork(ctx); err != nil {
+		stepLog.Errorf("Failed to create snapshot dir: %v", err)
+		rsp.Ret.RetCode = errorcode.ErrorCode_Unknown
+		rsp.Ret.RetMsg = fmt.Sprintf("failed to create snapshot dir: %v", err)
+		return rsp, nil
 	}
 
-	memoryObject, err = storage.CreateMemoryVolume(ctx, templateID, memorySizeBytes)
+	memoryObject, err = storage.CreateMemoryVolumeFor(ctx, backend, templateID, memorySizeBytes)
```
fslongjin, line 275: Both `CreateMemoryVolumeFor` error branches return without calling `cleanupSnapshotObjects()`, which is exactly what unmounts/deletes the S3 metadata volume that `prepareWork` just mounted and persisted (`ReleaseS3MetadataVolume`). A transient failure leaves an orphaned `s3-meta-<templateID>` mount/volume that is only reclaimed if Master later issues CleanupTemplate. Route every post-`prepareWork` error return through the same cleanup.

Cubelet/plugins/cbri/cubeboxcbri/cubebox.go
```diff
@@ -186,12 +188,51 @@ func (e *cubeboxInstancePlugin) CreateSandbox(ctx context.Context, flowOpts *wor
...
-			snapSpecPath = paths.Spec
+			if err := storage.EnsureShimSpecDirLink(paths.Base, paths.ResDir); err != nil {
+				return nil, ret.Err(errorcode.ErrorCode_AppSnapshotNotExist, err.Error())
+			}
+			backend := pauseResumeCatalogBackend(flowOpts)
+			if flowOpts.IsPauseResume() {
+				// Resume: clone sealed metadata snap → RW volume, then mount
+				// (MountS3MetadataAt never activates the package snap for IO).
+				metaMount := filepath.Join(paths.Base, storage.SnapshotMetadataDir)
+				if _, err := mountRestoreMetadata(ctx, flowOpts, backend, templateID, metaMount); err != nil {
```
fslongjin: Callers discard the path `mountRestoreMetadata` returns and instead use `paths.Base` derived from the run-template home. When the original template is present on the target node, the cross-node branch (`snapshotRestoreRawPath`) returns the template's home, so the shim restores the sandbox's imported memory but reads `{templateHome}/.../metadata.json` (the template's `app_snapshot_container_id`, or an empty dir after Finalize) instead of the sandbox's imported metadata. Derive `snapBasePath`/`snapSpecPath` from the imported metadata dir whenever `CrossNodeSandboxImport != nil`.

fslongjin (commented):
> **Review method note:** Because the GitHub CI diff for this PR is too large to fetch, I performed this code review locally using Claude Code. The comments below were written against the full PR diff and the working tree at head `b7cfa720`.

This PR adds an S3-backed CoW storage path with cross-node pause/snapshot resume, sandbox-owned volumes, a `remote_status`/`export_uuids` sync protocol, a Cubelet snapshot status RPC, and the matching Master-side pieces (placement in `restoreplace`, a `remotestatus` reconciler, new pause/snapshot tables, and CubeAPI query parameters). Overall the change is well-structured: the cross-node gate (`backend=s3` + `remote_status=ready`) is enforced from DB-persisted state rather than client input, proto/DB additions are backward compatible, the new subsystems ship with unit tests, and the CubeShim and Cubelet cgo/FFI layers verify against the native cubecow header. The items below are what I think deserve attention before merge.

## P1 - should fix / must answer before merge

1. **No data backfill for the new snapshot/pause tables** (migration + `pausesnap`/`templatecenter`). The new `t_cube_pause_snapshot` / `t_cube_snapshot` tables are created empty; legacy rows that lived in `t_cube_template_definition` (`kind=pause_snapshot` / `kind=snapshot`) are never copied over, and every management path now reads only the new tables. After a rolling upgrade, already-paused sandboxes become invisible and unresumable, and pre-existing snapshots become invisible, undeletable and unrollbackable. The Down migration drops the new tables, so a downgrade loses bindings created post-upgrade. Please backfill in the migration (or document a mandatory drain before upgrade).
2. **Cross-node placement omits the `kvm_module_taint` gate** (`restoreplace/placement.go`). `defaultListCompatible` never checks `HostFacts.KVMModuleTaint`, and `nodemeta.QueryHostFactCandidates` explicitly defers that gate to the caller. This contradicts `templatecenter/restore_compat.go`, which treats a tainted target as a hard blocker. A cross-node pause-resume / from-snapshot create can silently land on a host running a forced/out-of-tree `kvm.ko`. Skip candidates with a non-empty taint.
3. **S3 presence reconciler can destroy cluster-shared snapshot data on a transient probe failure** (`templatecenter/snapshot_reconciler.go`). A single `PreConditionFailed` from one node's catalog probe triggers `CleanupTemplate(s3)`, whose deterministic fallback deletes the shared cubecow objects, and stamps `status=FAILED`, which blocks `RollbackSandboxToSnapshot`. Skip S3 snapshots in the presence reconcile (their presence is already tracked via `remote_status` + the `remotestatus` poller).

## P2 - should fix

4. **Stale pause binding can poison the sandbox ID cache** (`sandbox/id_resolve.go`). Paused IDs are cached after the cluster scan, so a stale binding (origin node still recorded after a cross-node resume whose binding-delete failed) overwrites the fresher node-scan location on a cold cache, and a later Destroy/snapshot by ID targets the origin and leaks the live sandbox.
5. **Scheduler result is never verified against the compatible scope** (`restoreplace/placement.go`). `Decide` builds the candidate `scope` but does not check that `selectNodeFn`'s pick is inside it; `defaultSelectNode` drops the `scope` argument and `BackoffSelect` ignores `TemplateNodeScope`, so at capacity or with the locality filter disabled an arbitrary incompatible host can be selected.
6. **S3 metadata volume leaks on AppSnapshot failure** (`Cubelet/services/cubebox/appsnapshot.go`). Both `CreateMemoryVolumeFor` error branches return without calling `cleanupSnapshotObjects()`, leaving an orphaned `s3-meta-<templateID>` mount/volume that is only reclaimed if Master later issues CleanupTemplate.
7. **Cross-node restore can read the wrong metadata base** (`Cubelet/plugins/cbri/cubeboxcbri/cubebox.go`). Callers discard the path returned by `mountRestoreMetadata` and use the run-template home instead; when the original template is present on the target node the shim restores the sandbox's imported memory but reads the template's `metadata.json` (wrong container identity, or an empty dir after Finalize).
8. **Invalid backend is silently swallowed on direct sandbox create** (`sandboxspec/store.go`). The normalization error is dropped, so the spec is persisted with a poisoned `backend` while the column stays empty and the sandbox is silently created as xfs; the invalid value surfaces later as a confusing `unsupported backend` Pause failure. Template create validates at the boundary; direct sandbox create should too.
9. **CubeAPI `backend` filter is parsed but never applied** (`CubeAPI/src/models/mod.rs`). `ListSnapshotsQuery.backend` is documented as a filter but is not forwarded through `SnapshotService::list`/`CubeMasterClient`, so `GET /snapshots?backend=s3` silently returns snapshots of every backend.
10. **XFS resume regression when the origin is absent from the node cache** (`sandbox/sandbox_resume_pause.go`). Every resume now routes through `restoreplace.Decide`, which requires the origin to be in `localcache` and schedulable; after a Master restart or when the origin is cordoned, a previously-resumable XFS pause now hard-fails instead of falling back to `rec.NodeIP`.

## P3 - nits

11. **`remotestatus` reconciler has no retry/grace or timeout**: the first terminal probe is written as final (`failed` on any transient Cubelet/S3 read error, no re-probe), and an unreachable origin leaves rows `inprogress` forever. Both permanently block cross-node resume (gated on `remote_status=ready`).
12. **Duplicate-key detection is MySQL-only** (`templatecenter/snapshot_table.go`): the `"UNIQUE constraint"` needle never matches Postgres' lowercase `unique constraint` message, so a duplicate insert on Postgres returns the raw DB error instead of `ErrDuplicateTemplate`.
13. **PID-reuse race in destroy** (`Cubelet/services/cubebox/destroy_shim_wait.go`): `WaitProcessGone` only probes `kill(pid,0)`, so a recycled pid can hang the wait until the context deadline and skip the storage Destroy step, leaving the sandbox stuck in removal.
14. **Pause catalog unrecoverable after a Cubelet restart** (`cubeboxcbri/cubebox.go`): the S3 catalog probe only matches `tpl-<snap>-rootfs`, not the sandbox-owned `sb-<id>-rootfs-*` of a pause package, so resume silently falls back to the template path and mounts the pause metadata onto the shared template mount point.
15. **Paused sandboxes bypass list pagination** (`sandbox/sandbox_list.go`): pause bindings are appended after the node-window pagination, so the same paused sandbox appears on every page and `Size` no longer matches the returned count.
16. **`snapshot create --backend` CLI flag is a no-op** (`cmd/cubemastercli/.../snapshot.go`): the server deliberately uses the sandbox's persisted backend, so `--backend s3` on an xfs-backed sandbox silently produces an xfs snapshot.

## Verified clean (not flagged)

- Cross-node gating is genuinely enforced from DB-persisted backend/`remote_status`; client-supplied backends cannot flip placement.
- Proto field additions and the new snapshot RPC are wire-compatible; generated `.pb.go` matches the proto; the CubeShim guest-mount-restore logic and the cubecow cgo ABI verify against the native header.
- New packages are all wired (reconciler start, placement call sites, snapshot service registration); no orphaned config/features found.
- Migration helper functions exist and both MySQL/Postgres suites execute cleanly; the new tables match the Go models.

## Process notes

- This is a major user-facing feature (new `--backend` flags, `backend`/`remote_status` API fields, `[cow.s3]` Cubelet config) with no doc updates to the bilingual snapshot guide; please add or track them.
- No e2e coverage exercises the S3/cross-node path; also note the default xfs resume no longer deletes the local pause snapshot on disk (behavior change the existing pause/resume e2e may depend on).
- Minor: several new comments/proto docs use full-width Unicode punctuation (`／`, `｜`, `＋`, `§`); suggest ASCII to keep the English-comment gate happy.

CubeMaster/pkg/sandboxspec/store.go
```diff
@@ -108,6 +108,19 @@ func doPut(ctx context.Context, sandboxID string, req *sandboxtypes.CreateCubeSa
 	if canonical.Annotations != nil {
 		templateID = strings.TrimSpace(canonical.Annotations[constants.CubeAnnotationAppSnapshotTemplateID])
 	}
+	backend := ""
+	if normalized, ok, err := constants.OptionalSnapshotBackend(canonical.Backend); err == nil && ok {
```
fslongjin, line 112: `doPut` drops the `OptionalSnapshotBackend` normalization error and persists the raw invalid value into `request_json` while the `backend` column stays empty — the two become inconsistent and the sandbox is silently created as xfs. The poisoned value surfaces later: `pauseSandbox` reads it via `sandboxspec.Get` and Pause fails with `unsupported backend`, a confusing delayed failure. Template create validates/normalizes the backend at the boundary; the direct sandbox create path should too.

CubeAPI/src/models/mod.rs
```diff
@@ -410,6 +423,8 @@ pub struct ListSnapshotsQuery {
     /// Pagination cursor from previous response header x-next-token.
     #[serde(rename = "nextToken")]
     pub next_token: Option<String>,
+    /// Filter by CoW backend (xfs | s3).
+    pub backend: Option<String>,
```
fslongjin, line 427: The new `ListSnapshotsQuery.backend` field is documented as a filter but never read: `list_snapshots` forwards only `sandbox_id/limit/next_token`, `SnapshotService::list` builds a request without a `backend` field, and `CubeMasterClient::list_snapshots` doesn't pass it. A client calling `GET /snapshots?backend=s3` silently receives snapshots of every backend. Thread the parameter through (CubeMaster's own `GET /cube/snapshot` already supports `backend` filtering).

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
```diff
@@ -309,15 +330,40 @@ func resumeFromPauseSnapshot(ctx context.Context, req *types.UpdateRequest, host
...
 	instanceType := strings.TrimSpace(req.InstanceType)
+	if instanceType == "" {
+		instanceType = rec.InstanceType
+	}
 	if instanceType == "" {
 		instanceType = cubebox.InstanceType_cubebox.String()
 	}
 
+	createReq := loadResumeSandboxSpec(ctx, req.SandboxID)
+	placement, err := decidePauseResumePlacementFn(ctx, pauseResumePlacementInput(rec, instanceType, createReq))
```
fslongjin: Every resume (including local XFS) now routes through `restoreplace.Decide`, which requires the origin to be present in `localcache` and to pass `scheduler.Select`. After a Master restart (empty in-memory node cache) or when the origin is cordoned/under-resourced, a previously-resumable XFS pause now hard-fails with `SelectNodesFailed` instead of falling back to `rec.NodeIP`. Consider falling back to the old direct-target behavior for non-cross-node resumes when the origin can't be resolved.

CubeMaster/pkg/remotestatus/reconciler.go
```diff
@@ -0,0 +1,170 @@
...
+	})
+	if err != nil || st == nil {
+		log.G(ctx).Warnf("remote_status query %s %s on %s failed: %v", item.kind, snapshotID, hostIP, err)
+		return
+	}
+	next, ok := terminalRemoteStatus(st.GetState(), st.GetRemoteReady())
+	if !ok {
+		return
+	}
+	res := db.WithContext(ctx).Table(item.table).
```
fslongjin, line 141: The reconciler treats the first terminal probe as final and never re-probes `failed` rows; conversely there is no timeout for `inprogress` rows whose origin is unreachable (the status query fails and the row stays `inprogress` forever). Both cases permanently block cross-node resume (gated on `remote_status=ready`) with no operator-visible recovery short of delete-and-recreate. Consider a grace/backoff before committing `failed`, and an expiry for `inprogress` rows.

CubeMaster/pkg/templatecenter/snapshot_table.go
```diff
@@ -0,0 +1,181 @@
...
+	rec.Version = version
+	rec.RequestJSON = string(payload)
+	if rec.Status == "" {
+		rec.Status = StatusCreating
+	}
+	if rec.Backend == "" {
+		rec.Backend = constants.SnapshotBackendXFS
+	}
+	if err := tx.Table(constants.SnapshotTableName).Create(rec).Error; err != nil {
+		if strings.Contains(err.Error(), "1062") || strings.Contains(err.Error(), "Duplicate entry") || strings.Contains(err.Error(), "UNIQUE constraint") {
```
fslongjin: `createSnapshotTx` detects duplicates via `1062` / `Duplicate entry` / `UNIQUE constraint` — but PostgreSQL's message is lowercase `unique constraint`, so a duplicate insert on Postgres returns the raw DB error instead of `ErrDuplicateTemplate`. Add a lowercase match (or check the driver error code) to keep duplicate semantics on both dialects.

Cubelet/services/cubebox/destroy_shim_wait.go
```diff
@@ -0,0 +1,171 @@
...
+// Storage Destroy runs in the next workflow step and deletes S3 volumes;
+// doing that while the shim still holds the NVMe path produces host I/O
+// after delete_lvol.
+func waitSandboxRuntimeGone(ctx context.Context, sandboxID string, pids []int) error {
+	if len(pids) == 0 {
+		return nil
+	}
+	start := time.Now()
+	for _, pid := range pids {
+		if err := utils.WaitProcessGone(ctx, pid); err != nil {
```
fslongjin, line 160: `collectSandboxRuntimePIDs` snapshots host pids before kill, then `waitSandboxRuntimeGone` probes each with `WaitProcessGone` (`kill(pid,0)` only). If the kernel recycles a shim pid to an unrelated long-lived process before the wait runs, the wait blocks on that process until the context deadline and the storage Destroy step is skipped, leaving the sandbox stuck in removal. Bound the wait with its own short timeout and/or verify `/proc/<pid>/comm` still matches the shim/vmm before treating the pid as the runtime to wait on.

Cubelet/plugins/cbri/cubeboxcbri/cubebox.go
```diff
@@ -499,6 +544,109 @@ func resolveTemplateComponentPath(localTemplate *templatetypes.LocalRunTemplate,
...
+		return mountPath, storage.MountS3MetadataAt(ctx, backend, packageID, mountPath)
+	}
+	sandboxID := strings.TrimSpace(flowOpts.GetSandboxID())
+	if sandboxID == "" {
+		sandboxID = packageID
+	}
+	return mountPath, storage.CloneS3MetadataFromParent(ctx, backend, packageID, sandboxID, mountPath)
+}
+
+func pauseCatalogMetaDir(ctx context.Context, snapID, preferred string) string {
```
fslongjin, line 648: After a Cubelet restart, `pauseCatalogMetaDir` can't recover an S3 pause catalog: the `deriveS3CatalogEntry` probe only matches `tpl-<snap>-rootfs`, but a pause package's rootfs is the sandbox-owned `sb-<id>-rootfs-*`, so it returns empty and `snapshotRestoreRawPath` silently falls back to the template path — mounting the pause package's metadata onto the shared template's metadata mount point. On a catalog miss, derive the pause home from `SnapshotHome(backend, SnapshotKindPause, snapID)` instead of falling back to the template.

CubeMaster/pkg/service/sandbox/sandbox_list.go
```diff
@@ -112,9 +115,132 @@ func ListSandbox(ctx context.Context, req *types.ListCubeSandboxReq) (rsp *types
 		}
 		return rsp.Data[i].SandboxID < rsp.Data[j].SandboxID
 	})
+	enrichSandboxListBackends(ctx, rsp.Data)
+	mergePauseBindings(ctx, req, rsp)
```
fslongjin, line 119: `mergePauseBindings` appends pause rows after the `(StartIdx, Size)` node-window pagination and the sort, so paused sandboxes parked on nodes outside the requested window appear on every page and `rsp.Size` no longer matches `len(rsp.Data)`. Apply the pause bindings within the requested node scope when paginating (and adjust `Size`/dedupe across pages).

CubeMaster/cmd/cubemastercli/commands/cubebox/snapshot.go
```diff
@@ -132,6 +136,7 @@ var SnapshotCreateCommand = cli.Command{
 	Flags: []cli.Flag{
 		cli.StringFlag{Name: "sandbox-id", Usage: "sandbox id to snapshot"},
 		cli.StringFlag{Name: "display-name", Usage: "snapshot display name"},
+		cli.StringFlag{Name: "backend", Usage: "storage backend: xfs or s3 (default xfs)"},
```
fslongjin: `snapshot create --backend s3` is forwarded to `POST /cube/snapshot`, but the server (`SubmitSandboxSnapshot`) deliberately ignores the client-supplied backend and always uses the sandbox's persisted backend. On an xfs-backed sandbox an operator running `--backend s3` silently gets an xfs snapshot. Drop the flag or surface the effective backend after creation.

ls-ggg:
On the process notes in the review summary — not changing in this PR:

- User-facing bilingual snapshot/S3 guide was intentionally left out of this change. Tracking docs separately.
- S3 / cross-node coverage is cluster combo testing, not a GitHub Actions e2e job. Default XFS resume no longer deleting the local pause snapshot on disk is also intentional (pause-package lifetime is not tied to sandbox destroy).
- Full-width punctuation in a few comments is a nit; not blocking merge.

CubeMaster/pkg/remotestatus/reconciler.go
ls-ggg: Not changing this in this PR.

A transport / `queryStatus` error only logs and leaves the row `inprogress` — it does not stamp `failed`. Cubelet returning `failed` is written once on purpose (export really failed, including the ghost-uuid case). An unreachable origin staying `inprogress` is a known gap; we can add an expiry later, but it is not a merge blocker here.

Cubelet/services/cubebox/destroy_shim_wait.go
ls-ggg: Not changing this in this PR.

`WaitProcessGone` exists so Destroy does not tear down volumes while the shim/VMM is still reading them. A recycled pid blocking the wait is theoretically possible; on Linux that reuse is rare enough that we are not adding a `/proc/<pid>/comm` check here.

Cubelet/plugins/cbri/cubeboxcbri/cubebox.go
ls-ggg: Not changing this. The catalog derive path already matches how a sealed pause package is named.

`deriveS3CatalogEntry` probes the package-id object (`tpl-<snap>-rootfs`). That is the sealed pause/snapshot rootfs. `sb-<id>-rootfs-*` is the live sandbox volume after create/resume, not the pause package. Same-node pause → Cubelet restart → resume already works on this path.

CubeMaster/pkg/service/sandbox/sandbox_list.go
ls-ggg: Intentional, not changing.

`--index/--size` only windows the node scan. A paused sandbox has no shim, so `t_cube_pause_snapshot` is the source of truth and is merged in full. `--hostid` still applies, so `list --hostid X --quiet --delete` cannot touch a sandbox parked on another node. Paging pause bindings with the node window would hide paused sandboxes that the operator needs to see.

CubeMaster/pkg/restoreplace/placement.go
ls-ggg: Not changing this for now.

Cross-node placement stays on `cpuid_hash` + `host_kernel_release` only. `KVMModuleTaint` is not a restoreplace gate in this PR; we are not skipping tainted candidates here.

CubeMaster/pkg/service/sandbox/id_resolve.go
ls-ggg: Not changing this.

After a successful Resume the Master pause binding is deleted. A leftover row is not a valid state (only a crash or a failed `pausesnap.Delete` after Create). ID resolution should not paper over that. If the binding delete fails we already heal it on the next Pause when the sandbox is RUNNING; we are not going to let a stale pause row overwrite the node-scan cache as if that were normal.

CubeMaster/pkg/sandboxspec/store.go
ls-ggg: Not changing this.

`backend` is chosen at template create and later stages inherit it from the template / snapshot / pause row. A sandbox Create that sends an invalid `backend` is API misuse, not a supported path. We are not adding a second validation gate on `sandboxspec.Put`.

CubeAPI/src/models/mod.rs
ls-ggg: Not changing this.

`GET /snapshots` is not going to filter by `backend`. The query field can stay unused for now; clients should list and inspect `backend` on each item if they care.

CubeDB/migrate/migrations/mysql/20260814143000_create_snapshot_tables.sql
ls-ggg: Fixed.

Added a best-effort goose backfill (`20260824120000_backfill_snapshot_tables`) that copies `kind=snapshot` / `kind=pause_snapshot` lookup keys (`snapshot_id`, `origin_sandbox_id` / `sandbox_id`, `status`, `backend`). Missing tables/columns or insert errors are swallowed so Master/install still starts. goose_db_version records the version so later restarts do not run it again.

CubeMaster/pkg/templatecenter/snapshot_reconciler.go
ls-ggg: Fixed.

`reconcileSnapshotReplicaPresence` now skips S3 rows. S3 existence stays on `remote_status` + the remotestatus poller. A node catalog miss after Finalize is normal and must not `CleanupTemplate(s3)` or stamp FAILED.

CubeMaster/pkg/restoreplace/placement.go
ls-ggg: Fixed.

After cross-node `Select`, `Decide` rejects the pick if it is outside the cpuid/kernel `scope`. The scheduler is unchanged.

Cubelet/services/cubebox/appsnapshot.go
ls-ggg: Fixed.

Failures after `prepareWork`, including `CreateMemoryVolumeFor`, now call `cleanupSnapshotObjects()` so the S3 metadata volume is released.

Cubelet/plugins/cbri/cubeboxcbri/cubebox.go
ls-ggg: Fixed.

Cross-node restore now uses the directory `mountRestoreMetadata` returns for `snapBasePath` / `snapSpecPath`, not the template `paths.Base`.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
ls-ggg: Fixed.

Resume only calls `restoreplace.Decide` when S3 is `remote=ready` and the spec is not pinned to origin. XFS / not-ready S3 / host-mount use the pause row `NodeIP`/`NodeID` and do not consult the heartbeat cache.

CubeMaster/pkg/templatecenter/snapshot_table.go
ls-ggg: Fixed.

Duplicate-key detection is case-insensitive and covers PostgreSQL `unique constraint` / `duplicate key` as well as MySQL 1062.

CubeMaster/cmd/cubemastercli/commands/cubebox/snapshot.go
ls-ggg: Fixed.

`cubemastercli snapshot create --backend` is removed. The server already ignores a client backend and uses the sandbox spec.

ls-ggg:
To get this large PR validated in more test environments as soon as possible, I’ll go ahead and merge it.

There are a few known issues already, and we may uncover more during broader testing—I’ve logged them offline and will follow up with smaller, targeted PRs to address them

fslongjin:
## Post-merge adversarial review

Posting a follow-up review after merge. The PR introduces a large S3-backed cross-node pause/snapshot path. CI is green, and several prior review items were genuinely fixed in `271be770` — S3 replica-presence skip, the `scope` check in `Decide`, the cross-node metadata path, the XFS resume fallback, case-insensitive duplicate-key detection, and the removal of the no-op `snapshot create --backend` flag. The items below are what I still believe fall short of the project merge bar and should be addressed in follow-up PRs. Inline comments carry the details.

### Blocking — data / control plane

1. **Backfill is best-effort and leaves resumed sandboxes unreachable** — `CubeDB/migrate/migrations/{mysql,postgres}/20260824120000_backfill_snapshot_tables.sql`. Any SQL error is swallowed and goose still records the version, so a failed run is never retried. More importantly, the backfill does not copy `node_ip` (available on the replica table) or `instance_type` (present on `t_cube_template_definition`). `sandbox_update.go` resolves the resume host via `cache → proxy → resolvePauseHostIP`, and `resolvePauseHostIP` only accepts a non-empty `rec.NodeIP`. After a Master restart with an empty cache and an expired proxy map, Resume on a pre-upgrade paused sandbox returns `sandbox not found` *before* it ever reaches the `node_id` lookup. Empty `instance_type` also makes the row match every `pausesnap.List` instance-type filter.
2. **Cross-node Resume returns success when origin tombstone cleanup fails** — `CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go:455-462`. `DropOriginTombstone` failure is only logged, then `pausesnap.Delete` still runs and the request returns success. The origin keeps reporting PAUSED; `id_resolve` writes pause bindings after the node scan, so a later Destroy by ID can be routed to the origin and leak the live sandbox on the target. The code comment describes this exact consequence, but the implementation does not fail the request.
3. **Cubelet `Status` maps a transient store error to terminal `failed`** — `Cubelet/services/snapshot/service.go:94-98`. When `SnapshotUploadStatus` errors (S3 store not ready, s3lvol still in its 5s init retry), the RPC still succeeds with `State=failed`. The `remotestatus` poller writes `failed` once and never re-probes. Cross-node is gated on `remote_status=ready`, so a Cubelet restart during the s3lvol init window can permanently block cross-node resume/restore with no recovery path.
4. **E2B `network.rules` host-keyed compat removed in this PR** — `CubeAPI/src/models/mod.rs`. `deserialize_egress_rules` / `SandboxNetworkRulesInput` / `E2BNetworkRule` and their tests are deleted, and `openapi.yml` drops the schema. Clients still sending the host-keyed map shape will now fail deserialization. This is unrelated to the PR title and is a visible regression of an existing compat surface; please revert in a separate PR or confirm in writing that no caller uses this shape anymore.

### Should fix

5. **No `s3lvol` / `[cow.s3]` wiring in k8s / terraform.** `deploy/kubernetes/` and `deploy/one-click/terraform/` have zero references to `s3lvol` or `[cow.s3]`, while `Cubelet/storage/plugin.go` defaults to `/var/run/s3lvol.sock` and always starts `startS3CowInitLoop`. The headline cross-node path cannot be enabled on the official deployments, and nodes without s3lvol log an init failure every 5s. Mixed-version fleets also have no node-capability filter for S3.
6. **`restoreplace` does not enforce `KVMModuleTaint`** — `CubeMaster/pkg/restoreplace/placement.go:274-289`. `defaultListCompatible` filters only on `cpuid_hash` + `host_kernel_release`; `QueryHostFactCandidates` explicitly defers the taint gate to the caller, and `templatecenter/restore_compat.go` treats a non-empty taint as a hard blocker. Cross-node Resume / from-snapshot can land on a host running a forced/out-of-tree `kvm.ko`.
7. **`importOne` reuses a present object as success** — `Cubelet/storage/s3_sandbox_import.go:192-200`. The comment above the function states that "the object is here now" must NOT be accepted as proof that a failed import succeeded, because the leftover is an empty s3lvol shell that mounts as a corrupt filesystem later. The implementation does exactly that: `present → skip ImportLvol → Activate`. A retried Create can mount an empty disk.
8. **OpenAPI and a dead query filter** — `openapi.yml`. `CreateSnapshotRequest` / `SnapshotInfo` / `SnapshotListItem` / `RollbackRequest` do not list `backend` / `remoteStatus` even though the Rust models expose them. `GET /snapshots?backend=` is documented as a filter but `handlers/snapshots.rs` never forwards it to `SnapshotService::list`. Either thread it through or remove the field.
9. **CI green is false-green for this feature.** `tests/unittest/run.sh` (and `cubelet-pkg-test`) only run `go test -short ./pkg/...` for Cubelet; the S3 import/metadata/layout tests live under `Cubelet/storage` and are not in the gate. `pauseSandbox` / `resumeFromPauseSnapshot` orchestration (CrossNode annotation, `DropOriginTombstone`) is essentially untested, and the from-snap test does not assert `CubeAnnotationSnapshotCrossNode` — the only annotation Cubelet checks before importing. The CubeDB migration test only asserts columns exist, not that legacy rows are copied.

### Already correct (not flagging)

- Cross-node gate is enforced from DB-persisted `backend` + `remote_status`; a client cannot flip an xfs sandbox into s3-ready.
- S3 replica-presence reconciler skips S3 rows, so a single node catalog miss no longer deletes cluster-shared objects.
- XFS / not-ready / host-mount Resume no longer consults the heartbeat cache.
- AppSnapshot failure paths release the S3 metadata volume; cross-node metadata uses the `mountRestoreMetadata` return path.
- `remotestatus` uses `WHERE remote_status=inprogress` to avoid concurrent overwrites.
- S3 credentials are not logged; object keys use Master-generated `snap-` ids with `ValidateSafeID`.

### Suggested follow-up order

1. Hot-fix backfill: surface failures / make it re-entrant; backfill `node_ip` (join replica) and `instance_type`; add a seed→migrate→Resume test; provide a one-off compensation script for clusters that already ran the empty backfill.
2. Hot-fix Resume tombstone: fail the request (or retry to success) when origin cleanup fails; add orchestration tests that do not mock `Decide`.
3. Hot-fix Status state machine: return `pending`/`running` (or a Status RPC error) when the store is not ready; only treat an explicit export failure as terminal; add an `inprogress` expiry.
4. Restore E2B `network.rules` compat in a separate PR.
5. Wire `Cubelet/storage` tests into `tests/unittest/run.sh` / `cubelet-pkg-test`; add `prepareCrossNodeRestore`, `importOne` empty-shell, and `CubeAnnotationSnapshotCrossNode` coverage.
6. Regenerate `openapi.yml` with `backend` / `remoteStatus`; implement or remove `GET /snapshots?backend=`.
7. Surface `s3lvol` and `[cow.s3]` in Helm / terraform; skip the init loop on nodes without s3lvol.
8. kvm taint gate, `distribution_scope` semantics, snapshot storage水位, DeleteSnapshot in-use, Pause `Complete` failure state.

Happy to help with any of the follow-up PRs.

CubeDB/migrate/migrations/mysql/20260824120000_backfill_snapshot_tables.sql
```diff
@@ -0,0 +1,108 @@
...
+CALL cubemaster_acquire_migration_lock('cubemaster_migration_20260824120000_backfill_snapshot_tables', 60);
+
+-- +goose StatementBegin
+DROP PROCEDURE IF EXISTS cubemaster_try_backfill_snap_20260824;
+-- +goose StatementEnd
+
+-- +goose StatementBegin
+CREATE PROCEDURE cubemaster_try_backfill_snap_20260824()
+BEGIN
+  DECLARE CONTINUE HANDLER FOR SQLEXCEPTION BEGIN END;
```
fslongjin, line 26: **Backfill swallows all errors and still records the goose version.**

`CONTINUE HANDLER FOR SQLEXCEPTION BEGIN END;` (and the Postgres `EXCEPTION WHEN OTHERS` counterpart) wraps the INSERTs, so any failure — missing column, type drift, duplicate — is silently dropped and goose still marks this version as applied. A failed run is never retried on restart, and the legacy `kind=snapshot` / `kind=pause_snapshot` rows stay invisible to the new management paths (which only read `t_cube_snapshot` / `t_cube_pause_snapshot`).

On top of that, the backfill copies only lookup keys. `node_ip` is available on the replica table but is not joined in, and `instance_type` exists on `t_cube_template_definition` but is not copied. Concretely:

- `resolvePauseHostIP` only accepts a non-empty `rec.NodeIP`. After a Master restart with an empty cache and an expired proxy map, Resume on a pre-upgrade paused sandbox returns `sandbox not found` *before* it reaches the `node_id` fallback in `resumeFromPauseSnapshot`.
- `pausesnap.List` filters with `instance_type = ? OR instance_type = ''`, so a backfilled row with empty `instance_type` shows up in every instance-type list.

Suggested: make the migration fail (or re-entrant) on error; join the replica table to backfill `node_ip`; copy `instance_type`; add a test that seeds legacy rows, runs migrate, and asserts the new tables contain them with the fields Resume actually needs. Clusters that already ran the empty backfill will need a one-off compensation script.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
```diff
@@ -375,34 +441,55 @@ func resumeFromPauseSnapshot(ctx context.Context, req *types.UpdateRequest, host
...
+
+	// The sandbox now lives on the target, but the origin still holds the
+	// PAUSED CubeBox row Pause left behind. Same-node Resume replaces that
+	// row as part of Create; cross-node has to say so explicitly, or the
+	// origin keeps reporting a paused sandbox that no longer exists there —
+	// which shows up as a duplicate row in List and, worse, lets ID
+	// resolution send a later Destroy to the origin and leak the live one.
+	if placement != nil && placement.CrossNode {
+		if origin := strings.TrimSpace(rec.NodeIP); origin != "" && origin != targetIP {
+			if err := pausesnap.DropOriginTombstone(ctx, req.RequestID, req.SandboxID, origin); err != nil {
```
fslongjin, line 457: **Resume returns success even when the origin tombstone cleanup fails.**

The comment above this block correctly describes the consequence: the origin keeps reporting a paused sandbox that no longer exists there, which shows up as a duplicate List row and lets ID resolution route a later Destroy to the origin, leaking the live sandbox on the target. But the implementation only `Errorf`s the `DropOriginTombstone` error and then falls through to `pausesnap.Delete` + a success return.

`id_resolve` appends `collectPausedSandboxIDs` after the node scan and writes both into the cache, so a stale PAUSED row on the origin wins on a cold cache. The earlier review reply ("a leftover row is not a valid state, ID resolution should not paper over it") is inconsistent with this code: Resume itself is what creates that invalid state and then reports success.

Suggested: fail the Resume (or retry `DropOriginTombstone` to success) before deleting the binding and returning success. Add an orchestration test that does not mock `Decide` and covers the cross-node tombstone path.

Cubelet/services/snapshot/service.go
```diff
@@ -0,0 +1,132 @@
...
+		State:      cow.RemoteStatePending,
+	}
+	if backend != cow.BackendS3 {
+		rsp.Message = "xfs has no remote upload"
+		return rsp
+	}
+
+	st, err := storage.SnapshotUploadStatus(ctx, backend, snapshotID)
+	if err != nil {
+		rsp.State = cow.RemoteStateFailed
```
fslongjin, line 96: **A transient store error is mapped to terminal `failed`.**

When `storage.SnapshotUploadStatus` returns an error — S3 store not ready, s3lvol still in its 5s init retry, `requireCowStoreFor` failure — this RPC still succeeds with `State=failed`. The `remotestatus` poller writes `failed` once and never re-probes it. Cross-node Resume / from-snapshot is gated on `remote_status=ready`, so a Cubelet restart during the s3lvol init window can permanently block cross-node for that package with no automatic recovery.

The earlier reply to review item E assumed "Cubelet returning `failed` means export really failed". That holds for an explicit export failure, but not for this branch — here `failed` means "I could not query the store", which is transient.

Suggested: return `pending`/`running` (or return a Status RPC error so the poller keeps the row `inprogress`) when the store is not ready; reserve `failed` for a confirmed export failure (e.g. ghost uuid). Add a test for "RPC succeeds but State=failed from a transient store error" — the current tests only cover the RPC-error path.

Cubelet/storage/s3_sandbox_import.go
```diff
@@ -0,0 +1,209 @@
...
+	}
+	importer, ok := store.engine.(cowVolumeImporter)
+	if !ok || importer == nil {
+		return "", fmt.Errorf("cubecow engine cannot import %s %s", role, name)
+	}
+	present, err := cowObjectPresent(store.GetVolumeInfo(ctx, name))
+	if err != nil {
+		return "", fmt.Errorf("lookup %s %s: %w", role, name, err)
+	}
+	if !present {
```
fslongjin, line 196: **`importOne` reuses a present object as success — the comment says it must not.**

The comment above this function explicitly states that "the object is here now" must NOT be accepted as proof that a failed import succeeded, because the leftover is the empty shell s3lvol leaves behind, which mounts as a corrupt filesystem several steps later. The implementation does exactly the opposite: when `present` is true it skips `ImportLvol` and goes straight to `ActivateVolume`. A retried Create that finds its own leftover from a half-failed import will mount an empty/corrupt disk instead of re-importing.

Suggested: when the object is present but the prior import did not complete, delete it and re-import (or validate the export uuid / content before reusing). At minimum, do not treat `present` alone as success for a sandbox-owned import — the package-object race tolerance in `S3Cow.fetchOne` does not apply here, as the comment itself notes.

CubeMaster/pkg/restoreplace/placement.go
```diff
@@ -0,0 +1,339 @@
...
+	rows, err := queryHostFactCandidatesFn(ctx, origin.CPUIDHash, origin.HostKernelRelease, false)
+	if err != nil {
+		return nil, err
+	}
+	out := make([]compatibleCandidate, 0, len(rows))
+	for _, row := range rows {
+		if row == nil || row.HostFacts == nil {
+			continue
+		}
+		out = append(out, compatibleCandidate{NodeID: row.NodeID, NodeIP: row.HostIP})
```
fslongjin, line 287: **`defaultListCompatible` does not enforce the `KVMModuleTaint` gate.**

Candidates are filtered only on `cpuid_hash` + `host_kernel_release`. `nodemeta.QueryHostFactCandidates` explicitly defers the taint gate to the caller, and `templatecenter/restore_compat.go` treats a non-empty `KVMModuleTaint` as a hard blocker (the diagnostic `listCompatibleNodes` API does enforce it). The real scheduling path here does not, so a cross-node Resume / from-snapshot Create can land on a host running a forced/out-of-tree/staging `kvm.ko`, contradicting the documented "absolute kvm_module_taint security gate".

The earlier review reply ("we are not skipping tainted candidates here") confirms this is intentional, but it leaves the production restore path weaker than the diagnostic path. Suggested: apply the same taint check `buildHostFactDimensions` uses, or document why restoreplace is allowed to ignore a gate that `restore_compat` treats as absolute.

CubeDB/migrate/migrations/mysql/20260824120000_backfill_snapshot_tables.sql
ls-ggg: Won't change the backfill shape.

The migration is only there so operators can list and delete leftover pause/snapshot rows after upgrade. Those pre-upgrade pause sandboxes and snapshots are not usable on the new path; we are not trying to make Resume/FromSnap work for them.

Swallowing SQL errors so Master/install still starts stays as designed. Clusters that already applied this goose version do not need a compensation script for Resume.

Cubelet/services/snapshot/service.go
ls-ggg: Fixed: Status only reports `failed` when s3lvol itself says the export failed (`export_status` FAILED/ERROR, or an in-process `export_snapshot` RPC failure already recorded locally).

Store-not-ready, volume lookup errors, empty/`NONE`/`INPROGRESS`, and a uuid with no status leave `State` unset (or non-terminal). The remotestatus poller then keeps `remote_status=inprogress` and retries.

Cubelet/storage/s3_sandbox_import.go
ls-ggg: Agreed this is a real hole (empty s3lvol shell vs `present` treated as success). It is a low-probability retry path; we will fix it in a follow-up rather than here.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
ls-ggg: Won't change. The sandbox is already running on the target; failing Resume makes the caller retry, and that retry is 130400. Origin tombstone cleanup failure stays logged.

CubeMaster/pkg/restoreplace/placement.go
ls-ggg: Won't change. Cross-node restore stays `cpuid_hash` + `host_kernel_release` only; we are not applying the KVM taint gate here.

ls-ggg:
Thanks for the post-merge pass. Follow-up positions:

**1. Backfill** — won't expand it. It exists so operators can list/delete leftover pause and snapshot rows after upgrade. Those pre-upgrade pause sandboxes / snapshots are not usable on the new path; we are not trying to Resume them. Error swallowing so Master/install still starts stays as designed.

**2. Origin tombstone** — won't fail Resume. The sandbox is already running on the target; a failed RPC makes the caller retry (130400). Cleanup failure stays logged.

**3. Status → terminal `failed`** — fixing. Cubelet Status only reports `failed` when s3lvol itself says the export failed. Store-not-ready / lookup errors / empty status leave State unset so the poller keeps `inprogress`.

**4. E2B `network.rules`** — not removed. #1489 only added `backend` / `remoteStatus` on snapshot types. `deserialize_egress_rules` from #1483 is still on master.

**5. Helm/terraform s3lvol** — later. Known deploy gap, not a hotfix.

**6. KVM taint** — won't change. Cross-node restore stays `cpuid_hash` + `host_kernel_release`.

**7. `importOne` present=success** — agreed, low-probability retry path; follow-up later.

**8. OpenAPI `?backend=`** — list will not filter by backend. Docs/OpenAPI cleanup later.

**9. CI / storage tests in the gate** — later, not blocking this cycle.

PR #1490, opened 2026-08-24
feat(CubeS3lvol): add a remote COW based on S3

changpe1:
Add CubeS3lvol (s3lvol) as a standalone CubeSandbox component: a remote
Copy-On-Write block storage engine on S3/COS, fronted by an NVMe/TCP
target. Volume data lives in object storage; the local disk is used only
for the WAL and the metadata journal. Wiring it into the volume storage
path (cubecow / Cubelet / Volume API / k8s / terraform) is intentionally
out of scope for this change; the unit is delivered, built and installed
on its own.

What it does
  - Volume lifecycle over JSON-RPC (scripts/s3lvol_rpc.py):
    create_lvol / delete_lvol / resize_lvol (grow-only) / get_lvstores
  - Snapshots and clones: read-only snapshots, clones from snapshots;
    delete_lvol also clears the volume's activation record so a restart
    does not try to restore it
  - Cross-node snapshot export: rcow_export_snapshot returns an export
    uuid immediately, work continues in the background;
    rcow_get_snapshot_status reports INPROGRESS / DONE / NONE with an
    on-the-spot (never cached) deletable flag
  - Activation over NVMf: rcow_active_bdev / rcow_get_bdev (poll until
    device_path is non-empty) / rcow_deactive_bdev (idempotent)
  - Thin provisioning: capacity is fixed at first create
    (RCOW_CAPACITY_GB); unused space costs nothing
  - Host I/O without DMA: local WAL disk via bdev_aio syscalls, frontend
    via nvmf_tcp sockets, S3 traffic via CRT mallocs -- hence
    RCOW_NO_HUGE=1 is the default and hugepages are optional, never
    required

Design points
  - Write path: journal + WAL first, data flushed to S3 in background;
    read path serves from a local overlay cache; checkpoint advances the
    journal/WAL epoch. Object-store compaction today relies on WAL
    segment reclaim and deletion-path cleanup; a full orphan-object GC
    (lib/s3bsdev/s3_gc.c) is planned but not implemented yet
    (s3_list_objects() returns -ENOTSUP)
  - Crash-safe restart: attach replays the WAL tail; a stale S3 owner
    marker from a dead local pid is auto-forced. Recovery never
    disconnects the NVMf initiator: the target rebuilds the same
    NQN/NSID grid and nvme_tcp reconnects transparently, so in-flight
    business I/O is unaffected (only latency spikes)
  - Idempotent lifecycle: rcow_start.sh refuses to double-start and
    connect_all skips already-connected controllers; rcow_stop.sh
    reverses the start order (disconnect -> flush/unload -> SIGTERM)
  - Deterministic state layout: bstore.json decides attach vs create;
    wal_bdev.img size fixes the journal/WAL layout, so it is per-machine
    state that must never be copied or recreated after the fact
  - Credentials from cos.cfg are passed only via the target process
    environment -- never argv, never logs
  - Self-contained release: bin/s3lvol_tgt links SPDK/DPDK/AWS CRT
    statically; targets only need nvme-cli, python3, openssl 1.1,
    libuuid, libaio, libnuma
  - Tests: C unit tests under test/integration (with an SPDK thread
    stub), shell dataplane tests under test/dataplane (activation,
    control, dataplane, export, fs, guards, recovery, selfimport,
    snapshot)


fslongjin:
hello, this PR needs to add intergration with k8s. @changpe1

CubeS3lvol/test/integration/Makefile
fslongjin, line 1: needs to add a github CI workflow

.github/workflows/cube-s3lvol-test.yml
fslongjin, line 1: 请问这个能在builder image里面跑吗？这样也能确保开发者可以在本地使用builder image去跑测例。（类似其他的几个组件那样）

ls-ggg:
Cubelet accesses s3lvol​ via a Unix domain socket—are there any installation-time considerations we need to handle?

.github/workflows/cube-s3lvol-test.yml
changpe1: builder image是ubuntu 20.04吧，在https://github.com/TencentCloud/CubeSandbox/pull/1490/changes/63bc88be418e656af4f0cec7ad1522b876261073有增加build image支持

changpe1:
> Cubelet accesses s3lvol​ via a Unix domain socket—are there any installation-time considerations we need to handle?

I was thinking to add this check in CI using minIO as S3 storage, for now the s3lvol service is disabled by default in this PR.

deploy/one-click/install.sh
```diff
@@ -1543,12 +1593,30 @@ mkdir -p \
...
   /data/cube-shared \
   /data/cube-shared/volume \
   /data/shared
 
+# CubeS3lvol (s3lvol): per-machine WAL image. The package never ships it --
+# its size fixes the journal/WAL layout and it must not be copied between
+# hosts or recreated after the first activation (rcow_start.sh refuses to
+# self-create an empty image, so this is install-time only). Create only
+# when absent so an upgrade never clobbers existing WAL state.
+if [[ -f "${INSTALL_PREFIX}/CubeS3lvol/bin/s3lvol_tgt" ]]; then
```
fslongjin: This gate looks at “is `s3lvol_tgt` in the package?”, not “did the user turn s3lvol on?”.

`ONE_CLICK_ENABLE_S3LVOL` defaults to `0`, and that switch correctly skips starting the unit. This block does not look at the switch. The documented builder path always ships the binary, so a default install still tries to create `/data/cubelet/rcow/wal_bdev.img` (~33GiB).

Two problems follow:

1. A user who never enabled s3lvol still gets this local WAL file (including on control nodes, which copy the full package).
2. Nothing creates `/data/cubelet/rcow/` first. `truncate` will not create parent directories; on failure this `die`s and aborts the whole CubeSandbox install.

Please create the WAL only when `ONE_CLICK_ENABLE_S3LVOL=1`, and `mkdir -p` the parent directory before `truncate`. If the switch is off, this path should be a no-op.

changpe1: Done.

deploy/one-click/install.sh
```diff
@@ -1619,12 +1669,35 @@ mkdir -p \
...
+# CubeS3lvol (s3lvol): per-machine WAL image. The package never ships it --
+# its size fixes the journal/WAL layout and it must not be copied between
+# hosts or recreated after the first activation (rcow_start.sh refuses to
+# self-create an empty image, so this is install-time only). Create only
+# when absent so an upgrade never clobbers existing WAL state, and only
+# when the feature is enabled: a default install (ONE_CLICK_ENABLE_S3LVOL=0)
+# must not leave a ~33 GiB sparse file behind on nodes that never run s3lvol
+# (including control nodes, which copy the full package).
+if [[ "${ONE_CLICK_ENABLE_S3LVOL}" == "1" &&
+      -f "${INSTALL_PREFIX}/CubeS3lvol/bin/s3lvol_tgt" ]]; then
```
fslongjin, line 1693: **Blocking: WAL creation breaks default install.** Two problems:
1. `/data/cubelet/rcow` is never `mkdir`'d anywhere in one-click; `truncate` does not create parent dirs, so a fresh install `die`s here.
2. The guard is `[[ -f .../s3lvol_tgt ]]`, not `ONE_CLICK_ENABLE_S3LVOL == 1`. The official builder always ships the binary, so with the switch defaulting to `0`, a brand-new install fails before any role service starts.

Suggested:
```bash
if [[ "${ONE_CLICK_ENABLE_S3LVOL}" == "1" ]] \
   && [[ -f "${INSTALL_PREFIX}/CubeS3lvol/bin/s3lvol_tgt" ]]; then
  wal_img=/data/cubelet/rcow/wal_bdev.img
  mkdir -p "$(dirname "${wal_img}")"
  ...
fi
```
Also: the WAL image size is a permanent per-node contract (it backs an SPDK lvolstore). Re-running `install.sh` with a different `RCOW_WAL_MB` on an existing node corrupts the lvolstore. Guard with `[[ -f $wal_img ]] || truncate ...` and document the immutability.

deploy/one-click/install.sh
```diff
@@ -1827,12 +1900,26 @@ else
   remove_env_kv "${RUNTIME_ENV_FILE}" "CUBE_S3_S3FS_EXTRA_OPTS"
 fi
 
+# CubeS3lvol (s3lvol) runtime env: persist the resolved defaults (mirroring
+# rcow_common.sh) so the systemd unit picks them up via EnvironmentFile
+# without re-deriving them, and so `down.sh` / upgrade knows the intent.
+upsert_env_kv "${RUNTIME_ENV_FILE}" "ONE_CLICK_ENABLE_S3LVOL" "${ONE_CLICK_ENABLE_S3LVOL}"
+upsert_env_kv "${RUNTIME_ENV_FILE}" "RCOW_WAL_MB" "${RCOW_WAL_MB}"
```
fslongjin, line 1912: **Should fix: upgrade may overwrite merged `RCOW_*` values.** `install.sh` resolves `RCOW_*` with `${VAR:-default}` *before* the merged env is applied, then `upsert_env_kv` writes those resolved values back into `.one-click.env`. A user who only edited the old `.one-click.env` (without re-exporting) can have their settings replaced by installer defaults on upgrade. Resolve these vars *after* sourcing the merged env, or skip `upsert_env_kv` when the value equals the default.

deploy/one-click/lib/common.sh
```diff
@@ -292,6 +292,54 @@ validate_cubelet_cow_startup_deps() {
...
+
+# validate_cubelet_s3lvol_startup_deps: check the runtime deps of the
+# installed s3lvol_tgt binary. The binary statically links SPDK/DPDK/AWS
+# CRT, so `ldd` on it is the authoritative probe for exactly the system
+# libraries the target machine must provide (glibc is assumed present).
+# Runs only when ONE_CLICK_ENABLE_S3LVOL=1 and the binary is actually in
+# the package.
+validate_cubelet_s3lvol_startup_deps() {
+  local s3lvol_bin="$1" # <prefix>/CubeS3lvol/bin/s3lvol_tgt
+  [[ -n "${s3lvol_bin}" && -f "${s3lvol_bin}" ]] || return 0
```
fslongjin, line 310: **Should fix: preflight should die fast on a missing `cos.cfg`.** When `ONE_CLICK_ENABLE_S3LVOL=1`, the installer never creates or templates `/data/cubelet/cos.cfg`; the unit will hit `StartLimitBurst=5` and fail. Add a preflight check here that fails with a clear message and a template snippet in the docs.

deploy/one-click/env.example
```diff
@@ -190,6 +190,30 @@ CUBE_PROXY_GRPC_PORT=9090
...
+# RCOW_JOURNAL_MB=1024
+# Thin-provisioned capacity for newly created volumes (GiB).
+# RCOW_CAPACITY_GB=16384
+# Target process CPU affinity mask and memory (MiB). Align RCOW_TGT_CPUMASK
+# with the host's dedicated cores; 0x3 is the default mask on 4-core dev
+# boxes.
+# RCOW_TGT_CPUMASK=0x3
+# RCOW_TGT_MEM_MB=16384
+# NVMe/TCP listener the target serves on (initiator connects to this).
+# RCOW_LISTEN_ADDR=127.0.0.1
```
fslongjin, line 230: **Should fix: add a security note for `RCOW_LISTEN_ADDR`.** Default `127.0.0.1` is safe, but the comment should warn that binding to `0.0.0.0` exposes an unauthenticated NVMe/TCP target (`allow_any_host`) on the network. Also align this comment with the actual code defaults (`RCOW_CAPACITY_GB=16384`, not 1024).

CubeS3lvol/scripts/rcow_common.sh
```diff
@@ -0,0 +1,1445 @@
...
+# of something compiled in: the module resolves both paths through the environment
+# (S3LVOL_ACTIVE_FILE / S3LVOL_BSTORE_FILE, see vbdev_s3lvol_statefile.c) and falls
+# back to these same defaults when they are unset.
+#
+# Exported, not just set, because the target inherits them -- without that the
+# scripts would be reading one file while the module wrote another, which is worse
+# than having no setting at all. The names differ because the module has no business
+# knowing about the "rcow" deployment flavour; this is where the two meet.
+RCOW_ACTIVE_FILE="${RCOW_ACTIVE_FILE:-/var/tmp/rcow_active_lvols}"
+RCOW_BSTORE_FILE="${RCOW_BSTORE_FILE:-/var/tmp/bstore.json}"
```
fslongjin: **Should fix: state files default to `/var/tmp`.** `bstore.json` decides attach-vs-create and the active registry tracks running volumes. `/var/tmp` is sticky world-writable and commonly cleaned by `systemd-tmpfiles`. Losing `bstore.json` makes the next attach silently create an empty volume. Move both under `/data/cubelet/rcow/` and tighten perms (0600).

CubeS3lvol/scripts/rcow_common.sh
```diff
@@ -0,0 +1,1445 @@
...
+	if [ -z "${line}" ]; then
+		printf 'the target log does not say who holds the marker'
+		return 1
+	fi
+
+	node="${line#*node=}"; node="${node%% *}"
+	pid="${line##*pid=}"
+	me="$(hostname)"
+
+	if [ "${node}" != "${me}" ]; then
```
fslongjin, line 741: **Should fix: hostname-based staleness trusts duplicate hostnames.** `rcow_owner_is_stale` only allows force-take when the marker names *this* host and the pid is not running locally. That is safe against a different node, but two machines with the same hostname (containers, cloned disks) will each see the other's marker as "stale on this host" and force-take a volume the other side is still serving. S3 has no `If-None-Match`, so the owner lock cannot be made atomic. The code comments already acknowledge the collision risk; please surface it in the operator docs and recommend pinning `RCOW_LVS_NAME` from the orchestrator when real node ids exist.

CubeS3lvol/scripts/rcow_common.sh
```diff
@@ -0,0 +1,1445 @@
...
+		nqn="$(rcow_nqn "${i}")"
+		if printf '%s\n' "${existing}" | grep -qxF "${nqn}"; then
+			continue
+		fi
+
+		# -a: any host may connect. The initiator is this same machine over
+		#     loopback, and a hostnqn allowlist here would have to be kept
+		#     in step with /etc/nvme/hostnqn on every boot.
+		# -m: see the header comment -- without it nsid > 32 is refused.
+		script="${script}nvmf_create_subsystem ${nqn} -a -s $(printf 'RCOW%014d' "${i}") -m ${RCOW_NS_PER_SUBSYS}
```
fslongjin, line 850: **Should fix: `allow_any_host` opens the subsystem to any initiator.** `nvmf_create_subsystem ... -a` accepts any host. Combined with a `0.0.0.0` listen address this is an unauthenticated block device on the network. Default to an explicit host allowlist, or at least fail closed when `RCOW_LISTEN_ADDR` is not loopback.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_statefile.c
```diff
@@ -0,0 +1,263 @@
...
+	ssize_t written;
+
+	if (snprintf(tmp, sizeof(tmp), "%s.tmp", path) >= (int)sizeof(tmp)) {
+		SPDK_ERRLOG("statefile: path too long: %s\n", path);
+		return -ENAMETOOLONG;
+	}
+
+	/* 0644 to match what these files have always been: they hold no secrets
+	 * (credentials are environment-only) and the recovery scripts read them. */
+	fd = open(tmp, O_WRONLY | O_CREAT | O_TRUNC, 0644);
```
fslongjin: **Should fix: open state files with `O_NOFOLLOW` and tight perms.** `open(tmp, O_WRONLY|O_CREAT|O_TRUNC, 0644)` in a world-writable dir (`/var/tmp`) is a classic symlink-to-root attack surface. Use `O_NOFOLLOW`, mode `0600`, and write into a non-world-writable dir (see the `rcow_common.sh` comment).

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -0,0 +1,3266 @@
...
+	}
+	if (!ctx->overlay) {
+		SPDK_ERRLOG("WAL replay without an attached overlay\n");
+		return -EINVAL;
+	}
+
+	switch (hdr->type) {
+	case S3_WAL_WRITE:
+		if (!payload ||
+		    hdr->payload_len != hdr->nblocks * S3LVOL_BLOCK_SIZE) {
```
fslongjin: **Should fix: `nblocks * 4096` can wrap.** `hdr->nblocks` is `uint32_t`; a corrupted WAL header with `nblocks == 1048576` makes the product wrap to 0 and can pass the `payload_len` check. Not a default-path bug, but widen the multiply to `uint64_t` while the area is new.

CubeS3lvol/lib/s3bsdev/s3_gc.c
```diff
@@ -0,0 +1,93 @@
+/* Copyright (c) 2026 Tencent Inc.
+ * SPDX-License-Identifier: Apache-2.0 */
+/*
+ *   TODO: not implemented yet.
```
fslongjin, line 4: **Should fix: PR description overstates the feature.** The PR body says "checkpoint + GC keep the object store compact", but this file is a `TODO` stub and `s3_list_objects()` still returns `-ENOTSUP`. Either land a real GC or reword the PR body to say GC is future work. `env.example` is already honest ("until the backend is wired into the storage path"); the PR body should match.

CubeS3lvol/lib/s3bsdev/s3_client_aws.c
```diff
@@ -0,0 +1,2269 @@
...
+	if (event_loop_threads == 0) {
+		event_loop_threads = 4;
+	}
+
+	allocator = aws_default_allocator();
+	aws_s3_library_init(allocator);
+
+	/* Logger: errors only by default.
+	 *
+	 * S3LVOL_CRT_LOG_LEVEL raises it for debugging. TRACE dumps every
```
fslongjin, line 555: **Nit: `trace` log level leaks SigV4 material.** At `S3LVOL_CRT_LOG_LEVEL=trace` the AWS CRT logs the canonical request and the `Authorization` header. The README documents `trace` as a normal debug knob and also claims credentials "never [go] into logs". Either drop `trace` from the documented levels or redact SigV4 in the logger.

CubeS3lvol/README.md
```diff
@@ -0,0 +1,220 @@
...
+
+Everything in the scripts can be overridden from the environment. The commonly
+used ones:
+
+| Variable | Default | Meaning |
+|---|---|---|
+| `RCOW_COS_CFG` | `/data/cubelet/cos.cfg` | |
+| `RCOW_WAL_IMG` | `/data/cubelet/rcow/wal_bdev.img` | |
+| `RCOW_LVS_NAME` | `rcow` | lvstore name, also the prefix in S3 |
+| `RCOW_CAPACITY_GB` | `1024` | used only at first create; thin, unused space costs nothing |
```
fslongjin: **Should fix: documented defaults disagree with the code.** `RCOW_CAPACITY_GB` is documented as `1024` but the code default is `16384`; the WAL `truncate -s 512G` example does not match the installer's ~33 GiB; `RCOW_LVS_NAME` is documented as `rcow` but the script derives `rcow-<hostname-hash>`. The WAL image size is a permanent per-node contract, so these mismatches become irreversible node state once a user follows the README.

CubeS3lvol/README.md
```diff
@@ -0,0 +1,220 @@
...
+`RCOW_NO_HUGE=1` is a **choice, not a concession**: nothing on this data path
+needs DMA (the local disk goes through `bdev_aio` read/write syscalls, the
+network through nvmf_tcp sockets, and S3 traffic lives in memory CRT mallocs
+itself), and the only capability `--no-huge` drops is "DPDK resolving IOVAs for
+memory". Setting it to `0` switches to hugepages, in which case the script fails
+outright rather than silently degrading if `nr_hugepages=0`.
+
+## Operations
+
+All of this project's RPCs go through `scripts/s3lvol_rpc.py` (raw JSON-RPC);
```
fslongjin, line 154: **Nit: control-plane CLI is undocumented.** The README points users at `scripts/s3lvol_rpc.py` for `rcow_*` JSON-RPC, but does not list the methods, their args, or examples. Add a short method table (create/attach/detach/resize/delete/export) so operators do not have to read the script.

CubeS3lvol/mk/s3lvol.common.mk
```diff
@@ -0,0 +1,355 @@
+# Copyright (c) 2026 Tencent Inc.
+# SPDX-License-Identifier: Apache-2.0
+#
+#  s3lvol 公共构建变量
```
fslongjin: **Blocking: non-English comments.** This makefile contains Chinese commentary. Repo rule 11 requires code, comments, and log strings to be English (localized docs excepted). Translate before merge.

CubeS3lvol/.gitignore
```diff
@@ -0,0 +1,54 @@
+# 构建产物
```
fslongjin: **Blocking: non-English comments.** `.gitignore` contains Chinese comments. Repo rule 11 requires comments to be English. Translate before merge.

CubeS3lvol/test/run_all.sh
```diff
@@ -0,0 +1,541 @@
...
+FAILED_NAMES=""
+SKIPPED_NAMES=""
+XPASS_NAMES=""
+
+parse_counts()
+{
+	sed -n \
+	    -e 's/.*result:[[:space:]]*\([0-9]\+\)[[:space:]]*passed,[[:space:]]*\([0-9]\+\)[[:space:]]*failed.*/\1 \2/p' \
+	    -e 's/.*===[[:space:]]*\([0-9]\+\)[[:space:]]*passed,[[:space:]]*\([0-9]\+\)[[:space:]]*failed.*/\1 \2/p' \
+	    -e 's/.*结果:[[:space:]]*\([0-9]\+\)[[:space:]]*通过,[[:space:]]*\([0-9]\+\)[[:space:]]*失败.*/\1 \2/p' \
```
fslongjin: **Blocking: non-English content.** The `sed` pattern here matches Chinese test output. Repo rule 11 requires log strings to be English. Either make the test emit English or drop the localized match.

docker/Dockerfile.builder
```diff
@@ -136,6 +159,20 @@ RUN . /etc/buildenv \
        apt-get install -y gcc-x86-64-linux-gnu; \
     fi \
     && rm -rf /var/lib/apt/lists/*
+# Python build-deps for CubeS3lvol's SPDK/DPDK (populated by setup_dep.sh).
+# The pinned SPDK needs python >= 3.9 (genrpc.py uses
+# argparse.BooleanOptionalAction) and DPDK needs meson >= 0.57.2, neither of
+# which ubuntu 20.04's stock python3.8 / apt meson (0.53.2) provides. Install
+# python3.9 and shadow /usr/local/bin/python3 with it (PATH-first; the OS's
+# /usr/bin/python3 stays on 3.8), then install the Python toolchain into it.
+RUN ln -sf /usr/bin/python3.9 /usr/local/bin/python3 \
```
fslongjin: **Should fix: global `python3` / `meson` shadow.** `ln -sf /usr/bin/python3.9 /usr/local/bin/python3` plus pip `meson==1.10.0` changes the interpreter for every `#!/usr/bin/env python3` script in the shared builder, not just CubeS3lvol. Please verify the existing Go/Rust/kernel tracks still build green with the new Dockerfile, or scope the shadow to `setup_dep.sh` only.

fslongjin (changes requested):
**Verdict: Request changes. The engine direction is sound, but this PR breaks the default one-click install path and misses several repo gates.**

## What this PR does well

- The S3/COS remote COW data plane is coherent: journal + WAL first, background flush, overlay read cache, grow-only resize, idempotent activate/deactivate, async export with on-the-spot `deletable`.
- Credentials stay out of argv and RPC params; they only enter the target process environment. Default NVMe/TCP listener is `127.0.0.1`.
- DCO is signed; commits are atomic and well-described; `make all` is not coupled to s3lvol; the offline test workflow correctly separates SKIP from PASS.

## Blocks merge

1. **`install.sh` fails on a fresh install by default.** The official builder always ships `CubeS3lvol/bin/s3lvol_tgt`, and `install.sh` then unconditionally runs `truncate /data/cubelet/rcow/wal_bdev.img`. That directory is never created anywhere in one-click, and the block ignores `ONE_CLICK_ENABLE_S3LVOL` (which defaults to `0`). A brand-new compute or all-in-one install dies before any role service starts. Regression for existing one-click users, not an opt-in edge case. See inline on `install.sh`.
2. **Non-English characters in code/comments.** `CubeS3lvol/mk/s3lvol.common.mk`, `CubeS3lvol/.gitignore`, and `CubeS3lvol/test/run_all.sh:101` contain Chinese. Repo rule 11 requires code, comments, and log strings to be English. See inline on each.
3. **No `CODEOWNERS` entry for the new top-level component.** `.github/CODEOWNERS` has no `/CubeS3lvol/` line, so review falls through to the catch-all. Please add a module owner. (Not posted inline — `CODEOWNERS` is not in this PR's diff.)

## Should fix before merge (not strictly blocking, but the PR as written misleads reviewers)

4. **State files default to `/var/tmp`.** `bstore.json` (decides attach-vs-create) and the active registry live in `/var/tmp`, which systemd-tmpfiles commonly cleans and which is sticky world-writable. The Chinese README claims "metadata is preserved across restarts"; with the default path that is not true. Move them under `/data/cubelet/rcow/`, open with `O_NOFOLLOW`, tighten permissions. See inline on `rcow_common.sh` and `vbdev_s3lvol_statefile.c`.
5. **PR description overstates the feature.** It says "checkpoint + GC keep the object store compact", but `s3_gc.c` is a TODO stub and `s3_list_objects()` still returns `-ENOTSUP`. It also reads like a first-class volume backend, but cubecow / Cubelet / Volume API / k8s / terraform are not wired. `env.example` is honest ("until the backend is wired into the storage path"); the PR body should match. See inline on `s3_gc.c`.
6. **`CubeS3lvol/README.md` defaults disagree with the code.** `RCOW_CAPACITY_GB` documented as `1024` but is `16384`; WAL example `truncate -s 512G` but installer creates ~33 GiB; `RCOW_LVS_NAME` documented as `rcow` but the script derives `rcow-<hostname-hash>`. WAL image size is a permanent per-node contract, so these mismatches become irreversible node state. See inline on `README.md`.
7. **English one-click README has no s3lvol section.** `deploy/one-click/README_zh.md` documents stop/upgrade semantics and preflight deps; `deploy/one-click/README.md` does not mention s3lvol at all. (Not posted inline — README.md is not in this PR's diff.)
8. **Hostname-based owner staleness trusts duplicate hostnames.** `rcow_owner_is_stale` only allows force-take when the marker names this host and the pid is not running locally. That is safe against a different node, but two machines with the same hostname (containers, cloned disks) will each force-take a volume the other side is still serving. The code comments already acknowledge the collision risk; please surface it in the operator docs and recommend pinning `RCOW_LVS_NAME` when real node ids exist. See inline on `rcow_common.sh`.
9. **`ONE_CLICK_ENABLE_S3LVOL=1` has no `cos.cfg` story.** The installer never creates or templates `/data/cubelet/cos.cfg`; the unit will hit `StartLimitBurst=5` and fail. Preflight should die fast on a missing file, and the docs should show a template. See inline on `common.sh`.
10. **Upgrade may overwrite merged `RCOW_*`.** `install.sh` resolves `RCOW_*` with `${VAR:-default}` before the merged env is applied, then `upsert_env_kv` writes those values back. A user who only edited the old `.one-click.env` can have their settings replaced by installer defaults. See inline on `install.sh`.
11. **Builder image shadows `python3` / `meson` globally.** `ln -sf /usr/bin/python3.9 /usr/local/bin/python3` plus pip `meson==1.10.0` changes the interpreter for every `#!/usr/bin/env python3` script in the shared builder, not just CubeS3lvol. Please verify the existing Go/Rust/kernel tracks still build green with the new Dockerfile, or scope the shadow to `setup_dep.sh` only. See inline on `Dockerfile.builder`.
12. **`tests/unittest/run.sh` does not know this component.** At minimum register it in `GATED_TESTS` (or `NO_TESTS`) so the sweep stays honest. (Not posted inline — `run.sh` is not in this PR's diff.)
13. **WAL replay `nblocks * 4096` can wrap.** `nblocks` is `uint32_t`; a corrupted header with `nblocks == 1048576` makes the product wrap to 0 and can pass the length check. Not a default-path bug, but worth widening to `uint64_t` while the area is new. See inline on `s3_bs_dev.c`.

## Nit

- `Assisted-by: CodeBuddy` is missing the `MODEL_VERSION` required by `AGENTS.md`.
- `CubeS3lvol/README.md` has no `README_zh.md` sibling (repo rule 2).
- `install.sh:1329` comments say the unit wiring is done by `install-units.sh`; it is actually done in `start_systemd_target`.
- `S3LVOL_CRT_LOG_LEVEL=trace` leaks SigV4 material; the README documents `trace` as a normal debug knob and also claims credentials "never [go] into logs". See inline on `s3_client_aws.c` and `README.md`.

## Suggested merge order

1. Fix the WAL `mkdir` + gate on `ONE_CLICK_ENABLE_S3LVOL`.
2. Move state files off `/var/tmp`.
3. English-out the three Chinese files.
4. Add `CODEOWNERS`.
5. Align README defaults and add the English one-click section.
6. Correct the PR body re GC and "first-class backend".

After that the rest is hardening.

Please also follow the Issue & PR Close Policy in `CONTRIBUTING.md`: link related issues and state close semantics.

deploy/one-click/install.sh
changpe1: Done.

deploy/one-click/lib/common.sh
changpe1: Done.

CubeS3lvol/scripts/rcow_common.sh
changpe1: Done.

CubeS3lvol/scripts/rcow_common.sh
changpe1: Done.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_statefile.c
changpe1: Done.

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
changpe1: Done.

CubeS3lvol/lib/s3bsdev/s3_gc.c
changpe1: Done.

CubeS3lvol/mk/s3lvol.common.mk
changpe1: Done.

CubeS3lvol/test/run_all.sh
changpe1: Done.

fslongjin:
**s3lvol_tgt has no automatic log rotation.**

Logs default to `/data/log/rcow/s3lvol_tgt.log` (stdout/stderr redirect). The only existing “rotation” is a rename on restart (`s3lvol_tgt.log.YYYYMMDD-HHMMSS`). That is not enough for a long-lived data-plane process:

1. While the target stays up, the file grows without a size or time cap.
2. Timestamped archives are never compressed or deleted.
3. There is no `logrotate` policy. The process also keeps the fd open, so a host-side `rename` would keep writing to the old inode.

Please add a real rotation story (in-process size/time rotate, or logrotate + reopen/copytruncate + retention) before this ships as a production service.

fslongjin (approved):

PR #1494, opened 2026-08-24
fix(cubemaster): evict node cache only after sustained empty syncs

liciazhu:
Clear the local node cache only after emptySyncEvictThreshold consecutive empty CubeOps responses , absorbing transient jitter instead of wiping the scheduler view on a single empty response. Load errors reset the streak and never evict.


fslongjin (approved):
LGTM

PR #1496, opened 2026-08-25
cubecow: harden s3lvol JSON-RPC client against EINTR and frame desync

lisongqian:
The s3lvol JSON-RPC client occasionally surfaced errors of the form

    s3lvol rpc: read rcow_active_bdev: Interrupted system call (os error 4)

on the volume activation path (create_volume → rcow_create_lvol → rcow_active_bdev → rcow_get_bdev). Root cause: read_line() did the byte-at-a-time read with `?`, so a single EINTR (e.g. from SIGURG-based runtime preemption, GDB attach, cgroup freeze/thaw or SIGALRM timers) would abort the whole RPC even though the server had done nothing wrong. Worse, silently continuing on the same socket after a half-consumed response could desynchronise the JSON-RPC framing and let a stale response be returned as the result of the next request.

Fix, in three layers of defence:

  * read_line() now retries `ErrorKind::Interrupted` in place, matching the standard std::io convention, so transient signals no longer fail the RPC or corrupt the wire framing.
  * The write/read retry whitelists in call_raw() additionally include `Interrupted` as a belt-and-braces path: if EINTR ever escapes the inner reader, we drop the socket (`*guard = None`) and redial before retrying, since we can't trust the remaining bytes in the kernel buffer.
  * After decoding the response, we now assert that `resp.id` matches the request `id` as required by JSON-RPC 2.0. On mismatch we drop the connection and retry once; if it still mismatches we surface a PreconditionFailed rather than returning the wrong result.

The library builds clean (`cargo check -p cubecow --lib`); pre-existing errors in `src/bin/cubecow_cli.rs` are unrelated CLI/Engine drift.

### DIFF
--- a/cubecow/src/engine/s3.rs
+++ b/cubecow/src/engine/s3.rs
@@ -1296,9 +1296,17 @@ impl JsonRpcClient {
                     if attempt < 2
                         && matches!(
                             e.kind(),
-                            ErrorKind::BrokenPipe | ErrorKind::ConnectionReset
+                            ErrorKind::BrokenPipe
+                                | ErrorKind::ConnectionReset
+                                | ErrorKind::Interrupted
                         ) =>
                 {
+                    // BrokenPipe / ConnectionReset: peer closed the socket.
+                    // Interrupted (EINTR): the write was aborted by a signal
+                    // before any bytes were committed to the peer; the socket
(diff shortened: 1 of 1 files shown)


ls-ggg:
LGTM

PR #1499, opened 2026-08-25
fix(cubenet): reset stale TCP sessions on sandbox rollback

zhouxianping:
---
  Problem

  A sandbox rollback (restore snapshot + resume guest) rewinds the guest's TCP state, but the dataplane conntrack/NAT session table and the peer's TCP state do not rewind. In-flight connections then desync and hang: the guest waits for retransmits the peer already ACKed (recv
  hang), or keeps retransmitting data the peer already confirmed (send hang). Clearing the session table alone is not enough — the guest's stack won't error out on silent packet loss.

  Approach

  Detect stale (pre-rollback) sessions by generation and reset them so both endpoints get a clean ECONNRESET and reconnect promptly.

  - Generation = mvm_meta.version, carried in a new nat_session.gen field (stamped at creation) and in the session key for SNAT/L7. RollbackSandbox bumps it via a map read-modify-write +1 after the guest is restored and resumed.
  - RollbackSandbox bump is O(1): the TAP ifindex is resolved from the sandbox IP via the mvmip_to_ifindex map (LookupIfindexByIP) instead of scanning every TAP device.

  Per-connection-type behavior

  - SNAT: forward misses reuse the existing miss-reset; reverse (nodenic) resets the peer when the ingress entry (key version always 0, so it is found across a rollback) records a generation that differs from the current one. Host-bound traffic has no ingress entry and is passed 
  through, never reset (e.g. external SSH to the node).
  - L7 (localgw from_envoy): a stale L7 reply reflects a reset to the proxy and drops the session pair; a miss is non-blocking as before.
  - port_mapping: the inbound path tracks the connection in an egress session (fixed key, gen in value). A stale connection resets the peer; a reaped one is rebuilt (not reset); the guest side is reset via the mvmtap reply path.

  Unified RST builder

  A single tcp_send_reset(skb, out_ifindex, new_daddr) in tcp_reset.h serves both directions (removes the duplicated tcp_reply_reset):
  - peer/proxy-facing (nodenic / localgw) passes the trigger's source;
  - guest-facing (mvmtap) passes mvm_inner_ip, because from_cube rewrites the source to the sandbox's real IP before the reset decision.

  Verification

  - mvmtap / nodenic / localgw compile for amd64 + arm64 and pass the kernel verifier (bpftool prog loadall).
  - cubevs Go tests (incl. BumpMvmVersion RMW + field preservation) and Cubelet services/cubebox + network suites pass.
  - e2e (real snapshot/rollback with a long-lived connection observing ECONNRESET) should be run on a target cluster before merge.

### DIFF
--- a/CubeNet/src/localgw.bpf.c
+++ b/CubeNet/src/localgw.bpf.c
@@ -52,7 +52,7 @@ static __always_inline int update_l7_session_from_cubedev(struct __sk_buff *skb)
 
 	if (session_is_stale(sess)) {
 		delete_session_pair(&key, sess);
-		return tcp_reflect_reset(skb, skb->ingress_ifindex);
+		return tcp_send_reset(skb, skb->ingress_ifindex, l3->saddr);
 	}
 
 	now = bpf_ktime_get_ns();
--- a/CubeNet/src/localgw.bpf.c
+++ b/CubeNet/src/localgw.bpf.c
@@ -11,12 +11,22 @@
 #include "nat.h"
 #include "skb.h"
 #include "tcp.h"
+#include "tcp_reset.h"
 
 /* Update a CubeEgress TCP session from the original reply tuple before DNAT
  * rewrites the sandbox-assigned destination address to mvm_inner_ip. A missing
  * or non-L7 session is intentionally non-blocking.
  */
-static __always_inline void update_l7_session_from_cubedev(struct __sk_buff *skb)
+/* Update a CubeEgress TCP session from the original reply tuple before DNAT
+ * rewrites the sandbox-assigned destination address to mvm_inner_ip.
+ *
+ * Returns the reset-redirect action when the reply belongs to a stale
+ * (pre-rollback) L7 session — a RST is reflected to the proxy and the session
+ * pair removed so the proxied connection is torn down instead of desyncing —
+ * or 0 to continue normal delivery. A missing or non-L7 session is
+ * intentionally non-blocking (returns 0).
+ */
+static __always_inline int update_l7_session_from_cubedev(struct __sk_buff *skb)
 {
 	struct session_key key = {};
 	struct nat_session *sess;
@@ -26,9 +36,9 @@ static __always_inline void update_l7_session_from_cubedev(struct __sk_buff *skb
 	__u64 now;
 
 	if (!__pull_headers(skb, &l2, &l3, &l4))
-		return;
+		return 0;
 	if (l3->protocol != IPPROTO_TCP)
-		return;
+		return 0;
 
 	key.src_ip = l3->saddr;
 	key.dst_ip = l3->daddr;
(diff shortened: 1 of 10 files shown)


CubeNet/src/session.h
```diff
@@ -309,6 +376,17 @@ static __always_inline bool create_nat_session(struct __sk_buff *skb,
 	 * mvm_meta is what schedules this flow's next re-check.
 	 */
 	sess.policy_version = policy_version;
+
+	/* Stamp the creation generation (the VM's current mvm_meta->version) so a
+	 * reverse-path reader can tell a stale (pre-rollback) session from a
+	 * merely reaped one. For SNAT/L7 this equals ekey->version; for
+	 * port_mapping the key version is fixed at 0 while gen tracks rollback. */
+	{
```
chenhengqi, line 384: No need for an extra block.

CubeNet/src/nodenic.bpf.c
```diff
@@ -13,19 +13,82 @@
...
+	sess.access_time = now_ns;
+	sess.node_ifindex = node_ifindex;
+	sess.node_ip = node_ip;
+	sess.node_port = host_port;
+	sess.vm_ifindex = mvm_port->ifindex;
+	sess.vm_ip = mvm_inner_ip;
+	sess.vm_port = mvm_port->listen_port;
+	sess.state = initial_state;
+	sess.packet_class = PORT_MAPPING_PACKET;
+	{
```
chenhengqi, line 45: ditto

Cubelet/services/cubebox/rollback.go
```diff
@@ -335,6 +346,23 @@ func resolveRollbackTargets(ctx context.Context, backend string, req *cubebox.Ro
...
+// dataplane resets now-stale TCP sessions after a rollback. Best effort; the
+// caller logs a warning on failure. It resolves the TAP ifindex from the
+// sandbox IP in O(1) via the mvmip_to_ifindex map rather than scanning every
+// TAP device.
+func bumpRollbackNetworkGeneration(sandboxIP string) error {
+	ip := net.ParseIP(sandboxIP).To4()
+	if ip == nil {
+		return fmt.Errorf("invalid sandbox IP %q", sandboxIP)
+	}
+	ifindex, err := cubevs.LookupIfindexByIP(ip)
```
chenhengqi, line 359: We can use netlink to obtain ifindex.

CubeNet/src/localgw.bpf.c
```diff
@@ -26,9 +36,9 @@ static __always_inline void update_l7_session_from_cubedev(struct __sk_buff *skb
 	__u64 now;
 
 	if (!__pull_headers(skb, &l2, &l3, &l4))
-		return;
+		return 0;
 	if (l3->protocol != IPPROTO_TCP)
```
chenhengqi, line 40: This check is redundant.

PR #1512, opened 2026-08-25
feat(volume-s3): implement S3 volume plugin in Go

fslongjin:
## Summary

Rewrite the S3 volume plugin (`examples/volume/s3`) from a shell script into a self-contained Go binary.

### What changed

- **Go implementation**: replace `binary/cube-volume-s3.sh` with `cmd/cube-volume-s3` and internal packages:
  - `internal/config` — env-var configuration with validation
  - `internal/lockfile` — cross-process mount lock
  - `internal/s3api` — direct HTTP client to the S3-compatible endpoint (SigV4)
  - `internal/s3fsmnt` — s3fs mount/umount lifecycle
- **No AWS CLI dependency**: Create/Destroy now talk to the endpoint over HTTP, so the `awscli` install path and the bundled `awscli-bundle` in the one-click deploy package are removed.
- **CI coverage**: wire the standalone Go module into `build-check`, `fmt-check` and `unit-test-check` workflows (new `volume-s3` component), plus a top-level Makefile target (`cube-volume-s3`).
- **Image & docs**: update Cubelet/CubeMaster Dockerfiles, volume-deps install script, one-click deploy scripts and both EN/ZH volume-plugin docs.

### Testing

- Unit tests for all new packages (`config`, `lockfile`, `s3api`, `s3fsmnt`).
- Makefile target builds `_output/bin/cube-volume-s3`.

## Checklist

- [x] Unit tests pass
- [x] Build succeeds for amd64 and arm64 (CI `volume-s3` component)
- [x] Docs updated in English and Chinese


fslongjin:
Thanks for the review. Pushed fixes for the actionable items; the rest are addressed below.

## Fixed in the latest push

- **HIGH — global `CGO_ENABLED=0` on `build_go_binary`**: scoped to the `cube-volume-s3` build only via a new `static_linux` flag. `build_go_binary` now defaults to the host CGO setting, so `cubelet`/`cubecli` keep their cgo (nsenter, cubecow). `test_package_layout.sh` got static guards so this cannot regress.
- **MEDIUM — recommended one-click flow compiled the plugin on the host**: added a `track_volume_s3` builder track to `build-release-bundle-builder.sh` and pass `ONE_CLICK_VOLUME_S3_BIN` through, so the recommended entry point no longer needs a host `go`.
- **MEDIUM — `ParseEndpoint` drops the URL path**: `ParseEndpoint` now rejects path-prefixed endpoints (`https://host/s3proxy`) with an explicit error, since minio-go cannot carry a URL path and s3fs would silently split control/data planes. Default MinIO/AWS/COS/R2 endpoints have no path and are unaffected.
- **LOW — inline `#` comments kept in the value**: the config parser now strips an unquoted `#` comment before splitting `KEY=VALUE`, matching the old shell `source` semantics. Added `TestLoadStripsInlineComment`.
- **LOW — README Go version mismatch**: both READMEs now say ≥ 1.25 to match `go.mod` (`go 1.25.0`) and minio-go v7.3.0.
- **Observation — `test_package_layout.sh` CGO guard was weak**: tightened the grep to match the actual `go build` line carrying `CGO_ENABLED=0` rather than a nearby comment.
- **Observation — `go vet` not wired for the new module**: `make cube-volume-s3-test` now runs `go vet ./... && go test ./...`.

## Intentionally not changed

- **LOW — `flag.ExitOnError` bypasses JSON on bad flags**: the old shell plugin also emitted no JSON on unknown args (`die` wrote to stderr only). Cubelet treats both as a plugin call failure. Keeping `flag.ExitOnError` for parity with the stdlib default.
- **LOW — detach fallback ignores `--volume-base-dir`**: Cubelet's binary-plugin driver does not pass `--volume-base-dir` on detach (the `DetachRequest` has no such field); detach relies on the `mount_dir` recorded in attach metadata. The fallback to the default base dir matches the old shell behaviour. Changing it to `MountPoint(opts.volumeBaseDir, ...)` would make no difference since that flag is empty on detach.
- **LOW — `S3FS_EXTRA_OPTS` splitting is not quote-aware**: the old `read -r -a` also splits on whitespace (it is not shell-word-splitting), so `S3FS_EXTRA_OPTS='-o uid=1000,gid=1000'` becomes `["-o", "uid=1000,gid=1000"]` in both implementations. The example config uses space-free values. Documenting a "no embedded spaces" limitation rather than adding a custom splitter.
- **LOW — unrelated `cubebox_grpc.pb.go` regeneration**: keeping as-is.
- **Observation — `-trimpath` on all one-click Go builds**: harmless; affects only stack-trace paths. Will note in the release log if it survives the release cut.
- **Observation — config lookup resolves symlinks**: one-click and image installs place the binary as a regular file, not a symlink. No behaviour change for supported installs.
- **Observation — control nodes still install `s3fs`**: deliberate; a single-node control node also runs Cubelet and needs `s3fs` for attach. Docs already state only mount nodes need `s3fs`.

PR #1516, opened 2026-08-25
feat(clm): add active-standby failover

chenhengqi:
Run two warm CLM replicas in Kubernetes and use a Redis lease to gate singleton sweep and prune work while both replicas serve resume requests.

Use broadcast XREAD consumption, promotion catch-up, fencing epochs, and versioned state CAS to prevent stale replicas from overwriting newer state. Add Helm configuration, readiness observability, Redis transaction tests, and bilingual documentation. Keep one-click deployment single-replica.


cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -345,18 +682,22 @@ func handleEvent(ctx context.Context, ev redisstream.Event, push *proxypush.Clie
...
 			zap.String("sandbox_id", ev.SandboxID),
 			zap.Int("registry_size", reg.Len()))
-		if err := push.DeleteMeta(ctx, ev.SandboxID); err != nil {
-			log.Warn("delete event push failed",
-				zap.String("sandbox_id", ev.SandboxID), zap.Error(err))
+		if canWrite() {
+			if err := push.DeleteMeta(ctx, ev.SandboxID); err != nil {
+				log.Warn("delete event push failed",
+					zap.String("sandbox_id", ev.SandboxID), zap.Error(err))
+			}
```
FakeLearne, line 700: There is a potential data leak issue here during active-standby failover:

1. **Delete notifications are dropped**: During failover promotion catch-up and the subsequent 10s `HTTPTimeout` drain window, the new leader has not completed reconciliation yet, so `activeLeader.IsLeader()` remains `false`, making `canWrite()` evaluate to `false`. If an `OpDelete` event is consumed during this window, `reg.Delete()` removes the sandbox from the local in-memory registry, but `push.DeleteMeta` is skipped.
2. **Subsequent hydration cannot compensate**: When the replica is finally promoted and triggers `hydrateFleet`, `replayRegistryTo` only iterates over entries currently present in `reg.Snapshot()`. Since the deleted sandbox was already purged from `reg` in step 1, the hydrator has no tombstone record and will never push a delete request to CubeProxy.
3. **Impact**: Any sandbox destroyed while the old leader was crashing or during the promotion window will permanently leak in CubeProxy's 16MB `cube_sandbox_meta` shared dict, which can eventually exhaust shared memory or route traffic to ghost sandboxes.

**Suggestion**:
Deleting a metadata entry in CubeProxy is idempotent. If the replica already holds the Redis lease (`lease.IsLeader() == true`), we should consider allowing `push.DeleteMeta` to pass through immediately when consuming `OpDelete`. Alternatively, retain a short-lived tombstone in `reg` or perform a diff against CubeProxy during fleet hydration.

cube-lifecycle-manager/internal/leader/lease.go
```diff
@@ -0,0 +1,317 @@
...
+	for {
+		select {
+		case <-ctx.Done():
+			return ctx.Err()
+		case <-ticker.C:
+			if !l.IsLeader() {
+				return errLeaseExpired
+			}
+			ok, err := l.renew(ctx)
+			if err != nil {
```
FakeLearne, line 185: A single renewal failure or transient `WATCH` conflict here causes immediate demotion, leading to frequent leader flapping and a prolonged "no-leader" vacuum:

1. **Immediate demotion on transient glitches**: `hold()` exits on the first `err != nil` (e.g. transient network timeout) or `!ok` (single `TxFailedErr` conflict in `redisLeaseStore.Renew`). Once `hold()` exits, `Run()` immediately invokes `l.demote()`.
2. **Prolonged dead-lock vacuum (~7s + 10s drain)**: Although the leader locally stepped down, the lease key in Redis was written with a 10s TTL and remains valid for ~7 more seconds. During this window, neither the old leader nor standbys can acquire the lease via `SET NX`. The cluster enters a 7-second leader vacuum, followed by a 10-second `HTTPTimeout` drain upon new promotion—resulting in ~17 seconds of interrupted background maintenance.
3. **Lease deadline ignored**: The lease already computes a safe local deadline (`started + TTL - margin`), which represents the true mathematical validity of the held lease. Demoting before this deadline arrives violates standard lease semantics (e.g. k8s `leaderelection`).

**Suggestions**:
1. **Retry until local deadline expires (Primary fix)**: In `hold()`, if `renew()` fails due to transient network errors or `TxFailedErr`, do not immediately return and demote. As long as `time.Now().Before(*deadline)` holds true, keep the leader status active and retry renewal more frequently (e.g. every `RetryInterval = 1s`). Only demote when the local deadline has actually expired. Also consider adding a small retry loop (e.g. `maxAttempts = 3`) inside `redisLeaseStore.Renew` for `TxFailedErr`.
2. **Self-reclaim on acquire (Optimization)**: If an instance does demote but the key in Redis still holds its own token, `acquire()` could atomically renew the lease and resume leadership without waiting for the full 10s HTTP drain, since no peer could have become leader in the interim.

cube-lifecycle-manager/internal/resumer/resumer.go
```diff
@@ -247,50 +253,58 @@ func (r *Resumer) callCubeMasterResume(ctx context.Context, sandboxID, instanceT
...
-		_ = r.o.Redis.ClearStateNotify(ctx, sandboxID)
+		// A transport or timeout error has an unknown server-side result.
+		// Preserve ownership until its TTL so another replica cannot issue a
+		// duplicate resume while CubeMaster may already be completing this one.
+		if !errors.As(resumeErr, &apiErr) {
+			return errors.New("cubemaster resume result unknown: " + resumeErr.Error())
+		}
+		// A structured non-success response is definitive; release ownership
+		// with an independent context so request cancellation cannot strand it.
+		r.clearState(sandboxID)
```
FakeLearne, line 264: Retaining ownership here on unknown transport errors makes sense to prevent duplicate RPCs and protect against the sweeper pausing an in-flight resume. However, using the default `StateLockTTL` (60s) turns a transient timeout into a 1-minute DoS for that sandbox:

1. **The 60s lock-out issue**: When an ambiguous transport error occurs, leaving `"resuming"` with the full 60s TTL causes all subsequent client retries to enter `waitForRunning` in `acquireResumeOwnership`. Since no in-flight goroutine is actually working on the resume anymore, retries will stall until client context timeout, rendering the sandbox unresponsive for a whole minute.
2. **Why we shouldn't simply `clearState`**: Unconditionally deleting the key immediately is also dangerous: `resumer` errors out before updating `MergeLastActive`, so `LastActiveMs` remains stale. If the key is deleted, `sweeper.go`'s `tryPause` will no longer be blocked by `SETNX`, and the leader's next sweep tick (within 5s) could pause the sandbox that CubeMaster just resumed or is currently resuming, causing state flapping.
3. **The ambiguity window is bounded by `HTTPTimeout`**: The server-side execution window cannot exceed `HTTPTimeout` (10s). The problem is not the guard itself, but the parameter choice.

**Suggestion**:
Instead of leaving the key with `StateLockTTL` (60s), downgrade the marker to a short TTL corresponding to the RPC ambiguity window upon unknown error:
```go
default:
    if !errors.As(resumeErr, &apiErr) {
        // Retain ownership only for the ambiguity window (HTTPTimeout),
        // preventing sweeper races without locking the sandbox for 60s.
        _ = r.o.Redis.SetState(ctx, sandboxID, "resuming", r.o.HTTPTimeout)
        return errors.New("cubemaster resume result unknown: " + resumeErr.Error())
    }
    r.clearState(sandboxID)
    return errors.New("cubemaster resume: " + resumeErr.Error())
```
This preserves the guard against sweeper races and deduplicates in-flight RPCs, while allowing subsequent client retries to recover cleanly in ~10s instead of ~60s.

cube-lifecycle-manager/internal/sweeper/sweeper.go
FakeLearne, line 281: There is a critical timing discrepancy and ambiguous error-handling flaw here that contradicts the drain window guarantees stated in the README:

1. **The 1x `HTTPTimeout` drain window is mathematically insufficient**:
   - The README states that a newly promoted leader waits one `HTTPTimeout` drain window so in-flight writes from the previous leader can finish.
   - However, `sweeper.tryPause` only checks `IsLeader()` once before entry (`sweeper.go:180`). Its internal execution chain runs on the unbounded `rootCtx` with multiple serial network operations:
     `ProxyPush.SetState("pausing")` (up to 10s) $\to$ `CubeMaster.Pause` (up to 10s) $\to$ rollback `ProxyPush.SetState("running")` (up to 10s).
   - A critical section entered right before the old leader's local demotion can easily take up to 30s before dropping its final write—landing ~16s **after** the new leader has already completed its 10s drain and started active maintenance.
2. **Ambiguous pause errors cause ghost "running" sandboxes (traffic black hole)**:
   - When `CubeMaster.Pause` times out or encounters a transport reset, the server-side result is **unknown** (note that CubeMaster's sandbox lock TTL is 180s, confirming pause can be slow).
   - Treating unknown transport errors as definitive failures in `default:` by calling `ClearStateNotify` and broadcasting `"running"` to CubeProxy is dangerous: if the microVM was actually paused on the host, CubeProxy will route incoming traffic directly to a frozen VM without triggering auto-resume, resulting in a persistent traffic black hole.
   - Note that `resumer.go:256-265` already handles this exact ambiguity correctly by retaining ownership on non-`APIError` transport failures.

**Suggestions**:
1. **Align `tryPause` with `resumer.go`**: On timeout or transport failure, do not immediately clear state and broadcast `"running"`. Treat it as ambiguous to prevent misleading the proxy when a VM might actually be frozen.
2. **Bound `tryPause` / `tryKill` execution**: Bound each sweep decision with a timeout not exceeding a single `HTTPTimeout`.
3. **Clarify documentation**: Update the README to reflect that per-sandbox state locks (`SET NX` with TTL), rather than the HTTP drain, provide the actual hard fencing boundary for singleton actions.

PR #1528, opened 2026-08-26
s3lvol: record and retry deletes that cannot complete

changpe1:
## Summary

A snapshot delete that cannot go through — the snapshot is exported, has more than one clone, is being decoupled, or is still an active NVMe-oF namespace — is refused with `EBUSY`, and until now nothing recorded that a delete had ever been asked for. This PR records it and gives the operator a way to complete it once the blocker clears.

**This is a record plus a manual retry, not an automatic one.** There is no poller: nothing on the target acts on the marks by itself.

## What's in this PR

**1. `s3lvol: record and retry deletes that cannot complete`**

- A refused snapshot delete records a **pending-delete mark**, keyed by `(lvstore uuid, lvol uuid)` — not by name. A name is unique only inside one loaded lvstore and is reusable, so a name-keyed mark could outlive the object it was recorded for (delete and recreate under the same name, or unload and re-attach the lvstore) and a retry would then act on something nobody asked it to. `s3lvol_lvstore_free()` drops the marks of the lvstore going away, which covers the unload, delete and setup-error paths alike.
- `rcow_get_lvstores` reports per-lvol cluster counts and, per snapshot, `export_status`, `deletable` and `delete_pending`.
- `rcow_delete_lvol` accepts optional `lvol_uuid` / `lvs_uuid` and refuses when the named lvol is not the one the caller meant — so a retry issued for one snapshot cannot delete a later same-named one.
- `test/tools/s3lvol_rpc.py --retry-pending` reads `rcow_get_lvstores`, finds every snapshot that carries the mark **and** has become deletable, and deletes each **by uuid**. Snapshots without the mark are never touched.
- `export_snapshot_deletable()` corrected: the export-pin, decouple and NVMf-active checks now run **before** the "no open blob" shortcut, which used to report a deactivated-but-pinned snapshot as `deletable: YES` and had `--retry-pending` attempt a delete that then failed with `EBUSY`.
- New dataplane suite `run_pending_delete_test.sh` (21 assertions, wired into `run_all.sh`): refused → alive and marked → `--retry-pending` skips it while pinned → deletes it once the export is released → mark cleared; an unmarked control snapshot is never touched; a delete carrying a stale uuid refuses a same-named replacement while the current uuid succeeds.

**2. `docs(zh): document refused snapshot deletes in cross-node-snapshot`**

Known issue 1 read as "the delete always goes through, just not immediately". It now states that a referenced snapshot is refused, what the node records, how to act on it, and the limits below.

**3. `test(dataplane): resolve SPDK rpc.py from deps/spdk in the remaining scripts`** (companion fix)

`run_activation_test.sh`, `run_dataplane_test.sh`, `run_recovery_test.sh` still hardcoded `/data/home/cow/spdk` / `../spdk`; resolved through `SPDK_ROOT` / `deps/spdk` like the earlier batch.

## Scope and limits (documented, not implied)

- **No automatic retry** — no poller; `--retry-pending` has to be run.
- **In memory only** — the marks are gone on restart, and dropped with their lvstore on unload/delete.
- **No cancel** — completing the delete is the only way to clear a mark.
- **`delete_pending` is a snapshot notion** — `rcow_get_snapshot_status` reports it for `snapshot_name`; by `export_uuid` it is always `false`.
- **The cluster path is not wired to it** — Cubelet's `S3Cow.DeleteByKind` still treats a refused snapshot delete as success and does not run `--retry-pending`. This PR is the node-side operator tool; fixing that leak is a separate change (both README and the zh doc say so).

## Verified

Full regression against a real COS bucket: **30/30 suites, 1124 assertions passed, 0 failed** (offline + real-S3 integration + all dataplane suites, including the new `run_pending_delete_test.sh` at 21/21). Target builds clean under `-Werror`.

### DIFF
--- a/CubeS3lvol/README.md
+++ b/CubeS3lvol/README.md
@@ -220,16 +220,25 @@ keeps running in the background. The uuid is not a completion signal — poll it
 
 ```sh
 scripts/s3lvol_rpc.py rcow_get_snapshot_status '{"export_uuid":"<uuid>"}'
-# -> {"export_status":"INPROGRESS","deletable":"NO"}  ... then, eventually
-# -> {"export_status":"DONE","deletable":"NO"}          (still pinned while exported)
+# -> {"export_status":"INPROGRESS","deletable":"NO","delete_pending":false}
+#    ... then, eventually
+# -> {"export_status":"DONE","deletable":"NO","delete_pending":false}
+#    (still pinned while exported)
 ```
 
-`rcow_get_snapshot_status` returns two fields:
+`rcow_get_snapshot_status` returns three fields:
 
 - `export_status`: `INPROGRESS` / `DONE` / `NONE`.
 - `deletable`: `YES` / `NO`, computed on the spot each time (never cached) and
   mirroring what `rcow_delete_lvol` would actually refuse: a snapshot is **not
-  deletable** while an export is in progress or when it has more than one clone.
+  deletable** while an export is in progress or pins it, while a decouple is
+  running on it, while it is active over NVMf, or when it has more than one
+  clone.
+- `delete_pending`: whether a delete was asked for and refused — see
+  [Retrying a refused snapshot delete](#retrying-a-refused-snapshot-delete---retry-pending).
+  Only meaningful when the query names a `snapshot_name`; queried by
+  `export_uuid` it is always `false`, because an export names a snapshot but is
+  not one.
 
 It accepts exactly one of two keys: `export_uuid`, or `snapshot_name` for a
 snapshot that may never have been exported (such a snapshot reports `NONE`).
@@ -365,4 +374,87 @@ is not disturbed by scheduler contention.
   while the decouple still reads through it"*. Wait for the decouple to finish
   (`rcow_get_decouple` status -- its list emptying is the signal) before taking
   the snapshot or clone.
+
+## Retrying a refused snapshot delete (`--retry-pending`)
+
+A snapshot that is pinned — an export names it, it has clones, or a decouple is
+running — cannot be deleted: `rcow_delete_lvol` refuses, and the refusal records
+a **pending-delete mark** for the snapshot. The mark is the *only* record that a
+delete was asked for: from the target's side every delete arrives as the same
+RPC, so there is no way to tell a user's delete from an automation's. The mark
+therefore is what "the user asked for this delete" means.
+
+Marks are keyed by **(lvstore uuid, lvol uuid)**, not by name: a name is unique
+only inside one loaded lvstore and is reusable, so a name-keyed mark could end up
+pointing at a snapshot the delete was never refused for.
+
+`rcow_get_lvstores` reports the mark per snapshot (`delete_pending`), together
+with whether the snapshot is deletable *right now* (`deletable`). Once the
+blocker clears — the export is released or expires, the extra clone is deleted —
+the snapshot becomes deletable, but the delete still has to be carried out by
+hand:
+
+```sh
+test/tools/s3lvol_rpc.py --retry-pending
+```
+
+It reads `rcow_get_lvstores`, finds every snapshot that **both** carries the
+pending-delete mark **and** is currently deletable, and calls `rcow_delete_lvol`
+for each — passing the snapshot's uuid alongside its name, so a delete meant for
+one object cannot land on a later one that reused the name:
+
+- A snapshot without the mark is **never touched** — snapshots a test or another
+  node created, or that were never deleted on purpose, are left alone.
+- A snapshot that is still pinned reports `deletable: NO` and is deliberately
+  skipped; run `--retry-pending` again once the blocker clears.
+- Each successful delete clears the mark on the target; a delete refused again
+  is reported and does not stop the others.
+
+The command takes no method argument. Its exit status is 0 when every
+marked-and-deletable snapshot went through, 1 if any delete failed, and it
+prints `no pending snapshot deletes to retry` when there is nothing to do:
+
+```sh
+$ test/tools/s3lvol_rpc.py --retry-pending
+no pending snapshot deletes to retry
+```
+
+### What this is not
+
+The mark is a record and a manual retry, nothing more. Specifically:
+
+- **Not every refusal records a mark.** Recorded are the snapshot blockers
+  `s3lvol_lvol_destroy()` itself identifies — an export pin (published or still
+  in flight), more than one clone, a running decouple — plus an asynchronous
+  destroy failure. Not recorded: the RPC-layer refusals that run before it (an
+  NVMf-active volume, an unreadable active registry), a bdev unregister that
+  fails, and the case where `spdk_blob_get_clones()` answers an unknown error.
+  Those are either the caller's own precondition to fix (deactivate first) or a
+  failure that has to be looked at rather than retried blindly.
+- **No automatic retry.** Nothing on the target polls the marks; there is no
+  deferred-completion poller. `--retry-pending` is the only thing that acts on
+  them, and it has to be run.
+- **In memory only.** The marks live in the target process and are gone on
+  restart. A delete refused before a restart is afterwards indistinguishable
+  from one that was never asked for.
+- **Dropped with the lvstore.** Unloading or deleting an lvstore drops that
+  lvstore's marks, deliberately: past the teardown they would name lvols that no
+  longer exist, and an lvstore attached again can give the same names to
+  different objects.
+- **No cancel.** Completing the delete is the only way to clear a mark; a
+  refused delete stays recorded for as long as the target runs and its lvstore
+  stays attached.
+- **`delete_pending` is a snapshot notion.** `rcow_get_snapshot_status` reports
+  it when queried by `snapshot_name`; queried by `export_uuid` it is always
+  `false`, because an export names a snapshot but is not one.
+- **The cluster path does not use it.** Cubelet's own delete
+  (`S3Cow.DeleteByKind`) still treats a refused snapshot delete as success, and
+  it does not run `--retry-pending`. This is an operator tool for the node; the
+  object leak on the cluster path is a separate change.
+
+`deletable: YES` means every blocker the delete path checks is clear right now —
+export pins, a running decouple, an NVMf-active volume, and more than one clone.
+It is a snapshot of the current state, not a promise: something can pin the
+snapshot again between the query and the delete, and a retry that is refused
+again is reported by `--retry-pending` (exit status 1) rather than hidden.
 
--- a/CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
+++ b/CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
@@ -49,6 +49,121 @@
  * table, and a node with more live exports than this has a different problem. */
 #define S3LVOL_MAX_EXPORTS 64
 
+/* === Pending-delete marks ==================================================
+ * A delete that cannot complete because the snapshot is referenced (an export
+ * pins it, it has clones, or a decouple is running) records the intent here.
+ * rcow_get_lvstores reports it so the caller sees the snapshot was "tried to
+ * delete" and can retry once the blocker clears. The deferred-completion poller
+ * is a later step.
+ *
+ * Keyed by (lvstore uuid, lvol uuid), not by name. A name is unique only within
+ * one loaded lvstore and is reusable: delete a snapshot and create another one
+ * by the same name, or unload the lvstore and attach it again, and a name-keyed
+ * mark would point at a different object than the one the delete was refused
+ * for -- which --retry-pending would then delete. A lvol uuid is generated once
+ * and never reused, so the mark can only ever name the object it was recorded
+ * for. The lvstore uuid groups the marks so a teardown can drop them all
+ * (s3lvol_snapshot_pending_clear_lvs()); the name is kept for log lines only.
+ */
+
+struct s3lvol_pending_delete {
+	struct spdk_uuid           lvs_uuid;
+	struct spdk_uuid           lvol_uuid;
+	char                       name[SPDK_LVOL_NAME_MAX];
+	TAILQ_ENTRY(s3lvol_pending_delete) link;
+};
+
+static TAILQ_HEAD(, s3lvol_pending_delete) g_pending_deletes =
+	TAILQ_HEAD_INITIALIZER(g_pending_deletes);
+
+static struct s3lvol_pending_delete *
+pending_delete_find(const struct spdk_uuid *lvs_uuid,
+		    const struct spdk_uuid *lvol_uuid)
+{
+	struct s3lvol_pending_delete *pd;
+
+	TAILQ_FOREACH(pd, &g_pending_deletes, link) {
+		if (spdk_uuid_compare(&pd->lvs_uuid, lvs_uuid) == 0 &&
+		    spdk_uuid_compare(&pd->lvol_uuid, lvol_uuid) == 0) {
+			return pd;
+		}
+	}
+	return NULL;
+}
+
+void
+s3lvol_snapshot_pending_set(const struct spdk_uuid *lvs_uuid,
+			    const struct spdk_uuid *lvol_uuid,
+			    const char *name)
+{
+	struct s3lvol_pending_delete *pd;
+
+	if (!lvs_uuid || !lvol_uuid) {
+		return;
+	}
+	if (pending_delete_find(lvs_uuid, lvol_uuid)) {
+		return;
+	}
+	pd = calloc(1, sizeof(*pd));
+	if (!pd) {
+		SPDK_ERRLOG("failed to allocate pending-delete mark for '%s'\n",
+			    name ? name : "(unnamed)");
+		return;
+	}
+	spdk_uuid_copy(&pd->lvs_uuid, lvs_uuid);
+	spdk_uuid_copy(&pd->lvol_uuid, lvol_uuid);
+	snprintf(pd->name, sizeof(pd->name), "%s", name ? name : "");
+	TAILQ_INSERT_TAIL(&g_pending_deletes, pd, link);
+}
+
+bool
+s3lvol_snapshot_pending_test(const struct spdk_uuid *lvs_uuid,
+			     const struct spdk_uuid *lvol_uuid)
+{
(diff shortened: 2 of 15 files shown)


CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -49,6 +49,78 @@
...
 
+/* === Pending-delete marks ==================================================
+ * A delete that cannot complete because the snapshot is referenced (an export
+ * pins it, it has clones, or a decouple is running) records the intent here.
+ * rcow_get_lvstores reports it so the caller sees the snapshot was "tried to
+ * delete" and can retry once the blocker clears. The deferred-completion poller
+ * is a later step.
+ */
+
+struct s3lvol_pending_delete {
```
fslongjin, line 69: Pending-delete marks are keyed by bare `name`, and `pending_clear` is only called on successful delete (`vbdev_s3lvol_lvstore.c:2620`). lvstore unload/destroy/free never touch `g_pending_deletes`.

Reproducible mis-delete: lvstore A's `snap0` is refused -> A is unloaded/destroyed -> a new `snap0` is created on another lvstore (test rerun, recovery, different namespace) -> `--retry-pending` sees `is_snapshot && delete_pending && deletable=YES` and deletes the new `snap0`. The README invariant "a snapshot without the mark is never touched" no longer holds, because a stale mark can bind to a new object.

Suggest keying marks by `(lvstore uuid, lvol uuid)` and clearing entries on lvstore teardown.

CubeS3lvol/test/tools/s3lvol_rpc.py
```diff
@@ -123,31 +167,131 @@ def main():
...
+            if (l.get("is_snapshot") and l.get("delete_pending")
+                    and l.get("deletable") == "YES"):
+                targets.append(l.get("name", ""))
+
+    if not targets:
+        print("no pending snapshot deletes to retry")
+        return 0
+
+    done = 0
+    for name in targets:
```
fslongjin: `rcow_get_lvstores` already returns `uuid` and `lvs_name` per lvol, but `--retry-pending` drops both and sends only `{"lvol_name": name}`. `rcow_delete_lvol`'s decoder (`vbdev_s3lvol_rpc.c:845-847`) only accepts `lvol_name` — it doesn't even decode `lvs_name` that `rpc_lookup_lvol()` supports.

Within the per-snapshot delete window (default 300s timeout, sequential deletes), the original snapshot can be deleted and a same-name one recreated, and the delete lands on the new object. The pin/clone re-checks in `s3lvol_lvol_destroy` don't prevent this — the new object can be `deletable=YES`.

Suggest `rcow_delete_lvol` accept and validate `uuid` (or at least `lvs_name`), and `--retry-pending` pass and check it.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_lvstore.c
```diff
@@ -2717,6 +2724,9 @@ s3lvol_lvol_destroy(struct spdk_lvol *lvol,
 		SPDK_ERRLOG("lvol '%s' is being exported right now; wait for "
 			    "rcow_get_snapshot_status to stop reporting INPROGRESS "
 			    "before deleting it\n", lvol->name);
+		if (lvol->blob && spdk_blob_is_read_only(lvol->blob)) {
+			s3lvol_snapshot_pending_set(lvol->name);
```
fslongjin: The README and header say "the refusal records a pending-delete mark", but only 5 paths actually call `pending_set`: inflight pin (`:2728`), live pin (`:2743`), `clone_count>1` (`:2817`), `action_in_progress` (`:2834`), async destroy failure (`:2650`).

These refusals do NOT mark:
- RPC-layer NVMf-active / active-registry read failure (`vbdev_s3lvol_rpc.c:1085-1099`) — never reaches `s3lvol_lvol_destroy`.
- `spdk_blob_get_clones` unknown error (`:2799-2806`) — returns -EBUSY without `pending_set`.
- bdev unregister failure (`:2668-2674`) — calls `cb_fn` and `free(ctx)` directly, bypassing `s3lvol_lvol_destroyed()`.
- `calloc` failure in `pending_set` itself only logs and returns; caller still gets EBUSY thinking it was recorded.

Either mark on those paths too, or tighten the documented claim to "only snapshot blocks identified inside `s3lvol_lvol_destroy` and async destroy failures are recorded".

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -1723,6 +1725,7 @@ s3lvol_snapshot_query(const char *snapshot_name,
 
 	*state = snapshot_export_state(owner, lvol->name);
 	*deletable = export_snapshot_deletable(owner, lvol->name);
+	*pending = s3lvol_snapshot_pending_test(lvol->name);
```
fslongjin: Two issues converge here.

1. `export_snapshot_deletable` (`:1582-1627`) claims to mirror every refusal in `s3lvol_lvol_destroy`, but only checks inflight pin, live pin, `clone_count<=1`. Missing: `action_in_progress`, NVMf-active (RPC layer). `!lvol || !lvol->blob` returns `true` outright (`:1601-1602`), so a closed-but-pinned snapshot reports `deletable: YES` yet the delete fails with EBUSY. `--retry-pending` treats `deletable=="YES"` as a go condition, so it can fail-and-exit-1 on something that looked deletable.

2. `*pending` is only written in the `snapshot_name` branch of `rcow_get_snapshot_status`. When queried by `export_uuid`, `s3lvol_export_query` is used and `pending` stays `false` (`vbdev_s3lvol_rpc.c:1837,1862-1863,1899`), yet `delete_pending` is always emitted. So `delete_pending` by uuid is always false — a footgun for anyone asserting on it.

Suggest tightening `deletable` to match the destroy refusals (or documenting the gap), and documenting that `delete_pending` is only meaningful with `snapshot_name`.

CubeS3lvol/README.md
```diff
@@ -366,3 +366,42 @@ is not disturbed by scheduler contention.
   (`rcow_get_decouple` status -- its list emptying is the signal) before taking
   the snapshot or clone.
 
+## Retrying a refused snapshot delete (`--retry-pending`)
```
fslongjin, line 378: This new section is accurate on the `--retry-pending` mechanics, but the surrounding README is now out of sync:
- `:222-232` still says `rcow_get_snapshot_status` returns two fields and the example JSON has two; now three, and `delete_pending` is always false by `export_uuid`.
- `:230-232` says `deletable` mirrors what `rcow_delete_lvol` would refuse; it doesn't (misses `action_in_progress`, NVMf-active).

Also missing from the README (things operators will hit):
- Marks are in-memory only and lost on `s3lvol_tgt` restart; the snapshot stays but `--retry-pending` then prints "no pending snapshot deletes to retry".
- No cancel: a refused delete becomes an irrevocable deferred delete.
- `--ls` exists (the main way to see the `PEND` column) but isn't mentioned.
- `delete_pending` is always false when `rcow_get_snapshot_status` is queried by `export_uuid`.
- The cluster delete path (Cubelet) does not run `--retry-pending`; a snapshot delete refused as busy is still treated as success and the object leaks.

`test/README.md` tools section (`:468-479`) and status field description (`:230-233`) are also not updated.

fslongjin (changes requested):
Requesting changes. The direction is right — recording refused snapshot deletes and retrying them later is valuable. But three issues block merge as-is.

**P0-1: Pending marks keyed by name only, survive lvstore teardown → possible mis-delete of a same-name new snapshot.**
`g_pending_deletes` stores bare `name`; `pending_clear` is only called on successful delete (`vbdev_s3lvol_lvstore.c:2620`). lvstore unload/destroy/free never touch the list. Reproducible: lvstore A's `snap0` refused → A unloaded → new `snap0` elsewhere → `--retry-pending` deletes the new one. The README invariant "a snapshot without the mark is never touched" does not hold.

**P0-2: `--retry-pending` / `rcow_delete_lvol` use only the name, dropping the uuid that `rcow_get_lvstores` already returns.**
`--retry-pending` sends only `{"lvol_name": name}`; `rpc_delete_lvol_decoders` (`vbdev_s3lvol_rpc.c:845-847`) only decodes `lvol_name`. Within the per-snapshot delete window (default 300s timeout), a same-name recreate can be deleted by mistake.

**P0-3: The core new behavior has no tests.**
Marking, skipping, retrying, clearing — none asserted in CI. The 26/26 green only proves old suites weren't broken by extra JSON fields, not that the new semantics work.

**Also should-fix:**
- PR body says "retried automatically on a safe window" and lists `--delete-pending`; neither exists in code (no poller; argparse has only `--retry-pending` and `--ls`).
- README still says `rcow_get_snapshot_status` returns two fields (now three; `delete_pending` is always false when queried by `export_uuid` — `vbdev_s3lvol_rpc.c:1837,1862-1863,1899`).
- `deletable: YES` is not a sufficient condition for the delete to succeed: `export_snapshot_deletable` misses `action_in_progress` and NVMf-active, and `!lvol->blob` returns `true` outright (`vbdev_s3lvol_xfer.c:1601-1602`), so a closed-but-pinned snapshot reports YES yet the delete fails with EBUSY.
- No cancel RPC: a refused delete becomes an irrevocable deferred delete.
- The production delete path (Cubelet `S3Cow.DeleteByKind`, `Cubelet/storage/s3_volume_manager.go:330-346`) still treats snapshot EBUSY as success and leaks objects — not introduced by this PR, but this PR's capability isn't wired into it. If the scope is ops CLI only, README should say so; if it's to fix the leak, Cubelet still needs to change.
- `Assisted-by: CodeBuddy` is missing `MODEL_VERSION` per `AGENTS.md`.

**Before merge, please at least:**
1. Key marks by `(lvstore uuid, lvol uuid)` and clear them on lvstore teardown; have `--retry-pending` delete by uuid and verify.
2. Add one dataplane regression: refused → marked → skipped while pinned → retried after blocker clears → mark cleared; unmarked snapshots untouched.
3. Pull the docs and PR body back to what the code actually does (no automatic retry, no `--delete-pending`; in-memory only, lost on restart, no cancel, `delete_pending` meaningless by uuid, cluster path won't run `--retry-pending`).

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_lvstore.c
```diff
@@ -652,6 +652,16 @@ s3lvol_lvstore_free(struct s3lvol_lvstore *lvs)
...
 		return;
 	}
+
+	/* Pending-delete marks belong to this lvstore's lvols, all of which are
+	 * gone by now: unload closed them, delete destroyed them. Leaving the
+	 * marks behind is what would let --retry-pending act on a same-named
+	 * snapshot in the lvstore that gets attached next. The single funnel for
+	 * unload, delete and the setup error paths, so one call covers all. */
+	if (lvs->lvs) {
+		s3lvol_snapshot_pending_clear_lvs(&lvs->lvs->uuid);
```
fslongjin: The guard is the problem. The normal unload success path sets `lvs->lvs = NULL` at `vbdev_s3lvol_lvstore.c:2112` **before** `s3lvol_lvstore_free()` runs, so `if (lvs->lvs)` is false here and `clear_lvs` is skipped. `rcow_delete_lvstore` goes through the same unload chain. The setup-error paths that do reach this with `lvs->lvs` non-NULL almost never have pending marks yet.

Net effect: marks survive unload/destroy and only go away on process exit. README ("Dropped with the lvstore") and the zh doc say otherwise. After unloading and re-attaching the **same** lvstore, `--retry-pending` still deletes a previously-marked snapshot — the opposite of "dropped with the lvstore".

Suggest calling `s3lvol_snapshot_pending_clear_lvs(&lvs->lvs->uuid)` before the `lvs->lvs = NULL` assignment in `s3lvol_lvs_unload_cb` (and the matching destroy path), or storing the uuid on `s3lvol_lvstore` itself so `free()` can always clear. A test for unload → re-attach → `--retry-pending` reports nothing would lock it in.

docs/zh/guide/cross-node-snapshot.md
```diff
@@ -284,7 +284,27 @@ cubeopscli --address 127.0.0.1 --port 3010 node list --json
 
 ## 5. 已知问题
 
-1. **S3 快照对象由 S3lvol 异步删除。** 删除 S3 快照后，CubeS3lvol 在后台回收对象。删除接口返回时，对象不一定已经从 S3 上消失。
+1. **S3 快照对象由 S3lvol 异步删除，被引用的快照会拒绝删除。** 删除 S3 快照后，CubeS3lvol 在后台回收对象。
+   删除接口返回时，对象不一定已经从 S3 上消失。
+
+   此外，快照在被引用时**无法删除**，CubeS3lvol 会以 `EBUSY` 拒绝：正在或已经导出（其他节点可能正
+   在读取）、存在多个 clone、正在 decouple、或仍作为 NVMe-oF 命名空间处于 active。此时 CubeS3lvol 会
```
fslongjin: The sentence lists "仍作为 NVMe-oF 命名空间处于 active" alongside the refusals that record a mark, implying it also records one. It does not: `rpc_rcow_delete_lvol` refuses an active volume at `vbdev_s3lvol_rpc.c:1166-1170` before `s3lvol_lvol_destroy` is ever called, so `destroy_mark_pending()` is never reached for that case. The English README's `--retry-pending` section is accurate (it lists export pin / clones / decouple, not NVMf-active).

Either drop NVMf-active from this list, or add `destroy_mark_pending()` on the RPC-layer refusal. The former is consistent with the rest of the PR's "only snapshot blocks inside `s3lvol_lvol_destroy` mark" stance.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_rpc.c
```diff
@@ -1550,6 +1625,33 @@ rpc_rcow_get_lvstores(struct spdk_jsonrpc_request *request,
...
+				spdk_json_write_named_uint64(w, "allocated_clusters",
+					spdk_blob_get_num_allocated_clusters(lvol->blob));
+			} else {
+				spdk_json_write_named_uint64(w, "total_clusters", 0);
+				spdk_json_write_named_uint64(w, "allocated_clusters", 0);
+			}
+			/* Same answer as rcow_get_snapshot_status by name: NONE for a
+			 * volume with no export. delete_pending is the recorded "a
+			 * delete was attempted but could not complete" mark. */
+			if (s3lvol_snapshot_query(lvol->name, &state, &deletable,
```
fslongjin: The loop already holds `lvol`, but `s3lvol_snapshot_query(lvol->name, …)` goes through `s3lvol_lvol_find_any()` — a scan of every loaded lvstore's lvol list — plus `spdk_blob_get_clones` per lvol. For N lvols in one store that's N full scans. When two lvstores contain a same-named lvol, `find_any` returns NULL, the query fails, and `export_status` / `deletable` / `delete_pending` are silently omitted from **both** entries — so `--retry-pending` can't see those marks at all, even though they're keyed by uuid and exist.

The state can be computed directly from `lvol` and `s3lvol_lvstore_of_lvol(lvol)` without the by-name re-resolve. Not blocking, but worth fixing so the listing is consistent in exactly the ambiguous-name case the delete RPC itself can't address either.

fslongjin (changes requested):
Thanks for the update. The three P0s from the previous round are addressed: marks are keyed by `(lvstore uuid, lvol uuid)`, `--retry-pending` carries the uuids and the target refuses a name-uuid mismatch, and the new `run_pending_delete_test.sh` covers refused → marked → skipped while pinned → retried after release → mark cleared, plus an unmarked control and a stale-uuid refusal. The scope is also stated honestly now: no poller, in-memory only, no cancel, Cubelet's path not wired in.

One new issue is worth fixing before merge: the lvstore teardown **does not actually clear the pending marks**, despite the code and docs claiming it does. `s3lvol_snapshot_pending_clear_lvs()` is called from `s3lvol_lvstore_free()`, but the normal unload path sets `lvs->lvs = NULL` (`vbdev_s3lvol_lvstore.c:2112`) before reaching `free()`, so the `if (lvs->lvs)` guard is false and `clear_lvs` never runs. `rcow_delete_lvstore` goes through the same unload. So:

- README and the zh doc say "unloading or deleting an lvstore drops that lvstore's marks";
- the marks actually survive until the process exits;
- after unloading and re-attaching the **same** lvstore (uuids unchanged), `--retry-pending` still deletes the previously-marked snapshot — the opposite of "dropped with the lvstore".

This is not a mis-delete of a different object (uuids protect against that), but it contradicts the documented behavior, and an operator who treats unload as "cancel the pending delete" gets surprised.

Two ways to resolve:

1. Call `s3lvol_snapshot_pending_clear_lvs(&lvs->lvs->uuid)` **before** `lvs->lvs = NULL` in `s3lvol_lvs_unload_cb` (and the matching path in `rcow_delete_lvstore`'s chain), or stash the uuid on `s3lvol_lvstore` itself so `free()` can always clear. Add a test: unload → re-attach → `--retry-pending` reports nothing.
2. Or, if marks are meant to survive unload and only protect against same-name rebinding, drop the "dropped with the lvstore" wording in README and the zh doc, and state that re-attaching the same lvstore still retries the marked snapshot.

Either is fine; the current code-doc split is the only thing I'd block on.

A few smaller follow-ups (not blocking):

- The zh cross-node doc still lists "still an active NVMe-oF namespace" as a refusal that records a mark. The code refuses that in `rpc_rcow_delete_lvol` before `s3lvol_lvol_destroy`, so it does **not** mark. The English README is accurate here.
- `rcow_get_lvstores` still calls `s3lvol_snapshot_query(lvol->name)` per lvol, which goes through `s3lvol_lvol_find_any()` — a full scan plus `spdk_blob_get_clones` for each lvol, and when two lvstores hold a same-named lvol the query fails and `export_status`/`deletable`/`delete_pending` are silently omitted from both. The state can be computed directly from the `lvol` already in hand.
- `Assisted-by: CodeBuddy` on `42157954` and `62cef224` is missing `MODEL_VERSION` per `AGENTS.md`.
- English README doesn't mention `--ls`; the `--ls` argparse help lists `NAME/SIZE/ALLOC/EXPORT/DEL/BDEV` but the actual table also has `TYPE` and `PEND`; `test/README.md` isn't updated for `--retry-pending` / `--ls` / the new status field.

fslongjin (approved):

PR #1537, opened 2026-08-27
s3lvol: follow-up review fixes and expired-export recovery

changpe1:
## Summary

Follow-ups to the merged pending-delete work (#1528), plus a load-time recovery for over-grown export registries. One commit, 6 files under `CubeS3lvol/` (+107/-82).

## What's in this PR

**1. `--retry-pending` drives the delete by `lvs_name`, uuids as confirmation**

The earlier version sent only the uuids as a *check* on a name-resolved lookup. `rpc_lookup_lvol()` resolves across every loaded lvstore, and when two lvstores share a name it answers "not found in any lvstore" — so exactly the ambiguous case the uuid keying exists for could never be retried, and its mark could never be cleared. `rcow_delete_lvol` now accepts `lvs_name`, and `--retry-pending` sends it from the listing.

**2. `destroy_mark_pending()` no longer tests snapshot-ness first**

With the blob closed that is not reliably answerable — `spdk_blob_get_clones()` only finds blobs that something still references as a parent, so after a re-attach a snapshot whose last clone was deleted has no entry at all. The refusals that reach the mark helper are already limited to the self-clearing blockers (export pin, extra clones, running decouple); `rcow_get_lvstores`' `is_snapshot` is informational only.

**3. `exports_parse()` drops expired reference exports and rewrites the registry**

Reference exports that were never released stay in the registry until their lease lapses, and nothing removed them: the count grew without bound, and the day it crossed `S3LVOL_MAX_EXPORTS` the whole lvstore refused to attach — even though not one of the entries was live. A node whose lvstore had 70 all-expired template exports was bricked exactly this way.

- Expired entries are dropped as the registry is decoded; when any are dropped, the S3 object is rewritten without them so it heals back to the live set.
- `S3LVOL_MAX_EXPORTS` goes 64 → 256, so the cap defends against a pathological registry rather than being crossed by ordinary accumulation.

## Verified

- On the bricked node (70 all-expired exports): attach now succeeds (82 volumes restored), and the registry is rewritten to `{"version":2,"exports":[]}` on first load.
- `run_pending_delete_test.sh` 30/30; full regression 30/30 suites (1133 assertions) against a real COS bucket.

### DIFF
--- a/CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
+++ b/CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
@@ -46,8 +46,16 @@
 #include "vbdev_s3lvol_json.h"
 
 /* Same reasoning as the import cap: the registry is read into a fixed decoder
- * table, and a node with more live exports than this has a different problem. */
-#define S3LVOL_MAX_EXPORTS 64
+ * table, and a node with more live exports than this has a different problem.
+ *
+ * Raised from 64 because that was not true in practice. An export leaves the
+ * registry when it is released, and a template-heavy node accumulates entries
+ * whose importers are long gone but which nobody released -- crossing the cap
+ * then made the whole lvstore un-attachable, live exports and all. The entries
+ * do stop pinning their snapshots once their lease goes stale (see the lease
+ * machinery below), so this bounds a registry nobody is maintaining rather
+ * than a working set. */
+#define S3LVOL_MAX_EXPORTS 256
 
 /* === Pending-delete marks ==================================================
  * A delete that cannot complete because the snapshot is referenced (an export
@@ -904,14 +912,37 @@ exports_parse(struct s3lvol_lvstore *lvs, const void *json, size_t len)
 				  S3_EXPORT_LAYOUT_REF : S3_EXPORT_LAYOUT_DENSE;
 		snprintf(exp->uuid_str, sizeof(exp->uuid_str), "%s", e->export_uuid);
 		snprintf(exp->snapshot, sizeof(exp->snapshot), "%s", e->snapshot);
+
 		TAILQ_INSERT_TAIL(&g_exports, exp, link);
+
+		/* Watch the lease, exactly as s3lvol_export_add() does for an export
+		 * created here: the importers this registry describes may still be
+		 * reading, and only the lease can say so.
+		 *
+		 * Without this the entry sits at lease_checked == false for ever --
+		 * nothing ever performs the first GET -- and s3lvol_export_pinning()
+		 * takes its "assume an importer may be reading" branch on every
+		 * query. Safe, but permanent: an export whose importer went away
+		 * years ago still pins its snapshot, and the registry only grows.
+		 * With the poller running, a lease that is absent or stale lets the
+		 * delete through, and releasing the export is what takes the entry
+		 * out of the registry.
+		 *
+		 * Deliberately *not* a filter on expires_at. That stamp is written
+		 * once at export creation and never refreshed, while the importer
+		 * renews its lease past the manifest TTL for as long as it is
+		 * reading -- s3lvol_export_pinning() ignores expires_at whenever a
+		 * fresh lease exists for precisely that reason. Dropping entries by
+		 * TTL here would forget a live importer's pin, and a later delete of
+		 * that snapshot would leave it reading holes. */
+		s3lvol_export_lease_start(exp);
 
 		if (exp->layout == S3_EXPORT_LAYOUT_REF) {
 			SPDK_NOTICELOG("lvstore '%s' still owes export %s the snapshot "
 				       "'%s'%s\n", s3lvol_lvstore_get_name(lvs),
 				       exp->uuid_str, exp->snapshot,
 				       s3lvol_export_is_expired(exp) ?
-				       " (expired, no longer pinned)" : "");
+				       " (deadline passed; the lease decides)" : "");
 		}
 	}
 
--- a/CubeS3lvol/test/tools/s3lvol_rpc.py
+++ b/CubeS3lvol/test/tools/s3lvol_rpc.py
@@ -223,32 +223,37 @@ def retry_pending_deletes(args):
         if not isinstance(lvs, dict):
             continue
         lvs_uuid = lvs.get("uuid", "")
+        lvs_name = lvs.get("lvs_name", "")
         for l in lvs.get("lvols") or []:
             # delete_pending is the record that a delete was asked for and
             # refused; deletable=YES means the blocker has cleared. Together
             # they are exactly "a user asked, it failed, retry now".
             #
-            # delete_pending alone establishes snapshot-ness: the target only
-            # ever records a mark for a snapshot. is_snapshot is not required
-            # on top of it -- and must not be, since a snapshot deactivated
-            # while it waited for its blocker to clear would then never be
-            # retried.
+            # is_snapshot is deliberately not required on top: the target only
+            # records a mark for a volume something still referenced (an export
+            # pin, a sibling clone, a running decouple), and is_snapshot cannot
+            # be answered once the blob is closed -- requiring it would skip a
+            # snapshot that was deactivated while it waited.
             if l.get("delete_pending") and l.get("deletable") == "YES":
                 targets.append((l.get("name", ""), l.get("uuid", ""),
-                                lvs_uuid))
+                                lvs_uuid, lvs_name))
 
     if not targets:
         print("no pending snapshot deletes to retry")
         return 0
 
     done = 0
-    for name, lvol_uuid, lvs_uuid in targets:
-        # The uuids come from the same rcow_get_lvstores answer the mark was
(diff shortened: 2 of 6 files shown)


CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -904,14 +912,37 @@ exports_parse(struct s3lvol_lvstore *lvs, const void *json, size_t len)
 				  S3_EXPORT_LAYOUT_REF : S3_EXPORT_LAYOUT_DENSE;
 		snprintf(exp->uuid_str, sizeof(exp->uuid_str), "%s", e->export_uuid);
 		snprintf(exp->snapshot, sizeof(exp->snapshot), "%s", e->snapshot);
+
 		TAILQ_INSERT_TAIL(&g_exports, exp, link);
 
+		/* Watch the lease, exactly as s3lvol_export_add() does for an export
+		 * created here: the importers this registry describes may still be
+		 * reading, and only the lease can say so.
+		 *
```
fslongjin, line 921: This uses only `s3lvol_export_is_expired()` → registry `expires_at` (the TTL stamped at export creation, never refreshed by lease renewals). A reference export can be wall-clock-expired while a live importer still holds a fresh lease: `import_lease_start()` renews `meta/exports/<uuid>.lease` at `max(remaining_ttl/3, 1)s` **indefinitely, even past the manifest TTL**, and `s3lvol_export_pinning()` ignores `expires_at` whenever a fresh lease exists (see ~592–631 in this file).

After a source-node restart, this loop drops such an export and `s3lvol_export_registry_save()` rewrites the registry to forget it **permanently**. The snapshot loses its pin; a later delete (or `--retry-pending`) is allowed and the importer reads holes. That is the "live importer becomes a deletable snapshot" case the lease grace period exists to prevent.

Before this PR, the loaded entry stayed in `g_exports` with `lease_checked == false`, so the conservative "assume an importer may be reading" branch kept it pinned — safe, if registry growth was the problem. Dropping on TTL alone flips to the unsafe direction.

Also: `exports_parse()` inserts via `TAILQ_INSERT_TAIL` and never calls `s3lvol_export_lease_start()` (unlike `s3lvol_export_add()` at ~703), so even kept entries do not get a lease poll on attach.

**Suggested direction:** HEAD the `.lease` object before dropping; only drop when absent or stale. Or keep the entry and start the lease poller. Do not rewrite the registry until that decision is complete. Call `s3lvol_export_lease_start()` for ref exports that remain after load.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -904,14 +912,37 @@ exports_parse(struct s3lvol_lvstore *lvs, const void *json, size_t len)
...
+		 * that snapshot would leave it reading holes. */
+		s3lvol_export_lease_start(exp);
+
 		if (exp->layout == S3_EXPORT_LAYOUT_REF) {
 			SPDK_NOTICELOG("lvstore '%s' still owes export %s the snapshot "
 				       "'%s'%s\n", s3lvol_lvstore_get_name(lvs),
 				       exp->uuid_str, exp->snapshot,
 				       s3lvol_export_is_expired(exp) ?
-				       " (expired, no longer pinned)" : "");
+				       " (deadline passed; the lease decides)" : "");
```
fslongjin, line 945: Two issues:

**(a)** The save is not gated on `rc == 0`. If the parse loop breaks early on `calloc` failure (`rc = -ENOMEM`) while `trimmed` is already true, the registry object is rewritten containing only the entries parsed so far — silently and permanently discarding live exports that come later in the array. Attach then fails and `s3lvol_lvstore_free()` clears memory, but S3 already holds the truncated file.

**(b)** `s3lvol_export_registry_save()` returns immediately while the trim PUT is in flight. A successful attach can then run `rcow_export_snapshot` / `rcow_release_export`, which also PUT the same key. There is no generation / if-match; the slower trim PUT can overwrite a newer registry (including an empty trim overwriting a post-attach live export — consistent with the bricked-node rewrite to `{"exports":[]}`).

**Suggested direction:** `if (trimmed && rc == 0) { save; }`. Serialize or generation-guard all writes to `meta/exports.json`. Consider not marking the lvstore fully serviceable until trim save completes, or run trim after attach with a shared write lock.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -904,14 +912,37 @@ exports_parse(struct s3lvol_lvstore *lvs, const void *json, size_t len)
...
+		 * Deliberately *not* a filter on expires_at. That stamp is written
+		 * once at export creation and never refreshed, while the importer
+		 * renews its lease past the manifest TTL for as long as it is
+		 * reading -- s3lvol_export_pinning() ignores expires_at whenever a
+		 * fresh lease exists for precisely that reason. Dropping entries by
+		 * TTL here would forget a live importer's pin, and a later delete of
+		 * that snapshot would leave it reading holes. */
+		s3lvol_export_lease_start(exp);
+
 		if (exp->layout == S3_EXPORT_LAYOUT_REF) {
```
fslongjin, line 940: `exports_parse()` now drops expired entries and rewrites `meta/exports.json` on S3. That is externally observable (attach success, bucket contents) and irreversible on the trim-success path. This PR does not add a suite or extend `run_export_test.sh` / `run_pending_delete_test.sh` for:

- mixed live + TTL-expired registry → attach succeeds, registry rewritten to live set only;
- TTL expired but `.lease` still fresh → entry must **not** be dropped;
- parse failure (`-ENOMEM`) → registry must **not** be rewritten.

Existing export tests cover pin/release, not load-time trim. Green `run_pending_delete_test.sh` does not prove this path.

CubeS3lvol/test/tools/s3lvol_rpc.py
```diff
@@ -223,32 +223,37 @@ def retry_pending_deletes(args):
...
-        # its predecessor.
+    for name, lvol_uuid, lvs_uuid, lvs_name in targets:
+        # lvs_name drives the lookup, the uuids verify it. Without lvs_name the
+        # target resolves by name across every loaded lvstore and answers "not
+        # found in any lvstore" when two of them share the name -- so exactly
+        # the ambiguous case the uuid keying exists for could never be retried.
+        # With it, the lookup is scoped to one lvstore and the uuids then
+        # confirm the object is the one the mark was recorded for, rather than a
+        # replacement that reused the name.
         params = {"lvol_name": name}
```
fslongjin, line 254: The `lvs_name` + uuid confirmation fix is sound in code, but `run_pending_delete_test.sh` still uses a single lvstore. The failure mode this fixes — two lvstores, same `lvol_name`, mark on one, retry must delete that one — has no assertion. Worth adding a step: two stores, same name, export pin on one, mark + release + `--retry-pending` must target the marked store only (and would fail without `lvs_name` in the delete RPC).

fslongjin (changes requested):
Thanks for the follow-ups to #1528. The two pending-delete fixes look right:

- **`lvs_name` drives `rcow_delete_lvol` lookup**, with uuids as confirmation — this fixes the case where two loaded lvstores share an lvol name and `rpc_lookup_lvol()` without `lvs_name` answers "not found in any lvstore", so the ambiguous case the uuid keying exists for could never be retried.
- **Removing the snapshot pre-check in `destroy_mark_pending()`** is consistent with "blob closed ⇒ cannot reliably tell snapshot-ness", and `--retry-pending` already keys off `delete_pending`, not `is_snapshot`.

I would **not merge the expired-export recovery as written**. Dropping entries in `exports_parse()` keyed only on registry `expires_at` (TTL), then rewriting `meta/exports.json`, can permanently unpin a snapshot that a live cross-node importer is still reading via `.lease` renewals — the exact hazard the lease machinery exists to prevent. The rewrite also runs without gating on parse success and races with later export/release PUTs. Details in the inline comments below.

**Suggested path:** split this PR — land the two #1528 follow-ups first; fix lease-aware trimming, save ordering, and tests before merging registry rewrite.

**Before merge (blocking on the export-recovery part):**

1. Do not drop a ref export on TTL alone while a fresh `.lease` exists; gate on lease (HEAD/GET) or keep the entry and start the lease poller so `s3lvol_export_pinning()` decides.
2. Only call `s3lvol_export_registry_save()` when `exports_parse()` returns `rc == 0`; serialize overlapping PUTs to `meta/exports.json`.
3. Add a dataplane (or offline) regression: mixed live/expired registry → attach succeeds and live entries survive; TTL-expired but lease-fresh entry must **not** be dropped; `-ENOMEM` mid-parse must **not** rewrite.

**Non-blocking follow-ups:** dual-lvstore same-name `--retry-pending` test; `Assisted-by: CodeBuddy` missing `MODEL_VERSION`; README still says retry sends only name+uuid; `rcow_decouple_lvol` silently decodes but ignores the new `lvs_name` fields via shared `rpc_delete_lvol_decoders`.

fslongjin (approved):

PR #1541, opened 2026-08-27
fix(snapshot): reject exited template init tasks

purebluesong:
## Summary

Prevent an app-snapshot template with a declared readiness probe from being
captured after the application has become unready in the gap between Create
returning and `cube-runtime` freeze.

- Immediately before the snapshot transaction, Cubelet runs the template's
  declared TCP/Ping/HTTP readiness handler once.
- The snapshot recheck is fail-fast: it skips the original initial delay and
  uses one attempt rather than replaying Create's full retry/threshold loop.
- A failed recheck cleans the temporary snapshot objects and returns
  `PreConditionFailed`, so the template is not published as `READY`.
- The probe wait also observes context cancellation, avoiding a blocked
  temporary sandbox when `telnet.Telnet` exits without publishing a result.

This is intentionally a Cubelet-only change. It does **not** change CubeShim
process watching or container task lifecycle handling.

## Scope

This fix applies when a template declares a readiness probe. Templates without
a declared probe preserve their existing app-snapshot behavior; this PR does
not add a `Task.Status` fallback or a new requirement that every template have
a probe.

The change narrows the practical product-level gap immediately before capture.
It does not claim to make guest liveness and VM capture atomic; a capture-time
runtime/shim acknowledgement would be a separate improvement.

## Validation

- Linux builder: targeted `Cubelet/services/cubebox` tests cover the immediate
  one-shot probe, no-probe skip, cancellation safety, and current snapshot
  freeze/rootfs/resume compatibility helpers.
- Repository CI passed on the current head, including Cubelet build and unit
  tests for amd64 and arm64, format, DCO, and Docker smoke builds.
- Earlier STG behavioral validation reproduced the original failure shape with
  a template health probe that succeeded once and then exited; the repaired
  compatibility backport rejected the template rather than publishing it
  `READY`, while normal restore and command execution still succeeded.

Assisted-by: Codex:GPT-5


ls-ggg:
Thanks for the PR — the problem is real. App-snapshot can currently seal a template after the guest init task has already exited, and a later restore then fails because that task is already gone.

On the Cubelet side, I would rather not add the 100ms dual `Status` check. Create already probes the guest before we continue into template snapshot, and the remaining race is the gap between Create returning and cube-runtime freeze. I would prefer a single readiness probe immediately before we invoke cube-runtime, using the probe the template already declared. That also shrinks the race window, and it matches the product definition of “the sandbox is ready” better than sampling containerd `Status` twice.

The one-shot probe before snapshot should be fail-fast (not a replay of Create’s full readiness wait). Images with no probe still need another signal, or we should require a probe for template build.

purebluesong:
## Follow-up on readiness review

Thanks for the direction. This revision keeps the CubeShim init-task watcher, but replaces the 100ms dual `Task.Status` sampling with a snapshot-scoped readiness gate.

- `AppSnapshot` now runs the template's declared readiness handler once immediately before `cube-runtime` is invoked. The one-shot revalidation has no initial delay and uses success/failure threshold 1.
- Templates without a probe use one watcher-backed containerd `Task.Status` check at the same point.
- The normal `doProbe` path no longer performs the app-snapshot-only no-probe stability wait, so ordinary sandbox creation retains its prior latency and does not require a task handle in that branch.
- Snapshot rejection is now reported as `PreConditionFailed`.

This narrows the practical gap to the point immediately before VM freeze while preserving the correct watcher-based host lifecycle signal. The remaining capture-time race is documented as a future runtime/shim-level atomicity improvement rather than claimed closed by this PR.

Validation added: one-shot TCP readiness ignores initial delay and multi-success thresholds; no-probe normal creation does not access a containerd task; task-status fallback covers running/stopped/error states. Targeted Linux tests pass. The full repository CI has been triggered.

purebluesong:
## Validation on latest master rebase

Validated on the current rebased PR head (`7c5a4b4`):

```text
Linux builder
go test ./services/cubebox -run "TestVerifyTaskRunning|TestSnapshotReadinessProbeRunsOnceWithoutInitialDelay|TestProbeWithoutHandlerDoesNotRequireContainerTask" -count=1
PASS
```

Coverage exercised by this command:

- watcher-backed task status accepts `Running`, rejects `Stopped`, and propagates status errors;
- the snapshot readiness TCP probe performs one immediate check, without the template initial delay or multi-success threshold;
- ordinary no-probe sandbox creation does not require a containerd task handle and does not inherit the former 100ms app-snapshot delay.

The retained CubeShim log-forward test subset also passed in the Linux builder (3/3).

A full local `services/cubebox` run reaches an existing NUMA test that requires `/sys/devices/system/node`, unavailable in the containerized builder. The rebased PR has triggered the repository CI for full-environment validation.

purebluesong:
## Follow-up: probe liveness fixes

Addressed the two blocking findings in the latest head (`0ddfba1`):

1. `doSnapshotReadinessProbe` no longer performs a bare receive from `telnet.Telnet`. It now selects on the probe result or `ctx.Done()`, so a cancelled/expired AppSnapshot context cannot leave the temporary sandbox blocked before deferred cleanup.

2. Declared readiness probes now complement, rather than replace, watcher-backed task status. Snapshot readiness is evaluated as:

```text
Task.Status == Running
→ one-shot declared readiness probe (if any)
→ Task.Status == Running
→ cube-runtime snapshot
```

No-probe templates retain a single task-status check.

Added Linux tests cover context cancellation with no probe result, init task exit during the one-shot probe, and the no-probe single-check path. The targeted `services/cubebox` test command passes on this revision. Full CI has been retriggered.

purebluesong:
## STG behavioral E2E

In addition to the Linux CI, the app-snapshot failure mode was exercised against a deployed v0.6 compatibility backport in a private Kubernetes environment. The current upstream PR head uses a newer runtime architecture, so it was not mixed into that older deployment; this validates the same watcher + liveness algorithm with a real MicroVM.

The failure fixture starts a minimal HTTP server. Its first `/health` request returns HTTP 204, then the init process exits immediately. This guarantees that the original Create-time readiness probe has succeeded before the snapshot path observes the exit.

Observed result:

```text
dead-init template status: FAILED
error: init task ... not found
normal existing template: restore succeeded
Shell: appsnapshot-shell-ok
Python: 424242
```

Cleanup verification after the run:

```text
active sandbox count: 0
failed test template: absent from template list
compute node: Ready, original STG image restored/unchanged
```

This confirms the key failure is rejected rather than producing a READY template, while the normal template restore and command paths remain functional.

ls-ggg:
Please rebase onto current master first — the branch has drifted and review/CI should land on the latest tree.

On the approach: keep this to the **minimum Cubelet change**. Do not change the shim.

The CubeShim `wait_process` registration during `app_snapshot_create` should stay as it is today (skipped). Opening that watcher for template builds is a process-lifetime change: `TaskExit` goes through the normal exit path (`task.Delete` / `WithProcessKill`) with no AppSnapshot in-flight guard, and it can race cube-runtime while the VM is being frozen. That is more risk than this bug needs.

Rely on the template’s **declared readiness probe**, once, immediately before `cube-runtime` is invoked — fail-fast (no initial delay, no Create-style retries/thresholds). That is the product definition of “ready” and it closes the practical gap between Create returning and freeze.

Images with no probe: do not add a Cubelet/containerd `Task.Status` fallback and do not turn on the shim watcher to invent one. Either require a probe for template build, or leave those images on the existing Create path only.

So the intended diff is: Cubelet `verifyAppSnapshotReadiness` (or equivalent) = one-shot probe before snapshot, `PreConditionFailed` on failure. No `CubeShim/shim/src/container/mod.rs` change.

PR #1542, opened 2026-08-27
fix(cubelet): do not export template disks on AppSnapshot

ls-ggg:
## Summary
- Stop exporting template packages from `AppSnapshot`. Template disks stay node-local; Pause / CommitSandbox still export the sandbox snapshot package.
- Remove `UploadTemplateRootfs` and the tests that required a template rootfs export. That export held the object (`deletable=NO`) so template delete reported busy and leaked the rootfs snapshot.

## Test plan
- [x] `gofmt` on the four changed Go files
- [x] Cubelet storage tests that previously covered `UploadTemplateRootfs` are gone; remaining upload tests still pass
- [x] Create an S3 template and delete it: template list empty, template lvols gone except the node-local metadata base; new template rootfs stays `EXPORT=NONE`
- [x] Same-node Pause → Resume: guest `echo ok`, marker, and `dd` pass
- [x] Same-node Snapshot → FromSnap: guest marker and `dd` pass
- [x] Cross-node Pause → Resume: guest marker and `dd` pass
- [x] Cross-node Snapshot → FromSnap: guest marker and `dd` pass

### DIFF
--- a/Cubelet/services/cubebox/appsnapshot.go
+++ b/Cubelet/services/cubebox/appsnapshot.go
@@ -449,19 +449,8 @@ func (s *service) AppSnapshot(ctx context.Context, req *cubebox.AppSnapshotReque
 		stepLog.Warnf("s3 finalize package snapshots %s failed: %v", templateID, err)
 	}
 
-	// Export the rootfs snapshot only. Every sandbox rootfs is a child of
-	// it, and a child's export references only chunks that already have a
-	// committed S3 object — the parent's are silently skipped, so without
-	// this a cross-node restore imports a rootfs with no ext4 metadata at
-	// all (measured: 17 objects／7.2MB instead of 80／74.9MB; the imported
-	// disk has no root inode). Memory and metadata of a template are not in
-	// that chain and stay node-local. Best-effort: a template that fails to
-	// export still works on this node.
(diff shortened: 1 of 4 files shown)


ls-ggg:
Cluster regression after this change (template disks are not exported):

- Create an S3 template and delete it: the template is gone from the list; template volumes are gone except the node-local metadata base. New template rootfs stays `EXPORT=NONE`, so delete no longer hits busy or leaks the object.
- Same-node Pause → Resume: guest `echo ok`, a filesystem marker, and a 2MiB `dd` write all passed.
- Same-node Snapshot → FromSnap: guest marker and `dd` passed.
- Cross-node Pause → Resume: guest marker and `dd` passed.
- Cross-node Snapshot → FromSnap: guest marker and `dd` passed.

Re: review Finding 2 — cross-node restore of template-derived sandboxes is still supported. Pause / CommitSandbox export the sandbox snapshot package; the guest root is an overlay (full memory snapshot plus a writable upper). Those flows were retested after this PR and passed. Template disks stay node-local by design so template delete is not blocked by an export pin.

Finding 1 (`os` import) is fixed in this squashed commit. Finding 3 (`waitExportSettled` / serial-export comments) is left on purpose for a possible serial switch later.

PR #1549, opened 2026-08-27
feat(cubemastercli): support deleting multiple templates

luzhixing12345:
allow tpl delete to accept multiple positional template IDs. Continue deleting the remaining templates after an individual failure and return the combined errors afterward.

```bash
cubemastercli tpl delete <tpl-1> <tpl-2> ...
```


fslongjin:
For easier troubleshooting, it would be better to also include the template ID in the rolling failure logs, so we can tell right away which template failed.

luzhixing12345:
> For easier troubleshooting, it would be better to also include the template ID in the rolling failure logs, so we can tell right away which template failed.

done

fslongjin (approved):

PR #1552, opened 2026-08-28
fix(cubelet): name sandbox TAPs after IPv4 and reject oversize names

FakeLearne:
The z+IPv4 scheme produced 16-byte names (e.g. z192.168.195.183). vishvananda/netlink LinkAdd copies into ifr_name[:15] and truncates silently, so create could succeed while later open/query used a different string than the kernel device.

New creates use the dotted IPv4. Recover still accepts leftover z-prefixed names when the decoded address is in the configured sandbox CIDR. Create fails before LinkAdd if the name is >= IFNAMSIZ.

### DIFF
--- a/Cubelet/network/runtime/startup_recover.go
+++ b/Cubelet/network/runtime/startup_recover.go
@@ -777,16 +777,24 @@ func (s *NetworkController) claimRecoveredSuccessResources(state *managedState) 
 // uniquely from the IP, so no other goroutine can re-reference this tap between
 // the check and the destroy.
 //
-// Lookup is by deterministic tap name (RTM_GETLINK), not LinkList. A full dump
-// of every host interface on every pool-miss create races with concurrent TAP
-// churn and surfaces as netlink ErrDumpInterrupted under density load.
+// Lookup is by the current TAP name then the legacy z-prefixed name
+// (RTM_GETLINK), not LinkList. A full dump of every host interface on every
+// pool-miss create races with concurrent TAP churn and surfaces as netlink
+// ErrDumpInterrupted under density load.
 func (s *NetworkController) cleanupConflictingTap(ip net.IP) error {
-	tap, err := s.tapAdapter.GetByName(tapName(ip.String()))
-	if err != nil {
-		if isTapNotFound(err) {
-			return nil
+	var tap *tapDevice
+	for _, name := range candidateTapNames(ip.String()) {
+		found, err := s.tapAdapter.GetByName(name)
+		if err != nil {
+			if isTapNotFound(err) {
+				continue
+			}
+			return err
 		}
-		return err
+		if found != nil {
+			tap = found
+			break
+		}
 	}
 	if tap == nil {
 		return nil
--- a/Cubelet/network/runtime/tap_device.go
+++ b/Cubelet/network/runtime/tap_device.go
@@ -10,6 +10,7 @@ import (
 	"fmt"
 	"net"
 	"os"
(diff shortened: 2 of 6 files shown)


Cubelet/network/runtime/startup_recover.go
```diff
@@ -777,16 +777,24 @@ func (s *NetworkController) claimRecoveredSuccessResources(state *managedState)
...
-		if isTapNotFound(err) {
-			return nil
+	var tap *tapDevice
+	for _, name := range candidateTapNames(ip.String()) {
+		found, err := s.tapAdapter.GetByName(name)
+		if err != nil {
+			if isTapNotFound(err) {
+				continue
+			}
+			return err
```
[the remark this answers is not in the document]
FakeLearne: That makes sense.

Cubelet/network/runtime/tap_device.go
```diff
@@ -444,14 +457,53 @@ func deletePersistentTapByName(name string) error {
...
-// extractIP reverses tapName and rejects unrelated host TAP devices.
-func extractIP(name string) (string, error) {
-	if len(name) <= len(tapNamePrefix) || name[:len(tapNamePrefix)] != tapNamePrefix {
-		return "", fmt.Errorf("not cube tap: %s", name)
+// candidateTapNames is the lookup order for an allocated sandbox IP: the
+// current name first, then the leftover z-prefixed device after upgrade.
+func candidateTapNames(ip string) []string {
+	current := tapName(ip)
+	legacy := legacyTapName(ip)
+	if current == legacy {
```
chenhengqi: What does this check mean?

Cubelet/network/runtime/tap_device.go
```diff
@@ -128,6 +138,9 @@ func (realTapDeviceAdapter) Destroy(ifIdx int) error {
 func newTap(ip net.IP, mvmMacAddr string, mtu, cubeDevIdx int) (_ *tapDevice, retErr error) {
 	logger := CubeLog.WithContext(context.Background())
 	name := tapName(ip.String())
+	if err := validateTapName(name); err != nil {
```
chenhengqi: This seems unnecessary.

Cubelet/network/runtime/tap_device.go
```diff
@@ -444,14 +457,53 @@ func deletePersistentTapByName(name string) error {
...
+		return fmt.Errorf("tap name %q is %d bytes; kernel IFNAMSIZ is %d (max %d usable)",
+			name, len(name), unix.IFNAMSIZ, unix.IFNAMSIZ-1)
+	}
+	return nil
+}
+
+// parseCubeTapIP reverses tapName / legacyTapName and rejects unrelated host
+// TAP devices. Both "192.168.0.10" and "z192.168.0.10" map to the same IPv4.
+func parseCubeTapIP(name string) (net.IP, error) {
+	if ip := net.ParseIP(name).To4(); ip != nil {
```
chenhengqi: Always trim prefix and parse IP?

FakeLearne: done

Cubelet/network/runtime/tap_device.go
FakeLearne: done

Cubelet/network/runtime/tap_device.go
FakeLearne: done

chenhengqi (approved):

PR #1571, opened 2026-08-29
fix(sdk): honour e2b files.read formats across Node/Python/Go

### LINKED ISSUE #1570: [Bug Report] Node SDK files.read ignores format and always UTF-8-decodes (breaks e2b-compatible bytes/blob/stream)
## Summary

The Node SDK's `Filesystem.read` always calls `resp.text()` and returns a `string`. Callers that pass the e2b-compatible `format: 'bytes' | 'blob' | 'stream'` option are silently ignored, so binary files are corrupted before the application sees them.

This breaks drop-in / e2b-surface compatibility for any consumer that already writes:

```ts
const bytes = await sandbox.files.read(path, { format: 'bytes' })
```

## Measured

Against a live sandbox, a 4-byte PNG magic (`89 50 4E 47`) read through the current Node SDK arrives as a UTF-8 string whose bytes are `EF BF BD 50 4E 47` (the leading `0x89` became the U+FFFD replacement character). Length grows; the file is no longer a PNG.

The official e2b Node SDK returns an unaltered `Uint8Array` for the same call.

## Expected (e2b surface)

| `options.format` | Return type |
| --- | --- |
| omitted / `text` (default) | `string` |
| `bytes` | `Uint8Array` |
| `blob` | `Blob` |
| `stream` | `ReadableStream<Uint8Array>` |

## Environment

- CubeSandbox ≥ 0.7.0
- `@cubesandbox/sdk` Node package from `sdk/node` (still unpublished to npm; consumed as a build artifact)

## Why this matters

Deployments that speak CubeSandbox through the e2b-shaped client (and HamsterHQ in particular) serve panel image/binary downloads through `files.read({ format: 'bytes' })`. The docker/e2b simulation path works; the Cube path does not, for the same source line.

## Proposed fix

Honour `format` in `sdk/node/src/filesystem.ts` the same way e2b does (including the empty `content-length: 0` short-circuit). Unit tests covering `text` / `bytes` / `blob` / `stream` will land with the PR.

Happy to send the PR.

HuChundong:
## Summary

`files.read` in the CubeSandbox SDKs always treated the response as text, so the e2b-compatible binary formats were missing or ignored. Binary payloads (e.g. PNG `0x89`) were corrupted on the Node/Python UTF-8 text paths.

This PR aligns all three official SDKs with the e2b surface:

| SDK | Surface |
| --- | --- |
| **Node** | `format: 'text' \| 'bytes' \| 'blob' \| 'stream'` |
| **Python** | `format="text" \| "bytes" \| "stream"` → `str` / `bytearray` / `Iterator[bytes]` |
| **Go** | `Read` (text) + `ReadBytes` + `ReadStream` (e2b has no Go SDK; these are the idiomatic equivalents) |

Also mirrors e2b's empty `content-length: 0` short-circuit where applicable.

Fixes #1570

## Test plan

- [x] `cd sdk/node && npm test -- --run` (221 passed)
- [x] `cd sdk/go && go test ./...`
- [x] `cd sdk/python && pytest tests/test_sandbox.py -k read_` (7 passed, via container)
- [ ] Reviewer: optional live check — write a PNG into a sandbox and read it back with the binary API unaltered

### DIFF
--- a/sdk/go/README.md
+++ b/sdk/go/README.md
@@ -131,6 +131,7 @@ Consume output via **either** `Output()` **or** `Wait(onData)`, not both — the
 ```go
 // Read & write
 content, err := sb.Files().Read(ctx, "/etc/hosts")
+png, err := sb.Files().ReadBytes(ctx, "/tmp/chart.png")
 err = sb.Files().Write(ctx, "/tmp/hello.txt", []byte("hi"))
 
 // Execute all filesystem operations as a specific sandbox user.
@@ -166,7 +167,9 @@ for ev := range watcher.Events {
 | Method | Description |
 |---|---|
 | `ForUser(user)` | Return an immutable view that runs all filesystem operations as `user` |
-| `Read(ctx, path)` | Download file content via `GET /files` |
+| `Read(ctx, path)` | Download file content as text via `GET /files` (e2b `format=text`) |
+| `ReadBytes(ctx, path)` | Download raw bytes (e2b `format=bytes`) |
+| `ReadStream(ctx, path)` | Stream download; caller must `Close` (e2b `format=stream`) |
 | `Write(ctx, path, data)` | Upload via `POST /files` (octet-stream, multipart fallback) |
 | `WriteFiles(ctx, entries)` | Batch write, stops on first error, returns count |
 | `List(ctx, path)` | List directory entries via `ListDir` RPC |
--- a/sdk/go/envd.go
+++ b/sdk/go/envd.go
@@ -115,19 +115,17 @@ func (s *Sandbox) startProcess(ctx context.Context, payload processStartRequest,
 }
 
 func (s *Sandbox) readFile(ctx context.Context, path string, options ...fileRequestOption) (string, error) {
-	if err := s.ensureClient(); err != nil {
-		return "", err
-	}
-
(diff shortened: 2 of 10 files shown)


HuChundong:
Thanks for the review — follow-ups in 92f27b1:

1. **e2b call shape (finding 1):** Verified against the published e2b Node SDK (`files.read(path, opts?)` with `opts.format`). Format is **not** positional; `read(path, { format: 'bytes' })` is the e2b surface. Doc comment updated to say so explicitly. A JS caller writing `read(path, 'bytes')` would also silently fall back to text on e2b itself (`opts?.format ?? "text"`).

2. **`content-length: 0` (finding 2):** Left as-is — same short-circuit e2b uses for empty bodies. Observable only via `blob.type` on an empty file.

3. **Overload unions (finding 3):** Noted; left for a follow-up if dynamic `format` variables become common.

4. **Error path (finding 4):** Added a unit test for the non-200 JSON error body.

Also fixed the Code Format Check failure (extra blank line in `sdk/go/sdk_test.go`).

liciazhu:
LGTM, e2e verified.

liciazhu:
Suggest adding e2e test cases under /CubeSandbox/tests/e2e/sdk_compat, or supplementing with related test results. @HuChundong

PR #1575, opened 2026-08-31
refactor(scheduler): remove CubeMaster overcommit ratio

kinwin-ustc:
Remove CubeMaster-side `overcommit_ratio` / `overcommit_ratio_conf`. Filters and scorers now use Cubelet-reported `QuotaCpu` / `QuotaMem` directly. Leftover YAML keys are parsed but ignored with a startup warning.

Assisted-by: Cursor Grok 4.6:cursor-grok-4.6

### DIFF
--- a/CubeMaster/conf.yaml
+++ b/CubeMaster/conf.yaml
@@ -82,20 +82,9 @@ scheduler:
   # during scheduling. Defaults to false: account for the Redis allocation
   # records when computing schedulable capacity.
   ignore_redis_allocation: false
-  # Global CPU/Mem overcommit ratio applied to the node-reported quota capacity.
-  # Defaults to CPU=3, Mem=2.
-  overcommit_ratio:
-    cpu_ratio: 3.0
-    mem_ratio: 2.0
-  # Per-instance-type overcommit ratio (takes precedence over the global
-  # overcommit_ratio).
-  # overcommit_ratio_conf:
-  #   cubebox:
-  #     cpu_ratio: 6.0
-  #     mem_ratio: 4.0
-  #   cubebox_gpu:
-  #     cpu_ratio: 1.0
-  #     mem_ratio: 1.0
+  # Deprecated: overcommit_ratio / overcommit_ratio_conf are no longer used.
+  # Leftover keys here are ignored; CubeMaster logs a warning at startup
+  # if they are present.
   filter:
     enable_filters:
       - "cpu"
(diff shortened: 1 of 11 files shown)


fslongjin:
I verified the bot's findings against the code at `e35bca91`. These three are real, with some extra context:

## 1. The physical load guards' baseline is now also inflated by the overcommit ratio

Cubelet writes the ratio-multiplied value into `node.Status.Capacity` (`Cubelet/pkg/cubelet/node_status.go:636`), and `CpuTotal` / `MemMBTotal` are derived from that Capacity (`CubeOps/internal/nodemanagement/service/snapshot.go:240`). So the intentionally preserved physical backstops — `loadMemFree` (`CubeMaster/pkg/selector/filter/memfilter.go:50`), the CPU load sanity check (`CubeMaster/pkg/selector/prefilter/prefilter.go:73`), and the load-based scorers (`CubeMaster/pkg/selector/score/utils.go`) — all get an inflated baseline.

Concrete effect: on an 8 GiB machine with 4 GiB used, `loadMemFree` computes as 12 GiB, so a 10 GiB request can still pass the "physical" memory check.

It's less severe than the bot made it sound, though: for nodes without explicit `mcpu_limit` / `mem_limit`, Capacity was already CPU ×2 / mem ×1.25 before this PR (now ×3 / ×2), so that path is only further inflated. Only explicitly configured nodes actually go from physical values to overcommitted ones.

Ideally the load guards would use a separately reported physical total. At minimum the docs should state that this baseline no longer reflects physical resources — the current upgrade notes only cover the reserve side.

## 2. Default (host-derived) nodes lose schedulable capacity: CPU halved, memory −20%

For nodes with empty `mcpu_limit` / `mem_limit`, the effective capacity changes:

- CPU: was cores×2000m then ×3 on the master = 6000m/core; now cores×1000m×3 = **3000m/core**
- Memory: was host×1.25 then ×2 = 2.5×; now host×**2**

After overcommit moved into Cubelet, the host-derived base dropped from 2000m / ×5⁄4 to 1000m / ×1. This is documented, but existing clusters hit it without touching any config, and scheduling pressure will rise noticeably after upgrade. Worth a prominent release-note callout, or aligning the new defaults with the old effective capacity.

## 3. Affinity labels `memory_size` / `cpu_cores` now advertise overcommitted values

`CubeMaster/pkg/base/node/node.go:278` builds these labels straight from `n.QuotaMem` / `n.QuotaCpu`. A 3-physical-core machine now reports `cpu_cores: 9000m`, so an affinity request for an "8-core" node matches it. The labels used to describe hardware specs; now they describe overcommitted capacity. Either base them on physical values, or document the changed semantics.

---

All three trace back to the same root cause: the ratio was supposed to apply to quota only, but it now leaks into Capacity, the load-guard baselines, and the affinity labels — fields that were meant to stay physical.

fslongjin (approved):

PR #1578, opened 2026-08-31
fix(examples): improve code-sandbox-quickstart for local dev and crea…

AkatQuas:
## Summary
- Integrate optional `e2b-dev-sidecar` in `env_utils.py` when `CUBE_REMOTE_PROXY_BASE` is set, enabling data-plane access from outside the cluster without `*.cube.app` DNS
- Add `apply_create_time_envs()` workaround for cubebox/VNC templates where create-time `envs` do not survive entrypoint startup
- Update `create_with_envs.py` to use `printf` and re-apply envs after create; document sidecar env vars in `.env.example`
- Add `aiohttp`, `httpx`, and `cubesandbox` to quickstart requirements

## Test plan
- [x] `python3 cmd.py` with `.env` (`CUBE_REMOTE_PROXY_BASE=http://<node-ip>`)
- [x] `python3 create_with_envs.py` prints `session is user-session-test`
- [x] `python3 debug_local.py` (local helper, not included in this PR)


AkatQuas:
@lkml-likexu  This PR is a fork of https://github.com/TencentCloud/CubeSandbox/pull/1293 . Thanks for review.

luzhixing12345:
The new code still has a few minor issues, but they are acceptable for an example and do not need to block this PR.

examples/code-sandbox-quickstart/README_zh.md
luzhixing12345, line 111: @AkatQuas Could you help update the expected output of exec_code.py in README.md and README_zh.md? The actual output I obtained does not match what is shown here; this is a lingering issue.

AkatQuas: Cursor helps and updates the commit.

AkatQuas:
@luzhixing12345  , Could you please approve my PR when you get a chance? Thanks!

luzhixing12345:
@AkatQuas Thanks for the update. I think this is generally acceptable as an example, and the remaining robustness/style issues do not need to block the PR. There is one issue I recommend fixing before merge, though:

When `CUBE_REMOTE_PROXY_BASE` is not set, `ensure_dev_sidecar()` returns without patching the SDK. In that path, `sandbox.get_host(port)` returns a scheme-less host such as `49983-<sandbox-id>.cube.app`, and `apply_create_time_envs()` unconditionally turns it into `http://49983-<sandbox-id>.cube.app/init`

The request body contains the complete `envVars` map and may also carry `e2b-traffic-access-token`. Consequently, users who replace the demo values with real tokens or credentials can send them over plaintext HTTP. That's not correct as an example for new beginners.

Please distinguish the two cases explicitly: use HTTP only for the locally running dev sidecar, and default to HTTPS when the sidecar is inactive, with an explicit scheme override for deployments that intentionally use plaintext HTTP. For example

```python
host = sandbox.get_host(port)

if "://" in host:
  base = host
elif _SIDECAR_READY:
  base = f"http://{host}"
else:
  base = f"https://{host}"
```

AkatQuas:
@luzhixing12345  Change updated. 😄

luzhixing12345:
Merged. Thanks for your contribution.

PR #1582, opened 2026-08-31
fix(hotswap): use bytes.Equal and handle yaml.Marshal error in reload

### LINKED ISSUE #1470: [Bug Report] Config hot-reload uses bytes.EqualFold, so case-only edits are silently ignored
## Summary

`FileOperator.reload` compares the old and new marshalled config with `bytes.EqualFold`, so any edit
that only changes ASCII letter case is treated as "no change" and never applied. The
`yaml.Marshal` error on the same line is also unchecked.

## Environment

- Relevant component: CubeMaster and Cubelet (identical copies of the package)

## Steps to Reproduce

1. Start CubeMaster (or Cubelet) with a hot-swappable YAML config containing
   `log:\n  level: INFO`.
2. Edit the file in place to `log:\n  level: info` and save.
3. The file watcher fires, but no listener is notified and the running config is unchanged.

Any case-only edit reproduces it — for example a hostname `Foo.internal` → `foo.internal`, or a
driver name `MySQL` → `mysql`.

## Expected Behavior

The edit is detected and pushed to listeners, as it would be for any other content change.

## Actual Behavior

`reload()` returns `false` and nothing happens. Reproduced against the change-detection expression:

```
config edit                  reloaded?
log level INFO -> info       false       <== edit SILENTLY IGNORED
host Foo -> foo              false       <== edit SILENTLY IGNORED
driver MySQL -> mysql        false       <== edit SILENTLY IGNORED
genuine change (control)     true
length differs (control)     true
```

## Additional Context

`CubeMaster/pkg/base/hotswap/file.go:88-100` and the identical
`Cubelet/pkg/hotswap/file.go:88-100`:

```go
new, err := yaml.Marshal(config)
if len(o.backupOldCfg) == len(new) && bytes.EqualFold(o.backupOldCfg, new) {
    return false
}
```

`bytes.EqualFold` compares under Unicode simple case-folding — it is for case-insensitive *text*
matching, not byte identity.

Second defect on the same line: the `yaml.Marshal` error is never checked (`staticcheck SA4006`). The
sibling `Init()` at `:71-74` *does* check it, so the asymmetry within the same file looks like an
oversight. On a marshal failure `new` is nil, so `o.backupOldCfg` is set to nil and
`o.notify(config)` still fires — the change-detection baseline is wiped and every subsequent reload
reports "changed".

aniruddhaadak80:
Fixes #1470

## Summary
FileOperator.reload compared marshalled config with bytes.EqualFold, so any edit that only changes ASCII letter case was treated as "no change" and never applied. The yaml.Marshal error on the same line was also unchecked (staticcheck SA4006).

Repro:
- log level INFO -> info => reload() returns false (silently ignored)
- host Foo.internal -> foo.internal => same
- genuine content change => true (control)
- length differs => true (control)

Root cause: CubeMaster/pkg/base/hotswap/file.go:94 and identical Cubelet/pkg/hotswap/file.go:94

```
new, err := yaml.Marshal(config)
if len(o.backupOldCfg) == len(new) && bytes.EqualFold(o.backupOldCfg, new) { return false }
```

EqualFold is Unicode case-folding, not byte identity. Error from Marshal never checked -- on failure new is nil, so backupOldCfg is wiped to nil and notify still fires.

## Fix
- Check yaml.Marshal error, log via o.Errorf and return false
- Use bytes.Equal for exact comparison (handles length internally)
- Apply to both copies (CubeMaster and Cubelet -- they are vendored copies)

Mirrors Init() at 71-74 which already checks the marshal error.

## How tested
- git diff shows 10 insertions, 2 deletions across 2 files.
- Manual inspection: bytes.Equal is exact, case-sensitive; error path logs and aborts.
- No other files use EqualFold for config comparison (grep -r EqualFold confirms only these two).

Autonomously-by: Muse Spark:muse-spark-1.2-contributor-free

### DIFF
--- a/CubeMaster/pkg/base/hotswap/file.go
+++ b/CubeMaster/pkg/base/hotswap/file.go
@@ -91,7 +91,11 @@ func (o *FileOperator) reload(path string) bool {
 		return false
 	}
 	new, err := yaml.Marshal(config)
-	if len(o.backupOldCfg) == len(new) && bytes.EqualFold(o.backupOldCfg, new) {
+	if err != nil {
+		o.Errorf(context.Background(), "Marshal config fail:%v", err)
+		return false
+	}
+	if bytes.Equal(o.backupOldCfg, new) {
 		return false
 	}
 	o.backupOldCfg = new
--- a/Cubelet/pkg/hotswap/file.go
+++ b/Cubelet/pkg/hotswap/file.go
@@ -91,7 +91,11 @@ func (o *FileOperator) reload(path string) bool {
 		return false
 	}
 	new, err := yaml.Marshal(config)
-	if len(o.backupOldCfg) == len(new) && bytes.EqualFold(o.backupOldCfg, new) {
+	if err != nil {
+		o.Errorf(context.Background(), "Marshal config fail:%v", err)
+		return false
+	}
+	if bytes.Equal(o.backupOldCfg, new) {
 		return false
 	}
 	o.backupOldCfg = new


aniruddhaadak80:
Thanks for the thorough review @cubesandboxbot!

Acknowledged on the test suggestion: `hotswap` currently has no tests (both CubeMaster and Cubelet copies) and `reload` depends on `fsnotify`/`os.ReadFile` scaffolding, so a direct unit test needs a helper extraction. I kept this PR minimal to the exact parity fix (EqualFold -> Equal + marshal error) and fixed both vendored copies consistently to avoid one-sided drift, as you noted.

If maintainers prefer, I'm happy to follow up with a small pure helper (e.g. `shouldReload(old, new []byte) bool`) and a unit test asserting:
- case-only change (INFO -> info) returns true
- identical config returns false
in a separate PR.

DCO: per `AGENTS.md`/`CONTRIBUTING.md` AI MUST NOT add `Signed-off-by`, the human submitter will add it when merging. The `Autonomously-by` tag is present per policy.

Let me know if you'd like that test helper in this PR or as a follow-up!

PR #1590, opened 2026-08-31
fix(cubelet): add GetSandboxEvents gRPC to fix sandbox event 

liciazhu:
Fix: add a GetSandboxEvents RPC to the CubeboxMgr gRPC service. CubeMaster now resolves the sandbox's host IP via localcache and proxies the request to the owning Cubelet, which reads the shim log locally. This makes event viewing work across single-node, multi-node, and K8s deployments without any new network dependency.The WebUI now displays event timestamps converted to the browser's local timezone.


liciazhu:
Clipboard_Screenshot_1788332171

CubeMaster/pkg/service/httpservice/cube/sandbox_logs.go
```diff
@@ -5,54 +5,40 @@
...
 
 // SandboxLogsReq is the request body for POST /cube/sandbox/logs.
 type SandboxLogsReq struct {
 	types.Request
 	SandboxID    string `json:"sandboxID"`
 	InstanceType string `json:"instanceType,omitempty"`
-	// Cursor is a Unix millisecond timestamp; only return entries after this time.
-	Cursor int64 `json:"cursor,omitempty"`
-	// Limit is the maximum number of entries to return (default 200, max 2000).
+	// Limit is the maximum number of newest entries to return (default 200, max 1000).
```
fslongjin, line 38: This is a silent breaking change for existing callers.

**Before:** `/cube/sandbox/logs` supported `cursor`, returned `nextCursor` / `hasMore`, read from the **beginning** of the file, and capped at **2000** entries.

**Now:** always returns the **newest N** entries, no cursor fields, cap is **1000**. Unknown `cursor` is ignored instead of rejected.

That may be better for WebUI, but SDK/CubeOps callers can get the wrong window without noticing.

Please either:
- **A.** restore cursor pagination (or a compatible subset), or
- **B.** mark this as breaking in OpenAPI/changelog and reject unknown `cursor` with a clear error.

CubeMaster/pkg/service/httpservice/cube/sandbox_logs.go
```diff
@@ -115,94 +94,97 @@ func handleSandboxLogsAction(c *gin.Context) {
...
+	}
+	calleeEp := cubelet.GetCubeletAddr(hostIP)
+
+	cmReq := &cubebox.GetSandboxEventsRequest{
+		RequestID: req.RequestID,
+		SandboxID: req.SandboxID,
+		Limit:     int32(limit),
+	}
+	rsp, err := cubelet.GetSandboxEvents(c.Request.Context(), calleeEp, cmReq)
 	if err != nil {
```
fslongjin, line 120: When `hostIP == ""` you return `ConnHostFailed` (transient, retryable). But here any `GetSandboxEvents` transport failure is mapped to `MasterInternalError`, even though `actions.go` already wraps connection failures as `ConnHostFailed`.

Suggestion: use `ret.FromError(err)` and preserve retryable codes like `ConnHostFailed`.

Also, if the Cubelet is too old and returns gRPC `Unimplemented`, users currently only see `"failed to fetch sandbox events from cubelet"`, which is hard to diagnose during rolling upgrades.

CubeMaster/pkg/service/httpservice/cube/sandbox_logs.go
```diff
@@ -115,94 +94,97 @@ func handleSandboxLogsAction(c *gin.Context) {
...
+		common.WriteAPI(c, &SandboxLogsRes{Res: &types.Res{Ret: &types.Ret{
+			RetCode: int(publicCode),
+			RetMsg:  publicMessage,
+		}}})
+		return
+	}
+
+	entries := make([]SandboxLogEntry, 0, len(rsp.GetEvents()))
+	for _, e := range rsp.GetEvents() {
+		entries = append(entries, SandboxLogEntry{
```
fslongjin, line 152: Cubelet returns `SandboxEvent.module`, but the HTTP response drops it and only keeps `timestamp` / `message` / `level`.

If HTTP is not meant to expose `module`, please document that. Otherwise callers may assume the field was lost accidentally.

CubeMaster/pkg/service/httpservice/cube/sandbox_logs.go
```diff
@@ -115,94 +94,97 @@ func handleSandboxLogsAction(c *gin.Context) {
...
+}
 
-		entries = append(entries, SandboxLogEntry{
-			Timestamp: entry.Timestamp,
-			Message:   entry.LogContent,
-			Level:     "info",
-		})
-		nextCursor = tsMs
+// resolveSandboxHostIP returns the node host IP, skipping entries with an empty HostIP.
+func resolveSandboxHostIP(c *gin.Context, sandboxID string) string {
```
fslongjin, line 178: This duplicates `sandbox.resolveSandboxHostIP` in `sandbox_update.go`, and the behavior is not identical:

- here: skip cache/proxy entries with empty `HostIP`
- existing helper: may return empty IP from cache
- here: `pausesnap.GetBySandbox` directly
- existing helper: `resolvePauseHostIP`

Two concerns:
1. **Maintenance** — host-resolution rules can drift between exec/logs paths.
2. **False empty results** — on cache miss, falling back to an old pause node can route to a machine that no longer hosts the sandbox. Cubelet treats a missing log file as **success + empty list**, so users see "no events" instead of "wrong node".

Exec does not use pausesnap as a generic fallback. I'd align with that for live events, or make paused-sandbox history an explicit branch.

CubeMaster/pkg/cubelet/actions.go
```diff
@@ -157,6 +157,22 @@ func GetStorageMetrics(ctx context.Context, calleeEp string,
 	return c.GetStorageMetrics(ctx, req)
 }
 
+// GetSandboxEvents fetches events from the Cubelet owning the sandbox.
+func GetSandboxEvents(ctx context.Context, calleeEp string,
```
fslongjin, line 161: `GetSandboxEvents` is a new RPC. During rolling upgrade, an old Cubelet will return `Unimplemented` if Master is upgraded first.

Please:
1. detect `codes.Unimplemented` and return a clear message like `cubelet does not support GetSandboxEvents; upgrade cubelet first`
2. document upgrade order in release notes: **Cubelet first, then CubeMaster**

CubeMaster/pkg/cubelet/actions.go
```diff
@@ -157,6 +157,22 @@ func GetStorageMetrics(ctx context.Context, calleeEp string,
...
+	req *cubebox.GetSandboxEventsRequest) (*cubebox.GetSandboxEventsResponse, error) {
+	conn, err := grpcconn.GetWorkerConn(ctx, calleeEp)
+	if err != nil {
+		return nil, ret.Err(errorcode.ErrorCode_ConnHostFailed, err.Error())
+	}
+	defer conn.Close()
+	c := cubebox.NewCubeboxMgrClient(conn.Value())
+	ctx, cancel := context.WithTimeout(ctx,
```
fslongjin, line 169: Minor: `DestroyWithTimeout` in this file falls back when `timeout <= 0`, but here `WithTimeout` is applied directly. If `CommonTimeoutInsec == 0`, the call will time out immediately. Worth matching the other RPC helpers.

Cubelet/services/cubebox/sandbox_events_rpc.go
```diff
@@ -0,0 +1,191 @@
...
+	"github.com/tencentcloud/CubeSandbox/pkgs/proto/services/errorcode/v1"
+)
+
+const (
+	// defaultShimReqLogPath is CubeShim's node-local JSON log.
+	// Keep in sync with CubeShim/shim/src/log/mod.rs LOG_DIR + LOG_FILE.
+	defaultShimReqLogPath = "/data/log/CubeShim/cube-shim-req.log"
+
+	// defaultEventLimit / maxEventLimit mirror CubeMaster's old sandbox_logs.go
+	// constants to keep API pagination behaviour unchanged.
```
fslongjin, line 27: This comment says pagination behavior is unchanged, but pagination was removed. The semantics also changed from "read from file head" to "return newest N entries".

Please update the comment so future readers don't assume cursor support still exists.

Cubelet/services/cubebox/sandbox_events_rpc.go
```diff
@@ -0,0 +1,191 @@
...
+	limit := req.GetLimit()
+	if limit <= 0 {
+		limit = defaultEventLimit
+	}
+	if limit > maxEventLimit {
+		limit = maxEventLimit
+	}
+
+	events, err := readShimEventsNewest(ctx, shimReqLogPath(), sandboxID, limit)
+	if err != nil {
```
fslongjin, line 96: Non-`IsNotExist` read failures (permission denied, directory path, `ReadAt` error, cancelled context) all become `"failed to read shim log"`, and the real `err` is not logged on the Cubelet side. After Master sanitizes again, this is very hard to debug in production.

Please log `path + err` internally. Only `IsNotExist` should be treated as "no events yet".

Also, `readShimEventsNewest` may already have partial events when `ctx` is cancelled, but this handler discards them entirely.

Cubelet/services/cubebox/sandbox_events_rpc.go
```diff
@@ -0,0 +1,191 @@
...
+// non-standard installs) and falls back to defaultShimReqLogPath.
+func shimReqLogPath() string {
+	if p := strings.TrimSpace(os.Getenv("CUBELET_SHIM_REQ_LOG")); p != "" {
+		return p
+	}
+	return defaultShimReqLogPath
+}
+
+// readShimEventsNewest returns up to limit newest events, oldest first.
+func readShimEventsNewest(ctx context.Context, logPath, sandboxID string, limit int32) ([]*cubebox.SandboxEvent, error) {
```
fslongjin, line 121: Not a merge blocker, but the reverse chunk scanner is heavier than it needs to be. Since reads are already capped at 16MB, a simpler approach would be: read the last 16MB once, split lines, filter by `InstanceId`, take the newest N.

That would be easier to maintain and reason about. Fine to follow up later if you prefer.

pkgs/proto/services/cubebox/v1/cubebox.proto
```diff
@@ -1222,3 +1224,36 @@ message CleanupOrphanStorageFilesResponse {
...
 }
+
+// GetSandboxEventsRequest queries a sandbox's newest node-local events.
+message GetSandboxEventsRequest {
+  // requestID reqID
+  string requestID = 1;
+  // Sandbox ID whose events should be returned. Required.
+  string sandboxID = 2;
+  // Limit caps the number of entries returned (default 200, max 1000).
+  int32 limit = 4;
```
fslongjin, line 1235: `limit` is field 4 and field 3 is skipped with no `reserved 3;`. That looks like a removed cursor field that wasn't cleaned up.

For a new message, please either use `limit = 3` or add `reserved 3;`.

pkgs/proto/services/cubebox/v1/cubebox.proto
```diff
@@ -1222,3 +1224,36 @@ message CleanupOrphanStorageFilesResponse {
...
+}
+
+// SandboxEvent is one parsed line from the per-sandbox shim req log.
```
fslongjin, line 1238: Comment says "per-sandbox shim req log", but the implementation reads a **node-shared** `cube-shim-req.log` and filters by JSON `InstanceId`.

Please fix the comment so ops don't expect one file per sandbox.

pkgs/proto/services/cubebox/v1/cubebox.proto
```diff
@@ -1222,3 +1224,36 @@ message CleanupOrphanStorageFilesResponse {
...
+message SandboxEvent {
+  // RFC3339Nano timestamp as written by the shim.
```
fslongjin, line 1240: Comment says timestamp is "as written by the shim", but Cubelet parses it and re-emits **UTC** RFC3339Nano (`ts.UTC().Format(...)`).

Please document the normalized UTC output and that invalid timestamps are skipped.

CubeAPI/src/services/sandboxes.rs
```diff
@@ -399,12 +401,12 @@ impl SandboxService {
     pub async fn get_logs(
         &self,
         sandbox_id: &str,
-        start: Option<i64>,
+        _start: Option<i64>,
```
fslongjin, line 404: v1 still accepts `start`, but it is renamed to `_start` and ignored. OpenAPI may still document `start` as pagination.

Callers sending `start` will think they are paging through logs, but they will always get the newest N entries.

Please either restore `start` semantics or remove it from the public schema and note the breaking change in changelog.

web/src/pages/SandboxDetail.tsx
```diff
@@ -32,7 +32,8 @@ function isNotFoundError(error: unknown): boolean {
 function formatLogTime(ts: string): string {
```
fslongjin, line 32: `formatLogTime` only renders `HH:mm:ss.mmm` without a date. Events crossing midnight or timezone boundaries can look out of order.

If that's intentional, maybe note it in UI copy; otherwise consider showing the date or inserting a day separator.

web/src/pages/SandboxDetail.tsx
```diff
@@ -63,7 +64,7 @@ export default function SandboxDetailPage() {
   // ── Logs ────────────────────────────────────────────────────────────────
   const logs = useQuery({
     queryKey: ['sandbox-logs', sandboxID],
-    queryFn: () => sandboxApi.logs(sandboxID),
+    queryFn: () => sandboxApi.logs(sandboxID, { limit: 1000 }),
```
fslongjin, line 67: This panel polls every 10 seconds (`limit: 1000`), not a realtime stream. The in-page `logsDesc` text is accurate, but `docs/guide/webui.md` still says users can "watch its logs stream in real time" — worth updating in docs.

fslongjin (changes requested):
Thanks for the PR — proxying shim event reads to the owning Cubelet is the right fix for multi-node/K8s. Reusing the existing Cubelet gRPC path without adding new ports looks good to me, and the error sanitization is an improvement over the old implementation.

I don't see obvious new security issues in this change. Before merge, though, there are a few user-visible / operability problems worth fixing:

1. **Breaking API change (silent)**: cursor pagination was removed, but CubeOps/OpenAPI still expose `cursor`/`start`. Callers may think pagination still works.
2. **Rolling upgrade**: old Cubelets will return `Unimplemented` for the new RPC; the error message should make that obvious.
3. **Host resolution**: duplicated `resolveSandboxHostIP` plus pausesnap fallback can route to the wrong node and return an empty success page.
4. **Docs / API drift**: user guides still say logs cannot be fetched remotely; CubeAPI still has "endpoint pending" fallbacks.

Architecture looks good overall. Happy to re-review once the inline comments below are addressed.

---

**Additional notes (files/lines outside this diff):**

- `CubeAPI/src/services/sandboxes.rs` (`get_logs` / `get_logs_v2`): `is_endpoint_missing()` still returns fake placeholder log lines even though CubeMaster now implements `/cube/sandbox/logs`. Please remove those fallbacks.
- `CubeAPI/src/models/mod.rs`: default `limit` is **1000** in CubeAPI vs **200** in CubeMaster/Cubelet — please align or document.
- `CubeOps/internal/handler/sdk.go:476-479` — still forwards `?cursor=` even though Master no longer reads it.
- `CubeAPI/src/cubemaster/mod.rs:22-25, 194-195` — still says the logs endpoint is pending and describes stdout/stderr.
- `web/src/locales/en/sandboxDetail.json` / `zh` — panel title is Events, but empty/error strings still say "log(s)".
- `docs/guide/sandbox-logs.md` (+ zh) — still says logs cannot be fetched remotely via API.
- `docs/guide/webui.md` (+ zh) — still says realtime log streaming.
- `docs/changelog/` (+ zh) — no entry for this feature / API change / upgrade order.
- `openapi.yml` / `web/src/api/generated/schema.ts` — still document `start` / `cursor` / `direction`.

PR #1595, opened 2026-09-01
refactor(cubemaster): consolidate CPU/mem overcommit into Cubelet

### LINKED ISSUE #1565: [Bug Report] Overcommit ratios not enforced under concurrent creation — node can be massively over-scheduled
## Summary

Under concurrent batch creation, cubemaster admits far more sandboxes onto a node than its configured overcommit ratio should allow, because scheduling decisions are made against a stale, 1s-polled resource snapshot. This is compounded by cubelet applying its own default overcommit factor on top of whatever ratio is configured on cubemaster, and by the fact that cubelet's creation path never re-checks local resources before booting a VM — so nothing catches the over-allocation until the node is already oversubscribed, risking OOM on pre-existing sandboxes.

## Environment

- CubeSandbox version / commit: `master` branch (as of this report)
- Deployment mode: single-node
- Relevant component: CubeMaster, Cubelet

## Steps to Reproduce

1. Provision a node with 64 vCPU / 128GB memory, sandbox spec 2 vCPU / 2.0GB, overcommit ratios left at default (`cpu_ratio: 3`, `mem_ratio: 2`).
2. Run `./cube-bench -c 10 -n 300 -w 3 -mode create-only` against it (10 concurrency, 300 create requests).
3. Observe how many requests succeed vs. the theoretical ceiling.

## Expected Behavior

With the compounded default overcommit (cubelet's own default factor — CPU ×2, memory ×1.25 — multiplied by cubemaster's `overcommit_ratio` of CPU ×3, memory ×2), the node's true ceiling is **160 sandboxes** (memory-bound: `128GB × 1.25 × 2 ÷ 2.0GB`). Requests beyond that should be rejected with `no more resource`.

## Actual Behavior

286 / 300 requests succeeded — well beyond the 160 ceiling. Only 14 were rejected:

```
HTTP 500: {"code":500,"message":"CubeMaster returned error code 130597: no more resource"}
```

Bench summary: Completed 286/300, Errors 14, QPS 75.0, Avg Create 136ms, Elapsed 3.9s.

**Root causes, with code references:**

1. **Scheduler reads a stale snapshot.** `cpuFilter.Select`/`memFilter.Select` (`CubeMaster/pkg/selector/filter/{cpufilter,memfilter}.go`) each run a quota check and a physical-load check, but both read from the same `inList[i]` node object, refreshed only every ~1s from Redis. Concurrent requests within that window all see the same pre-allocation numbers and all pass the filter independently.

2. **Cubelet's default overcommit compounds with cubemaster's.** In `Cubelet/pkg/cubelet/node_status.go`, `resolveHostQuotaCPUMilli`/`resolveHostQuotaMemMB` apply a default factor (CPU ×2, mem ×1.25) to the node's *reported* quota when `hostCfg.Quota.Cpu/Mem` isn't explicitly set. Cubemaster's `EffectiveQuotaCpu`/`EffectiveQuotaMem` (`CubeMaster/pkg/base/config/config.go`) then multiplies that already-inflated value by its own `overcommit_ratio` — so the effective overcommit an operator gets by default is 6× CPU / 2.5× memory, not the 3×/2× implied by `overcommit_ratio` alone.

3. **No local admission check in the cubelet creation path.** `checkResource` (check.go:152) only validates request format, the cgroup step only computes quota (not physical usage), and `prepare_resource`/`create_vm`/`boot_vm` (sb.rs:750, 918-919) never check local `/proc/meminfo` before booting. Over-allocation is only ever caught if/when cloud-hypervisor's mmap allocation itself fails — and under Linux memory overcommit + demand paging, freshly-created sandboxes that haven't written their full declared memory yet don't trigger that failure at all, so the over-allocation stays invisible until multiple sandboxes become active simultaneously, at which point the node can hit real OOM and take down pre-existing sandboxes.

## Additional Context

Two design questions worth discussing:

- Would the community consider a K8s-style `requests`/`limits` split (with QoS tiers) instead of a single ratio scaling the whole node, so overcommit risk lands on lower-priority sandboxes rather than uniformly?
- Would it make sense for cubelet to add a local admission check (similar to kubelet's) that re-validates actual available resources right before `create_vm`/`boot_vm`, as a last line of defense against stale scheduling decisions?

Happy to help test or discuss further.

devincd:
## Summary

Fixes #1565

## Motivation

When sandboxes are created concurrently in batches, CubeMaster admits far more sandboxes onto a node than the configured overcommit ratios should permit (issue #1565). One of the root causes is that CPU/mem overcommit is applied **twice**, so the effective overcommit is much larger than what an operator configures:

- **Cubelet** bakes a hardcoded overcommit factor into the node-reported quota when deriving it from host hardware   (CPU ×2, Mem ×1.25).
- **CubeMaster** then multiplies that already-inflated value by its own
  `overcommit_ratio` (default CPU ×3, Mem ×2).

The compounded default becomes **6× CPU / 2.5× memory** instead of the 3× / 2× the operator expects, so the scheduler admits roughly twice the intended number of sandboxes onto a node.

Overcommit is fundamentally a per-node decision — it depends on the node's hardware, workload, and local policy — so it belongs at the node agent (Cubelet), not in the centralized scheduler. Consolidating it in one place also removes the double-application footgun and leaves room for future Cubelet-side resource reservation.

  ## What this PR does

Remove CPU/mem overcommit from CubeMaster entirely. The scheduler now treats the Cubelet-reported `quota_cpu` / `quota_mem_mb` as the final schedulable capacity, since that value already carries Cubelet's overcommit factor.

- `CubeMaster/pkg/base/config/config.go`: remove the `OvercommitRatio` / `OvercommitRatioByType` fields, the `OvercommitRatioConf` type, the `GetEffectiveOvercommitRatio` / `sanitized` / `isValidRatio` / `EffectiveQuotaCpu` / `EffectiveQuotaMem` / `floatToInt64Clamped` helpers, the default ratio constants, and the default-value injection in `preHandleScheduler`. `EffectiveAllocated` (the `ignore_redis_allocation` path) is kept untouched.
- `CubeMaster/pkg/selector/filter/cpufilter.go`, `memfilter.go`: compute `quotaCpuFree` / `quotaMemFree` directly from the node-reported `QuotaCpu` / `QuotaMem` instead of the overcommit-adjusted effective quota.
- `CubeMaster/pkg/selector/score/utils.go`, `realtimescore.go`: score against `n.QuotaCpu` / `n.QuotaMem` directly.
- `CubeMaster/pkg/base/config/config_test.go`: drop the overcommit and `floatToInt64Clamped` tests; keep and rename the `EffectiveAllocated` and ignore-defaults coverage; drop the now-unused `math` import.
- `CubeMaster/conf.yaml`: remove the `overcommit_ratio` / `overcommit_ratio_conf` block.
- `docs/guide/cubemaster-scheduler-config.md` (+ zh): remove the overcommit field reference, clarify that the reported quota already carries Cubelet's overcommit, and drop stale overcommit guidance in the multi-node deploy guide.

  ## Notes

- Cubelet is intentionally untouched; its existing quota-derivation overcommit strategy is preserved as the single source of overcommit.
- Existing deployments with a stale `overcommit_ratio` key in `cubemaster.yaml` are unaffected: the YAML loader ignores unknown fields, so the stale key is silently dropped rather than failing startup. Operators can safely remove it from their config.

### DIFF
--- a/CubeMaster/conf.yaml
+++ b/CubeMaster/conf.yaml
@@ -82,20 +82,6 @@ scheduler:
   # during scheduling. Defaults to false: account for the Redis allocation
   # records when computing schedulable capacity.
   ignore_redis_allocation: false
-  # Global CPU/Mem overcommit ratio applied to the node-reported quota capacity.
-  # Defaults to CPU=3, Mem=2.
-  overcommit_ratio:
-    cpu_ratio: 3.0
-    mem_ratio: 2.0
-  # Per-instance-type overcommit ratio (takes precedence over the global
-  # overcommit_ratio).
-  # overcommit_ratio_conf:
-  #   cubebox:
-  #     cpu_ratio: 6.0
-  #     mem_ratio: 4.0
-  #   cubebox_gpu:
-  #     cpu_ratio: 1.0
-  #     mem_ratio: 1.0
   filter:
     enable_filters:
       - "cpu"
--- a/CubeMaster/pkg/base/config/config.go
+++ b/CubeMaster/pkg/base/config/config.go
@@ -251,12 +251,6 @@ type SchedulerConf struct {
 	// 0). A pointer is used so an unset value can default to false while still
 	// allowing operators to explicitly enable it. Defaults to false.
 	IgnoreRedisAllocation *bool `yaml:"ignore_redis_allocation"`
-	// OvercommitRatio is the global CPU/Mem overcommit ratio applied to the
-	// node-reported quota during scheduling. Defaults to CPU=3, Mem=2.
-	OvercommitRatio *OvercommitRatioConf `yaml:"overcommit_ratio"`
-	// OvercommitRatioByType overrides OvercommitRatio for specific instance
-	// types and takes precedence over the global ratio.
-	OvercommitRatioByType map[string]OvercommitRatioConf `yaml:"overcommit_ratio_conf"`
 }
 
 var defaultNodeAffinitySelectorAllowedKeys = []string{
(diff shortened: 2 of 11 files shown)


fslongjin:
please add some verification results on a real cluster. And, this PR is duplicated with https://github.com/TencentCloud/CubeSandbox/pull/1575

devincd:
> please add some verification results on a real cluster. And, this PR is duplicated with #1575

Got it, thanks for pointing that out — I'll close this PR then, since #1575 already covers it.

PR #1597, opened 2026-09-01
fix(cubelog): accept case-insensitive log levels

### LINKED ISSUE #1593: [Bug Report] CubeMaster silently downgrades shipped log level "info" to DEBUG
## Summary

CubeMaster's shipped configuration files all use lowercase `level: "info"`, but `CubeLog.StringToLevel` matches only uppercase names and falls back to `DEBUG` for anything else. Every default deployment therefore runs at DEBUG instead of INFO, which writes request signature material to disk, enables caller reporting, and inflates request-log volume.

## Environment

- CubeSandbox version / commit: `9270cabfc048f25b8ce411895b2dc5443956a6b0`
- Host OS and kernel version: Ubuntu 22.04.4 LTS, Linux `6.6.69-opencloudos9.cubesandbox.pvm.host-gb85200d80fa2`
- KVM info (`modinfo kvm`): `/lib/modules/6.6.69-opencloudos9.cubesandbox.pvm.host-gb85200d80fa2/kernel/arch/x86/kvm/kvm.ko`, `intree: Y`, vermagic matches the running kernel
- Deployment mode: affects **all shipped modes** — single-node, Kubernetes chart, and Terraform
- Relevant component: CubeMaster (parsing defect in `pkgs/CubeLog`)
- Build/test environment: builder image `cube-sandbox-cubemaster-builder:ubuntu2004`, Go `1.25.7`

## Steps to Reproduce

1. Check out `9270cabf` and use any shipped CubeMaster configuration unchanged, e.g. `configs/single-node/cubemaster.yaml`:

   ```yaml
   log:
     module: "cubemaster"
     path: "/data/log/CubeMaster"
     file_size: 100
     file_num: 10
     level: "info"
   ```

2. Start CubeMaster so that `log.Init()` runs.
3. Observe that `log.IsDebug()` is true and that DEBUG-only request logging is emitted.

The parse chain can be verified in isolation with the probe under **Test Report** below.

## Expected Behavior

The shipped lowercase value `info` should resolve to the `INFO` level.

Log levels should be accepted case-insensitively, or unsupported values should be rejected with a clear configuration error. They should not silently fall back to `DEBUG`.

## Actual Behavior

The YAML decoder preserves the value verbatim — `Conf.UnmarshalYAML` in `CubeMaster/pkg/base/log/types.go` only reconciles snake_case/camelCase *key* spellings and never normalizes the value. It is then passed straight through:

- `CubeMaster/pkg/base/log/log.go:23` — `CubeLog.SetLevel(CubeLog.StringToLevel(conf.Level))` at startup
- `CubeMaster/pkg/base/log/log.go:64` — the same conversion during configuration hot reload

`CubeLog.StringToLevel` (`pkgs/CubeLog/logger.go:146-159`) matches only uppercase:

```go
switch level {
case "DEBUG":
    return DEBUG
case "INFO":
    return INFO
case "WARN":
    return WARN
case "ERROR":
    return ERROR
case "FATAL":
    return FATAL
default:
    return DEBUG
}
```

so `StringToLevel("info") == DEBUG`.

This has three consequences.

**1. Request signature material is written to disk.** `CubeMaster/pkg/service/httpservice/middleware/middleware.go:86` serializes the full `auth.SignatureParams` struct, which includes the signature and nonce:

```go
if log.IsDebug() {
    log.G(ctx).Debugf("http_request_comming: %v", utils.InterfaceToString(sgnp))
}
```

`auth.SignatureParams` (`CubeMaster/pkg/base/auth/auth.go:21-29`) carries `cube_nonce` and `cube_signature`.

Separately, `CubeMaster/pkg/service/httpservice/middleware/gin_middleware.go:55` dumps the request once DEBUG is on:

```go
var dump []byte
if log.IsDebug() {
    dump, _ = httputil.DumpRequest(c.Request, config.GetConfig().Common.DebugDumpHttpBody)
}
```

logged at `gin_middleware.go:82` as `http_request_coming`. Since `debug_dump_http_body` (`CubeMaster/pkg/base/config/config.go:64`) is not set in any shipped config, it defaults to `false`, so the body is not dumped — but all request headers are, and the auth parameters are read from headers.

**2. Caller reporting is enabled.** `CubeMaster/pkg/base/log/log.go:26-28` gates on the resulting level:

```go
if IsDebug() {
    CubeLog.SetReportCaller(true)
}
```

adding a `runtime.Caller` stack walk per log line.

**3. Request log volume grows well beyond what INFO would produce**, since both middleware paths log per request.

## Test Report

Probe added temporarily at `CubeMaster/pkg/base/log/zz_probe_test.go` (removed after the run), reading each shipped config and reporting the resolved level:

```go
package log

import (
	"os"
	"regexp"
	"testing"

	"github.com/tencentcloud/CubeSandbox/pkgs/CubeLog"
	"gopkg.in/yaml.v3"
)

type probeRoot struct {
	Log Conf `yaml:"log"`
}

var levelName = map[CubeLog.LogLevel]string{
	CubeLog.DEBUG: "DEBUG", CubeLog.INFO: "INFO", CubeLog.WARN: "WARN",
	CubeLog.ERROR: "ERROR", CubeLog.FATAL: "FATAL",
}

func report(t *testing.T, file, raw string) {
	t.Helper()
	lvl := CubeLog.StringToLevel(raw)
	CubeLog.SetLevel(lvl)
	t.Logf("%-56s level=%-8q -> %-5s  IsDebug()=%v", file, raw, levelName[lvl], IsDebug())
}

func TestProbeShippedLevels(t *testing.T) {
	for _, f := range []string{
		"../../../conf.yaml",
		"../../../../configs/single-node/cubemaster.yaml",
	} {
		b, err := os.ReadFile(f)
		if err != nil {
			t.Fatalf("read %s: %v", f, err)
		}
		var root probeRoot
		if err := yaml.Unmarshal(b, &root); err != nil {
			t.Fatalf("parse %s: %v", f, err)
		}
		report(t, f, root.Log.Level)
	}

	// Helm template and HCL are not plain YAML; extract the literal.
	re := regexp.MustCompile(`level\s*[:=]\s*"([^"]+)"`)
	for _, f := range []string{
		"../../../../deploy/kubernetes/chart/files/cube-master/conf.yaml",
		"../../../../deploy/one-click/terraform/tencentcloud/tke-addons.tf",
	} {
		b, err := os.ReadFile(f)
		if err != nil {
			t.Fatalf("read %s: %v", f, err)
		}
		m := re.FindSubmatch(b)
		if m == nil {
			t.Fatalf("no level literal in %s", f)
		}
		report(t, f, string(m[1]))
	}
}

func TestProbeCaseVariants(t *testing.T) {
	for _, in := range []string{"info", "Info", "INFO", " INFO ", "warn", "typo"} {
		t.Logf("StringToLevel(%-8q) = %s", in, levelName[CubeLog.StringToLevel(in)])
	}
}
```

Output:

```text
=== RUN   TestProbeShippedLevels
    ../../../conf.yaml                                       level="info"   -> DEBUG  IsDebug()=true
    ../../../../configs/single-node/cubemaster.yaml          level="info"   -> DEBUG  IsDebug()=true
    ../../../../deploy/kubernetes/chart/files/cube-master/conf.yaml level="info" -> DEBUG  IsDebug()=true
    ../../../../deploy/one-click/terraform/tencentcloud/tke-addons.tf level="info" -> DEBUG  IsDebug()=true
--- PASS: TestProbeShippedLevels (0.00s)
=== RUN   TestProbeCaseVariants
    StringToLevel("info"  ) = DEBUG
    StringToLevel("Info"  ) = DEBUG
    StringToLevel("INFO"  ) = INFO
    StringToLevel(" INFO ") = DEBUG
    StringToLevel("warn"  ) = DEBUG
    StringToLevel("typo"  ) = DEBUG
--- PASS: TestProbeCaseVariants (0.00s)
PASS
```

The existing `pkgs/CubeLog` suite passes unchanged:

```text
go test ./... -count=1
ok  	github.com/tencentcloud/CubeSandbox/pkgs/CubeLog	0.014s
```

It passes because `TestSetLevelGetLevelAndStringToLevel` (`pkgs/CubeLog/contract_test.go:55-75`) asserts only the five uppercase spellings plus `"unknown"` — the lowercase case that every shipped config actually uses is not covered, which is why this went unnoticed.

## Additional Context

All four shipped configurations use the lowercase spelling at `9270cabf`:

- `CubeMaster/conf.yaml:21`
- `configs/single-node/cubemaster.yaml:23`
- `deploy/kubernetes/chart/files/cube-master/conf.yaml:20`
- `deploy/one-click/terraform/tencentcloud/tke-addons.tf:283` (`level = "info"`)

**Interaction with #1470.** Editing a running deployment's config from `info` to `INFO` does not take effect, because the hot-reload guard at `CubeMaster/pkg/base/hotswap/file.go:94` compares case-insensitively:

```go
if len(o.backupOldCfg) == len(new) && bytes.EqualFold(o.backupOldCfg, new) {
```

`bytes.EqualFold` treats `info` and `INFO` as equal, so the reload is skipped. Working around this bug therefore requires a process restart, not a config reload. Tracked separately as #1470.

**Relationship to #1289.** #1289 covered a separate `file_size` / `file_num` key mismatch affecting log rotation and is closed. This report is limited to log-level parsing; the two together could compound into unbounded log growth on older releases.

## Suggested Fix

Normalize before matching in `StringToLevel`:

```go
switch strings.ToUpper(strings.TrimSpace(level)) {
```

and extend `TestSetLevelGetLevelAndStringToLevel` to cover `info`, `Info`, `INFO`, `" INFO "`, and an invalid value.

**Open question, worth deciding separately:** the `default: return DEBUG` fallback means any typo in the level name silently enables the most verbose setting, including credential-bearing request logs. A safer default would be `INFO`, or an explicit startup error on an unrecognized value. That is a behavior change and probably belongs in its own PR rather than being folded into the case-insensitivity fix.

Young-Allen:
## Summary

Fixes #1593
Accept case-insensitive log levels and ignore leading/trailing whitespace.

## Root cause

CubeMaster configurations use values such as `level: "info"`.
CubeMaster passes this value to `StringToLevel`, which previously matched only exact uppercase strings, causing valid lowercase values to fall back to `DEBUG`.

## Changes

- Normalize input with `strings.TrimSpace` and `strings.ToUpper`.
- Add contract coverage for uppercase, lowercase, mixed-case, whitespace-padded, unknown, and empty values.
- Keep the existing `DEBUG` fallback behavior.
- No configuration templates were modified.

## Tests

- CubeLog tests passed.
- CubeMaster tests passed in the builder environment.

Issue #1289 is already fixed and is outside this PR.
Issue #1470 is outside this PR scope.

### DIFF
--- a/pkgs/CubeLog/contract_test.go
+++ b/pkgs/CubeLog/contract_test.go
@@ -65,12 +65,20 @@ func TestSetLevelGetLevelAndStringToLevel(t *testing.T) {
 		{"WARN", WARN},
 		{"ERROR", ERROR},
 		{"FATAL", FATAL},
+		{"debug", DEBUG},
+		{"info", INFO},
+		{"warn", WARN},
+		{"error", ERROR},
+		{"fatal", FATAL},
+		{"Info", INFO},
+		{" INFO ", INFO},
 		{"unknown", DEBUG},
+		{"", DEBUG},
 	}
 	for _, tc := range cases {
 		got := StringToLevel(tc.in)
 		if got != tc.want {
-			t.Fatalf("StringToLevel(%q) = %v, want %v", tc.in, got, tc.want)
+			t.Errorf("StringToLevel(%q) = %v, want %v", tc.in, got, tc.want)
 		}
 	}
 
--- a/pkgs/CubeLog/logger.go
+++ b/pkgs/CubeLog/logger.go
@@ -144,7 +144,7 @@ func EnableLongFilePath() {
 }
 
 func StringToLevel(level string) LogLevel {
-	switch level {
+	switch strings.ToUpper(strings.TrimSpace(level)) {
 	case "DEBUG":
 		return DEBUG
 	case "INFO":


fslongjin (approved):

luzhixing12345:
nice catch

Issue #1354, opened 2026-08-17
[Feature Request] Extract cross-component shared Go libraries into a top-level libs/ directory


fslongjin:
## Problem / Motivation

There are multiple **compile-time dependencies between sibling components** in the Go modules of this monorepo, implemented via go.mod `replace` directives pointing at sibling directories. As a result, "deployable components" and "shared libraries" are indistinguishable in the directory layout, which leads to a series of downstream costs in builds, CI, and releases.

### Current dependency graph

(The dashed box marks pure shared libraries — no cmd/main, not deployable.)

```mermaid
graph TD
    subgraph components ["Deployable components"]
        CubeMaster["CubeMaster"]
        CubeOps["CubeOps"]
        Cubelet["Cubelet"]
        ExampleRPC["examples/volume/cos/rpc"]
    end

    subgraph libs ["Pure shared libraries (no cmd/main, not deployable)"]
        CubeDB["CubeDB"]
        cubelog["cubelog"]
        cubevs["CubeNet/cubevs"]
    end

    CubeMaster --> CubeDB
    CubeMaster --> cubelog
    CubeMaster --> Cubelet
    CubeMaster -.-> cubevs

    CubeOps --> CubeDB
    CubeOps --> cubelog

    Cubelet --> cubevs
    Cubelet --> cubelog

    ExampleRPC --> Cubelet

    classDef lib fill:#f6f8fa,stroke:#999,stroke-dasharray: 5 5;
    class CubeDB,cubelog,cubevs lib;
```

How each dependency is actually used (all via go.mod `replace` pointing at sibling directories):

| Dependent | Depends on | Actual usage | Evidence |
|---|---|---|---|
| CubeMaster | CubeDB (entire module) | The whole data layer: dao / driver / migrate, 7 files | `CubeMaster/pkg/base/db/db.go:21` etc. |
| CubeMaster | cubelog | Logging / tracing, 60 files | — |
| CubeMaster | Cubelet (entire module) | Only 1 file uses 2 lightweight packages: volumeplugin pb + grpctarget | `CubeMaster/pkg/volume/plugin/rpc/driver.go:19-20` |
| CubeMaster | CubeNet/cubevs | Dead config: replaced but never required | `CubeMaster/go.mod:191` |
| CubeOps | CubeDB (entire module) | 5 files importing `CubeDB/dao` etc. | `CubeOps/internal/store/db.go:10-13` |
[...]

PR #1600, opened 2026-09-01
refactor(proto): unify gRPC protos into shared pkgs/proto module

fslongjin:
## Summary

Introduce `pkgs/proto`, a single shared Go module that becomes the sole
source of truth for the CubeMaster ↔ Cubelet gRPC contracts, and migrate
every consumer onto it. This removes the duplicated proto trees that
historically lived under `CubeMaster/api` and `Cubelet/api` and had to be
kept in sync by hand (git history shows 12 protocol changes, each touching
two copies).

Related: #1354, #1023.

## What changed

- **New `pkgs/proto` module**: 6 proto groups (`cubebox` / `errorcode` /
  `images` / `snapshot` / `volumeplugin` / `types`) with checked-in
  generated `.pb.go`, the `volumeplugin/grpctarget` address-normalization
  helper, and a `protoc-gen-doc` API reference (`doc/api.md`).
- **Migration**: all consumers (CubeMaster, Cubelet,
  cube-lifecycle-manager, volume plugins, the cos/rpc example) now import
  the module via a `replace` directive. CubeMaster no longer replaces the
  whole Cubelet module just to reach the proto imports.
- **Safety hardening**:
  - `Cubelet/pkg/critypes`: nil-guarded CRI conversion helpers plus
    table-driven tests (previously exported from `Cubelet/api/types/v1`).
  - `SafePrintImageSpec` now masks registry credentials
    (token/username annotations) before specs are logged.
  - Fixed `cos/rpc` `isMountPoint`: an existing directory that is not a
    mount point is reported as "not mounted" instead of an error.
- **CI & tooling**: `pkgs/proto/*` changes trigger the real consumers'
  test sets (CubeMaster, cubelet, volume-cos-rpc); a new
  `volume-cos-rpc` component and a `make cube-volume-cos-rpc-test` target
  were added; docs updated for the new layout.

## Wire compatibility

Non-negotiable identifiers — proto package names, message names, field
numbers and full gRPC service method names — are unchanged. Only file
locations, `go_package` paths and import paths moved, so old and new
components can still interoperate on the wire.

## Testing

- Unit tests: updated consumer packages pass (CubeMaster, Cubelet,
  cos/rpc), including the new critypes / grpctarget / images-utils tests.
- CI: unit-test-check now covers `pkgs/proto` changes via its consumers.
- **E2E**: full `tests/e2e/sdk_compat` suite passed against a live
  one-click deployment (all-in-one control node) running this branch's
  build — **229 passed, 16 skipped, 1 xfailed, 0 failed**. The 16 skips
  are documented opt-in gates (platform lifecycle, key auth, L7 HTTPS CA
  trust, external-resolver connectivity); the xfail is the known
  host-mount symlink TOCTOU case.

## Diff stat

284 files changed, +5,641 / −17,015 (net ≈ −11k lines, mostly duplicated
generated code).


ls-ggg:
LGTM

PR #1602, opened 2026-09-01
fix(terraform): query cube-ops nodes via CLB and fix jq syntax

liciazhu:
Node verification/template creation was skipped ("Registered healthy nodes: 0") because:
1. The node query exec'd into the cube-ops pod to curl localhost, but the alpine runtime image ships no curl, so nodes_json held the exec error text and jq fell back to 0. Now curl cube-ops' CLB from the jumpserver .
2. The jq filters used wrong shapes.


liciazhu:
Clipboard_Screenshot_1788249512

fslongjin (approved):

PR #1607, opened 2026-09-01
fix(cubeops): support configurable Redis database selection

zyl1121:
## Motivation

When CubeMaster is configured to read node metrics from a non-zero Redis logical database, CubeOps still writes those metrics to DB 0 when it uses split Redis settings. CubeMaster therefore cannot see the metrics, causing scheduler metric updates to time out.

The same mismatch affects Sentinel connections because CubeOps does not select the configured logical database after discovering the Redis master.

## What Changed

* Add `redis_db` YAML and `REDIS_DB` environment configuration to CubeOps, defaulting to database `0`.
* Use the configured database for split standalone Redis connections.
* Use the configured database for Sentinel master connections.
* Treat `REDIS_URL` as the complete, highest-priority Redis connection configuration, including its database path, without combining it with Sentinel or split settings.
* Add focused coverage for Redis connection source priority, Sentinel master database selection, and YAML/environment precedence.
* Update CubeOps documentation.

## Validation

* `go test ./... -count=1` passed for CubeOps.
* CubeOps binaries built successfully.
* `REDIS_URL` priority and Sentinel database selection tests passed for 20 consecutive runs.
* Runtime Redis selection validation confirmed that a metric written with database `7` is readable from DB 7 and absent from DB 0.

## Scope

This PR is intentionally limited to CubeOps logical Redis database selection. It does not change the Helm chart or CubeMaster.

The chart already supports additional CubeOps environment variables through `cubeOps.env`, so Kubernetes installations can set `REDIS_DB` without adding a Chart-level `redis_db` value. CubeMaster continues using its existing `conf.yaml` Redis database configuration.

### DIFF
--- a/CubeOps/README.md
+++ b/CubeOps/README.md
@@ -125,7 +125,8 @@ defaults.
 | `DATABASE_URL` | *(required)* | MySQL connection URL. If unset, built from `CUBE_SANDBOX_MYSQL_{HOST,PORT,USER,PASSWORD,DB}` env vars. |
 | `CUBE_MASTER_ADDR` | `http://127.0.0.1:8089` | CubeMaster base URL |
 | `CUBE_API_SANDBOX_DOMAIN` | `cube.app` | Sandbox domain (used by SDK handler for sandbox URL construction) |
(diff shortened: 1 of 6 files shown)


fslongjin (approved):
LGTM

PR #1609, opened 2026-09-01
cubelet plugin init failure when gateway ARP cache is not yet populated  

### LINKED ISSUE #1608: [Bug Report]  cubelet  plugin init failure when gateway ARP cache is not yet populated
During Kubernetes cubenode  Cubelet plugin initialization, plugin fails to load with:
failed to get internal plugin: gateway mac for ens3 via 192.168.0.1 not found

Image
Manually pinging the gateway prior to initialization resolves the issue, confirming that ARP cache population is the missing prerequisite.

Image

GetGatewayMacAddr reads the kernel neighbor table exactly once. At early boot the gateway's ARP entry may not exist yet (or is in NUD_INCOMPLETE state), so the lookup fails immediately with no recovery mechanism. This is a timing-dependent issue — it does not reproduce on every boot.
Root Cause
GetHostDevice → GetGatewayMacAddr performs a single NeighList call. If no usable neighbor entry (REACHABLE / STALE / DELAY / PROBE / PERMANENT) is found for the gateway IP, it returns an error. There is no ARP trigger or retry logic.
Fix
Modified GetGatewayMacAddr in Cubelet/network/runtime/systemnet/device.go to:
Fast path — first attempt reads the neighbor table as before; if the entry exists, return immediately (no behavior change for the common case).
Retry with ARP trigger — on cache miss, retry up to 5 times with 200ms backoff. Each retry calls triggerARPResolution which uses a two-layer approach:
Layer 1: NeighSet(NUD_PROBE) — precisely targets the gateway on the correct interface index, causes the kernel to send an ARP Request. Side-effect-free but only works when a neighbor entry already exists.
Layer 2: UDP probe fallback — sends a small UDP packet to the gateway, forcing the kernel to create a neighbor entry and issue an ARP request. This handles the case where no entry exists yet (the exact scenario during node init).
Files Changed
Cubelet/network/runtime/systemnet/device.go — added retry loop, triggerARPResolution, extracted lookupGatewayMac helper
Cubelet/network/runtime/systemnet/device_test.go — added tests for lookupGatewayMac and the retry-on-cache-miss path

Gu0YuXianG:
Breaking Change Notice: Error message format for GetGatewayMacAddr has been updated to include retry and probe error context. Consumers matching on the exact error string may be affected.

fix #1608 
During Kubernetes cubenode Cubelet plugin initialization, plugin fails to load with:
failed to get internal plugin: gateway mac for ens3 via 192.168.0.1 not found
image

Image Manually pinging the gateway prior to initialization resolves the issue, confirming that ARP cache population is the missing prerequisite. Image
image

GetGatewayMacAddr reads the kernel neighbor table exactly once. At early boot the gateway's ARP entry may not exist yet (or is in NUD_INCOMPLETE state), so the lookup fails immediately with no recovery mechanism. This is a timing-dependent issue — it does not reproduce on every boot.
Root Cause
GetHostDevice → GetGatewayMacAddr performs a single NeighList call. If no usable neighbor entry (REACHABLE / STALE / DELAY / PROBE / PERMANENT) is found for the gateway IP, it returns an error. There is no ARP trigger or retry logic.
Fix
Modified GetGatewayMacAddr in Cubelet/network/runtime/systemnet/device.go to:
Fast path — first attempt reads the neighbor table as before; if the entry exists, return immediately (no behavior change for the common case).
Retry with ARP trigger — on cache miss, retry up to 5 times with 200ms backoff. Each retry calls triggerARPResolution which uses a two-layer approach:
Layer 1: NeighSet(NUD_PROBE) — precisely targets the gateway on the correct interface index, causes the kernel to send an ARP Request. Side-effect-free but only works when a neighbor entry already exists.
Layer 2: UDP probe fallback — sends a small UDP packet to the gateway, forcing the kernel to create a neighbor entry and issue an ARP request. This handles the case where no entry exists yet (the exact scenario during node init).
Files Changed
Cubelet/network/runtime/systemnet/device.go — added retry loop, triggerARPResolution, extracted lookupGatewayMac helper
Cubelet/network/runtime/systemnet/device_test.go — added tests for lookupGatewayMac and the retry-on-cache-miss path


chenhengqi (commented):
Please squash your commits and add your signed-off-by.

This seems like a transient error. Can we add a preflight check instead? BTW, what CNI plugin do you use?

Gu0YuXianG:
> Please squash your commits and add your signed-off-by.
> 
> This seems like a transient error. Can we add a preflight check instead? BTW, what CNI plugin do you use?

@chenhengqi 

We're using Calico.

PR #1619, opened 2026-09-02
feat(cubemaster): add extensible scheduler plugin system and scheduling-quality evaluation

shsaihdsaiudh:
## Summary

This draft adds two complementary pieces to CubeMaster scheduling, developed on top of each other:

1. **Scheduling-quality evaluation framework** — Prometheus metrics for the scheduling pipeline (`scheduler_attempts_total`, `scheduler_duration_seconds`, `scheduler_decisions_total`, `scheduler_reschedules_total`, `sandbox_create_*`, cluster gauges), a seeded workload generator + A/B compare report in `examples/cube-bench`, and `cmd/schedsim`, an in-process simulator that replays cube-bench traces through the real `scheduler.Select` over a simulated node fleet and reports packing/fragmentation/balance/locality metrics.
2. **Extensible scheduler plugin system** (merged from fork PR shsaihdsaiudh/CubeSandbox#1) — an explicit plugin registry replacing the reflect-based maps, named strategy **profiles** with per-plugin config and failure policies, three picker modes (`random`/`spread`/`highest`), CEL expression plugins, and external gRPC plugins, with legacy config translated to an equivalent `default` profile so existing behavior is unchanged.

The merge also resolves the `TODO(profile-router)` in the scheduler metrics: all scheduler/sandbox-create Prometheus series are now stamped with the profile that actually routed the request.

Design background: see `SCHEDULER_PLUGIN_DESIGN.md` and `SCHEDULER_CODE_READING_GUIDE.md` on the branch root.

## Changes

- `ab2ddfd7` feat(cubemaster): add scheduler Prometheus metrics
- `4967c66c` feat(cube-bench): add seeded workload generator and A/B compare report
- `675bb864` feat(cubemaster): add schedsim scheduling simulator
- `de1a7af3`/`be808e09` docs(scheduler): evaluation and plugin design
- `4209e1f7` feat(scheduler): add extensible plugin system (from shsaihdsaiudh/CubeSandbox#1)
- `c45dda12` feat(scheduler): merge extensible plugin system from PR #1
- `a8c8b415` feat(cubemaster): wire profile labels into scheduler metrics and add schedsim compare mode (`--compare name=config.yaml,...` renders a markdown A/B report across scheduler configs)

## Test plan

- [x] `go build ./...` in CubeMaster
- [x] `go test` green for `pkg/scheduler/...`, `pkg/selector/...`, `pkg/base/...`, `pkg/service/sandbox/...`, `pkg/server`, `pkg/localcache`, `cmd/schedsim`, and the `examples/cube-bench` module
- [x] `go vet ./...` — no new warnings
- [x] schedsim e2e: `burst` / `template_storm` / `mixed_spec` traces × (legacy config vs profiles config) × 3 rounds on a 96-core box; all rounds `success_rate=1.0`, comparison reports render as expected
- [ ] Real-cluster cube-bench run against a live CubeAPI (needs a multi-node deployment)

## Notes

- Per the repo `AGENTS.md` AI policy, AI-authored commits carry `Assisted-by:` / `Autonomously-by:` tags and no `Signed-off-by`; the human submitter will add their `Signed-off-by` (DCO) before this leaves draft.
- Opened as a **draft** to collect automated review feedback first.


CubeMaster/pkg/localcache/reservation.go
```diff
@@ -0,0 +1,296 @@
...
+	// reservation Hash and rolls back if any dimension exceeds the headroom
+	// computed by the caller from its freshest node view:
+	//
+	//	KEYS[1] = reservation hash key
+	//	ARGV[1] = safety TTL (seconds)
+	//	ARGV[2..5] = deltas: cpu_milli, mem_mb, mvm, creating
+	//	ARGV[6..9] = maxima: cpu_milli, mem_mb, mvm, creating
+	//
+	// Returns 1 on success, 0 on conflict.
+	reservationAcquireScript = `
```
fslongjin, line 90: This script is not idempotent, but it is invoked through `wrapredis.Do`, which retries transport failures. If Redis executes the first `EVAL` and the client times out before receiving the reply, the second `EVAL` applies `HINCRBY` again.

For example, a 1000 milli-core acquire can leave Redis at 2000 while the local registry records only 1000. If the second execution crosses a limit, it rolls back only the second delta and leaves the first delta leaked in Redis until the safety TTL expires.

Could we make the operation idempotent with a unique reservation token and per-reservation state, or bypass transport retries for this specific `EVAL`? A regression test should simulate "`EVAL` committed, client read timed out, request retried" and assert both Redis and local accounting remain consistent.

CubeMaster/pkg/localcache/reservation.go
```diff
@@ -0,0 +1,296 @@
...
+return 1
+`
+	// reservationReleaseScript subtracts the reservation again, clamping each
+	// field at zero so a release after a TTL expiry cannot drive the Hash
+	// negative.
+	//
+	//	KEYS[1] = reservation hash key
+	//	ARGV[1] = safety TTL (seconds)
+	//	ARGV[2..5] = deltas: cpu_milli, mem_mb, mvm, creating
+	reservationReleaseScript = `
```
fslongjin, line 112: Release has the same retry/replay problem as acquire. If the first `EVAL` commits but the client sees a transport error, `wrapredis` retries and subtracts the same CPU/memory/MVM/creating deltas a second time. The clamp-to-zero behavior can also erase reservations held by other replicas.

Could release also use the reservation token/idempotency approach? Ideally the regression test would cover both directions: a retried acquire must not double charge, and a retried release must not double subtract.

CubeMaster/pkg/localcache/reservation_test.go
```diff
@@ -0,0 +1,281 @@
...
+		t.Fatalf("HGETALL error: %v", err)
+	}
+	if fields["cpu_milli"] != "1000" || fields["mem_mb"] != "1024" || fields["mvm"] != "1" || fields["creating"] != "1" {
+		t.Fatalf("redis hash=%v", fields)
+	}
+	if ttl := server.TTL(rediskey.NodeReservation("node-r4")); ttl <= 0 {
+		t.Fatalf("reservation hash has no safety TTL: %v", ttl)
+	}
+
+	// Cross-master style conflict: the Redis check rejects what would
```
fslongjin, line 207: This test is labelled as a cross-master-style conflict, but the `200000` milli-core request is rejected by `checkReservationCapacity` before it reaches the Redis Lua script because the test node quota is `64000` milli-cores. The current assertion therefore does not prove that Redis-side atomic conflict detection and rollback work.

Could we prefill the Redis reservation hash close to the available quota, then request an amount that passes local validation but exceeds the Redis-side headroom? That would exercise the Lua conflict and rollback path directly.

CubeMaster/pkg/scheduler/selctx/random_select.go
```diff
@@ -6,23 +6,29 @@ package selctx
...
 func (r *randomSelect) Add(item interface{}, weight int) {
 	r.items = append(r.items, item)
 }
 
+// All 随机选择器不维护权重表，返回 nil
 func (r *randomSelect) All() map[interface{}]int {
 	return nil
 }
 
+// RemoveAll 随机选择器无需清空权重表，空实现
```
fslongjin, line 30: `Reset` and `RemoveAll` are empty, but `randomSelect.items` accumulates entries across calls to `LeastRandomSelect`. Reservation conflicts reuse the same `SelectorCtx` and add the conflicting node to `lastBadFilters`; prefilter can remove it from the new candidate list, but it can still remain in `rSelect.items` from the previous attempt and be returned by `Next()`.

Could we clear the accumulated items before repopulating the selector, or implement `Reset`/`RemoveAll` consistently? A regression test that reuses one `SelectorCtx`, marks a node bad, and calls `LeastRandomSelect` again would lock this behavior in.

CubeMaster/pkg/scheduler/schedule.go
```diff
@@ -176,64 +468,158 @@ func parallelRunFilters(selCtx *selctx.SelectorCtx, filters []filter.Selector) (
...
-
-						resultMap[n.ID()] = n
+						seen[scored.ID()] = struct{}{}
 					}
 				}
+				if result.err != nil {
+					break
+				}
+			}
+			if result.err == nil && binding.ForceEnabled && len(seen) != len(candidates) {
```
fslongjin, line 558: `ForceEnabled` treats an empty score result as a plugin failure. However, `template_local_pressure` intentionally returns an empty list when the request has no `TemplateID`, and its comment says the pipeline will ignore the empty result.

With `failure: score-default-score`, this becomes a full default-score result and still contributes to `totalWeight`. A zero default score does not change relative ordering, so this is not necessarily a placement bug, but the implementation and the plugin contract are inconsistent.

Could we either allow an explicit "not applicable / skip" result, or require every score plugin to return a score for every candidate? Option 1 seems clearer for plugins whose dimension genuinely does not apply.

CubeMaster/pkg/localcache/reservation.go
```diff
@@ -0,0 +1,296 @@
...
+
+	reservation := &NodeReservation{NodeID: nodeID, CpuMilli: cpuMilli, MemMB: memMB}
+	if config.GetConfig().RedisConf == nil {
+		// No Redis configured: local-only by design (single-replica
+		// deployments never coordinate through Redis).
+		return reservation, nil
+	}
+	acquired, err := redisAcquireReservation(ctx, current, cpuMilli, memMB)
+	switch {
+	case err != nil:
```
fslongjin, line 181: When Redis is configured but the acquire fails, the code keeps the local reservation and continues. This is reasonable for a single CubeMaster, but in a partial-failure or multi-replica deployment some replicas may still be writing Redis while this replica falls back to local-only accounting. The `realtime_create_num` guard does not fully protect CPU and memory quota.

Could we make the degradation policy explicit, for example:
- local-only for single-replica mode;
- fail the reservation in multi-replica mode, or require an operator opt-in to accept the fail-open behavior?

At minimum, the multi-replica degradation behavior should be called out in the upgrade documentation.

CubeMaster/pkg/base/config/config.go
```diff
@@ -987,6 +1074,66 @@ func preHandOverhead(config *Config) error {
...
+//go:embed scheduler_factory.yaml
+var schedulerFactoryYAML []byte
+
+// injectFactorySchedulerProfiles 在用户完全没有配置调度策略时注入出厂策略：
+// 要求 scheduler.profiles 与 legacy 的 scheduler.filter / scheduler.score /
+// scheduler.postscore 全部为空，且未显式设置 scheduler.disable_factory_profiles。
+// 一旦用户显式配置了其中任意一项，出厂策略整体不注入（all-or-nothing），
+// 避免覆盖存量部署依赖的 legacy 编译路径（profile.Compile 有 profiles 时忽略 legacy）。
+//
+// 出厂 YAML 同时携带 legacy score 子树：profile 模式下它不参与流水线编译，
```
fslongjin, line 1093: The factory YAML carries a legacy `scheduler.score` subtree solely because built-in scorers such as `real_time_weighted_average` and `image_score` read their factor switches from the global legacy config. This makes profile configuration split across two models and can surprise operators who modify profile weights without also changing the legacy `plugin_conf`.

Could we either move these per-plugin settings into profile `args` and validate required configuration at compile time, or document this coupling prominently and validate that the required legacy fields exist when a profile references these scorers? Silent no-op scorers should not be possible in the new profile schema.

docs/dev/scheduler-strategy-mapping.md
```diff
@@ -0,0 +1,179 @@
...
+
+**Trade-offs**: tight packing raises utilization but enlarges hotspots and
+single-node blast radius; under 3× overcommit at moderate occupancy the per-node
+quota deltas are tiny, so `resource_fit_score` values cluster and the effective
+spread is governed by top_n and the `real_time_weighted_average` weight — tune
+with single-variable A/B runs.
+
+## Metric definitions and sources
+
+The simulator (schedsim, `--workload burst|template_storm|mixed_spec`) and real
```
fslongjin, line 165: This section says `schedsim` accepts `--workload`, but that flag belongs to `cube-bench`; `schedsim` consumes a trace file. It also says simulator and production use one set of metric definitions, but several definitions differ: Jain for an empty set, fragmentation shape and dimensions, herding window, and whether reservations are modeled.

Could we correct the command and describe the metric differences explicitly? That would prevent users from treating `schedsim` and Prometheus numbers as directly interchangeable.

docs/guide/scheduler-plugin.md
```diff
@@ -0,0 +1,84 @@
...
+- **Template locality scoring is a boolean factor.** The legacy `template_id` image score used to scale linearly with the template size present on the node; it is now an exact-match 100/0 factor (a node either has the template locally or it does not), matching the `template_local` fact exposed to CEL and gRPC plugins. Legacy-configured clusters may see different placement for template requests; in practice the difference is small because the spread/binpack scorers dominate the weighted aggregate.
```
fslongjin, line 12: The claim that the practical difference is small is too broad. For the shipped `template_reuse` profile, `image_score` is the dominant `0.7`-weight factor, and a legacy cluster that enables only `image_score` can see a material placement change.

Could we rewrite this as a real behavior change, explain the old linear score and new `100/0` locality score, and note that the impact depends on which scorers and weights are enabled?

docs/zh/guide/scheduler-plugin.md
```diff
@@ -0,0 +1,84 @@
...
+        filter: fail-closed
+        score: default-score
+        no_candidate: fail
+```
+
+`selection.method` 决定如何从评分结果中选出最终节点：`highest` 严格选取评分最高的节点；`spread` 在评分最高的前 `top_n` 个候选中确定性选取当前运行沙箱数与在途预留数之和最少的节点（占用相同时保持评分顺序），用于把放置摊开；`random`（缺省）在前 `top_n` 个候选中按分数加权随机。`top_n: -1` 表示候选范围为全部通过过滤的节点。
+
+自定义 Profile 固定执行 `node_safety`、`cpu`、`mem`、`disk`、`template_locality` 和 `realtime_create_num` Guards，配置不能关闭或重复声明这些安全约束。其中 `node_safety` 会在正常路径和 backoff 路径检查健康度、指标新鲜度、MVM 上限及 CPU load 合法性。
+
+选定节点后，CubeMaster 会重读该节点并原子预留本次请求的 CPU、内存配额、一个 MVM 槽位和一个创建并发槽位，避免并发创建在节点指标更新前反复落到同一节点。预留冲突会换节点有限重选；Cubelet 创建调用返回后（无论成功或失败）即释放预留，成功场景由下一次节点指标上报接管记账。多 CubeMaster 副本通过按节点的 Redis 计数协调预留，Redis 不可用时退化为本地预留。本副本持有的在途预留数以 `node.reserved` 暴露给插件。
```
fslongjin, line 46: This operations guide still tells users to tune `scheduler.priority_select_num`/`least_select_name` without mentioning profile selection, the factory-profile injection condition, or how an explicit `scheduler.filter` block suppresses the factory profiles. It also does not describe the new reservation behavior or the new Prometheus metrics.

Could we update the main deployment guide before merge so operators do not have to discover the new model from the plugin page? It should cover:
- profile routing and `profile_route_label_keys`;
- when factory profiles are injected or suppressed;
- the global reservation behavior and retry/degradation semantics;
- a user-facing metrics reference.

fslongjin (changes requested):
Request changes.

Thanks for the extensive work here. The overall direction is strong: the explicit plugin registry, named profiles, mandatory guards, CEL validation, and the evaluation stack are a significant improvement over the implicit legacy registration model. The gRPC snapshot design and several earlier concurrency fixes also look good.

The main blocker is the Redis reservation implementation. Both the acquire and release Lua scripts are non-idempotent, but they are executed through `wrapredis.Do`, which retries transport failures. If Redis commits an `EVAL` and the client times out before reading the reply, the retried `EVAL` can apply the same `HINCRBY` twice. That can either leak a phantom reservation until the safety TTL expires or subtract a reservation twice and remove another replica's valid reservation. This is reachable on the default reservation path for every non-direct-host create.

Please make both acquire and release idempotent (for example, use a unique reservation token and per-reservation state) or disable transport retries for these specific `EVAL` calls. Please also add regression coverage for the "command committed but client timed out and retried" case, covering both acquire and release, and assert that the Redis hash and local accounting remain consistent.

Two implementation issues are strongly recommended for this PR:

1. The "cross-master conflict" reservation test currently fails before reaching the Lua path because `200000` milli-cores exceeds the local `64000` quota. It should prefill Redis near the quota and request an amount that passes local validation but fails the Redis-side atomic check.
2. `randomSelect.Reset/RemoveAll` are empty while `LeastRandomSelect` accumulates items. Because reservation conflicts reuse the same `SelectorCtx` and mark the conflicting node as bad, a later random pick can still select the stale node from `rSelect.items`. Resetting the selector state before repopulating it would avoid wasting the bounded retry budget.

Separately, documentation and configuration consistency must be corrected before merge. In particular, the factory profiles still depend on a legacy `scheduler.score` subtree, several docs overstate metric comparability, the template-locality behavior change is understated, and the main operations guide still describes only the legacy selection knobs. Please update the docs and configuration model so operators can determine which profile is active, where scorer factor switches live, what the new reservation behavior is, and how the evaluation metrics differ between `schedsim` and Prometheus.

The remaining architecture follow-ups are not merge blockers in my view, but they should be tracked explicitly.

CubeMaster/pkg/service/sandbox/sandbox_run.go
```diff
@@ -227,14 +242,60 @@ func (c *createSandboxContext) handleCubelet() {
...
+		// Reserve CPU / memory / MVM / create-concurrency slots on the
+		// selected node so concurrent creates stop piling onto the same
+		// node before Cubelet metrics catch up. A conflict marks the node
+		// bad and reselects, a bounded number of times, with a short
+		// exponential backoff in between: the conflicting state is another
+		// in-flight create that typically clears within hundreds of
+		// milliseconds, so backing off lets a burst degrade into queueing
+		// instead of re-conflicting on the same stale view and burning the
+		// conflict budget within microseconds.
+		if err := c.reserveSelectedHost(); err != nil {
```
fslongjin, line 261: I am not convinced the cross-replica half of the reservation should be strongly
consistent on the create path. The local half clearly earns its keep: it closes
the "node picked but Cubelet metrics have not flowed back" window, which is the
herding this PR measured, at roughly half a microsecond per create. The Redis
half only narrows the cross-replica view, which today is carried by the node
metric/allocation sync (SyncMetricDataInterval defaults to 1s) and by
realtime_create_num - and the implementation itself already treats
"local-only + realtime_create_num" as an acceptable posture whenever Redis is
unset or errors, described in the file header as "matching the pre-reservation
behavior".

What it costs is a blocking external dependency in a phase that had none. Before
this change the create path only touched Redis after the Cubelet create
succeeded (the proxy-map HSET in dealSuccResult, deliberately parallelized with
the MySQL write so the wall-clock cost collapses to max(redis, spec)). Blocking
in that earlier phase also is not bounded by the request: wrapredis.Do uses
Pool.Get(), which is GetContext(context.Background()) in redigo, so with
Wait: true and max_active: 32 a saturated pool parks the create goroutine
indefinitely; newConn sets no write deadline; and CreateTimeoutInsec cannot
interrupt the call because the context is only used for logging.

In a large cluster, per-create global accounting caps scheduling throughput at
the Redis round trip, pushes Redis tail latency straight into create P99, makes
the reservation key hot exactly when a burst targets one node, and turns
overload into explicit SelectNodesNoRes failures instead of the softer
realtime_create_num degradation.

Suggestions, in order of preference:
- keep selection authoritative on the local registry and publish reservation
  deltas asynchronously (for example from the existing metric sync loop), then
  treat the cross-replica aggregate as an advisory input;
- gate the Redis half on HealthyMasterNodes() > 1, since a single-replica
  deployment gains nothing from it while paying two round trips per create;
- if the EVAL stays, make it advisory (only when the local check is marginal)
  with a hard timeout that fails open.

If the synchronous check is kept for a specific production failure mode, could
you describe that failure mode and the measured impact? Otherwise the docs
should state that each create now performs synchronous reservation EVALs, since
operators currently have no way to know their create latency budget depends on
Redis.

fslongjin (commented):
Follow-up to my earlier review: the reservation idempotency blocker still stands. This is a separate point on the same code path.

One more point on the same code path, independent of the idempotency bug: I do
not think cross-master state should be made strongly consistent on the create
path. The local half of the reservation earns its keep - it closes the herding
window this PR measured, at roughly 0.5us per create. The Redis half only
narrows a window that the node metric/allocation sync (SyncMetricDataInterval
defaults to 1s) and realtime_create_num already cover, and the implementation
itself treats "local-only + realtime_create_num" as an acceptable posture
whenever Redis is unset or errors.

What it costs is a blocking external dependency in a phase that previously had
none: before this change the create path only touched Redis after the Cubelet
create succeeded (the proxy-map HSET in dealSuccResult, deliberately
parallelized with the MySQL write so the cost collapses to max(redis, spec)).
That blocking is also not bounded by the request today - wrapredis.Do uses
Pool.Get(), which is GetContext(context.Background()) in redigo, so with
Wait: true and max_active: 32 a saturated pool parks the create goroutine
indefinitely; newConn sets no write deadline; and CreateTimeoutInsec cannot
interrupt it because the context is only used for logging.

Please consider publishing reservation deltas asynchronously (for example from
the existing metric sync loop) and treating the cross-replica aggregate as an
advisory input to selection, or at least gating the Redis half on
HealthyMasterNodes() > 1 with a hard timeout. In a large cluster, per-create
global accounting caps scheduling throughput at the Redis RTT, pushes Redis
tail latency straight into create P99, and turns overload into explicit
SelectNodesNoRes failures instead of a softer degradation.

CubeMaster/pkg/service/sandbox/sandbox_run.go
fslongjin: Keeping this synchronous means the answer to "what would actually go wrong
without it" should be on the record. The path itself is unchanged from the
round-5 head: TryReserveNode performs the Redis EVAL here, before callCubelet,
and the blocking is still unbounded - wrapredis.Do uses Pool.Get(), which is
GetContext(context.Background()) in redigo, so with Wait: true and
max_active: 32 a saturated pool parks the create goroutine indefinitely, newConn
sets no write deadline, and CreateTimeoutInsec cannot interrupt it because the
context is only used for logging. With reservation_redis_error_policy now
recommending fail_closed for multi-replica deployments, a Redis stall becomes a
create failure rather than a slow create.

Could you describe the production failure mode this prevents (local accounting +
realtime_create_num + the ~1s metric sync being insufficient), with whatever
measurement you have, or the numbers for the alternative (async publish of
reservation deltas, or gating on HealthyMasterNodes() > 1)? If there is no such
data yet, saying so is fine - I would just like the trade-off stated explicitly
so the docs can carry it.

fslongjin (commented):
Two items from my earlier reviews are still open after round 6. I want to
acknowledge the rest first, because it landed well.

Confirmed in 3272a813 / 38095470: the acquire and release scripts are now
idempotent on a per-reservation token, with regression tests for the
"committed but the reply was lost" case; the cross-master conflict test now
actually reaches the Lua check; randomSelect.RemoveAll/Reset clear the
accumulated items; score.ErrNotApplicable gives the template-locality case a
contract-sanctioned skip instead of an empty result; the profile compiler now
rejects built-in scorers whose legacy scheduler.score block is missing;
reservation_redis_error_policy turns the Redis failure mode into an explicit
operator choice; and the schedsim/metric documentation was corrected in both
languages. Thank you - that covers everything I raised about correctness and
documentation.

What is still open are the two items below. Both are requests for a written
answer, not for more code in this PR.

A. The multi-node operations guide is still unchanged

The reservation behavior and reservation_redis_error_policy are now documented
in docs/guide/cubemaster-scheduler-config.md and docs/guide/scheduler-plugin.md,
which is good. But docs/guide/multi-node-deploy.md is untouched. That is the
document an operator follows when deploying several CubeMasters, and it still
describes only the legacy selection knobs; nothing there says that creates now
take a synchronous Redis dependency, or what to set for
reservation_redis_error_policy in a multi-replica deployment. I cannot leave
this as an inline comment because the file is not part of this PR's diff, which
is why it is here instead.

Could you either add a short "Reservations and Redis" note (plus the policy
knob) there, or confirm that you consider that guide out of scope for this PR?
If it is out of scope, where should an operator deploying multiple CubeMasters
read about the new dependency?

B. The synchronous cross-replica reservation is still on the create path

Round 6 made the failure policy explicit, which I read as a deliberate decision
to keep the synchronous Redis check in the create path. I would like that
decision argued rather than assumed, because the trade-off is unchanged and, if
anything, sharper now:

- the blocking is still not bounded by the request. wrapredis.Do uses
  Pool.Get(), which is GetContext(context.Background()) in redigo, so with
  Wait: true and max_active: 32 a saturated pool parks the create goroutine
  indefinitely; newConn sets no write deadline; and CreateTimeoutInsec cannot
  interrupt it because the context is only used for logging. With fail_closed
  recommended for multi-replica deployments, a Redis stall now converts
  directly into create failures rather than slow creates;
- there is still no HealthyMasterNodes() > 1 gate, so a single-replica
  deployment (DefaultHeadlessServiceNodesNum defaults to 1) pays two round
  trips per create for zero cross-replica benefit;
- the base create path performed no synchronous Redis work in the
  select/dispatch phase at all - Redis was touched only after the Cubelet
  create succeeded, deliberately parallelized with the MySQL spec write so the
  cost collapses to max(redis, spec).

To make that decision reviewable, could you answer in writing:

1. Which production failure mode does the synchronous cross-replica check
   prevent that local accounting plus realtime_create_num and the ~1s node
   metric/allocation sync does not? What is its measured impact (incident or
   test data, magnitude, frequency)?
2. What does it cost at load? Specifically, the p50/p99 of the select+dispatch
   phase and of create with and without the reservation EVAL, at a couple of
   concurrency levels (e.g. 200/500/1000 concurrent creates), and what happens
   when Redis latency is inflated or the connection pool is saturated.
3. For single-replica deployments, why is the synchronous pair kept instead of
   gating it on HealthyMasterNodes() > 1?
4. If it stays, will the blocking be bounded - a write deadline, a
   context-aware acquire path, and a pool separate from the node metric writes
   - so that a Redis stall cannot park a create goroutine indefinitely?

If the honest answer to (1) is "no such failure mode was observed and we have no
data", that is a usable answer. It would just mean the synchronous EVAL buys an
unquantified improvement at the cost of a hard dependency in the create path,
and that the asynchronous variant (publish reservation deltas from the existing
metric sync loop and treat the cross-replica aggregate as advisory) or the
single-replica gate should be preferred. I am not asking for that rewrite in
this PR; I am asking which option we are choosing and on what evidence, so the
record shows it was a decision rather than a default.

CubeMaster/pkg/service/sandbox/sandbox_run.go
JunWang666: Thanks for following up on these two points. After checking the implementation, I agree that round 6 clarified how Redis errors are handled, but did not establish whether the benefits of synchronous coordination justify its cost.

### A. Multi-node deployment documentation

`multi-node-deploy.md` is indeed missing this explanation. I will add a short “Reservations and Redis” note explaining the synchronous Redis dependency on the create path, the trade-offs of `reservation_redis_error_policy`, and links to the detailed configuration documentation.

Until then, the relevant explanations are in `docs/guide/cubemaster-scheduler-config.md` and `docs/guide/scheduler-plugin.md`.

### B. Synchronous cross-replica reservation

#### 1. What failure mode is the synchronous reservation intended to prevent?

The intended failure mode is concurrent over-admission across multiple Masters during the metric propagation window.

Without cross-replica coordination, multiple Masters can observe the same apparent remaining capacity and independently admit requests whose combined CPU and memory requirements exceed that capacity. The existing `realtime_create_num` guard and the “local concurrency × Master count” estimate provide pressure protection, but they do not atomically coordinate the actual CPU and memory requirements of in-flight requests across replicas.

That describes the design intent, however. I do not currently have production incident data quantifying the frequency or impact of this problem. The existing tests verify cross-replica capacity conflicts and retry idempotency, but I do not have evidence showing how often this failure mode occurs in production or what its measured impact is.

#### 2. What is the cost of the synchronous reservation at load?

I do not currently have the requested comparative measurements.

The existing scheduler benchmarks do not establish that the incremental `EVAL` cost is negligible. I do not have p50/p99 measurements for the select+dispatch phase or the full create path with and without reservations at 200/500/1000 concurrent creates. I also do not have measurements under increased Redis latency or connection-pool saturation.

I therefore cannot draw a performance conclusion from the available evidence, or claim that the benefit of synchronous reservations currently outweighs their cost.

I also do not currently have comparative measurements for synchronous reservations, asynchronous aggregation, and a single-replica bypass.

#### 3. Why is the synchronous pair still used for single-replica deployments?

There is currently no replica-count gate. A single-replica deployment with Redis configured performs the same pair of synchronous `EVAL`s.

For a deployment guaranteed to have only one active Master, I do not have a cross-replica coordination justification for that cost. A single-replica bypass therefore deserves priority for evaluation.

#### 4. If synchronous coordination remains, will the Redis blocking be bounded?

The blocking is not fully bounded by the request today.

The normal request entry point can return when the create timeout expires, but that context cannot interrupt the Redis wait in the background goroutine executing the create. The current implementation has connect and read timeouts, but no write timeout, and pool acquisition is not tied to the request context.

`fail_closed` only determines the behavior after Redis returns an error; it does not bound the wait itself.

If synchronous coordination is retained for multiple replicas, request-bounded connection acquisition and command execution need to be addressed. A write timeout is also needed, and connection-pool isolation should be assessed so that reservation traffic and node metric writes do not create undesirable failure coupling.

Making the Redis failure policy configurable does not resolve those blocking and failure-isolation concerns.

### Current conclusion

The original design prioritized coordinating concurrent admissions across replicas, but did not sufficiently evaluate the effects of putting Redis synchronously on the create path, including create latency, failure propagation, and the cost for single-replica deployments.

Synchronous coordination provides cross-replica resource checks before dispatch, but adds a Redis dependency to the critical path. Asynchronous aggregation reduces that dependency, but retains the possibility of concurrent over-admission during the propagation window. Bypassing Redis for explicitly single-replica deployments avoids overhead that provides no cross-replica benefit.

I do not currently have enough comparative data to determine which trade-off best fits the actual workload. I will document these trade-offs and current limitations. A single-replica bypass deserves priority for evaluation, and if synchronous coordination remains for multiple replicas, the Redis wait and connection acquisition need to be request-bounded and the connection-pool isolation needs to be assessed.

shsaihdsaiudh: I will conduct a test first, but I prefer to split off the redundant functions and only retain the parts that need to be implemented according to the task specification

shsaihdsaiudh: After discussing with my teammates, I feel that the current PR has exceeded the scope outlined in the task specification. Therefore, we plan to remove unnecessary content. I would like to discuss which content should be retained first

JunWang666: The original purpose of the synchronized Redis reservation mechanism was to coordinate in-flight resource consumption across multiple CubeMaster instances before that usage is reflected in node metrics. This prevents multiple replicas from accepting requests based on the same view of remaining capacity. The existing real-time creation count and local concurrency estimates can limit creation pressure, but they cannot provide atomic cross-replica coordination based on the actual CPU and memory requirements of each request.

However, after further review, I believe the cost and complexity of this design were underestimated. It introduces a synchronous Redis dependency into the critical creation path, while even a single-replica deployment pays unnecessary overhead. Comparative testing shows that as Redis latency increases, connection-pool queuing significantly amplifies tail latency. These tests have not yet covered realistic fully loaded multi-replica scenarios, so we currently cannot quantify the benefit in terms of failures and retries avoided.

In addition, the current implementation still has a gap between reservation release and metric takeover, and there is no guarantee that stale reservations left behind by abnormal failures will be reclaimed in time. Adding timeouts and isolating the connection pool would not resolve these accounting-consistency issues.

Given these considerations, I am inclined to remove the synchronous Redis reservation from the Master creation path, retain local reservations and the existing concurrency safeguards, and move node-level atomic resource admission control to Cubelet.

All creation requests from different Masters targeting the same node eventually pass through that node’s Cubelet. Cubelet can therefore perform a unified check against:

allocated resources + in-flight reservations + the current request’s resource demand

If the request is admissible, Cubelet creates a reservation. Once creation succeeds, the reservation is converted into allocated usage; if creation fails, the reservation is released. The Master remains responsible for node selection and may perform a limited number of retries with alternative nodes if Cubelet rejects the request.

This keeps capacity protection closed-loop at the node where the resources actually reside, without relying on synchronous Redis coordination across Masters.

One important caveat is that Cubelet currently cannot be assumed to reject every request that would exceed CPU or memory capacity.

shsaihdsaiudh:
**Proposed scope cut for this PR**

@fslongjin Thanks for the patient reviews across 6 rounds. After internal discussion, we agree this PR has grown well beyond the task specification (evaluation system + plugin framework + 3 built-in profiles + benchmark/simulator), and the reservation subsystem is the source of every remaining blocker.

We propose to cut this PR back to the task scope:

**Keep (in scope):**
- Scheduler plugin framework (profile compile, factory YAML, CEL/gRPC plugin machinery)
- The three built-in profiles and their scorers
- Evaluation metrics (jain / herding / plugin duration) + schedsim + cube-bench + both reports

**Remove (out of scope, to be split or dropped):**
- The reservation subsystem entirely (local ledger + Redis coordination + `sandbox_run.go` wiring)
- `create_concurrency_score` / `template_local_pressure` (they depend on reservation counters)
- Any Cubelet-side admission control — this deserves its own issue and design discussion, not this PR

Before we start cutting: does this keep-list match your expectation? Once you confirm the boundary, we will push the trimmed branch promptly.

CubeMaster/pkg/service/sandbox/sandbox_run.go
fslongjin: Since an actual production cluster may have thousands of nodes and carry extremely high concurrent creation load, if we imposed strong-consistency requirements on a given node's resource limits on the cubemaster side, it would impact the cluster's sandbox creation throughput under high-concurrency scenarios. Therefore, we previously designed it so that cubelet rejects over-quota resources and returns a creation failure, after which cubemaster excludes the node that just failed and then retries.

JunWang666:
This implements the proposed scope cut from TencentCloud/CubeSandbox#1619 — the reservation subsystem and its two dependent scorers are fully removed while the plugin framework, three factory profiles, and the evaluation tooling are kept intact — could you review when you have a chance?

docs/dev/scheduler-sim-report.md
```diff
@@ -0,0 +1,328 @@
...
+appears only marginally (≤0.03%) because utilization is low. No success-rate
+cost at any setting in this workload. Recommendation: keep 3.0 for maximal
+consolidation; if node-level burst risk matters more than empty-node count,
+1.5 buys 2× herding relief for ~2 extra active nodes.
+
+## template_reuse weight comparison (300 nodes, 5 seeds, seeds 42–46)
+
+> Added 2026-09-10 on the merged baseline (`fix2-template-reuse` worktree).
+> Arbitrates the shipped `template_reuse` weights (image_score /
+> template_local_pressure / real_time_weighted_average = **0.7/0.2/0.3**)
```
fslongjin, line 213: Two documentation leftovers from the scope cut:

- this record still lists `template_local_pressure` in the shipped `template_reuse` weights (0.7/0.2/0.3), and the later sections reason about its behaviour. The scorer is gone, and both `pkg/base/config/scheduler_factory.yaml` and `cmd/schedsim/template_reuse.profiles.sim.yaml` now ship `image_score` 0.7 / `real_time_weighted_average` 0.3. Please update this file or mark it as historical, and say whether the numbers were re-taken without it.
- `docs/guide/multi-node-deploy.md` has the new section, but not the two caveats the Chinese version carries: the rolling-upgrade window, and the fact that the create path still depends on Redis. `docs-bilingual-check.yml` does not cover that file, so nothing catches the drift.

fslongjin (commented):
One more pass. I traced the "node is full" path end to end, and the failover it now relies on is not actually wired up in this branch.

With the master-side reservation gone, over-admission on a chosen node is covered only by the assumption that "Cubelet rejects the create, CubeMaster marks the node bad and reselects". That does not hold today:

- The failover decision is made entirely by the error code. `errorCodeRetry()` (CubeMaster/pkg/service/sandbox/sandbox_run.go:442-479) sets loopRetry/reschedule only for codes in `initLoopRetryMap` (CubeMaster/pkg/errorcode/error.go:176-201). That map is eight codes plus `cubelet_conf.loop_retry_code`, and neither CubeMaster/conf.yaml nor configs/single-node/cubemaster.yaml sets `loop_retry_code` (nor `retry_code` / `circuit_break_code` / `backoff_retry_code`).
- Only three of those eight codes have a producer anywhere in the tree: `MasterRateLimitedError` (master-side rate limiting), `ConcurrentFailed` (the Cubelet create-flow limiter, whose `creation_concurrent_num` is 0, i.e. disabled, in both shipped Cubelet configs), and `NoSpaceLeftOnDevice` (a "no space left" string match at Cubelet/services/cubebox/service.go:372, so disk only). `ConcurrentLimited`, `DeployLayerTimeout`, `DeployCodeTimeout`, `PullImageTimeOut` and `HostDiskNotEnough` appear nowhere outside the enum, the generated API doc and the retry maps themselves.
- Cubelet does no CPU/memory/MVM admission on the create path. `Create` (Cubelet/services/cubebox/service.go:236) reaches `checkResource` (Cubelet/services/cubebox/check.go:148-168), which only calls `resource.ParseQuantity` to validate the *format* of resources.cpu / resources.mem; no quota is compared. QuotaCPU / QuotaMemMB / MaxMvmNum are computed in Cubelet/pkg/cubelet/node_status.go:516-518 and reported to the master for cpufilter/memfilter, but Cubelet does not enforce them locally. The error enum has no CPU/memory/MVM over-quota code at all.
- A genuinely full node therefore fails later and returns codes that are all outside the set: `NewTaskFailed` (130556, Cubelet/services/cubebox/check.go:253 - the fallthrough of `transformError`, where a hypervisor-side allocation failure lands), `StartTaskFailed` (130558, cube_container_create.go:1382), `CreateCgroupFailed` (130562, Cubelet/plugins/cube/internals/cgroup/local.go:352), `SetCubeboxCgroupLimitFailed` (130549, cube_container_create.go:184), `CreateContainerFailed` (130507, cube_container_create.go:328), and `Unknown` via the panic recovery.
- For any of those, `errorCodeRetry()` returns false, so `handleCubelet()` (line 210) leaves the loop and the create fails on the first attempt. The node is not even added to the bad set, because `AddLastBadNode` only runs for `IsCircutBreakCode` (sandbox_run.go:252-255), and that set holds only `HostDiskNotEnough`, `NoSpaceLeftOnDevice`, `ConnHostFailed` and `ReqCubeAPIFailed`.
- Separately, when the master itself cannot find a node that fits, `schedule()` returns `SelectNodesNoRes` (130597) at sandbox_run.go:492, and that is returned immediately - it is in no retry set either.

So in the shipped default configuration "a node cannot fit this request -> fail over to another node" is effectively unreachable, and disk exhaustion is the only capacity-related failure that can actually trigger a retry. The first rejection fails the request outright, which is exactly the window the removed reservation used to cover.

Could you answer in writing:

(a) Is making this a real failover a configuration item - `loop_retry_code` shipped in both sample configs and covered by a test - or a new retryable over-quota code that Cubelet should return?
(b) If it is neither, please withdraw the "Cubelet rejects, CubeMaster reselects" premise from the PR description and the new docs, and state plainly that a full node surfaces as a create failure rather than as a retry.

One small item as well: pkgs/proto/services/schedulerplugin/v1/plugin.pb.go records `protoc v3.21.12` while the other generated files under pkgs/proto record v5.28.3. Please regenerate it with the repo toolchain; the `reserved 24;` handling itself is correct.

Once (a)/(b) is answered and that is regenerated I am happy to approve.

JunWang666:
Thanks for the trace — we verified it against the tree and it holds: `checkResource` only validates quantity format, the quota values are computed solely for status reporting, the shipped configs set no `loop_retry_code`, and the codes a genuinely full node produces are outside every retry and bad-node set. So in the shipped default configuration the "Cubelet rejects, CubeMaster reselects" premise does not hold today.

Our answer is (b), with (a) deferred to the follow-up:

* In this PR we withdraw the premise: the "Cubelet rejects, CubeMaster reselects" wording will be removed from the PR description and the docs, replaced with a plain statement that a full node currently surfaces as a create failure on the first attempt rather than a retry.
* The follow-up PR will make the failover real in the form you outline in (a): Cubelet-side CPU/memory/MVM admission on the create path returning a new retryable over-quota error code, plus the master-side retry and bad-node wiring (shipped `loop_retry_code` defaults and regression tests included). We will open an issue first to discuss that design.

CubeMaster/pkg/scheduler/schedule.go
```diff
@@ -122,118 +313,253 @@ func runPreFilter(selCtx *selctx.SelectorCtx) (err error) {
...
 	if scheduler.postScore != nil {
-		_ = scheduler.postScore.PostedScore(selCtx, resultMap)
+		_ = scheduler.postScore.PostedScore(selCtx, aggregated)
 	}
-
-	result.AllSortByScore()
+	sort.SliceStable(result, func(i, j int) bool { return result[i].Score > result[j].Score })
 	if log.IsDebug() {
-		log.G(selCtx.Ctx).Debugf("runScoreFilter:%v", result.String())
+		log.G(selCtx.Ctx).Debugf("runScoreFilter profile=%s:%v", selCtx.GetProfileName(), result.String())
```
fslongjin: Dead-code leftover from this round: `runScoreFilter` was deleted here, but `runProfileScores` still logs under that old name — this line and line 561. Could you rename the tag to `runProfileScores` (or just `score`)? As it stands, grepping the logs for the scoring path points at a function that no longer exists.

PR #1627, opened 2026-09-02
feat(scheduler): add extensible plugin system

shsaihdsaiudh:
## Summary

Split out from #1619 to fit the auto-review time budget; #1619 remains as the overall design overview. This change originates from fork PR shsaihdsaiudh/CubeSandbox#1 (author @JunWang666, `Signed-off-by` preserved on the commit; original authorship trailer `Autonomously-by: Codex:GPT-5.6` kept).

Add an extensible scheduler plugin system:

- **Explicit plugin registry** (`pkg/selector/plugin`): factory args per plugin instance, fail-fast on unknown plugins, errors instead of panics; Selector interfaces unchanged.
- **Two out-of-tree plugin kinds**: CEL expressions (`plugin/expr`, via `google/cel-go`) and gRPC plugins (`plugin/grpcplugin`) speaking the versioned, batch-oriented `SchedulerPlugin` API (Handshake / SyncSnapshot / Filter / Score; per-request calls carry only a snapshot version, request context and candidate IDs).
- **Named profiles** (`pkg/scheduler/profile`): per-plugin args + named Profiles with a legacy translator mapping the existing config onto the default profile.
- **Builtins migrated onto the same mechanism**: new `node_safety` filter, `template_locality` converted to a builtin plugin.
- **Proto** lives at `pkgs/proto/services/schedulerplugin/v1` following the unified proto module layout (#1600); generated code regenerated with the pinned toolchain (protoc-gen-go v1.36.11 / protoc-gen-go-grpc v1.6.1).
- Docs: `docs/guide/scheduler-plugin.md` (en + zh), runnable example under `CubeMaster/examples/scheduler-plugin/`.

## Test plan

- `cd pkgs/proto && go build ./... && go test ./...` — passing.
- `cd CubeMaster && go build ./...` — passing.
- `cd CubeMaster && go test ./pkg/selector/... ./pkg/scheduler/... ./pkg/base/config/... ./pkg/base/node/...` — passing (all 12 test-bearing packages ok, incl. new `selector/plugin`, `plugin/expr`, `plugin/grpcplugin`, `scheduler/profile`).
- The combined branch this was split from was additionally validated end-to-end on a single-node dev cluster.


shsaihdsaiudh:
Superseded by a two-way split — the auto-review job (15-min timeout) could not finish a pass over this 36-file diff three times in a row, so the PR is split by file with authorship preserved:

- **#1631** — framework: schedulerplugin/v1 proto, plugin registry (Go/CEL/gRPC), node_safety guard, SelectorCtx snapshot API, Node.Clone rework. Self-contained; builds and tests pass standalone.
- **#1632** — integration: profiles, schedule-loop wiring, legacy config adapter, example plugin, docs. Stacked on #1631 (does not build until it merges).

The union of the two branches is byte-identical to this PR's head (b04b1b64) for all 36 files. All review rounds received here (r1-r4) are already incorporated.

PR #1631, opened 2026-09-02
feat(scheduler): add selector plugin framework and schedulerplugin/v1 proto

shsaihdsaiudh:
## What

Framework half of the extensible scheduler plugin system, **split out from #1627** (36 files / +4770) so the auto-review window can finish a pass — that PR's review hit the 15-minute timeout three times in a row.

This PR is self-contained and compiles/tests clean on its own:

- `pkgs/proto/services/schedulerplugin/v1`: plugin gRPC service proto + generated stubs (snapshot sync, filter, score RPCs)
- `pkg/selector/plugin`: unified registry resolving built-in Go / CEL-expression / external-gRPC plugins through one factory API, with a circuit breaker and snapshot-version sync for external plugins
- `pkg/selector/filter/node_safety.go`: guard filter repeating the non-negotiable health checks after backoff candidate expansion
- `pkg/scheduler/selctx`: per-request snapshot surface plugins read (`FreezeSnapshot`/`SnapshotNodes`/`SnapshotFacts`, request labels, profile name)
- `pkg/base/node`: explicit-field `Node.Clone` rework (the old struct copy carried the atomic `noCopy` marker by value)
- `pkg/base/config`: plugin/profile config schema types

The pipeline wiring (profiles in the schedule loop, legacy config adapter, docs) is the companion PR: **integration half follows as a stacked draft**.

## Attribution

Content split by file from #1627, preserving authorship: the feature commit keeps JunWang666 as author with its original trailers; the review-round hardening fixes are a separate commit.

## Verification

- `go build ./...` in `CubeMaster` and `pkgs/proto`: clean
- `go test ./pkg/selector/... ./pkg/base/config/... ./pkg/base/node/... ./pkg/scheduler/selctx/...`: all pass


shsaihdsaiudh:
Closing this split-out review PR: the remaining review pass is moving to a local review. The branch and all review history are retained; this can be reopened at any time.

PR #1632, opened 2026-09-02
feat(scheduler): wire the plugin system into the select pipeline

shsaihdsaiudh:
## What

Integration half of the extensible scheduler plugin system, **split out from #1627** — stacked on the framework half **#1631**, which adds the `pkg/selector/plugin` registry, the `schedulerplugin/v1` proto stubs, and the `SelectorCtx` snapshot API this PR imports. **This branch does not build until #1631 merges** (the plugin packages do not exist on master yet); please review the diff, and see #1631 for the framework itself.

This PR wires the framework into the scheduling pipeline:

- `pkg/scheduler/profile`: declarative profiles — named filter/score chains resolved through the unified registry, with instance-type routing and label-controlled visibility
- `pkg/scheduler/schedule.go` + `init.go`: profile executor in the select loop; per-request `FreezeSnapshot` before plugins run; `StartAsyncScore` keeps the legacy background refresher an explicit hook
- legacy `enable_filters`/`enable_scorers` config adapter (`compileLegacy`), skipping entries with non-positive/NaN/Inf weights instead of wiring degenerate scorers
- `template_locality` filter reads frozen snapshot facts when available
- `examples/scheduler-plugin`: example external gRPC plugin
- bilingual docs (`docs/guide/scheduler-plugin.md` + zh)

## Operational notes (deliberate behavior changes, called out for release notes)

- unknown names in legacy `enable_filters`/`enable_scorers` now fail fast at boot instead of being silently skipped
- an unreachable external gRPC plugin blocks scheduling startup for profiles that reference it

## Attribution

Content split by file from #1627, preserving authorship: the feature commit keeps JunWang666 as author with its original trailers; the review-round hardening fixes are a separate commit.


shsaihdsaiudh:
Closing this split-out review PR: the remaining review pass is moving to a local review. The branch and all review history are retained; this can be reopened at any time.

PR #1638, opened 2026-09-03
feat(chart): wire top-level redis.db for Master, Proxy, and LCM

xiaojunxiang2023:
## Summary

- CubeMaster Helm conf previously hard-coded `db_no: 0`, while CubeProxy / cube-lifecycle-manager read nested `*.redis.db` values. Sharing one Redis/Sentinel across Cube clusters therefore could not select a non-zero logical DB without risking Master vs Proxy/LCM split-brain.
- Add top-level `redis.db` (default `0`) and render it into Master conf plus Proxy / LCM / Ops Redis env via `cube.redisDB`.
- Remove nested `cubeProxy.redis.db` / `lifecycleManager.redis.db` from values. `validate.yaml` fails at render when a non-zero leftover is set (points to top-level `redis.db`); nested `db: 0` is tolerated so ordinary `helm upgrade` of existing releases is not blocked by chart-default leftovers.

## Test plan

- [x] `helm lint deploy/kubernetes/chart --set-string mysql.password=test --set-string mysql.rootPassword=test --set-string redis.password=test` — passed
- [x] `bash deploy/kubernetes/chart/scripts/test-redis-db-guard.sh` — passed (default `0`, override `3`, non-zero nested fails even with Ops off, nested `db:0` upgrade leftover renders)
- [x] `bash deploy/kubernetes/chart/scripts/test-redis-sentinel-guard.sh` — passed

### DIFF
--- a/deploy/kubernetes/chart/files/cube-master/conf.yaml
+++ b/deploy/kubernetes/chart/files/cube-master/conf.yaml
@@ -72,7 +72,7 @@ redis:
   sentinel_nodes: {{ .Values.redis.sentinelNodes | default "" | quote }}
   sentinel_password: {{ .Values.redis.sentinelPassword | default "" | quote }}
   password: {{ .Values.redis.password | quote }}
-  db_no: 0
+  db_no: {{ include "cube.redisDB" . }}
   max_idle: 8
   max_active: 32
   idle_timeout: 30
--- a/deploy/kubernetes/chart/scripts/test-redis-db-guard.sh
+++ b/deploy/kubernetes/chart/scripts/test-redis-db-guard.sh
@@ -60,7 +60,7 @@ helm template redis-db-default "$CHART_DIR" $COMMON_SETS \
   --set cubeOps.enabled=true \
   > "$TMP_DIR/default.yaml"
 
-grep -q 'db_no: 0' "$TMP_DIR/default.yaml" || {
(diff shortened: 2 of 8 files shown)


fslongjin:
**Please remove the nested `db` fields instead of deprecating them**

Right now `cubeProxy.redis.db` and `lifecycleManager.redis.db` are still in `values.yaml` with a Deprecated comment, but the chart ignores them. That's easy to miss on upgrade — people will keep setting those keys and think they work.

This chart usually handles removed knobs by failing at render time (e.g. `cubeProxy.sidecar`), not by leaving the key around and silently ignoring it.

**Suggested change in this PR:**

1. Delete `cubeProxy.redis.db` and `lifecycleManager.redis.db` from `values.yaml` — no Deprecated placeholder.
2. In `validate.yaml`, fail if someone still passes them (same style as other removed fields), and point them to top-level `redis.db`.
3. Update the guard test so it checks that the old keys **fail render**, not that they're silently ignored.

That way upgrades break loudly with a clear message, instead of quietly changing behavior.

xiaojunxiang2023:
@fslongjin Thanks for the review — agreed.

Removed `cubeProxy.redis.db` and `lifecycleManager.redis.db` from values (no Deprecated placeholders). `validate.yaml` now fails at render if either key is still set, pointing to top-level `redis.db`. The guard script asserts those keys fail render (instead of silent ignore) and checks every env occurrence.

PR #1639, opened 2026-09-03
feat(pause): incremental memory dump like snapshot and keep XFS pause package

ls-ggg:
## Summary
- Pause now writes memory the same way CommitSandbox does: clone a dest, then incremental-overlay dirty pages (`snapshot_type` on the shim Pause path).
- After Resume, the next Pause overlays onto the last-restore image (XFS keeps the pause package on disk; S3 already cloned onto `sb-*-memory` and still drops the package).
- Master still always sends `CleanupTemplate`. Cubelet no-ops that RPC on XFS while a live sandbox holds `cube.master.pause.snapshot.id`; the next Pause or Destroy actually deletes the package.

## Test plan
- [x] `make cubelet` / `make cubemaster` on builder
- [x] Cubelet `go test ./pkg/constants/ ./services/cubebox/`
- [x] Cubelet storage unit tests for the S3 volume-source clone path
- [x] CubeMaster `go test ./pkg/service/sandbox/ -run Resume|Pause`
- [x] Shim `cargo test pause_snapshot_config`
- [x] Cluster: XFS keep-package loop, Pause1–4 + write load, S3 same-node / cross-node Resume; wiki combo minus known Device-busy skips; B6 after cubelet restart once cubelet is ready

### DIFF
--- a/CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
@@ -446,11 +446,14 @@ func resumeFromPauseSnapshot(ctx context.Context, req *types.UpdateRequest, host
 	// describes a sandbox that is by now running whatever the proxy thinks.
 	purgeErr := cubeproxy.InvalidateBackendCache(ctx, req.SandboxID, targetIP)
 
-	// After Create the sandbox runs on private disks. Drop the pause package
-	// on the node that still holds it: origin when that IP is known and not
-	// the target (also the PAUSED tombstone), otherwise this node. Do not
-	// key this off CrossNode — a cross-node placement with a blank origin
-	// would otherwise skip both cleanup paths and leak the package.
+	// After Create the sandbox runs on private disks. Ask Cubelet to drop
+	// the pause package on the node that still holds it: origin when that
+	// IP is known and not the target (also the PAUSED tombstone), otherwise
+	// this node. Cubelet decides whether the catalog actually goes away
+	// (S3 already cloned onto sb-*-memory; XFS Resume still mmaps the
+	// package and no-ops until the next Pause or Destroy). Do not key this
+	// off CrossNode — a cross-node placement with a blank origin would
+	// otherwise skip both cleanup paths and leak the package.
 	origin := strings.TrimSpace(rec.NodeIP)
 	if origin != "" && origin != targetIP {
 		if err := pausesnap.DropOriginTombstone(ctx, req.RequestID, req.SandboxID, origin, rec.SnapshotID, rec.Backend); err != nil {
--- a/CubeShim/shim/src/service/update_ext.rs
+++ b/CubeShim/shim/src/service/update_ext.rs
@@ -5,6 +5,7 @@
 use crate::log::Log;
 use crate::{common::CResult, errf, infof, sandbox::sb::SandBox};
 use cube_hypervisor::config::RestoreConfig;
+use cube_hypervisor::SnapshotType;
 use serde::{Deserialize, Serialize};
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
@@ -104,6 +105,20 @@ struct PauseSnapshotConfig {
     /// CommitSandbox layout (`--memory-vol`).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub memory_vol_url: Option<String>,
+
+    /// Same values as cube-runtime `--snapshot-type`: `full`, `incremental`,
+    /// `soft-dirty`. Missing / empty / unknown defaults to Full so older
+    /// Cubelets keep the historical full dump.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub snapshot_type: Option<String>,
+}
+
+fn parse_pause_snapshot_type(raw: Option<&str>) -> SnapshotType {
+    let s = raw.map(str::trim).unwrap_or("");
+    if s.is_empty() {
+        return SnapshotType::Full;
+    }
+    s.parse().unwrap_or(SnapshotType::Full)
 }
 
 /// Outcome of an extended update action.
@@ -204,14 +219,16 @@ async fn do_pause_to_snapshot(
         .into());
     }
 
+    let snapshot_type = parse_pause_snapshot_type(pause_cfg.snapshot_type.as_deref());
     infof!(
         log,
-        "pause to snapshot: destination={} memory_vol_url={:?}",
+        "pause to snapshot: destination={} memory_vol_url={:?} snapshot_type={}",
         destination_path,
-        pause_cfg.memory_vol_url
+        pause_cfg.memory_vol_url,
+        snapshot_type
     );
 
-    sb.pause_vm_to_snapshot(&destination_path, pause_cfg.memory_vol_url)
+    sb.pause_vm_to_snapshot(&destination_path, pause_cfg.memory_vol_url, snapshot_type)
         .await
         .map_err(|e| {
             errf!(log, "pause to snapshot failed: {}", e);
@@ -273,5 +290,28 @@ mod tests {
             cfg.memory_vol_url.as_deref(),
             Some("file:///dev/cubecow/mem1")
         );
+        assert_eq!(cfg.snapshot_type, None);
+        assert_eq!(
+            parse_pause_snapshot_type(cfg.snapshot_type.as_deref()),
+            SnapshotType::Full
+        );
+    }
+
+    #[test]
+    fn pause_snapshot_config_parses_soft_dirty() {
+        let raw = r#"{"destination_url":"/data/snap/pause-1","memory_vol_url":"file:///dev/cubecow/mem1","snapshot_type":"soft-dirty"}"#;
+        let cfg: PauseSnapshotConfig = serde_json::from_str(raw).unwrap();
+        assert_eq!(cfg.snapshot_type.as_deref(), Some("soft-dirty"));
+        assert_eq!(
+            parse_pause_snapshot_type(cfg.snapshot_type.as_deref()),
+            SnapshotType::SoftDirty
+        );
+        assert_eq!(
+            parse_pause_snapshot_type(Some("incremental")),
+            SnapshotType::Incremental
+        );
+        assert_eq!(parse_pause_snapshot_type(Some("weird")), SnapshotType::Full);
+        assert_eq!(parse_pause_snapshot_type(Some("")), SnapshotType::Full);
+        assert_eq!(parse_pause_snapshot_type(None), SnapshotType::Full);
     }
 }
--- a/Cubelet/pkg/constants/const.go
+++ b/Cubelet/pkg/constants/const.go
@@ -208,12 +208,12 @@ const (
 	// the local catalog under this id; Kind=pause_snapshot.
 	MasterAnnotationPauseSnapshotID = "cube.master.pause.snapshot.id"
 	// MasterAnnotationLaunchMemorySnapshotID is the template or customer
-	// snapshot the sandbox was first started from. Pause always clones
-	// this memory image and writes an incremental overlay. Resume and
-	// Commit must not overwrite it.
+	// snapshot the sandbox was first started from. Pause may clone it only
+	// when it is still the VM's last restore (first Pause after
+	// Create-from-template). Resume and Commit must not overwrite it.
 	MasterAnnotationLaunchMemorySnapshotID = "cube.master.launch.memory.snapshot.id"
 	// MasterAnnotationStorageBackend is the CoW backend Master passes on
-	// Pause / Commit (xfs｜s3). Empty means xfs.
+	// Pause / Commit (xfs/s3). Empty means xfs.
 	MasterAnnotationStorageBackend = "cube.master.storage.backend"
 	// MasterAnnotationSnapshotRemoteUUIDs is the JSON blob of remote
 	// volume uuids (rootfs/memory/metadata) for cubecow_import_lvol.
--- a/Cubelet/pkg/constants/const.go
+++ b/Cubelet/pkg/constants/const.go
@@ -207,6 +207,11 @@ const (
 	// id (same snap-* format as normal Commit snapshots). Cubelet only stores
 	// the local catalog under this id; Kind=pause_snapshot.
 	MasterAnnotationPauseSnapshotID = "cube.master.pause.snapshot.id"
+	// MasterAnnotationLaunchMemorySnapshotID is the template or customer
+	// snapshot the sandbox was first started from. Pause always clones
+	// this memory image and writes an incremental overlay. Resume and
+	// Commit must not overwrite it.
+	MasterAnnotationLaunchMemorySnapshotID = "cube.master.launch.memory.snapshot.id"
 	// MasterAnnotationStorageBackend is the CoW backend Master passes on
 	// Pause / Commit (xfs｜s3). Empty means xfs.
 	MasterAnnotationStorageBackend = "cube.master.storage.backend"
--- a/Cubelet/services/cubebox/cube_container_create.go
+++ b/Cubelet/services/cubebox/cube_container_create.go
@@ -270,6 +270,7 @@ func (l *local) createContainers(ctx context.Context, flowOpts *workflow.CreateC
 	}
 
 	l.storeNumaQueues(ctx, sandBox, flowOpts)
+	stampLaunchMemoryAncestorOnce(sandBox, resolveLaunchAncestorSnapshotID(sandBox))
 	if snapshotID, ok := flowOpts.GetSnapshotTemplateID(); ok && flowOpts.IsRetoreSnapshot() {
 		now := time.Now().UTC()
 		setRuntimeSnapshotBindingLabels(sandBox, snapshotID, now)
@@ -280,9 +281,10 @@ func (l *local) createContainers(ctx context.Context, flowOpts *workflow.CreateC
 		// reflinkable base after the most recent commit's snapshot is
 		// deleted.
 		setRuntimeRestoreBaseLabels(sandBox, snapshotID, now)
-		// Resume-from-pause stamps pause snapshot id so Destroy can GC a
-		// leftover pause catalog if Pause-time CleanupTemplate of the
-		// previous live snap missed it.
+		// Resume-from-pause stamps pause snapshot id. XFS Resume still
+		// mmaps that package; CleanupTemplate no-ops while this label
+		// is live. Next Pause or Destroy GCs it. S3 already cloned onto
+		// sb-*-memory and CleanupTemplate deletes the package.
 		if pauseID := strings.TrimSpace(realReq.GetAnnotations()[constants.MasterAnnotationPauseSnapshotID]); pauseID != "" {
 			if sandBox.Labels == nil {
 				sandBox.Labels = map[string]string{}
--- a/Cubelet/services/cubebox/cube_container_create.go
+++ b/Cubelet/services/cubebox/cube_container_create.go
@@ -281,10 +281,12 @@ func (l *local) createContainers(ctx context.Context, flowOpts *workflow.CreateC
 		// reflinkable base after the most recent commit's snapshot is
 		// deleted.
 		setRuntimeRestoreBaseLabels(sandBox, snapshotID, now)
-		// Resume-from-pause stamps pause snapshot id. XFS Resume still
-		// mmaps that package; CleanupTemplate no-ops while this label
-		// is live. Next Pause or Destroy GCs it. S3 already cloned onto
-		// sb-*-memory and CleanupTemplate deletes the package.
+		// Resume-from-pause stamps pause snapshot id. Master strips this
+		// key from user Create; only the thin Resume request carries it.
+		// XFS Resume still mmaps that package; CleanupTemplate no-ops
+		// while this label is live. Next Pause or Destroy GCs it. S3
+		// already cloned onto sb-*-memory and CleanupTemplate deletes
+		// the package.
 		if pauseID := strings.TrimSpace(realReq.GetAnnotations()[constants.MasterAnnotationPauseSnapshotID]); pauseID != "" {
 			if sandBox.Labels == nil {
 				sandBox.Labels = map[string]string{}
--- a/Cubelet/services/cubebox/pause_cow.go
+++ b/Cubelet/services/cubebox/pause_cow.go
@@ -31,6 +31,20 @@ import (
 type pauseSnapshotConfig struct {
 	DestinationURL string  `json:"destination_url"`
 	MemoryVolURL   *string `json:"memory_vol_url,omitempty"`
+	// SnapshotType is the same wire value CommitSandbox passes to
+	// cube-runtime (--snapshot-type): soft-dirty, incremental, or full.
+	SnapshotType string `json:"snapshot_type,omitempty"`
+}
+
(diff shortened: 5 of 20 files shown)


Cubelet/pkg/constants/const.go
```diff
@@ -207,6 +207,11 @@ const (
 	// id (same snap-* format as normal Commit snapshots). Cubelet only stores
 	// the local catalog under this id; Kind=pause_snapshot.
 	MasterAnnotationPauseSnapshotID = "cube.master.pause.snapshot.id"
+	// MasterAnnotationLaunchMemorySnapshotID is the template or customer
+	// snapshot the sandbox was first started from. Pause always clones
```
fslongjin: The comment says "Pause always clones this memory image and writes an incremental overlay", but `preparePauseMemoryArtifact` only clones the launch ancestor when `launchAncestorIsLastRestore` is true. After Resume, restore-base is the pause package; the actual paths are last-restore catalog or `sb-*-memory`.

Suggested wording:
"Records the template/customer snapshot this sandbox was first started from. Pause may clone it only when it is still the VM's last restore (first Pause after Create-from-template)."

Cubelet/services/cubebox/snapshot_base_memory.go
```diff
@@ -115,6 +115,87 @@ func resolveRestoreBaseMemoryObject(ctx context.Context, cb *cubeboxstore.CubeBo
 	return resolveMemoryObjectFromSnapshotID(ctx, backend, resolveRestoreBaseSnapshotID(cb))
 }
 
+// resolveLaunchAncestorSnapshotID is the memory base Pause copies from:
+// the template or customer snapshot this sandbox was first started from.
+// It ignores last-commit / last-restore labels and the pause snap itself
+// (those change across Pause／Resume／Commit).
```
fslongjin: [nit / policy] This PR introduces fullwidth `／` in new comments, which violates the repo non-English character policy. Please use ASCII `/` instead, e.g. `Pause/Resume/Commit`.

Same issue on L143, L152, and L237 in this file.

Cubelet/services/cubebox/snapshot_base_memory.go
```diff
@@ -142,9 +223,97 @@ func resolveMemoryObjectFromSnapshotID(ctx context.Context, backend, snapshotID
...
+//     S3 Resume deletes the pause package after cloning onto this disk.
+//  2. Launch ancestor, but only when that image is the VM's last restore
+//     (first Pause after Create-from-template). See
+//     [launchAncestorIsLastRestore].
+//  3. Last-restore catalog. XFS Resume keeps the pause package so this
+//     Pause can incremental-overlay onto that image; Cubelet GCs the
+//     old package after this Pause succeeds (or on Destroy).
+//  4. Full dump into an empty volume when no catalog／volume base is
+//     left (template gone, or an S3 own-volume clone that failed).
+func preparePauseMemoryArtifact(
```
fslongjin, line 239: `preparePauseMemoryArtifact` is the core behavioral change, but only helpers like `launchAncestorIsLastRestore` / `resolveLaunchAncestorSnapshotID` have unit tests today — no table-driven coverage for the four-tier fallback (own vol → ancestor → last-restore → full).

Suggested cases:
1. S3 Resume with own vol → incremental
2. First Pause after Create-from-template → clone ancestor
3. XFS Pause after Resume → clone last-restore (pause catalog)
4. Base unavailable → full dump

Non-blocking, but this is exactly the holey-memory fix path.

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -78,9 +92,57 @@ func stampedPauseSnapshotID(sb *cubeboxstore.CubeBox) string {
...
+	}
+	return s.cubeboxMgr.cubeboxManger.List()
+}
+
+// keepLiveXFSPausePackage is true when Master's Resume-time CleanupTemplate
+// must leave this pause catalog on disk. XFS Resume mmaps the package file;
+// S3 already cloned onto sb-*-memory and must drop the package. PAUSED /
+// EXITED / UNKNOWN do not hold a live mmap, so DelPaused and leftover GC
+// still delete.
+func keepLiveXFSPausePackage(boxes []*cubeboxstore.CubeBox, snapID, backend string) bool {
```
fslongjin, line 114: **[security / medium]** `keepLiveXFSPausePackage` only checks `stampedPauseSnapshotID(sb) == snapID` and CREATED/RUNNING/PAUSING, but the pause label can be injected via user Create requests:

- Master `checkAndGetAnnotation` forwards all `cube.master.*` annotations (see `CubeMaster/pkg/service/sandbox/util.go` — not in this diff)
- `cube_container_create.go` Resume path copies annotations into Labels

An attacker on a shared XFS node can run a sandbox with `cube.master.pause.snapshot.id=<victim catalog id>` and cause `CleanupTemplate` for that id to succeed as a no-op, blocking template/snapshot GC.

Suggestions:
1. Master should strip user-supplied `cube.master.pause.snapshot.id` / `cube.master.launch.memory.snapshot.id`
2. The keep gate should also verify catalog `Kind=pause_snapshot` and a real restore binding, not label match alone.

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -122,6 +184,7 @@ func (s *service) updateWithPauseCow(
 	}
 	stepLog = stepLog.WithFields(CubeLog.Fields{"backend": backend})
 	prevLiveSnap := stampedPauseSnapshotID(sb)
+	stampLaunchMemoryAncestorOnce(sb, resolveLaunchAncestorSnapshotID(sb))
 	stampPauseSnapshotID(sb, snapID)
```
fslongjin, line 218: Pause stamps `stampPauseSnapshotID(sb, snapID)` and Syncs before the shim finishes writing the snapshot. If a delayed Resume `CleanupTemplate` for the *previous* pause package arrives here, the keep gate evaluates against the **new** snapID and may allow deleting the old package.

I believe XFS unlink is still safe while mmap is open; worst case is catalog loss and the next Pause falls back to full dump. Worth a one-line comment if you've seen this race in cluster tests; otherwise feel free to ignore.

Cubelet/services/cubebox/pause_cow.go
```diff
@@ -172,7 +235,8 @@ func (s *service) updateWithPauseCow(
 		rsp.Ret.RetMsg = fmt.Sprintf("failed to resolve sandbox rootfs: %v", err)
 		return rsp, nil
 	}
-	memoryObject, err := storage.CreateMemoryVolumeFor(ctx, backend, snapID, memorySizeBytes)
+	// Own volume if this sandbox has one; otherwise the start template／snapshot.
```
fslongjin: [nit / policy] `template／snapshot` → `template/snapshot` (fullwidth slash).

Cubelet/services/cubebox/cube_container_create.go
```diff
@@ -280,9 +281,10 @@ func (l *local) createContainers(ctx context.Context, flowOpts *workflow.CreateC
...
 		// deleted.
 		setRuntimeRestoreBaseLabels(sandBox, snapshotID, now)
-		// Resume-from-pause stamps pause snapshot id so Destroy can GC a
-		// leftover pause catalog if Pause-time CleanupTemplate of the
-		// previous live snap missed it.
+		// Resume-from-pause stamps pause snapshot id. XFS Resume still
+		// mmaps that package; CleanupTemplate no-ops while this label
+		// is live. Next Pause or Destroy GCs it. S3 already cloned onto
+		// sb-*-memory and CleanupTemplate deletes the package.
 		if pauseID := strings.TrimSpace(realReq.GetAnnotations()[constants.MasterAnnotationPauseSnapshotID]); pauseID != "" {
```
fslongjin, line 290: Resume-from-pause copies `realReq.Annotations[cube.master.pause.snapshot.id]` into Labels without verifying the id was issued by Master's pause/resume flow.

Combined with `keepLiveXFSPausePackage`, a user-injected pause id on Create can trigger the XFS Cleanup no-op (see `template_ops.go` L512).

Suggestion: only accept Master-issued thin Create annotations here, or at least bind to `desired.sandbox.id` / internal restore state.

Cubelet/services/cubebox/template_ops.go
```diff
@@ -501,6 +509,11 @@ func (s *service) CleanupTemplate(ctx context.Context, req *cubebox.CleanupTempl
 			}
 		}
 	}
+	if honorLiveXFSPause && keepLiveXFSPausePackage(s.listCubeboxes(), rsp.TemplateID, backend) {
```
fslongjin: The XFS keep-package no-op is reasonable (VM still mmaps the pause file after Resume), but the implementation only scans local CubeBox pause labels — no catalog kind or restore binding check.

Please harden together with `pause_cow.go` / `cube_container_create.go`. If Master changes are deferred, at least require catalog `Kind=pause_snapshot` here to reduce the chance of blocking unrelated tpl/snap deletes.

Cubelet/storage/s3_volume_manager.go
```diff
@@ -173,11 +173,14 @@ func (m *S3Cow) CreateMemoryVolume(ctx context.Context, templateID string, sizeB
...
 
 // CommitTemplateMemory derives a writable memory volume for templateID from
-// a RO memory snapshot (create_volume_from_snapshot). Source must be a
-// snapshot — seal prior package memory before using it as a clone base.
+// a RO memory snapshot (create_volume_from_snapshot). A live sandbox volume
+// (imported / cloned sb-*-memory) is snapshotted first, then cloned.
 func (m *S3Cow) CommitTemplateMemory(ctx context.Context, sourceName, templateID string, sizeBytes uint64) (*cowVolume, error) {
 	volumeName := cowTemplateMemoryName(templateID)
 	devPath, err := m.createOrResolveVolumeFromSnapshot(ctx, sourceName, volumeName)
+	if err != nil && isS3CloneSourceMustBeSnapshot(err) {
```
fslongjin, line 181: `isS3CloneSourceMustBeSnapshot` matches substrings in `err.Error()` to detect "source is a volume, not a snapshot". If cubecow changes the message, the live-volume clone path after S3 Resume→Pause silently falls back to full dump.

Prefer a typed/sentinel error from cubecow or `errors.Is` here. `s3_pause_memory_src_test.go` only locks the string, not the real clone path.

Cubelet/storage/s3_volume_manager.go
```diff
@@ -215,6 +218,17 @@ func (m *S3Cow) CloneSandboxMemory(ctx context.Context, sandboxID, sourceName st
 	return cloneSandboxMemory(ctx, m, sandboxID, sourceName)
 }
 
+// cloneVolumeFromWritableSource snapshots a live volume, clones that snap
+// into dest, then deletes the snap. The snap exists only inside this call.
+func (m *S3Cow) cloneVolumeFromWritableSource(ctx context.Context, sourceVol, destVol string) (string, error) {
+	srcSnap := destVol + "-src"
```
fslongjin, line 224: nit: temp snap name `destVol + "-src"` assumes dest is globally unique. Fine today with Master-allocated `snap-*` ids; if naming rules change, consider a sandbox or request id prefix.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
```diff
@@ -446,11 +446,14 @@ func resumeFromPauseSnapshot(ctx context.Context, req *types.UpdateRequest, host
...
 
-	// After Create the sandbox runs on private disks. Drop the pause package
-	// on the node that still holds it: origin when that IP is known and not
-	// the target (also the PAUSED tombstone), otherwise this node. Do not
-	// key this off CrossNode — a cross-node placement with a blank origin
-	// would otherwise skip both cleanup paths and leak the package.
+	// After Create the sandbox runs on private disks. Ask Cubelet to drop
+	// the pause package on the node that still holds it: origin when that
+	// IP is known and not the target (also the PAUSED tombstone), otherwise
+	// this node. Cubelet decides whether the catalog actually goes away
```
fslongjin, line 452: The comment correctly notes Cubelet decides whether the catalog goes away. Related security concern (upstream, `util.go` not in this diff): Master forwards user-supplied `cube.master.*` annotations on Create, which can combine with the new XFS keep gate to block `CleanupTemplate` for unrelated catalogs. Please add a denylist on Master for platform-only keys like `cube.master.pause.snapshot.id`.

CubeShim/shim/src/service/update_ext.rs
```diff
@@ -104,6 +105,20 @@ struct PauseSnapshotConfig {
     /// CommitSandbox layout (`--memory-vol`).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub memory_vol_url: Option<String>,
+
+    /// Same values as cube-runtime `--snapshot-type`: `full`, `incremental`,
```
fslongjin, line 109: Nice: defaulting missing/unknown `snapshot_type` to `SnapshotType::Full` keeps older Cubelets working. `pause_snapshot_config_parses_soft_dirty` covers the wire format well.

fslongjin (changes requested):
## Review: Request changes

Thanks for tackling a real correctness issue — after Resume, Pause onto the wrong memory base can produce holey memory (agent health may still respond, then internal RPCs like `set_guest_date_time` / reseed hang). This PR aligns Pause with Commit (clone base + incremental overlay) and keeps the XFS pause catalog on disk across Resume via keep-package. The overall design looks sound to me.

**What works well**
- `launchAncestorIsLastRestore` + `preparePauseMemoryArtifact` four-tier fallback is clear; comments explain why pagemap_anon vs soft-dirty windows differ.
- Master still always sends `CleanupTemplate`; Cubelet decides whether the catalog is actually removed — rollout-friendly.
- Shim `snapshot_type` defaults to Full when missing/unknown — backward compatible with older Cubelets.
- Unit tests for branch predicates (`TestLaunchAncestorIsLastRestore`, `TestKeepLiveXFSPausePackage`, etc.) cover the easiest mistakes.

**Please address before merge**
1. **Docs** — no `docs/` changes, but Pause is no longer full-copy; `lifecycle` / `cross-node-snapshot` guides and `examples/snapshot-rollback-clone/bench_pause_resume_concurrency.py` still describe full-memory-copy. User-visible XFS behavior (pause catalog remains on disk after Resume until next Pause/Destroy) is also undocumented.
2. **Security** — `keepLiveXFSPausePackage` only matches `cube.master.pause.snapshot.id` labels, while Master `checkAndGetAnnotation` (`CubeMaster/pkg/service/sandbox/util.go`) forwards all user-supplied `cube.master.*` annotations. On a shared XFS node, a forged label can make `CleanupTemplate` succeed as a no-op and block another tenant's catalog GC (see inline comments).
3. **Repo policy** — this PR introduces fullwidth `／` in new comments (non-English character policy); `const.go` says Pause always clones the launch ancestor (implementation differs); commit lacks `Assisted-by` (PR body mentions Made with Cursor).

**Follow-ups (non-blocking but valuable)**
- e2e: Pause → Resume → write state → Pause → Resume (`tests/e2e/sdk_compat/cases/lifecycle/test_pause_resume.py` still only covers one round).
- Table-driven tests for `preparePauseMemoryArtifact` (one case per tier).
- Replace S3 `isS3CloneSourceMustBeSnapshot` string matching with a typed/sentinel error.

**Docs to add (not in this diff)**
| File | Gap |
|------|-----|
| `docs/guide/lifecycle.md` + `docs/zh/guide/lifecycle.md` | Pause is incremental; XFS running sandbox still holds pause catalog after Resume |
| `docs/guide/cross-node-snapshot.md` + zh | S3 drops package after Resume; XFS keeps it |
| benchmark blog posts (optional) | footnote that numbers predate incremental Pause |

**Commit metadata (nit)**
- `feat(pause):` prefix looks good; please add `Assisted-by: ...` per AGENTS.md.
- Single commit spans Cubelet/Shim/Master — minor deviation from one-component-per-commit.

Core logic direction is right; happy to LGTM after docs, security boundary, and policy nits are fixed.

ls-ggg:
Same-node XFS retest after deploying this PR's cubelet and cubemaster (shim already had `snapshot_type`). Per-instance latency, not batch-wall / concurrency. Warmup discarded; formal 5 rounds × 10 = n=50. After each resume: guest `echo ok`, pause-front marker, and `dd` 2MiB. Finished with 0 sandboxes.

Also retested same-node S3 Pause/Resume (two cycles, one export at a time, wait remote=ready). S3 Resume still drops the pause package (keep is XFS-only). Both Pauses were incremental; the second Pause after Resume stayed incremental (private `sb-*-memory`, not a full dump). Guest marker + `dd` 2MiB both cycles. Finished with 0 sandboxes; package-object busy leak on S3 delete is by design.

Follow-up commit on this branch tightens the XFS keep gate: only Cubelet Labels count, catalog must be `Kind=pause_snapshot`, restore-base still keeps the previous package while Pause stamps a new id, and user Create can no longer forward pause / launch / restore-base annotations.

**Functional**

- Resume keeps the XFS pause package; the next Pause incremental-overlays onto that last-restore image and GCs the previous package; Destroy removes the remaining one.
- First Pause after create-from-template is incremental (`snapshot_type=Incremental`). VMM pagemap_anon wrote ~10.4MiB of a 2GiB image (~99.5% savings).
- Pause1–4 with guest write load (touch 48MiB RAM + write 8MiB between pauses), 1 round: pause1 268ms (~26 extents) → pause2 805ms (~6849) → pause3 956ms → pause4 932ms (~10462). Resume stayed ~80–110ms aside from one 663ms outlier. Same shape as before: first Pause clones the warm template; later Pauses after Resume are fragmented 4K overlays.

**c=10 first Pause / Resume (n=50)**

| Op | avg | p95 | min | max |
|---|---:|---:|---:|---:|
| Pause | 222 | 263 | 146 | 267 |
| Resume | 212 | 223 | 196 | 225 |

**c=10 two-cycle Pause / Resume (n=50)**

| Op | avg | p95 | min | max |
|---|---:|---:|---:|---:|
| Pause1 | 218 | 277 | 149 | 294 |
| Resume1 | 213 | 224 | 194 | 226 |
| Pause2 | 4213 | 4389 | 3942 | 4420 |
| Resume2 | 224 | 252 | 197 | 260 |

Pause2 is still ~4s because 10 sandboxes concurrently clone the last-restore package then overlay. Resume stays ~210–220ms on both cycles.

**S3 same-node (two cycles)**

| Op | time |
|---|---:|
| Create | 470ms |
| Pause1 | 2.3s (`remote=ready`, Incremental, ~69MiB / 2GiB, 96.8% savings) |
| Resume1 | 693ms (package directory dropped) |
| Pause2 | 2.6s (new snap id, Incremental, ~54MiB, 97.5% savings) |
| Resume2 | 736ms (package dropped again) |

Did not run S3 c=10 or concurrent export (one export at a time).

ls-ggg:
@fslongjin Thanks for the review. Follow-up `a91ab1b8` covers the blocking security / comment / policy items. Same-node XFS and S3 were retested after that commit (numbers in the previous comment).

**Security (keep gate + Master denylist)** — done

- `keepLiveXFSPausePackage` only reads Cubelet-stamped **Labels**, not Create annotations.
- Keep also requires catalog `Kind=pause_snapshot`, so a forged label cannot pin a template or customer snap.
- Master `checkAndGetAnnotation` strips user Create keys: `pause.snapshot.id`, `launch.memory.snapshot.id`, `runtime.restore.snapshot.id` / `attached_at`. Resume thin Create does not go through this path and still stamps Labels itself.
- Resume-from-pause still copies `pause.snapshot.id` into Labels; that id now has to come from Master, not a client Create.

**Delayed Cleanup vs new pause id** — done

Pause stamps the new id before the overlay finishes. Keep now also matches the **restore-base** Label (Pause does not advance that), so a late `CleanupTemplate` for the previous package still no-ops while the VM mmaps / incrementally overlays it. Covered by `TestKeepLiveXFSPausePackage`.

**Comment / policy nits** — done

- `const.go`: Pause clones the launch ancestor only when it is still last restore (first Pause after Create-from-template).
- New comments use ASCII `/` only.

**Leaving as follow-up (as you marked non-blocking)**

- Docs (`lifecycle` / `cross-node-snapshot`): Pause is incremental; XFS keeps the catalog after Resume, S3 drops it. Can land in a docs PR.
- Table-driven `preparePauseMemoryArtifact` four-tier cases.
- Typed/sentinel error instead of `isS3CloneSourceMustBeSnapshot` string match.
- sdk_compat two-round Pause/Resume.

**Assisted-by**

Skipping the trailer. Standing rule on this repo is no `Assisted-by` / `Autonomously-by` on commits or PR text.

Happy to take another look if anything above still looks short.

Cubelet/pkg/constants/const.go
ls-ggg: Fixed in a91ab1b8: comment now says Pause may clone the launch ancestor only when it is still last restore (first Pause after Create-from-template).

Cubelet/services/cubebox/snapshot_base_memory.go
ls-ggg: Fixed in a91ab1b8: new comments use ASCII / only (this file and the other callouts).

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Fixed in a91ab1b8: keep reads Labels only (not Create annotations), requires catalog Kind=pause_snapshot, and Master checkAndGetAnnotation strips pause / launch / restore-base keys on user Create. TestKeepLiveXFSPausePackage covers a forged annotation.

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Addressed in a91ab1b8: keep also matches the restore-base Label, which Pause does not advance, so a late Cleanup of the previous package still no-ops. Added a keep-gate comment and a unit case.

Cubelet/services/cubebox/pause_cow.go
ls-ggg: Fixed: ASCII /.

Cubelet/services/cubebox/cube_container_create.go
ls-ggg: Master now strips those keys on user Create (a91ab1b8). This Resume path still copies pause.snapshot.id into Labels, but that annotation is Master-issued. The keep gate no longer honors a Create annotation.

Cubelet/services/cubebox/template_ops.go
ls-ggg: Fixed in a91ab1b8: CleanupTemplate keep also requires catalog Kind=pause_snapshot.

CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
ls-ggg: Done in a91ab1b8: checkAndGetAnnotation drops user-supplied pause.snapshot.id, launch.memory.snapshot.id, and restore-base keys. TestCheckAndGetAnnotationStripsPlatformPauseKeys covers it.

Cubelet/services/cubebox/snapshot_base_memory.go
ls-ggg: Leaving as follow-up as marked. The four-tier order is unchanged; cluster/XFS retest after the keep harden still hit own-vol (S3 pause2 incremental) and last-restore (XFS pause2).

Cubelet/storage/s3_volume_manager.go
ls-ggg: Agreed, leaving as follow-up. No change in this commit; a cubecow message tweak would fall back to full dump rather than a wrong incremental.

fslongjin (approved):

PR #1644, opened 2026-09-04
feat(one-click): support external PostgreSQL via CUBE_DATABASE_DRIVER

xiaojunxiang2023:
## Summary
- Add `CUBE_DATABASE_DRIVER` (`mysql` | `postgres`) to one-click, mirroring Helm `database.driver` / `postgres.*`.
- When `postgres` is selected (or implied by a this-run `CUBE_EXTERNAL_POSTGRES_HOST`), require/use `CUBE_EXTERNAL_POSTGRES_*`, patch CubeMaster `instance_db_config` with **line-anchored** sed (so `common.cube_ops_addr` is not clobbered), persist a `postgresql://` `DATABASE_URL`, and skip/mask the bundled MySQL container (one-click never ships local PostgreSQL).
- Keep MySQL and PostgreSQL credential keys separate; opposite-engine markers are scrubbed on switch. If `CUBE_DATABASE_DRIVER` is omitted, a this-run external host implies the matching driver.
- Drop unreachable `restore_bundled_mysql` (PKG_ROOT is a fresh unpack every run; the packaged template is already clean).
- Compute nodes skip external DB connectivity preflight for both PostgreSQL and MySQL (control-plane DB is not used on compute).

## Test plan
- [x] `bash deploy/one-click/tests/test_external_postgres.sh` (incl. conf.yaml `cube_ops_addr` regression + intent/persist fixtures)
- [x] `bash deploy/one-click/tests/test_toggle_inputs.sh`
- [x] Docker self-test on ubuntu22 via `cube-compiler:ubuntu22`
- [ ] CI: DCO + fmt/validate/shellcheck

### DIFF
--- a/deploy/one-click/env.example
+++ b/deploy/one-click/env.example
@@ -75,18 +75,33 @@ CUBE_SANDBOX_REDIS_IMAGE=cube-sandbox-image.tencentcloudcr.com/opensource/redis:
 # CUBE_SANDBOX_MINIO_IMAGE=cube-sandbox-int.tencentcloudcr.com/cube-sandbox/minio:RELEASE.2025-09-07T16-13-09Z
 WEB_UI_IMAGE=cube-sandbox-image.tencentcloudcr.com/opensource/openresty:1.21.4.1-6-alpine-fat
 
-# ---- External MySQL / Redis ----
+# ---- External MySQL / PostgreSQL / Redis ----
 # By default install.sh starts a local MySQL and Redis container (managed by
-# systemd). To use an existing MySQL/Redis server instead, set the *_HOST
-# variables below. When set, install.sh patches CubeMaster conf.yaml, points
-# CubeAPI (DATABASE_URL) and cube-proxy at the external endpoint, masks the
-# corresponding local systemd service, and the up/down/quickcheck helpers stop
-# managing the local container. Leave unset to keep the bundled local services.
+# systemd). To use an existing server instead, set the *_HOST variables below.
+# CUBE_DATABASE_DRIVER mirrors Helm database.driver (set once):
+#   mysql    (default) -> bundled MySQL, or CUBE_EXTERNAL_MYSQL_* when HOST is set
+#   postgres           -> always external; requires CUBE_EXTERNAL_POSTGRES_HOST
+#                         (one-click never ships a local PostgreSQL)
+# Each engine keeps its own host/port/database/user/password keys. If DRIVER is
+# omitted, a non-empty CUBE_EXTERNAL_POSTGRES_HOST (or MYSQL_HOST) implies the
+# matching driver on upgrade/install. When an external SQL host is set,
+# install.sh patches CubeMaster conf.yaml (instance_db_config.driver), points
+# CubeAPI/CubeOps (DATABASE_URL) at the external endpoint, masks the local
+# MySQL systemd service, and the up/down/quickcheck helpers stop managing the
+# local container. Leave unset to keep the bundled local services.
+# CUBE_DATABASE_DRIVER=mysql
+# --- external MySQL (optional) ---
 # CUBE_EXTERNAL_MYSQL_HOST=10.0.0.20
 # CUBE_EXTERNAL_MYSQL_PORT=3306
 # CUBE_EXTERNAL_MYSQL_USER=cube
 # CUBE_EXTERNAL_MYSQL_PASSWORD=cube_pass
 # CUBE_EXTERNAL_MYSQL_DB=cube_mvp
+# --- external PostgreSQL (optional; set DRIVER=postgres or rely on HOST imply) ---
+# CUBE_EXTERNAL_POSTGRES_HOST=10.0.0.20
+# CUBE_EXTERNAL_POSTGRES_PORT=5432
+# CUBE_EXTERNAL_POSTGRES_USER=cube
+# CUBE_EXTERNAL_POSTGRES_PASSWORD=cube_pass
+# CUBE_EXTERNAL_POSTGRES_DB=cube_mvp
 # CUBE_EXTERNAL_REDIS_HOST=10.0.0.21
 # CUBE_EXTERNAL_REDIS_PORT=6379
 # CUBE_EXTERNAL_REDIS_PASSWORD=ceuhvu123
--- a/deploy/one-click/install.sh
+++ b/deploy/one-click/install.sh
@@ -47,11 +47,14 @@ ENV_FILE="${ONE_CLICK_ENV_FILE:-${SCRIPT_DIR}/.env}"
 # sourced: the upgrade merge later loads the old .one-click.env, which would
 # otherwise clobber both `VAR=x ./install.sh` and toggle keys carried in .env.
 snapshot_one_click_toggles "${ENV_FILE}"
+snapshot_one_click_database_intent "${ENV_FILE}"
 if [[ -f "${ENV_FILE}" ]]; then
   load_env_file "${ENV_FILE}"
   # CLI flags must win over .env values: load_env_file uses `set -a; source`,
   # which would otherwise clobber the CLI-provided values set above.
   apply_cli_overrides
+  # Shell-interpret .env DB values (quotes stripped) into the dotenv snapshot.
+  capture_one_click_database_dotenv_values
   case "${ONE_CLICK_MODE}" in
     ""|install|upgrade|auto) ;;
     *) die "unsupported --mode: ${ONE_CLICK_MODE} (expected install|upgrade|auto)" ;;
@@ -61,11 +64,15 @@ fi
 DEPLOY_ROLE="$(one_click_deploy_role)"
 
 # ---- External MySQL / Redis support ----
-# Set CUBE_EXTERNAL_MYSQL_HOST / CUBE_EXTERNAL_REDIS_HOST to use external
-# services instead of the bundled local Docker containers. Defaults are filled
-# after the optional upgrade env merge so they are based on the final runtime
-# configuration.
+# Set CUBE_EXTERNAL_MYSQL_HOST / CUBE_EXTERNAL_POSTGRES_HOST /
+# CUBE_EXTERNAL_REDIS_HOST to use external services instead of the bundled
+# local Docker containers. Defaults are filled after the optional upgrade env
+# merge so they are based on the final runtime configuration.
+# CUBE_DATABASE_DRIVER mirrors Helm database.driver (mysql|postgres); postgres
+# is always external (one-click never ships a local PostgreSQL).
 init_external_dep_defaults() {
+  CUBE_DATABASE_DRIVER="${CUBE_DATABASE_DRIVER:-mysql}"
+
   CUBE_EXTERNAL_MYSQL_HOST="${CUBE_EXTERNAL_MYSQL_HOST:-}"
   CUBE_EXTERNAL_MYSQL_PORT="${CUBE_EXTERNAL_MYSQL_PORT:-3306}"
   CUBE_EXTERNAL_MYSQL_USER="${CUBE_EXTERNAL_MYSQL_USER:-cube}"
@@ -75,6 +82,14 @@ init_external_dep_defaults() {
   # CUBE_SANDBOX_MYSQL_DB (without an explicit CUBE_EXTERNAL_MYSQL_DB) would make
   # the persisted .one-click.env and the seed step disagree on the database name.
   CUBE_EXTERNAL_MYSQL_DB="${CUBE_EXTERNAL_MYSQL_DB:-${CUBE_SANDBOX_MYSQL_DB:-cube_mvp}}"
+
+  # External PostgreSQL (CUBE_DATABASE_DRIVER=postgres). Separate keys from
+  # MySQL so nothing is shared or inferred across engines (same as Helm).
+  CUBE_EXTERNAL_POSTGRES_HOST="${CUBE_EXTERNAL_POSTGRES_HOST:-}"
+  CUBE_EXTERNAL_POSTGRES_PORT="${CUBE_EXTERNAL_POSTGRES_PORT:-5432}"
+  CUBE_EXTERNAL_POSTGRES_USER="${CUBE_EXTERNAL_POSTGRES_USER:-cube}"
+  CUBE_EXTERNAL_POSTGRES_PASSWORD="${CUBE_EXTERNAL_POSTGRES_PASSWORD:-cube_pass}"
+  CUBE_EXTERNAL_POSTGRES_DB="${CUBE_EXTERNAL_POSTGRES_DB:-cube_mvp}"
 
   # Mirrors the MySQL behaviour above (patch conf.yaml, persist env, mask local
   # redis unit).
@@ -114,6 +129,10 @@ warn_default_external_credentials() {
   if [[ -n "${CUBE_EXTERNAL_MYSQL_HOST}" && "${CUBE_EXTERNAL_MYSQL_PASSWORD}" == "cube_pass" ]]; then
     log "WARNING: external MySQL (${CUBE_EXTERNAL_MYSQL_HOST}) configured with the default password 'cube_pass'."
     log "WARNING: set CUBE_EXTERNAL_MYSQL_PASSWORD to a strong value in your .env before exposing this deployment."
+  fi
+  if [[ -n "${CUBE_EXTERNAL_POSTGRES_HOST}" && "${CUBE_EXTERNAL_POSTGRES_PASSWORD}" == "cube_pass" ]]; then
+    log "WARNING: external PostgreSQL (${CUBE_EXTERNAL_POSTGRES_HOST}) configured with the default password 'cube_pass'."
+    log "WARNING: set CUBE_EXTERNAL_POSTGRES_PASSWORD to a strong value in your .env before exposing this deployment."
   fi
   if [[ -n "${CUBE_EXTERNAL_REDIS_HOST}" && "${CUBE_EXTERNAL_REDIS_PASSWORD}" == "ceuhvu123" ]]; then
     log "WARNING: external Redis (${CUBE_EXTERNAL_REDIS_HOST}) configured with the default password 'ceuhvu123'."
@@ -257,8 +276,16 @@ fi
 # above preserves the old runtime values for keys whose .env value equals the
 # env.example default, which would otherwise ignore an explicit flip back.
 apply_one_click_toggles
+# Re-apply this-run DB engine intent so upgrade merge cannot keep a stale
+# opposite-engine CUBE_EXTERNAL_* marker from .one-click.env.
+apply_one_click_database_intent
 
 init_external_dep_defaults
+# Compute nodes never open the control-plane DB; skip driver/host validation
+# so a mirrored CUBE_DATABASE_DRIVER=postgres without local reachability works.
+if [[ "${DEPLOY_ROLE}" != "compute" ]]; then
+  validate_one_click_database_config
+fi
 if [[ "${DEPLOY_ROLE}" == "compute" ]]; then
   if [[ "${CUBE_SANDBOX_MINIO_ENABLED}" == "1" ]]; then
     log "compute role does not deploy MinIO; ignoring CUBE_SANDBOX_MINIO_ENABLED (volume plugin uses CUBE_S3_*)"
@@ -448,9 +475,10 @@ generate_cubemaster_config_ports() {
     "${cfg}"
 }
 
-# When external MySQL/Redis is configured, patch CubeMaster conf.yaml to replace
-# the default 127.0.0.1 endpoints with the external connection details. Must run
-# after generate_cubemaster_config_ports so the port placeholders are resolved.
+# When external MySQL/PostgreSQL/Redis is configured, patch CubeMaster conf.yaml
+# to replace the default 127.0.0.1 endpoints with the external connection
+# details. Must run after generate_cubemaster_config_ports so the port
+# placeholders are resolved.
 patch_cubemaster_external_deps() {
   [[ "${DEPLOY_ROLE}" != "compute" ]] || return 0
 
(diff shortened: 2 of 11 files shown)


deploy/one-click/install.sh
```diff
@@ -475,20 +503,55 @@ patch_cubemaster_external_deps() {
...
+  if [[ -n "${CUBE_EXTERNAL_POSTGRES_HOST}" ]]; then
+    log "patching conf.yaml for external PostgreSQL: ${CUBE_EXTERNAL_POSTGRES_HOST}:${CUBE_EXTERNAL_POSTGRES_PORT}/${CUBE_EXTERNAL_POSTGRES_DB}"
+    local pg_addr_esc pg_user_esc pg_pwd_esc pg_db_esc
+    pg_addr_esc="$(escape_sed "${CUBE_EXTERNAL_POSTGRES_HOST}:${CUBE_EXTERNAL_POSTGRES_PORT}")"
+    pg_user_esc="$(escape_sed "${CUBE_EXTERNAL_POSTGRES_USER}")"
+    pg_pwd_esc="$(escape_sed "${CUBE_EXTERNAL_POSTGRES_PASSWORD}")"
+    pg_db_esc="$(escape_sed "${CUBE_EXTERNAL_POSTGRES_DB}")"
+    sed -i \
+      -e "s|driver: \".*\"|driver: \"postgres\"|" \
+      -e "s|addr: \".*\"|addr: \"${pg_addr_esc}\"|" \
```
fslongjin: **[Blocker] This `addr:` substitution is unanchored and clobbers `common.cube_ops_addr`.**

`configs/single-node/cubemaster.yaml:10` has `cube_ops_addr: "__CUBEMASTER_CUBE_OPS_ADDR__"`. After `generate_cubemaster_config_ports` (install.sh:474) expands the placeholder to `http://127.0.0.1:3010`, that line **contains the substring `addr: "..."`** and is matched by this rule. Reproduced locally:

```yaml
common:
  cube_ops_addr: "10.0.0.20:5432"      # was "http://127.0.0.1:3010"
instance_db_config:
  addr: "10.0.0.20:5432"
```

`cube_ops_addr` is the base URL CubeMaster uses to call CubeOps for node management (list / isolate / unisolate), consumed at `CubeMaster/pkg/nodemeta/service.go:44`. After the rewrite the value has no scheme and points at the database port, so node metadata sync fails or silently falls back to the local DB loader.

**Any external SQL configuration hits this on every install/upgrade.** The MySQL branch (L572) is the pre-existing path; the new PostgreSQL branch (L549) and the new `restore_bundled_mysql` branch (L590) inherit the same problem.

Please anchor all five patterns and preserve indentation via a back-reference:

```bash
sed -i \
  -e "s|^\([[:space:]]*\)driver: \".*\"|\1driver: \"postgres\"|" \
  -e "s|^\([[:space:]]*\)addr: \".*\"|\1addr: \"${pg_addr_esc}\"|" \
  -e "s|^\([[:space:]]*\)user: \".*\"|\1user: \"${pg_user_esc}\"|" \
  -e "s|^\([[:space:]]*\)pwd: \".*\"|\1pwd: \"${pg_pwd_esc}\"|" \
  -e "s|^\([[:space:]]*\)db_name: \".*\"|\1db_name: \"${pg_db_esc}\"|" \
  "${cfg}"
```

With `^[[:space:]]*` immediately followed by `addr:`, a line like `  cube_ops_addr:` no longer matches, and `\1` keeps the indentation. I verified this locally: `cube_ops_addr` is left untouched and the five keys are rewritten correctly. Please apply the same change to L570-576 and L588-594.

deploy/one-click/install.sh
```diff
@@ -505,6 +568,25 @@ patch_cubemaster_external_deps() {
     # a trailing 'g' flag each line is patched exactly once and Redis fields
```
fslongjin: The premise of this comment (which starts at L564) is wrong, and it is the reason the blocker above has survived this long:

> These keys only appear in the MySQL section (instance_db_config), so without
> a trailing 'g' flag each line is patched exactly once

`common.cube_ops_addr` also contains the substring `addr: "`, so "each line is patched exactly once" does not imply "only instance_db_config is patched". Please restate this as a requirement to **anchor at line start**, otherwise the next person will keep writing unanchored patterns based on this comment.

deploy/one-click/install.sh
```diff
@@ -475,20 +503,55 @@ patch_cubemaster_external_deps() {
     fi
   fi
 
+  # External SQL → bundled MySQL: restore driver/addr when leaving postgres or
+  # an external MySQL host so CubeMaster does not keep the previous engine.
+  local restore_bundled_mysql=0
```
fslongjin: **[P1] This `restore_bundled_mysql` branch is unreachable.**

The predicate reads `${PKG_ROOT}/CubeMaster/conf.yaml`, and `PKG_ROOT` is freshly unpacked package content on every run:

- `install.sh:227` - `WORK_DIR="$(mktemp -d)"`
- `install.sh:1678` - `tar -xzf "${PACKAGE_TAR}" -C "${WORK_DIR}"`
- `install.sh:1762-1806` - on upgrade, `rm -rf "${INSTALL_PREFIX}/CubeMaster"` then `cp -a "${PKG_ROOT}/." "${INSTALL_PREFIX}/"`

So conf.yaml is always the pristine template: `driver` is always `"mysql"` and `addr` is always `127.0.0.1:${CUBE_SANDBOX_MYSQL_PORT}` (the placeholder is expanded by `generate_cubemaster_config_ports` from the same variable this predicate reads, so they always agree). Neither condition can ever fire.

I also checked: on upgrade `backup_before_upgrade` (`lib/common.sh:2091`) does back up `CubeMaster/conf.yaml`, but nothing in install.sh restores it into `PKG_ROOT`.

Pick one:

1. Delete this predicate and the branch at L577-595 - "the packaged template is always clean" already guarantees correctness;
2. If the intent is "preserve operator-edited conf.yaml across upgrades", add a restore mechanism for conf.yaml first (see `restore_volume_plugin_config_from_upgrade_backup` for a precedent).

Either way, please say which one in the commit message so the next reader does not have to guess. (`restore_bundled_redis` / `scrub_stale_sentinel` have the same unreachability problem, but that is pre-existing - worth a separate issue.)

deploy/one-click/lib/common.sh
```diff
@@ -566,6 +566,166 @@ apply_one_click_toggles() {
   return 0
 }
 
+# Database engine keys are commented out of env.example, so merge_env_three_way
```
fslongjin, line 570: **[P1 / abstraction] This snapshot+replay duplicates the existing `ONE_CLICK_TOGGLE_KEYS` mechanism by roughly 90%.**

`snapshot_one_click_database_intent` / `capture_one_click_database_dotenv_values` / `apply_one_click_database_intent` (L569-727, ~160 lines) are structurally almost identical to `snapshot_one_click_toggles` / `apply_one_click_toggles` (L502-567). The real differences are only three:

1. Two-phase value capture, avoiding the `read_env_key` quoting problem;
2. An extra "no DB intent this run -> no-op entirely" gate;
3. Driver inference + opposite-engine scrubbing.

(1) is actually the **more correct** implementation (see the related note at the end), which suggests the duplication comes from a defect in the existing toggle implementation rather than from a genuinely new requirement.

Suggestion: generalize the `ONE_CLICK_TOGGLE_KEYS` snapshot/replay (parameterize by key list + snapshot name, or simply let toggles use the two-phase capture too), and keep only (2) and (3) as DB-specific post-hooks. That removes ~80 lines of duplication and fixes the toggle quoting bug along the way.

If that feels too invasive, at least add a cross-reference comment on both sides: "this is the same mechanism as apply_one_click_toggles - change both together."

Related, can be a follow-up: the quoting bug you fixed with the two-phase capture still exists on the `ONE_CLICK_TOGGLE_KEYS` path. `snapshot_one_click_toggles` (L543) reads values with `read_env_key`, which returns the raw RHS with quotes intact, while `load_env_file` `source`s the file and strips them. A value such as `KEY='p@ss word#'` is therefore still corrupted there. It has not blown up only because `ONE_CLICK_TOGGLE_KEYS` currently holds just two boolean keys (`ONE_CLICK_ENABLE_S3LVOL`, `CUBE_PVM_ENABLE`), but the two-phase capture is the fix whenever that changes.

deploy/one-click/lib/common.sh
```diff
@@ -975,6 +1135,135 @@ _remove_env_keys() {
   done
 }
 
+# one_click_skip_local_mysql is true when the bundled MySQL container must not
+# start: external MySQL, or database.driver=postgres (always external).
+one_click_skip_local_mysql() {
```
fslongjin, line 1169: **[P1] This predicate is hand-copied in three places, in three different shapes.**

- `lib/common.sh:1140` - `one_click_skip_local_mysql()` (this helper)
- `install.sh:1580` - calls the helper
- `scripts/one-click/quickcheck.sh:345-349` - copy, variable `SKIP_LOCAL_MYSQL`
- `scripts/one-click/up-support.sh:115-119` - copy, variable `skip_local_mysql`
- `scripts/systemd/mysql-postcheck.sh:8-9` - copy, inlined in an `if`

I checked: `scripts/one-click/common.sh` and `scripts/systemd/common.sh` are independently maintained runtime-trimmed copies that do not contain this helper, so those three scripts genuinely cannot call it. The duplication is caused by the existing repo layout, not by cutting corners here.

The risk is that one three-way condition (external mysql / external pg / driver=postgres) now exists in three different forms; adding another engine or changing the rule will miss one. Suggestion: add the same helper to both runtime common.sh files. Failing that, add a consistency test in `test_external_postgres.sh` asserting the condition text matches across call sites - better than testing nothing.

deploy/one-click/install.sh
```diff
@@ -448,9 +475,10 @@ generate_cubemaster_config_ports() {
...
 }
 
-# When external MySQL/Redis is configured, patch CubeMaster conf.yaml to replace
-# the default 127.0.0.1 endpoints with the external connection details. Must run
-# after generate_cubemaster_config_ports so the port placeholders are resolved.
+# When external MySQL/PostgreSQL/Redis is configured, patch CubeMaster conf.yaml
+# to replace the default 127.0.0.1 endpoints with the external connection
+# details. Must run after generate_cubemaster_config_ports so the port
+# placeholders are resolved.
 patch_cubemaster_external_deps() {
```
fslongjin, line 482: **[P2 / test gap] This function has zero test coverage, and it is the most dangerous one in the PR.**

`patch_cubemaster_external_deps` rewrites `CubeMaster/conf.yaml` with sed, yet `test_external_postgres.sh` does not cover it at all - every test targets the "pure" functions (`persist_*`, `validate_*`, intent replay) and none asserts what conf.yaml actually looks like afterwards. That is exactly how the `cube_ops_addr` defect I flagged got through.

Please add a fixture test; `tests/test_cubeops_addr.sh` is a good template:

```bash
test_patch_conf_postgres() {
  # build a conf.yaml from configs/single-node/cubemaster.yaml with placeholders expanded
  # call patch_cubemaster_external_deps (driver=postgres + host)
  # assert:
  #   instance_db_config.driver == postgres
  #   instance_db_config.addr   == host:port
  #   common.cube_ops_addr      == its original value (regression for the clobbering bug)
}
```

Plus a postgres -> built-in restore case. With those two, the clobbering bug could not have been merged.

deploy/one-click/install.sh
```diff
@@ -580,7 +662,40 @@ patch_cubemaster_external_deps() {
...
+          -t "${connect_timeout}" >/dev/null 2>&1; then
+        die "cannot reach external PostgreSQL at ${CUBE_EXTERNAL_POSTGRES_HOST}:${CUBE_EXTERNAL_POSTGRES_PORT} as user '${CUBE_EXTERNAL_POSTGRES_USER}'.
+  Verify CUBE_EXTERNAL_POSTGRES_HOST / _PORT / _USER / _PASSWORD / _DB and that the server is reachable from this host."
+      fi
+      log "external PostgreSQL server reachable (credentials not verified; install psql for a full check)"
+    else
+      log "psql/pg_isready not found; skipping external PostgreSQL connectivity preflight"
+    fi
+  fi
+
```
fslongjin, line 649: **[P2] This adds the compute exemption to the MySQL branch too, which changes existing behavior.**

The pre-existing code was `if [[ -n "${CUBE_EXTERNAL_MYSQL_HOST}" ]]; then`. Doing it for the PostgreSQL branch (L665) is fine since that is new, but for MySQL it is a behavior change that the PR description does not mention.

I agree with the direction (compute nodes do not talk to the control-plane DB, so they should not require a reachable MySQL client). Just please note it in the description or split it out, to keep bisecting sane later.

deploy/one-click/lib/common.sh
```diff
@@ -2169,6 +2458,13 @@ install_s3_volume_host_deps() {
   local deploy_role="$2"
   local script
 
+  # No S3 backend configured (MinIO off and CUBE_S3_ENDPOINT empty): skip
```
fslongjin: **[P2] This change is unrelated to PostgreSQL support; please split it out.**

The comment says "MinIO off **and** CUBE_S3_ENDPOINT empty" but the code only checks `CUBE_S3_ENDPOINT` and never looks at `CUBE_SANDBOX_MINIO_ENABLED` - please align the comment with the code.

The logic itself checks out: `fill_s3_from_local_minio` (install.sh:1631) runs before `prepare_volume_plugin_install` (install.sh:1811), so when MinIO is enabled `CUBE_S3_ENDPOINT` is already populated and this will not skip by mistake.

deploy/one-click/env.example
```diff
@@ -75,18 +75,31 @@ CUBE_SANDBOX_REDIS_IMAGE=cube-sandbox-image.tencentcloudcr.com/opensource/redis:
...
+#   mysql    (default) -> bundled MySQL, or CUBE_EXTERNAL_MYSQL_* when HOST is set
+#   postgres           -> always external; requires CUBE_EXTERNAL_POSTGRES_HOST
+#                         (one-click never ships a local PostgreSQL)
+# Each engine keeps its own host/port/database/user/password — nothing is shared
+# or inferred across sections. When an external SQL host is set, install.sh
+# patches CubeMaster conf.yaml (including instance_db_config.driver), points
+# CubeAPI/CubeOps (DATABASE_URL) at the external endpoint, masks the local
+# MySQL systemd service, and the up/down/quickcheck helpers stop managing the
+# local container. Leave unset to keep the bundled local services.
+# CUBE_DATABASE_DRIVER=mysql
```
fslongjin, line 92: **[P3] `CUBE_DATABASE_DRIVER` appears twice here (L91 and L97).**

If a user uncomments both the MySQL and the PostgreSQL block, `source .env` makes the latter win, which is easy to trip over. Suggest keeping a single `CUBE_DATABASE_DRIVER` entry near the top of the section and leaving only the engine-specific `CUBE_EXTERNAL_*` lines inside each block.

deploy/one-click/env.example
```diff
@@ -75,18 +75,31 @@ CUBE_SANDBOX_REDIS_IMAGE=cube-sandbox-image.tencentcloudcr.com/opensource/redis:
...
-# variables below. When set, install.sh patches CubeMaster conf.yaml, points
-# CubeAPI (DATABASE_URL) and cube-proxy at the external endpoint, masks the
-# corresponding local systemd service, and the up/down/quickcheck helpers stop
-# managing the local container. Leave unset to keep the bundled local services.
+# systemd). To use an existing server instead, set the *_HOST variables below.
+# CUBE_DATABASE_DRIVER mirrors Helm database.driver:
+#   mysql    (default) -> bundled MySQL, or CUBE_EXTERNAL_MYSQL_* when HOST is set
+#   postgres           -> always external; requires CUBE_EXTERNAL_POSTGRES_HOST
+#                         (one-click never ships a local PostgreSQL)
+# Each engine keeps its own host/port/database/user/password — nothing is shared
```
fslongjin: **[P3] The comment says nothing is inferred across engines, but the code does infer.**

> nothing is shared or inferred across sections

`common.sh:676-697` implements "host implies driver": declaring `CUBE_EXTERNAL_POSTGRES_HOST` infers `driver=postgres`.

That behavior is **good** (one less key for the user to fill in), so the docs should describe the convenience instead of denying it.

deploy/one-click/tests/test_external_postgres.sh
```diff
@@ -0,0 +1,402 @@
...
+    unset CUBE_EXTERNAL_MYSQL_HOST CUBE_EXTERNAL_POSTGRES_HOST CUBE_DATABASE_DRIVER
+    # shellcheck disable=SC1091
+    source "${ONE_CLICK_DIR}/lib/common.sh"
+    one_click_skip_local_mysql && fail "default must not skip"
+    export CUBE_EXTERNAL_POSTGRES_HOST=pg.example.com
+    one_click_skip_local_mysql || fail "postgres host must skip"
+  ) || fail "skip helper checks failed"
+}
+
+test_contract_wiring() {
```
fslongjin, line 182: **[P3] `test_contract_wiring` is a source-grepping test anti-pattern.**

It asserts on the source text of `install.sh` / `up-support.sh` / `quickcheck.sh`, and even has `grep -A2 'Prefer psql' | grep -q 'command -v psql'`. These assertions test "does this string exist in the source", not behavior:

- Reword the "Prefer psql" comment and the test goes red;
- Extract the sed into a helper or change the quoting style and it goes red;
- Conversely, change `s|addr: ".*"|` to `s|xaddr: ".*"|` (logic now broken) and it stays green.

Suggest dropping it and converting the valuable bits into behavior tests - `driver: "postgres"` is covered by asserting the actual conf.yaml output, and the psql precedence can be asserted by putting a fake `psql` / `pg_isready` on PATH.

deploy/one-click/lib/common.sh
```diff
@@ -566,6 +566,166 @@ apply_one_click_toggles() {
...
+    if [[ -n "${CUBE_EXTERNAL_POSTGRES_HOST:-}" ]]; then
+      CUBE_DATABASE_DRIVER="postgres"
+    else
+      CUBE_DATABASE_DRIVER="mysql"
+    fi
+  fi
+
+  # Fail fast on this-run contradictions before scrubbing so validate_*'s
+  # mutually-exclusive / wrong-driver die paths remain reachable for .env input.
+  if _one_click_db_intent_declared CUBE_EXTERNAL_MYSQL_HOST       && _one_click_db_intent_declared CUBE_EXTERNAL_POSTGRES_HOST       && [[ -n "${CUBE_EXTERNAL_MYSQL_HOST:-}" && -n "${CUBE_EXTERNAL_POSTGRES_HOST:-}" ]]; then
```
fslongjin, line 703: **[P3] These three `die` messages are duplicated verbatim in `validate_one_click_database_config` (L1148-1178)** (mutually exclusive hosts / postgres cannot combine with a MySQL host / PG host requires driver=postgres).

Validation rules should have a single source of truth. Suggest extracting one shared function both call sites use, otherwise changing a message or a rule will inevitably miss one of them.

deploy/one-click/install.sh
```diff
@@ -580,7 +662,40 @@ patch_cubemaster_external_deps() {
...
+          -p "${CUBE_EXTERNAL_POSTGRES_PORT}" \
+          -U "${CUBE_EXTERNAL_POSTGRES_USER}" \
+          -d "${CUBE_EXTERNAL_POSTGRES_DB}" \
+          -t "${connect_timeout}" >/dev/null 2>&1; then
+        die "cannot reach external PostgreSQL at ${CUBE_EXTERNAL_POSTGRES_HOST}:${CUBE_EXTERNAL_POSTGRES_PORT} as user '${CUBE_EXTERNAL_POSTGRES_USER}'.
+  Verify CUBE_EXTERNAL_POSTGRES_HOST / _PORT / _USER / _PASSWORD / _DB and that the server is reachable from this host."
+      fi
+      log "external PostgreSQL server reachable (credentials not verified; install psql for a full check)"
+    else
+      log "psql/pg_isready not found; skipping external PostgreSQL connectivity preflight"
```
fslongjin: **[P3] When neither `psql` nor `pg_isready` exists the check is skipped silently; suggest WARNING level.**

This only emits an info line, but skipping means a wrong password surfaces only when CubeMaster/CubeAPI start, which is expensive to diagnose. Please promote it to a warning.

fslongjin (changes requested):
## Review verdict: Request changes

The abstraction is right and the 7 issues from the previous bot review are all
correctly fixed (see the table below), but I found one blocker that has to be
addressed before this can go in.

### Overall

Mirroring Helm's `database.driver` / `postgres.*` is the right call. I verified the
PostgreSQL backend is genuinely wired end to end: `CubeDB/dao/driver/postgres`
exists, `DBConfig.Driver` -> `db.ConfigFromDBConfig` -> `dao.Config.Driver` is
complete, PG and MySQL each have 28 migrations, and `CubeOps.parseDatabaseURL`
already handles `postgresql://`. The packaging path is closed too:
`build-release-bundle.sh:786` copies `configs/single-node/cubemaster.yaml` over
`CubeMaster/conf.yaml`, and `install.sh:531-533` back-fills `driver:` for older
packages. "Declaring just the HOST implies the driver" is a nice usability touch.

### Previous round: 7/7 fixed

| Issue raised previously | Status |
|---|---|
| `pg_isready` took precedence, bad credentials only surfaced at runtime | OK - `psql -c 'SELECT 1'` first, log says "credentials not verified" |
| Upgrade merge re-injected old engine markers, making switches inexpressible | OK - intent snapshot + replay after merge |
| Both engines were cleared, undeclared credentials reset to defaults | OK - only the opposite engine is scrubbed |
| Snapshot read a single channel, dropping CLI overrides | OK - per-key: dotenv wins, process env fills gaps |
| Empty host produced `postgresql://cube:cube_pass@:5432/...` | OK - all markers + `DATABASE_URL` removed when host is empty |
| `read_env_key` returned raw RHS, corrupting quoted values | OK - two-phase capture, covered by a dedicated test |
| A stale driver marker scrubbed the newly declared other-engine host | OK - "host implies driver" supported |

### Blocker: unanchored `addr:` sed clobbers `cube_ops_addr`

The substitution `s|addr: ".*"|` in `patch_cubemaster_external_deps` has no line
anchor. `configs/single-node/cubemaster.yaml:10` contains
`cube_ops_addr: "__CUBEMASTER_CUBE_OPS_ADDR__"`, and once the placeholder is
expanded to `http://127.0.0.1:3010` that line **contains `addr: "`** and gets hit by
the same rule. Reproduced locally:

```yaml
common:
  cube_ops_addr: "10.0.0.20:5432"      # was "http://127.0.0.1:3010"
instance_db_config:
  driver: "postgres"
  addr: "10.0.0.20:5432"
```

Any external SQL setup hits this on every install/upgrade. The MySQL branch is the
pre-existing path; the new PostgreSQL branch and the new `restore_bundled_mysql`
branch inherit the same problem. `cube_ops_addr` is the base URL CubeMaster uses to
call CubeOps for node management (list / isolate / unisolate), consumed at
`CubeMaster/pkg/nodemeta/service.go:44`; after the rewrite it has no scheme and
points at the database port, so node metadata sync fails or silently degrades.

Root cause of it surviving this long: the comment at `install.sh:564-569` wrongly
claims "these keys only appear in the MySQL section (instance_db_config)" —
`cube_ops_addr` is the counterexample. See the inline comment for the fix.

### Worth fixing in the same PR

- **P1** The `restore_bundled_mysql` branch is unreachable (`PKG_ROOT` is a fresh
  `mktemp -d` + unpack on every run, so conf.yaml is always pristine). Either drop it
  or add an upgrade-restore mechanism for conf.yaml.
- **P1** The DB intent mechanism duplicates `apply_one_click_toggles` by ~90%; the
  "skip local MySQL" predicate is hand-copied in 3 places with 3 different shapes.
- **P2** Unrelated `install_s3_volume_host_deps` change is bundled in (its comment
  mentions MinIO but the code never checks it).
- **P2** `patch_cubemaster_external_deps` has zero test coverage - it is the most
  dangerous function in this PR, and that is exactly how the blocker slipped
  through. Please add a fixture test.
- **P2** Skipping the external MySQL preflight on compute nodes is a good change, but
  it is a behavior change; please note it in the description.

### Verification

- All 17 `deploy/one-click/tests/test_*.sh` scripts pass (CI picks up new tests by
  glob).
- CI: DCO / fmt+validate+shellcheck (amd64 & arm64) all green.
- Escaping, URL encoding, special-character passwords and quoted `.env` values are
  covered by tests.

Once the blocker is fixed and the conf.yaml fixture test is added, I have no other
blocking concerns.

xiaojunxiang2023:
@fslongjin Thanks for the thorough review — addressed on 228aeaff.

**Blocker** — `patch_cubemaster_instance_db_config` now uses line-anchored `^[[:space:]]*addr:` (and the same for driver/user/pwd/db_name), so `common.cube_ops_addr` is not clobbered. Added a conf.yaml fixture test that asserts `cube_ops_addr` stays unchanged after a PostgreSQL patch.

**P1**
- Dropped unreachable `restore_bundled_mysql` (PKG_ROOT is a fresh unpack; the packaged template is already clean).
- DB intent vs toggles: left as a parallel snapshot/replay with a cross-reference comment; unifying them is a follow-up (shared helper would touch more than this PR).
- Skip-local-mysql: kept `one_click_skip_local_mysql` as the install-time helper and added a consistency test against the runtime predicates; extracting a shared runtime helper is also follow-up.

**P2**
- conf.yaml fixture coverage added (see above).
- Compute skipping MySQL preflight: intentional symmetry with PostgreSQL (control-plane DB unused on compute); noted in the PR description.
- Unrelated `install_s3_volume_host_deps` early-return: removed from this PR.

**P3** — env.example (single DRIVER), docs (host-implies-driver), WARNING when neither psql nor pg_isready is present: fixed in the same commit.

deploy/one-click/lib/common.sh
```diff
@@ -975,6 +1138,150 @@ _remove_env_keys() {
...
+}
+
+# one_click_skip_local_mysql is true when an external DB endpoint is configured.
+# Key off host presence (not bare CUBE_DATABASE_DRIVER): driver=postgres with an
+# empty host must not skip bundled MySQL if this env is later reused on control.
+one_click_skip_local_mysql() {
+  [[ -n "${CUBE_EXTERNAL_MYSQL_HOST:-}" || -n "${CUBE_EXTERNAL_POSTGRES_HOST:-}" ]]
+}
+
+# validate_one_click_database_config mirrors Helm database.driver / postgres.*:
```
fslongjin, line 1173: **[nit] Dropping the bare-DRIVER condition is the right call** - it removes the three-way copy-paste drift I flagged, and keying off host presence means `driver=postgres` with an empty host can no longer skip bundled MySQL.

I checked the role safety: `cube-sandbox-compute.target` does not `Wants=` `cube-sandbox-mysql.service` (only `control.target` does), so returning false on a compute node and hitting the `unmask` branch cannot pull MySQL up there. No regression.

deploy/one-click/env.example
```diff
@@ -75,18 +75,33 @@ CUBE_SANDBOX_REDIS_IMAGE=cube-sandbox-image.tencentcloudcr.com/opensource/redis:
...
-# ---- External MySQL / Redis ----
+# ---- External MySQL / PostgreSQL / Redis ----
 # By default install.sh starts a local MySQL and Redis container (managed by
-# systemd). To use an existing MySQL/Redis server instead, set the *_HOST
-# variables below. When set, install.sh patches CubeMaster conf.yaml, points
-# CubeAPI (DATABASE_URL) and cube-proxy at the external endpoint, masks the
-# corresponding local systemd service, and the up/down/quickcheck helpers stop
-# managing the local container. Leave unset to keep the bundled local services.
+# systemd). To use an existing server instead, set the *_HOST variables below.
+# CUBE_DATABASE_DRIVER mirrors Helm database.driver (set once):
```
fslongjin, line 81: **[nit] Now that host implies driver, `CUBE_DATABASE_DRIVER` is effectively optional.**

After this round, every decision point keys off host presence: `one_click_skip_local_mysql`, the conf.yaml patch branches, and the up/quickcheck helpers. `CUBE_DATABASE_DRIVER` is left with validation (legal values, mutual exclusion, postgres requires a host) and persistence. Combined with host-implies-driver, a user can configure external PostgreSQL without ever setting it.

That is a nice reduction in required knobs, but this file still presents it as the first thing in the section and the PR title leads with it. Suggest stating its role explicitly, e.g. "optional type check - omit it and the engine is implied by whichever `CUBE_EXTERNAL_*_HOST` you set". Same note for the README section.

deploy/one-click/tests/test_external_postgres.sh
```diff
@@ -0,0 +1,479 @@
...
+  local up_support="${ONE_CLICK_DIR}/scripts/one-click/up-support.sh"
+  local env_example="${ONE_CLICK_DIR}/env.example"
+  local cfg="${ONE_CLICK_DIR}/../../configs/single-node/cubemaster.yaml"
+  assert_contains "${env_example}" "CUBE_DATABASE_DRIVER=mysql"
+  assert_contains "${env_example}" "CUBE_EXTERNAL_POSTGRES_HOST"
+  # Single DRIVER knob (not duplicated inside each engine block).
+  local driver_lines
+  driver_lines="$(grep -cE '^#?[[:space:]]*CUBE_DATABASE_DRIVER=' "${env_example}" || true)"
+  [[ "${driver_lines}" -eq 1 ]] || fail "env.example should declare CUBE_DATABASE_DRIVER once (got ${driver_lines})"
+  assert_contains "${install_sh}" "CUBE_EXTERNAL_POSTGRES_HOST"
```
fslongjin, line 193: **[nit] These source-grepping assertions can go now that the fixture tests exist.**

Good call dropping `grep -A2 'Prefer psql' | grep -q 'command -v psql'` - that was the most fragile one. The remaining `assert_contains` calls still assert that specific identifiers (`patch_cubemaster_instance_db_config`, `apply_one_click_database_intent`, `PGCONNECT_TIMEOUT`, ...) appear in the source, which tests "does this string exist" rather than behavior: renaming a helper or extracting a function turns them red while everything still works, and breaking the sed logic does not turn them red.

`test_patch_conf_postgres_preserves_cube_ops_addr` / `_mysql_` already cover the behavior, so I would delete these five and keep only the `driver_lines -eq 1` check below, which does have real meaning (env.example structure).

fslongjin (approved):
## Re-review: approving

The blocker and every P1/P2 from my first review are addressed. Thanks for the quick
turnaround - the fix is exactly what I suggested and you went further by also routing
the MySQL branch through the same helper.

### Verified

- **Blocker fixed.** `patch_cubemaster_instance_db_config` anchors all five sed
  patterns at line start, so `common.cube_ops_addr` survives. I re-ran my original
  reproduction against both engines:
  - external PostgreSQL: `cube_ops_addr` stays `http://127.0.0.1:3010`
  - external MySQL with a `|` in the password: stays `http://127.0.0.1:3010`,
    `pwd: "p@ss|word"` escaped correctly
- **`restore_bundled_mysql` gone**, replaced by a comment explaining PKG_ROOT is a
  fresh unpack every run. Correct call.
- **Skip-local-MySQL predicate unified** to "host present", with
  `test_skip_local_mysql_predicate_consistency` guarding it. I checked the role
  safety: `cube-sandbox-compute.target` does not `Wants=` the mysql unit (only
  `control.target` does), so the unmask on compute cannot pull MySQL up. No regression.
- Unrelated `install_s3_volume_host_deps` change reverted; env.example driver key
  de-duplicated; psql-missing log promoted to WARNING; docs now describe host-implies-
  driver instead of denying it.
- All 17 `deploy/one-click/tests/test_*.sh` pass; CI green.

### Non-blocking suggestions (inline)

1. `CUBE_DATABASE_DRIVER` is now effectively only a validation knob (everything else
   keys off host presence, and host implies driver). Worth stating that explicitly in
   env.example / README so users know it is optional.
2. A few source-grepping assertions remain in `test_contract_wiring`; now that the
   fixture tests exist, they add little and will fail on harmless refactors.

Nice work - approving.

PR #1649, opened 2026-09-04
feat(sdk): add E2B-compatible Sandbox.fork to python/node/go

### LINKED ISSUE #1648: [Bug Report] E2B Sandbox.fork logic cannot be migrated to the CubeSandbox SDK without a rewrite
## Summary

When using the CubeSandbox SDK, there is a compatibility issue with Sandbox.fork. The clone interface provided by the CubeSandbox SDK behaves differently from fork, so it cannot be directly substituted. The fork logic needs to be rewritten and a fork interface should be provided.

## Environment

- CubeSandbox version / commit:0.7.0
- Relevant component: SDK（python/go/node）

## Steps to Reproduce

# E2B 
from e2b import Sandbox
sb = Sandbox.create()
results = sb.fork(count=5)          # always returns a list, one entry per fork
forks  = [r for r in results if isinstance(r, Sandbox)]
errors = [r for r in results if not isinstance(r, Sandbox)]  # per-fork failures
1. Create a sandbox in CubeSandbox.
2. Try to fork it with `count` as you would in E2B.
3. Attempt to handle per-fork partial success/failure the E2B way.
## Expected Behavior

An E2B-compatible `Sandbox.fork` should be provided on the CubeSandbox SDK(python/node/go)

## Actual Behavior

```
paste logs here
```

liciazhu:
fixes #1648 
Sandbox.fork now maps to a CubeAPI /fork endpoint (new), which forwards to a new CubeMaster fork action — so a single call derives N running copies from one snapshot.
  - CubeMaster: new POST /cube/sandbox/:id/fork (snapshot once, derive count  copies with bounded concurrency, per-fork independent results).
  - CubeAPI: thin-proxy POST /sandboxes/:id/fork (count+timeout).
  - SDK (python/node/go): Sandbox.fork calls /fork and parses [{sandbox}|{error}];                                                                    - tests(e2e): add sdk_compat fork coverage


liciazhu:
e2e tests passed.

PR #1651, opened 2026-09-04
fix(one-click): skip s3fs/jq host install when CUBE_S3_ENDPOINT is unset

xiaojunxiang2023:
## Summary

- `install_s3_volume_host_deps` now returns early when `CUBE_S3_ENDPOINT` is empty, so offline/private hosts are not forced through `dnf`/`yum` for jq/s3fs.
- When bundled MinIO is enabled, `fill_s3_from_local_minio` still runs earlier and sets the endpoint, so those installs keep installing host deps as before.
- Added unit coverage in `test_minio_s3_guard.sh` for the skip and the still-run paths.

## Test plan

- [x] `bash deploy/one-click/tests/test_minio_s3_guard.sh` — passed (ubuntu22 host)
- [x] Same suite in `cube-compiler:ubuntu22` Docker — passed
- [ ] CI: DCO + one-click deployer tests / fmt+validate+shellcheck

### DIFF
--- a/deploy/one-click/lib/common.sh
+++ b/deploy/one-click/lib/common.sh
@@ -2476,6 +2476,14 @@ install_s3_volume_host_deps() {
   local deploy_role="$2"
   local script
 
+  # No S3 backend configured (CUBE_S3_ENDPOINT empty; MinIO path fills this
+  # earlier when enabled): skip jq/s3fs package installs so offline/private
+  # hosts are not forced onto dnf/yum.
+  if [[ -z "${CUBE_S3_ENDPOINT:-}" ]]; then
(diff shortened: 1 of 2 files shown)


fslongjin:
Thanks for the change. I have a concern about the direction though: quite a few upcoming Cube features will depend on S3 — including the template center service that's about to be merged — so S3 is effectively becoming a required component rather than an optional one. Given that, I don't think we should add this special-case early return here. I'd lean toward treating S3 as mandatory in the environment (i.e. `CUBE_S3_ENDPOINT` should always be set), which keeps the install path simpler and avoids silently producing hosts that later turn out to be unable to mount S3 volumes. Curious to hear what others think before we lock in the optional behavior.

PR #1652, opened 2026-09-04
feat(deploy): opt-in softlink cubebox_os_image onto the data disk

xiaojunxiang2023:
## Summary

- Keep the historical `<toolbox>/cubebox_os_image` path by default for forward compatibility with existing installs.
- Opt in to softlink `<toolbox>/cubebox_os_image` → `CUBEBOX_OS_IMAGE_DATA_DIR` (default `/data/cubebox_os_image`) via `CUBE_CUBEBOX_OS_IMAGE_ON_DATA=1` / `hostPaths.cubeboxOsImageOnData=true` so large template rootfs can stay on the data disk without changing cubelet's toolbox path.
- One-click: default `CUBE_CUBEBOX_OS_IMAGE_ON_DATA=0`; persisted to `.one-click.env`; wipe/assert treat the managed symlink as allowed when enabled.
- Helm chart: `hostPaths.cubeboxOsImageOnData: false` by default. The `data-cubebox-os-image` hostPath volume remains mounted either way (`DirectoryOrCreate`) so a leftover symlink cannot resolve into the image overlay.
- Empty destination dirs (e.g. `DirectoryOrCreate`) use `mv` for first migration instead of recursive `cp`.

## Test plan

- [x] `bash deploy/one-click/tests/test_install_mode.sh` — passed (ubuntu22 + cube-compiler container)
- [x] `bash deploy/one-click/tests/test_toggle_inputs.sh` — passed
- [x] `bash deploy/kubernetes/images/scripts/test-cubebox-os-image.sh` — passed (ubuntu22 + cube-compiler container)
- [x] `helm template` with default — renders `CUBE_CUBEBOX_OS_IMAGE_ON_DATA=0`
- [x] `helm template --set hostPaths.cubeboxOsImageOnData=true` — renders `CUBE_CUBEBOX_OS_IMAGE_ON_DATA=1`


fslongjin:
Thanks for your contribution! I think the feature in this PR that allows setting the cubebox OS image path via an environment variable is quite nice. However, from the standpoint of forward compatibility for existing users, I'd prefer to keep the default path unchanged and simply expose an environment variable here to let users customize the path if needed.

xiaojunxiang2023:
@fslongjin Thanks for the review!

Agreed on forward compatibility. The softlink is now opt-in and the historical toolbox path stays the default:

- one-click: `CUBE_CUBEBOX_OS_IMAGE_ON_DATA=0` by default; set `1` (+ optional `CUBEBOX_OS_IMAGE_DATA_DIR`) to redirect onto the data disk
- Helm: `hostPaths.cubeboxOsImageOnData: false` by default; set `true` to enable

Force-pushed.

PR #1655, opened 2026-09-04
fix(cube-egress): raise transparent proxy timeouts to 2h

FakeLearne:
LLM prefill TTFB often exceeds the baked-in 60s proxy_read_timeout and gets a 504. Default both read/send to 2h.

### DIFF
--- a/CubeEgress/nginx.conf
+++ b/CubeEgress/nginx.conf
@@ -124,8 +124,8 @@ http {
 
             proxy_pass http://$server_addr:$server_port;
             proxy_connect_timeout 10s;
-            proxy_send_timeout 60s;
-            proxy_read_timeout 60s;
+            proxy_send_timeout 2h;
+            proxy_read_timeout 2h;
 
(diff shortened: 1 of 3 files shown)


CubeEgress/nginx.conf
```diff
@@ -193,8 +194,9 @@ http {
             proxy_ssl_name $cube_original_sni;
             proxy_pass https://$server_addr:$server_port;
             proxy_connect_timeout 10s;
-            proxy_send_timeout 60s;
-            proxy_read_timeout 60s;
+            # Overridden at start by CUBE_EGRESS_PROXY_{SEND,READ}_TIMEOUT (default 300s).
+            proxy_send_timeout 300s;
```
chenhengqi: We can simply raise these timeouts to 2hr like CubeProxy and don't expose timeout options.

FakeLearne: Done

chenhengqi (approved):

PR #1663, opened 2026-09-07
s3lvol: pending-delete, derived exports, and faster same-bucket ingest

### LINKED ISSUE #1634: [Bug Report] s3lvol: an export whose lease is confirmed absent and whose TTL has lapsed is neither released nor stops polling, degrading into a 1 req/s 404 storm after restart
## Summary

**Affected versions**: v0.7.0 (`s3lvol_git d0081641c59822e4e5653b7462e914410b81910a`, built 2026-08-28), and current `master`

**Severity**: High — a single node reliably saturates its 2 dedicated busy-poll cores, writes 200+ MB of error log in under two days, and the export registry grows without ever shrinking

**Component**: `CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c`

Once a reference export has both (a) a lease object confirmed absent (HEAD returns 404 → `-ENOENT`) and (b) a lapsed manifest TTL, it is in a terminal state that no importer can possibly be reading. Yet at that point:

1. `lease_poller` is never unregistered. It keeps issuing a HEAD once per second against a key already known not to exist, each one leaving an aws-c-s3 ERROR line in the log.
2. The registry entry is never removed, so a restart makes `exports_parse()` load it back and re-arm the poller.

The resulting request rate is exactly `leaked_entries × 1 req/s`, grows linearly with the number of entries, and accumulates permanently across restarts.

## Reproduction

Preconditions: `enable_s3lvol: true`, snapshot backend on S3 (ours is Baidu BOS via its S3-compatible API, plaintext HTTP + path style).

1. Pause a sandbox / resume it on another node, producing a reference export. The log looks healthy at this point:

   ```
   s3_export_run_ref: Export <uuid> references 2000 object(s) holding 2097152000 byte(s)
                      of a 2147483648-byte snapshot across 1 layer(s)
   s3lvol_export_lease_start: watching lease of export <uuid> every 1199 second(s)
   ```

2. Wait out the manifest TTL (~1200 s). The importer's `/meta/exports/<uuid>.lease` does not exist (never created, or already cleaned up), so every lease check returns 404:

   ```
   [ERROR] [S3MetaRequest] - Meta request cannot recover from error 14343
           (Invalid response status from request). (response status=404)
   ```

   At this stage the interval is still 1200 s, so it is nearly invisible — **but the entry has already leaked**.

3. `systemctl restart cube-sandbox-s3lvol`. On attach:

   ```
   exports_parse: lvstore '<lvs>' still owes export <uuid> the snapshot '<name>'
                  (deadline passed; the lease decides)
   s3lvol_export_lease_start: watching lease of export <uuid> every 1 second(s)
   ```

   The poll interval goes from 1200 s to **1 s**, and the 404 storm begins.

**Is a restart a guaranteed reproducer?** Yes, deterministically, provided the registry already holds at least one such entry. The restart does not create the leak; it amplifies a silent one by a factor of 1200. The entries themselves also accumulate deterministically — see the data below.

## Observed data

Four machines across two clusters reproduced this independently, and the leaked-entry count matches the measured 404 rate exactly 1:1:

- Node A: 57 entries → ~157 req/s; 1.35M error lines and 279 MB of log in 12.5 h; both SPDK reactor cores consumed by lease checks
- Node B: 27 entries → 810 hits in 30 s = 27 req/s; 411 MB log
- Node C: 6 entries → 180 hits in 30 s = 6 req/s; 219 MB in under two days
- Node D: 3 entries → 92 hits in 30 s = 3 req/s; 115 MB

Node A also suffered a whole-machine freeze (no panic, no OOM, clean IPMI SEL, stable memory) which has not recurred since s3lvol was stopped there. We cannot establish causation (journald was not persistent, no kdump), so this is recorded as background only.

**An export is never released on its own.** Across the complete log history of two machines, the number of spontaneous `Released export` events is **0**:

- Node B: 42 exports created in total = 27 owed + 15 still live. Not one ever left the registry.
- Node A: 57 created, all 57 owed.

Consequently, operational workarounds such as "avoid deleting the snapshot inside the TTL window" do not help: the leak is established the moment the deadline passes, independent of when the snapshot is deleted.

## Root cause

The handling of 404 in `export_lease_head_done()` is itself correct:

```c
if (status == -ENOENT) {
        exp->lease_checked = true;
        exp->lease_updated_at = 0;
        exp->lease_failures = 0;
        export_lease_get_done(ctx);
        return;
}
```

`lease_checked = true` together with `lease_updated_at = 0` lets the delete path through, which matches what we observe: `rcow_get_lvstores` reports these snapshots as `deletable: "YES"` throughout. The problem is what happens after this branch.

- **The poller is never unregistered.** `-ENOENT` is terminal: the lease object does not exist and `expires_at` has passed, so there is no way for an importer to show up later (an importer must write its lease before it can read). Yet `export_lease_renew()` keeps sending a HEAD every second. The `S3LVOL_LEASE_MAX_FAILURES` backoff only covers the `status != 0 && status != -ENOENT` branch; a 404 takes the "successfully determined" path, which *resets* `lease_failures` to 0, so backoff can never engage.

- **The registry entry is never reclaimed.** The comment in `exports_parse()` states this is deliberate:

  > Deliberately *not* a filter on `expires_at`. That stamp is written once at export creation and never refreshed, while the importer renews its lease past the manifest TTL for as long as it is reading.

  That reasoning holds completely for "TTL passed but the lease is still being renewed". But "lease confirmed 404 **and** TTL passed" is a different case: with both conditions true, no importer can be reading and the entry is safe to drop. The current code does not distinguish the two,   so the entry persists forever and re-arms a 1 s poller on every attach.

As an aside: #1537 previously dropped expired entries by `expires_at` at parse time (motivated by a registry growing past `S3LVOL_MAX_EXPORTS` and bricking the lvstore's attach), and was later superseded by the lease approach above. The lease approach fixed a real correctness problem — forgetting a live importer's pin — but reintroduced the unbounded-entry resource problem, this time surfacing as a 404 storm rather than a failed attach.

One secondary factor leaves the leak unowned. Per `CubeS3lvol/README.md`, Cubelet's `S3Cow.DeleteByKind` treats a refused snapshot delete as success. The control plane therefore believes the snapshot is gone, stops tracking it, and nobody ever comes back to release the export. The logs show a large number of:

```
s3lvol_lvol_destroy: *ERROR*: lvol '<snap>' is the snapshot behind export <uuid>,
    which another node may be reading through. Release that export, or wait for
    it to expire, before deleting this.
```

On one machine 54 distinct snapshots hit this refusal, so it is the normal path rather than an edge case.

## Expected behaviour

When `lease_checked == true && lease_updated_at == 0 && now > expires_at`, the export should:

1. unregister `lease_poller` — the state is terminal, there is nothing left to check;
2. be released automatically, or at minimum be dropped from the registry in `exports_parse()` with the registry object rewritten, so it does not come back to life across restarts.

Failing that, the 404 branch should at least be subject to backoff instead of holding a constant 1 req/s.

## Current operational workaround

`rcow_release_export` clears these precisely. `export_uuid` is the only required parameter (`lvs_name` is not needed):

```sh
python3 scripts/s3lvol_rpc.py rcow_release_export '{"export_uuid":"<uuid>"}'
```

Each call deletes exactly one S3 object (the manifest) and leaves the local lvol untouched: after release, `export_status` flips from `DONE` to `NONE` and the lvol count is unchanged. After running it over 84 entries, the 404 rate dropped to exactly zero, and a restart confirmed the registry rewrite with `0 export(s) in the registry`.

Two things worth flagging for anyone hitting this:

- **`s3lvol_rpc.py --retry-pending` does not solve this.** It only acts on snapshots with `delete_pending == true && deletable == YES`, whereas the snapshots behind these leaked exports are invariably `delete_pending: false`. Using it here would instead delete a different set of snapshots that an operator is supposed to judge individually.
- **The uuid of an owed export is only obtainable from the log.** `rpc_get_methods` exposes no equivalent of `rcow_get_exports` (only the import-side `rcow_get_imports`), and `rcow_get_lvstores` yields `export_status` but not the uuid. The sole source is the `still owes export <uuid> the snapshot '<name>'` line emitted by `exports_parse()` at attach. An RPC to query the export registry would make this far easier to operate.

## Environment

- CubeSandbox v0.7.0 (in-place upgrade from v0.6.0)
- CubeS3lvol v0.7.0, `s3lvol_git d0081641c59822e4e5653b7462e914410b81910a`, built 2026-08-28
- SPDK `d64c4fa89233397460e2e4ff55a1c69b8e498598`, `spdk_target_arch: haswell`, aws-c-s3 v0.8.7
- Object store: Baidu BOS via its S3-compatible API, `path_style = true`, `no_tls = true` (plaintext HTTP)
- s3lvol started with `-m 0x3`: 2 reactors pinned to 2 dedicated cores

changpe1:
## Summary
- Complete pending-delete with a poller, export reaper, lease-renewal floor, and `rcow_get_exports` (fixes #1634). Allow shared OpenSSL when `libssl.a` is missing so local `make` works on RHEL/TencentOS.
- Add manifest v3 / chained derived exports, snapshot cancel+converge, Cubelet-facing RPC logging, and a dataplane suite that drives cubecow's `rcow_*` order (including not-found `get_bdev`).
- Speed same-bucket restore: decouple via CopyObject ingest (prefetch + chunk-map bind), parse `CopyObjectResult` instead of HEADing every cluster, and wait for a live nvme node after reactivate. Ingest `end` now waits for in-flight binds so `s3_ingest` is not freed while journal callbacks still hold slot pointers.

## Test plan
- [x] `run_export_test.sh` (194 passed) — same-bucket import + background decouple
- [x] `run_decouple_queue_test.sh` (22 passed)
- [x] `run_agent_template_test.sh` (33 passed)
- [x] `run_cubecow_client_test.sh` (27 passed)
- [ ] Remaining CubeS3lvol dataplane suites as environment allows (`make check` in `CubeS3lvol/`)

Assisted-by: Cursor:Grok-4.6

### DIFF
--- a/CubeS3lvol/README.md
+++ b/CubeS3lvol/README.md
@@ -209,18 +209,27 @@ scripts/s3lvol_rpc.py rcow_delete_lvol     '{"lvol_name":"clone0"}'
 scripts/s3lvol_rpc.py rcow_get_lvstores    # list all lvstores and lvols
 ```
 
+(`rcow_pending_load_hold` exists only so tests can park pending-delete registry
+HEAD/GET around unload; it is not an operations RPC.)
+
 The parameter names are deliberately chosen; the easy mistakes to make when
 copying them:
 
 - `create_lvol` has **no `lvs_name`** — it operates on the single lvstore that
   exists.
 - `create_snapshot` takes `lvol_name` as the source and `snapshot_name` as the
-  target; snapshots are read-only.
+  target; snapshots are read-only. If the source is still decoupling from an
+  import, the snapshot keeps that external parent and the reply includes
+  `decouple_cancelled: true`.
 - `create_clone` takes `snapshot_name` (which **must be a read-only snapshot**)
   as the source and `clone_name` as the target.
 - `resize_lvol` can only grow, never shrink.
 - `delete_lvol` also clears the volume's activation record, so the next restart
-  does not try to restore it.
+  does not try to restore it. A snapshot that is still pinned is recorded as an
+  intent rather than carried out. When the blocker is one the poller can finish
+  on its own, the reply is the usual success envelope plus `deferred: true`. A
+  lease-less export (`export_legacy`) still returns EBUSY; the mark is recorded
+  either way. `rcow_cancel_pending_delete` withdraws it.
 
 **Exporting a snapshot** (cross-node transfer):
 
@@ -402,10 +411,11 @@ only inside one loaded lvstore and is reusable, so a name-keyed mark could end u
 pointing at a snapshot the delete was never refused for.
 
 `rcow_get_lvstores` reports the mark per snapshot (`delete_pending`), together
-with whether the snapshot is deletable *right now* (`deletable`). Once the
-blocker clears — the export is released or expires, the extra clone is deleted —
-the snapshot becomes deletable, but the delete still has to be carried out by
-hand:
+with whether the snapshot is deletable *right now* (`deletable`).
+
+Most recorded intents are finished by a **background poller** (~60 s, 16 deletes
+per tick) once the blocker is gone. `--retry-pending` is for the cases the
+poller will not touch, and for not waiting a minute:
 
 ```sh
 test/tools/s3lvol_rpc.py --retry-pending
@@ -432,38 +442,84 @@ $ test/tools/s3lvol_rpc.py --retry-pending
 no pending snapshot deletes to retry
 ```
 
-### What this is not
+### What the poller completes, and what still needs a human
 
-The mark is a record and a manual retry, nothing more. Specifically:
+The poller finishes deletes whose blockers clear without anyone deciding
+anything:
+
+- a **leased** export (`export`) — the importer stops renewing, the lease goes
+  stale, that is evidence nobody is reading;
+- an export still **publishing** (`export_inflight`);
+- extra **clones** (`clone_count`);
+- a running **decouple** (`decouple`).
+
+It will **not** complete:
+
+- an export with **no lease at all** (`export_legacy`). Its TTL lapsing only
+  means time passed, not that an importer stopped reading; auto-completing that
+  would delete objects out from under a live reader. Those marks stay until a
+  human (or `--retry-pending`) decides;
+- a destroy that already **failed asynchronously** (`failed`). Retrying it every
+  minute would only repeat the same failure.
+
+If the poller itself failed to register, queued deletes also wait for an
+explicit retry (a log line says so).
+
+### Persistence and the crash window
+
+Marks are written to `<prefix>/meta/pending-deletes.json` and restored when the
+lvstore attaches, so a restart does not forget a delete the caller was told not
+to reissue.
+
+The RPC answers **before** that PUT. If the process crashes (or the PUT fails)
+in the window after the mark is in memory and before the object lands, the
+intent is gone after the next attach and the delete has to be asked for again
+— the same class of window the exports registry accepts, and cheaper: the
+snapshot is still there.
+
+Unload or destroy drops the in-memory queue for that lvstore (marks must not
+outlive the lvols they name). The S3 object is reloaded on the next attach of
+the same prefix; a failed or missing registry is logged and attach continues.
+
+### Cancelling a pending delete (`rcow_cancel_pending_delete`)
+
+The mark is an intent, not a lock. Withdraw it with:
+
+```sh
+scripts/s3lvol_rpc.py rcow_cancel_pending_delete '{"lvol_name":"snap0"}'
+# optional lvs_name when the same snapshot name exists on more than one lvstore
+scripts/s3lvol_rpc.py rcow_cancel_pending_delete '{"lvs_name":"vs0","lvol_name":"snap0"}'
+```
+
+`lvol_name` is required; `lvs_name` is optional and only disambiguates. The
+RPC is **idempotent**: cancelling a name that was never queued, or that the
+poller has already finished, still succeeds — the caller asked for "this
+snapshot will not be deleted behind my back", and that is the state they get.
+The snapshot itself was never touched, only the intent.
+
+The same crash window applies as for recording: the RPC answers, then the
+registry PUT runs. A crash before that PUT lands can restore the mark on the
+next attach.
+
+If the poller has **already submitted** `s3lvol_lvol_destroy`, cancel only
+drops the mark. It does not abort the in-flight destroy; the snapshot may
+still go away. Check `rcow_get_lvstores` afterwards if that race matters.
+
+### What this is not
 
 - **Not every refusal records a mark.** Recorded are the blockers
   `s3lvol_lvol_destroy()` itself identifies — an export pin (published or still
   in flight), more than one clone, a running decouple — plus an asynchronous
-  destroy failure. Those are the ones that clear on their own, which is what
-  makes coming back to them worthwhile. Not recorded: the RPC-layer refusals
-  that run before it (an NVMf-active volume, an unreadable active registry), a
-  bdev unregister that fails, and the case where `spdk_blob_get_clones()`
-  answers an unknown error. Those are either the caller's own precondition to
-  fix (deactivate first) or a failure that has to be looked at rather than
-  retried blindly.
+  destroy failure. Not recorded: the RPC-layer refusals that run before it (an
+  NVMf-active volume, an unreadable active registry), a bdev unregister that
+  fails, and the case where `spdk_blob_get_clones()` answers an unknown error.
+  Those are either the caller's own precondition to fix (deactivate first) or a
+  failure that has to be looked at rather than retried blindly.
 
   The recording does not re-check that the volume is a snapshot: with the blob
   closed that is not reliably answerable, and only the three blockers above get
   that far anyway. In practice a mark therefore means "a delete was asked for
   and refused because something still referenced the volume".
-- **No automatic retry.** Nothing on the target polls the marks; there is no
-  deferred-completion poller. `--retry-pending` is the only thing that acts on
-  them, and it has to be run.
-- **In memory only.** The marks live in the target process and are gone on
-  restart. A delete refused before a restart is afterwards indistinguishable
-  from one that was never asked for.
-- **Dropped with the lvstore.** Unloading or deleting an lvstore drops that
-  lvstore's marks, deliberately: past the teardown they would name lvols that no
-  longer exist, and an lvstore attached again can give the same names to
-  different objects.
-- **No cancel.** Completing the delete is the only way to clear a mark; a
-  refused delete stays recorded for as long as the target runs and its lvstore
-  stays attached.
 - **`delete_pending` is a snapshot notion.** `rcow_get_snapshot_status` reports
   it when queried by `snapshot_name`; queried by `export_uuid` it is always
   `false`, because an export names a snapshot but is not one.
--- a/CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
+++ b/CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
@@ -2346,7 +2346,7 @@ s3_flush_put(struct s3_flush_ctx *fc)
 	/* Overlay goes on last so it wins over whatever the old object held. */
 	s3_overlay_flush_merge(ctx->overlay, fc->view, fc->chunk_buf);
 
-	/* create-once (P2): a fresh uuid every time, never an in-place update. */
+	/* create-once: a fresh uuid every time, never an in-place update. */
 	spdk_uuid_generate(&fc->new_uuid);
 	s3_data_key(ctx, &fc->new_uuid, key, sizeof(key));
 
@@ -3215,23 +3215,19 @@ s3_bs_dev_get_stats(struct spdk_bs_dev *bs_dev, struct s3_bs_dev_stats *out)
 	out->ckpt_lsn    = ctx->ckpt_lsn_done;
 	out->ckpt_gen    = ctx->ckpt_gen;
 	out->ckpt_interval_sec = ctx->ckpt_interval_sec;
-	{
-		struct s3_journal *j = s3_chunk_map_get_journal(ctx->chunk_map);
 
-		out->journal_used_bytes = s3_journal_get_used_bytes(j);
-		out->journal_capacity_bytes = s3_journal_get_capacity_bytes(j);
-	}
+	struct s3_journal *j = s3_chunk_map_get_journal(ctx->chunk_map);
+	out->journal_used_bytes = s3_journal_get_used_bytes(j);
+	out->journal_capacity_bytes = s3_journal_get_capacity_bytes(j);
 
 	out->overlay_bytes       = s3_overlay_get_bytes(ctx->overlay);
 	out->overlay_live_chunks = s3_overlay_get_live_chunks(ctx->overlay);
-	{
-		struct s3_overlay_stats ostats = {};
 
-		s3_overlay_get_stats(ctx->overlay, &ostats);
-		out->overlay_flushed_full   = ostats.flushed_full;
-		out->overlay_flushed_aged   = ostats.flushed_aged;
-		out->overlay_flushed_forced = ostats.flushed_forced;
-	}
+	struct s3_overlay_stats ostats = {};
+	s3_overlay_get_stats(ctx->overlay, &ostats);
+	out->overlay_flushed_full   = ostats.flushed_full;
+	out->overlay_flushed_aged   = ostats.flushed_aged;
+	out->overlay_flushed_forced = ostats.flushed_forced;
 
 	if (ctx->flusher) {
 		s3_flusher_get_stats(ctx->flusher, &fstats);
--- a/CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
+++ b/CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
@@ -2142,7 +2142,6 @@ struct s3_ingest_slot {
 	struct spdk_uuid          uuid;
 	uint32_t                  valid_bytes;
 	char                      dest_key[S3_KEY_MAX];
-	uint64_t                  head_size;
 	struct s3_ingest_op      *waiter;
 };
 
@@ -2150,6 +2149,9 @@ struct s3_ingest {
 	char                      src_bucket[128];
 	s3_ingest_src_fn          src_fn;
 	void                     *src_arg;
+	/* CopyObjects plus in-flight chunk-map binds. End must not free this
+	 * struct until both have drained: bind callbacks keep a pointer into
+	 * slots[], which lives here. */
 	uint32_t                  inflight;
 	struct s3_ingest_slot     slots[S3_INGEST_SLOTS];
 	s3_bs_dev_cb              end_cb;
@@ -2160,18 +2162,56 @@ struct s3_ingest {
 struct s3_ingest_op {
 	struct s3_ctx             *ctx;
 	struct spdk_bs_dev_cb_args *cb_args;
+	struct spdk_thread       *submit_thread;
 	uint64_t                   dst_lba;
 	uint64_t                   src_lba;
 	uint64_t                   lba_count;
 	uint64_t                   pos;
+	int                        status;
+};
+
+struct ingest_prefetch_msg {
+	struct s3_ctx *ctx;
+	uint64_t        src_lba;
+	uint64_t        lba_count;
 };
 
 static void s3_bs_dev_copy(struct spdk_bs_dev *dev, struct spdk_io_channel *channel,
 			   uint64_t dst_lba, uint64_t src_lba, uint64_t lba_count,
 			   struct spdk_bs_dev_cb_args *cb_args);
 static void ingest_op_next(struct s3_ingest_op *op);
+static void ingest_op_complete(struct s3_ingest_op *op, int status);
 static void ingest_try_finish_end(struct s3_ctx *ctx);
 static int ingest_start_slot(struct s3_ingest_slot *slot, uint64_t src_chunk);
+static void ingest_bound(void *cb_arg, const struct spdk_uuid *old_uuid, int status);
+static void ingest_start_bind(struct s3_ingest_slot *slot, uint64_t dst_chunk);
+
+static void
+ingest_op_deliver(void *arg)
+{
+	struct s3_ingest_op *op = arg;
+
+	op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, op->status);
+	free(op);
+}
+
+static void
+ingest_op_complete(struct s3_ingest_op *op, int status)
+{
+	int rc;
+
+	op->status = status;
+	if (!op->submit_thread || op->submit_thread == spdk_get_thread()) {
+		ingest_op_deliver(op);
+		return;
+	}
+	rc = spdk_thread_send_msg(op->submit_thread, ingest_op_deliver, op);
+	if (rc != 0) {
+		SPDK_ERRLOG("ingest copy complete bounce failed: %s\n",
+			    spdk_strerror(-rc));
+		ingest_op_deliver(op);
+	}
+}
 
 static struct s3_ingest_slot *
 ingest_find_slot(struct s3_ingest *in, uint64_t src_chunk, bool alloc_free)
@@ -2211,6 +2251,20 @@ ingest_abandon_copy(struct s3_ingest_slot *slot)
 	ingest_slot_reset(slot);
 }
 
+static bool
+ingest_has_async_slots(struct s3_ingest *in)
+{
+	uint32_t i;
+
+	for (i = 0; i < S3_INGEST_SLOTS; i++) {
+		if (in->slots[i].state == S3_INGEST_COPYING ||
+		    in->slots[i].state == S3_INGEST_BINDING) {
+			return true;
+		}
+	}
+	return false;
+}
+
 static void
 ingest_try_finish_end(struct s3_ctx *ctx)
 {
@@ -2219,7 +2273,10 @@ ingest_try_finish_end(struct s3_ctx *ctx)
 	void *arg;
 	uint32_t i;
 
-	if (!in || !in->ending || in->inflight != 0) {
+	/* Do not free while a CopyObject or a chunk-map bind is still using a
+	 * slot. inflight should already cover both; the slot walk is in case a
+	 * bind was issued without a matching increment. */
+	if (!in || !in->ending || in->inflight != 0 || ingest_has_async_slots(in)) {
 		return;
 	}
 
@@ -2241,30 +2298,48 @@ ingest_try_finish_end(struct s3_ctx *ctx)
 }
 
 static void
+ingest_start_bind(struct s3_ingest_slot *slot, uint64_t dst_chunk)
+{
+	struct s3_ctx *ctx = slot->ctx;
+	struct s3_ingest *in = ctx->ingest;
+
+	slot->state = S3_INGEST_BINDING;
+	in->inflight++;
+	s3_chunk_map_insert(ctx->chunk_map, dst_chunk, &slot->uuid,
+			    slot->valid_bytes, ingest_bound, slot);
+}
+
+static void
 ingest_bound(void *cb_arg, const struct spdk_uuid *old_uuid, int status)
 {
 	struct s3_ingest_slot *slot = cb_arg;
-	struct s3_ingest_op *op = slot->waiter;
 	struct s3_ctx *ctx = slot->ctx;
+	struct s3_ingest *in;
+	struct s3_ingest_op *op;
 	uint32_t bpc;
 
+	if (!ctx || !ctx->ingest) {
+		return;
+	}
+	in = ctx->ingest;
+	if (in->inflight > 0) {
+		in->inflight--;
+	}
+
+	op = slot->waiter;
 	slot->waiter = NULL;
 
 	if (status != 0) {
 		/* Insert did not take: dest_key is unreferenced. */
 		ingest_abandon_copy(slot);
 		if (op) {
-			op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, status);
-			free(op);
+			ingest_op_complete(op, status);
 		}
+		ingest_try_finish_end(ctx);
 		return;
 	}
 
-	if (!ctx && op) {
-		ctx = op->ctx;
-	}
-
-	if (old_uuid && !spdk_uuid_is_null(old_uuid) && ctx) {
+	if (old_uuid && !spdk_uuid_is_null(old_uuid)) {
 		char old_key[S3_KEY_MAX];
 
 		s3_data_key(ctx, old_uuid, old_key, sizeof(old_key));
@@ -2273,23 +2348,25 @@ ingest_bound(void *cb_arg, const struct spdk_uuid *old_uuid, int status)
 
 	ingest_slot_reset(slot);
 	if (!op) {
+		ingest_try_finish_end(ctx);
 		return;
 	}
 
 	bpc = s3_blocks_per_chunk(ctx);
 	op->pos += bpc;
 	ingest_op_next(op);
+	ingest_try_finish_end(ctx);
 }
 
 static void
-ingest_slot_headed(void *cb_arg, int status)
+ingest_slot_copied(void *cb_arg, int status)
 {
 	struct s3_ingest_slot *slot = cb_arg;
 	struct s3_ctx *ctx = slot->ctx;
 	struct s3_ingest *in;
 
 	if (!ctx || !ctx->ingest) {
-		ingest_slot_reset(slot);
+		/* End already freed the parent struct that contains this slot. */
 		return;
 	}
 	in = ctx->ingest;
@@ -2298,18 +2375,13 @@ ingest_slot_headed(void *cb_arg, int status)
 		in->inflight--;
 	}
 
-	if (status == 0 && slot->head_size < slot->valid_bytes) {
-		status = -EIO;
-	}
-
 	if (status != 0) {
 		struct s3_ingest_op *op = slot->waiter;
 
 		slot->waiter = NULL;
 		ingest_abandon_copy(slot);
 		if (op) {
-			op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, status);
-			free(op);
+			ingest_op_complete(op, status);
 		}
 		ingest_try_finish_end(ctx);
 		return;
@@ -2321,34 +2393,9 @@ ingest_slot_headed(void *cb_arg, int status)
 					     slot->waiter->dst_lba + slot->waiter->pos,
 					     ctx->chunk_shift);
 
-		slot->state = S3_INGEST_BINDING;
-		s3_chunk_map_insert(ctx->chunk_map, dst_chunk, &slot->uuid,
-				    slot->valid_bytes, ingest_bound, slot);
+		ingest_start_bind(slot, dst_chunk);
 	}
 	ingest_try_finish_end(ctx);
-}
-
-static void
-ingest_slot_copied(void *cb_arg, int status)
-{
-	struct s3_ingest_slot *slot = cb_arg;
-	struct s3_ctx *ctx = slot->ctx;
-	int rc;
-
-	if (status != 0) {
-		ingest_slot_headed(slot, status);
-		return;
-	}
-	if (!ctx) {
-		ingest_slot_reset(slot);
-		return;
-	}
-
-	rc = s3_head(ctx->client, slot->dest_key, &slot->head_size,
-		     ingest_slot_headed, slot);
-	if (rc != 0) {
-		ingest_slot_headed(slot, rc);
-	}
 }
 
 static int
@@ -2392,30 +2439,26 @@ ingest_op_next(struct s3_ingest_op *op)
 	int rc;
 
 	if (op->pos >= op->lba_count) {
-		op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, 0);
-		free(op);
+		ingest_op_complete(op, 0);
 		return;
 	}
 
 	if (!in) {
-		op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, -EINVAL);
-		free(op);
+		ingest_op_complete(op, -EINVAL);
 		return;
 	}
 
 	dst_off = op->dst_lba + op->pos;
 	if (s3_lba_offset_in_chunk(dst_off, ctx->chunk_shift) != 0 ||
 	    s3_lba_offset_in_chunk(op->src_lba + op->pos, ctx->chunk_shift) != 0) {
-		op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, -EINVAL);
-		free(op);
+		ingest_op_complete(op, -EINVAL);
 		return;
 	}
 
 	src_chunk = s3_lba_to_chunk_index(op->src_lba + op->pos, ctx->chunk_shift);
 	slot = ingest_find_slot(in, src_chunk, true);
 	if (!slot) {
-		op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, -ENOMEM);
-		free(op);
+		ingest_op_complete(op, -ENOMEM);
 		return;
 	}
 
@@ -2424,16 +2467,14 @@ ingest_op_next(struct s3_ingest_op *op)
 		slot->waiter = op;
 		rc = ingest_start_slot(slot, src_chunk);
 		if (rc != 0) {
-			op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, rc);
-			free(op);
+			ingest_op_complete(op, rc);
 		}
 		return;
 	}
 
 	if (slot->state == S3_INGEST_COPYING) {
 		if (slot->waiter) {
-			op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, -EBUSY);
-			free(op);
+			ingest_op_complete(op, -EBUSY);
 			return;
 		}
 		slot->waiter = op;
@@ -2444,14 +2485,17 @@ ingest_op_next(struct s3_ingest_op *op)
 		uint64_t dst_chunk = s3_lba_to_chunk_index(dst_off, ctx->chunk_shift);
 
 		slot->waiter = op;
-		slot->state = S3_INGEST_BINDING;
-		s3_chunk_map_insert(ctx->chunk_map, dst_chunk, &slot->uuid,
-				    slot->valid_bytes, ingest_bound, slot);
+		ingest_start_bind(slot, dst_chunk);
 		return;
 	}
 
-	op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, -EBUSY);
-	free(op);
+	ingest_op_complete(op, -EBUSY);
+}
+
+static void
+ingest_op_on_owner(void *arg)
+{
+	ingest_op_next(arg);
 }
 
 static void
@@ -2461,6 +2505,7 @@ s3_bs_dev_copy(struct spdk_bs_dev *dev, struct spdk_io_channel *channel,
 {
 	struct s3_ctx *ctx = (struct s3_ctx *)dev;
 	struct s3_ingest_op *op;
+	int rc;
 
 	(void)channel;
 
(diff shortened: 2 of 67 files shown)


fslongjin (changes requested):
Thanks for the large and thoughtful change. The pending-delete poller, export reaper, lease-renewal floor, and the overall direction for #1634 look right to me, and the in-code commentary is unusually clear.

That said, I do not think this is ready to merge yet. The PR bundles three largely independent changes (pending-delete lifecycle, manifest v3 / derived exports, and same-bucket CopyObject ingest). The ingest path introduces a silent data-corruption window, and the manifest refetch `-EALREADY` handling can turn a normal race into a spurious read failure. Neither path has a regression test in the default gate. User-facing docs also still describe the old pending-delete behavior.

## What looks good

- Pending-delete marks keyed by `(lvstore uuid, lvol uuid)` — avoids same-name mistakes.
- Lease-less exports are **not** auto-deleted by the poller — correct safety boundary.
- Export reaper + 20s lease floor — good fix for the post-restart 404 storm (#1634).
- `rcow_get_bdev` waits for a `/dev` node whose `major:minor` matches sysfs — avoids stale nodes after deactive/reactivate.
- Ingest `end` waits for in-flight binds before freeing ingest state.
- Same-bucket CopyObject speedup is a sensible optimization when scoped correctly.

## Must fix before merge

### 1. [HIGH] Sibling esnap volumes in the same lvstore can get wrong data during CopyObject ingest

To speed up same-bucket restore, decouple installs `copy()` on the **lvstore-wide shared** `bs_dev` and binds ingest to **one** decouple’s manifest. Patch `0006` also allows **any** esnap clone’s first write to a new cluster to use `dest->copy` when the parent’s `translate_lba` returns true (`export_translate_lba()` now always returns true).

Scenario:
- Volume **X** is in a CopyObject decouple (ingest active).
- Another esnap clone **Y** in the **same lvstore** (second import, sibling clone, queued decouple, etc.) takes its **first CoW write**.
- The write is routed through **X’s ingest/manifest**, not Y’s parent export.

Outcome: silent wrong objects if the chunk index lands in X’s manifest, or spurious I/O errors on Y if it does not.

`decouple_find_blocker()` serializes decouples per lvstore, but that does **not** block “X is decoupling while Y is writable.” `import(decouple=true)` activates the volume before decouple finishes, so this is normal traffic, not an edge case.

**Relevant code:** `patches/0006-blob-allow-esnap-dev-copy.patch`, `s3_export_bs_dev.c` (`export_translate_lba`), `s3_bs_dev.c` (`s3_bs_dev_ingest_begin` / `s3_bs_dev_copy`), `vbdev_s3lvol_xfer.c` (same-bucket ingest).

**Suggested fix (pick one or equivalent):**
1. Have `copy()` reject destinations that are not the lvol currently being decoupled; or
2. Install `copy()` only for that blob’s materialise window and keep `translate_lba` false otherwise; or
3. Prove and enforce that no other esnap clone in the lvstore can take a first write during ingest (document the new user-visible constraint).

**Test:** Please add a failing regression test: while lvstore volume A is in CopyObject ingest decouple, sibling esnap volume B’s first write must still be correct. Do not rely only on `probe_*.sh` scripts that are not wired into `run_all.sh`.

---

### 2. [MEDIUM] Manifest refetch treats `-EALREADY` as “source deleted the snapshot”

When a refetch finds the remote manifest is not newer than the one already installed, `swap_manifest` returns `-EALREADY`. `export_refetch_got_manifest()` currently treats that as proof the source deleted objects without materialising, releases waiters with `retry=false`, and completes reads with `-ENOENT`.

`-EALREADY` also means **another refetch already swapped this device to the current generation**. A GET issued against the pre-swap manifest can 404 after the swap; a second refetch then hits `-EALREADY` even though a retry against the new manifest would succeed. This produces spurious I/O errors and misleading “data loss” logs.

**Code:** `s3_export_bs_dev.c` — `-EALREADY` branch in `export_refetch_got_manifest()`.

**Fix:** Distinguish “generation is already current because another refetch handled it” (`retry=true`) from “generation is current and the objects are genuinely gone.” `s3_export_swap_test.c` should cover the late-404-after-swap case, not only “same generation => -EALREADY”.

---

### 3. [MEDIUM] Orphan S3 objects on chunk-map bind failure

On `s3_chunk_map_insert()` failure, `ingest_bound()` resets the slot and completes the waiter but does **not** delete `slot->dest_key`. The copy-failure path does call `s3_delete()`. Failed inserts leave unreferenced uuid objects in the bucket.

**Code:** `s3_bs_dev.c` — `ingest_bound()` error path vs `ingest_slot_copied()`.

---

### 4. [MEDIUM] User docs contradict the implementation

The code now has a ~60s pending-delete poller and S3 persistence (`pending_persist`), but these still say marks are in-memory only with no automatic retry:
- `CubeS3lvol/README.md` (“Retrying a refused snapshot delete”)
- `docs/guide/cross-node-snapshot.md` and `docs/zh/guide/cross-node-snapshot.md`
- `CubeS3lvol/test/README.md` (`--retry-pending` section)

Please update EN/ZH docs to describe: which reasons the poller auto-completes, which still need a human (especially lease-less exports), and the crash window before persist lands.

---

### 5. [MEDIUM] No gate tests for the riskiest paths

CI (`cube-s3lvol-test.yml`) runs offline tests only. The PR test plan does not list results for `run_pending_delete_test.sh`, `run_derived_test.sh`, `run_snapshot_cancel_test.sh`, or `run_snapshot_converge_test.sh`. The 12 `probe_*.sh` scripts are not in `run_all.sh` and are diagnostic only.

After fixing #1 and #2, please add tests that fail without the fix and run from the default test entry point, and update the PR test plan with full dataplane results.

## Nice to have

- Brief upgrade note for manifest v3 / pending registry (rolling upgrade, mixed versions).
- Confirm target stores always return `CopyObjectResult` on HTTP 200 (stricter than before).
- Add `Assisted-by:` to commits 1–4; fix typos in commit 4 subject (`Englinsh`, `unecessary`).
- Refresh the outdated `blob_can_copy` comment in `vbdev_s3lvol_lvstore.c`.
- Consider splitting this PR to reduce regression surface.

## Merge checklist

- [ ] CopyObject ingest `copy()` cannot affect sibling esnap first CoW in the same lvstore
- [ ] Refetch `-EALREADY` retries late 404s instead of failing with `-ENOENT`
- [ ] Bind failure deletes orphan `dest_key`
- [ ] README + cross-node-snapshot docs (EN/ZH) updated
- [ ] Regression tests for #1 and #2 in the default gate
- [ ] PR test plan lists full dataplane results

**Verdict:** Good direction and valuable #1634 work, but please fix the sibling-esnap ingest misroute and the refetch `-EALREADY` handling before merge.

---
*Review based on static analysis of the PR diff at `76e583b6`; I did not re-run the full dataplane suite in this environment.*

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -338,6 +288,41 @@ struct export_lease_get_ctx {
...
+ * the deadline. No importer can appear later -- they must write the lease
+ * before they can read -- so watching it, and keeping the registry entry, is
+ * a leak. Distinct from lease_updated_at == 0: that is also the fallback
+ * after S3LVOL_LEASE_MAX_FAILURES, which must not reap. */
+static bool
+export_lease_is_terminal(const struct s3lvol_export *exp)
+{
+	return exp->layout == S3_EXPORT_LAYOUT_REF
+	       && exp->lease_aware
+	       && exp->lease_absent
```
fslongjin: **[Critical] This combines a historical 404 with the current expiry state.** `lease_absent` only proves that the lease was missing during the most recent HEAD; it does not prove that the check happened after `expires_at`. An importer can create its lease after that HEAD but before the deadline, and the reaper can run after the deadline but before the source observes the PUT. This then removes the manifest/registry entry for an active import and drops its protection.

Please track when absence was observed and require an after-expiry miss, or issue a dedicated final HEAD after expiry before reaping. A regression test should create the lease after an earlier 404 but before expiry and exercise both HEAD/reaper completion orders.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -0,0 +1,1007 @@
...
+
+	if (!lvs) {
+		return;
+	}
+
+	ctx = calloc(1, sizeof(*ctx));
+	if (!ctx) {
+		return;
+	}
+	ctx->lvs = lvs;
```
fslongjin: **[Critical] This raw lvstore pointer can outlive the object.** The pending registry load is fire-and-forget after attach, but it is not included in teardown draining and holds no lvstore lifetime reference. If unload/delete completes before the HEAD or GET callback, `pending_load_done()` and `pending_parse()` dereference freed memory.

Please make this an owned lvstore operation that teardown waits for/cancels, or retain stable identity/client state and re-resolve the live lvstore by UUID in the callback. Tests should delay both HEAD and GET, unload during each stage, and ensure a late callback cannot populate a newly attached same-name lvstore.

CubeS3lvol/lib/s3bsdev/s3_client_aws.c
```diff
@@ -2065,6 +2071,80 @@ s3_delete_batch(struct s3_client *client, const char **keys, uint32_t count,
...
+	for (i = 0; i + 1 + nlen <= len; i++) {
+		char c;
+
+		if (xml[i] != '<') {
+			continue;
+		}
+		if (memcmp(xml + i + 1, name, nlen) != 0) {
+			continue;
+		}
+		c = xml[i + 1 + nlen];
```
fslongjin: **[High] One-byte heap over-read.** The loop permits `i + 1 + nlen == len`, then this line reads `xml[len]`. I reproduced this with ASan using an 8192-byte response ending in `<Error`.

Please require the delimiter index to be strictly below `len`. The body callback also silently truncates responses above 8192 bytes, so an opening-tag search alone can accept incomplete XML. Please add boundary, truncated-body, callback-fragmentation, and incomplete-XML tests.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -1250,16 +1559,39 @@ export_ref(struct export_ctx *ctx)
...
+		struct s3lvol_import *imp = import_find(ctx->lvs, esnap_uuid);
+
+		if (!imp) {
+			SPDK_NOTICELOG("Snapshot reads through export %s, which this "
+				       "lvstore has no manifest for; exporting by copying "
+				       "instead.\n", esnap_uuid);
+			export_copy(ctx);
+			return;
+		}
+		opts.parent = imp->m;
```
fslongjin, line 1628: **[High] This can publish a derived manifest from a stale parent generation.** A 404 refetch swaps only `s3_export_dev::m`; `s3lvol_import::m` and the persisted imports registry remain unchanged. Normal reads can therefore use the new generation while this path inherits old object references that materialisation has already removed. Same-bucket decouple has the same problem through `d->m = imp->m`.

Please make the import own one current manifest that is updated, persisted, and shared with the data plane, or refetch/validate generation before derive and decouple. Add refetch → derive, refetch → same-bucket decouple, and refetch → restart tests after deleting the old source objects.

CubeS3lvol/lib/s3bsdev/s3_export_ref.c
```diff
@@ -307,12 +343,44 @@ s3_export_run_ref(const struct s3_export_ref_opts *opts, s3_export_cb cb, void *
 			      " chunk(s)\n", opts->uuid_str, i, contributed);
 	}
 
+	/* Then whatever the chain inherited from an export rather than from a blob.
+	 * Last, so every local layer has already claimed what it owns -- a chunk
+	 * rewritten since the import must resolve to the rewritten cluster. */
+	if (opts->parent) {
+		uint64_t bytes = 0;
+
+		rc = s3_export_manifest_inherit(m, opts->parent, &inherited, &bytes);
```
fslongjin, line 362: **[High] Inheritance cannot distinguish an untouched hole from an explicit local zero.** `walk_layer()` recognizes a locally allocated chunk whose mapping was removed by zero/unmap and skips it as zero, but it does not mark that chunk as resolved. This inheritance pass then sees `m->present == 0` and restores the parent's old reference, so a derived export can silently expose stale data where the current volume contains zeros.

Please represent an explicit-zero/tombstone state, or maintain a separate resolved bitmap that prevents parent inheritance. Add a test that writes nonzero parent data, zeroes or unmaps one chunk locally, then snapshot/export/imports and verifies that chunk remains zero.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -0,0 +1,1007 @@
...
+
+	rc = pending_serialize(lvs, lvs_uuid, &ctx->json, &len);
+	if (rc != 0) {
+		free(ctx);
+		return;
+	}
+	ctx->iov.iov_base = ctx->json;
+	ctx->iov.iov_len = len;
+
+	rc = s3_put(s3lvol_lvstore_get_client(lvs), ctx->key, &ctx->iov, 1, false,
```
fslongjin: **[High] Full-registry PUTs can complete out of mutation order.** Every set/clear submits an independent PUT to the same key, with no per-lvstore serialization or generation check. A slow `set(A)` can land after a later `cancel(A)`; after restart A reappears and the background poller may execute a deletion the user cancelled. Other orderings can lose newly queued entries.

Please serialize writes per lvstore and use a mutation generation/dirty flag so completion schedules the newest snapshot. Tests should force reverse completion order for set/set, set/clear, and clear/set, then reload and verify last-mutation-wins behavior.

CubeS3lvol/lib/s3bsdev/s3_export_bs_dev.c
```diff
@@ -501,14 +638,411 @@ export_destroy(struct spdk_bs_dev *bs_dev)
...
+	 *
+	 * A reader that loads the new manifest must not find geometry belonging to
+	 * the old one. The other order would leave a window where it does, and the
+	 * result is a key built for the wrong chunk -- which resolves to an object
+	 * that exists and holds something else. Writing them first is safe because
+	 * they are only read *after* the pointer, and the old manifest's values are
+	 * what they already were unless the chunk size changed. */
+	dev->chunk_shift      = (uint32_t)spdk_u32log2(m->chunk_size);
+	dev->blocks_per_chunk = m->chunk_size / S3LVOL_BLOCK_SIZE;
+	dev->m                = m;
```
fslongjin: **[High] This is an unsynchronized cross-thread publication.** Other SPDK reactors read `dev->m`, `chunk_shift`, and `blocks_per_chunk` with ordinary loads while this path writes them with ordinary stores. `spdk_for_each_channel()` protects the old manifest's reclamation time, but it does not remove the C data race or ensure that a multi-chunk read uses one generation. The swap also permits the same UUID/size with a different chunk size.

Please atomically publish one manifest snapshot and have each I/O capture and use only that manifest's geometry through completion; at minimum reject chunk-size changes. The current swap test creates one SPDK thread, so this needs a multi-reactor concurrent read/swap test and preferably TSan coverage.

CubeS3lvol/README.md
```diff
@@ -432,38 +433,61 @@ $ test/tools/s3lvol_rpc.py --retry-pending
...
-  restart. A delete refused before a restart is afterwards indistinguishable
-  from one that was never asked for.
-- **Dropped with the lvstore.** Unloading or deleting an lvstore drops that
-  lvstore's marks, deliberately: past the teardown they would name lvols that no
-  longer exist, and an lvstore attached again can give the same names to
-  different objects.
-- **No cancel.** Completing the delete is the only way to clear a mark; a
-  refused delete stays recorded for as long as the target runs and its lvstore
-  stays attached.
+- **No cancel.** Completing the delete is the only way to clear a mark.
```
fslongjin: **[Medium] This contradicts the implementation in this PR.** `rcow_cancel_pending_delete` is registered in `vbdev_s3lvol_rpc.c`, so pending deletes can be cancelled.

Please document its parameters, idempotent behavior, persistence window, and the race boundary if the poller has already submitted destroy. The English and Chinese guides should also cover the new `deferred=true` delete response and `cancelled_decouple` snapshot behavior.

fslongjin (changes requested):
The overall direction for fixing #1634 is sound, especially stopping terminal lease polling, reclaiming dead registry entries, and exposing `rcow_get_exports`. However, I found blocking correctness and memory-safety issues in the new lifecycle, persistence, derived-export, and manifest-swap paths.

The inline comments cover:

1. reaping based on a lease 404 observed before expiry;
2. a pending-registry load that can outlive its lvstore;
3. an ASan-reproduced one-byte heap over-read in CopyObject response parsing;
4. split manifest authority after refetch;
5. explicit zero/unmap being overwritten by inherited parent data;
6. out-of-order pending-registry PUTs;
7. unsynchronized cross-thread manifest replacement;
8. documentation contradicting the new cancellation RPC.

Before merge, these paths need deterministic ordering/lifetime regression tests in addition to the happy-path dataplane suites. In particular, please cover lease creation between an earlier 404 and expiry, unload during delayed pending-registry HEAD/GET, reverse completion order of registry PUTs, refetch followed by decouple/derive/restart, zero/unmap over inherited data, and multi-reactor read/swap. The remaining existing integration and dataplane suites should also be run because this PR changes several independent production data paths.

CubeS3lvol/lib/s3bsdev/s3_export_bs_dev.c
```diff
@@ -212,44 +328,69 @@ export_read(struct spdk_bs_dev *bs_dev, struct spdk_io_channel *channel,
 		return;
 	}
 	io->dev = dev;
+	io->m = __atomic_load_n(&dev->m, __ATOMIC_ACQUIRE);
+	s3_export_manifest_ref(io->m);
```
fslongjin, line 333: **[High] This new per-I/O reference hands a non-atomic counter to arbitrary threads.**

The `home_thread` comment in this file states that everything else is read-only after construction, "which is what lets reads run on any thread with no serialisation" — nvmf runs one thread per poll group, and blobstore calls into the bs_dev from whichever thread holds the channel. So this new `s3_export_manifest_ref(io->m)` — and the matching unref on the completion path — now execute **concurrently on multiple threads**, while `s3_export_manifest_ref`/`s3_export_manifest_unref` are still plain `m->refcnt++` / `--m->refcnt` (`s3_export.c:265-281`).

A lost increment under-counts the manifest; the swap's unref then frees it while reads are still in flight, and the `io->m` dereference after the await in `export_io_complete` (line 224) becomes a use-after-free. The pointer itself was made atomic (ACQUIRE load / RELEASE store) — but not the counter that protects it.

The fix is small: `__atomic_fetch_add(&m->refcnt, 1, __ATOMIC_RELAXED)` in `ref`, `__atomic_fetch_sub(..., __ATOMIC_ACQ_REL)` in `unref`, freeing when the returned value is 1. The manifest's other fields are immutable — refcnt is the only one that needs this. The `dev->reads++`-style cross-thread counters can get the same treatment in passing.

`s3_export_swap_test.c` is single-threaded and its device never reads, so it cannot catch this; a two-thread concurrent-read + swap test would be ideal.

CubeS3lvol/module/bdev/s3lvol/pending_persist_ctl.h
```diff
@@ -0,0 +1,51 @@
...
+/* True: caller should serialise the current marks and PUT again. */
+static inline bool
+pending_persist_end(struct pending_persist_ctl *c)
+{
+	c->in_flight = false;
+	if (!c->dirty) {
+		return false;
+	}
+	c->dirty = false;
+	c->in_flight = true;
```
fslongjin: **[High] The `end()` preset of `in_flight = true` deadlocks the restart path: after the first coalesced write, this lvstore permanently stops persisting.**

Sequence: PUT#1 is in flight, a second mutation arrives → `begin()` only sets dirty → PUT#1 completes → `end()` consumes dirty, **presets `in_flight = true`**, returns true → the caller `pending_save_done` (`vbdev_s3lvol_pending.c:763-765`) goes through `pending_persist()` → which calls `begin()` again (line 804) → sees `in_flight == true` → sets dirty, returns false → **PUT#2 is never submitted**. From then on `in_flight=true` is a lie: every subsequent mark/cancel only touches memory, until unload resets it. After a restart the registry reflects the old state — a cancelled pending delete comes back, which is exactly the scenario this state machine was built to prevent.

I reproduced this by driving the real header through the production call sequence (`begin → [mutation: begin] → end → begin`): after the first coalesced write the state is `puts=1 in_flight=1 dirty=1`, and two more mutations later it is still `puts=1`. The trigger is merely "a second mutation arrives while a PUT is in flight" — with WAN S3 PUT latencies of 50-500ms, back-to-back mark/cancel hits it.

The two sync-failure restarts inside `pending_persist()` (serialize failure, and `s3_put` submission failure, around lines 818 and 835) have the same shape — please cover them too.

The unit test models the resubmission as `puts++` directly (`s3_pending_persist_test.c:40-44`), bypassing the `begin()` gate, which is why 15 green assertions still miss this. Please add a case asserting that `begin()` immediately after a `true`-returning `end()` submits.

Simplest fix: drop this preset. The restart is a same-thread synchronous call; there is no window between `end` and the next `begin` for it to guard.

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -2116,6 +2120,462 @@ s3_bs_dev_destroy(struct spdk_bs_dev *dev)
...
+		ingest_start_bind(slot, dst_chunk);
+		return;
+	}
+
+	op->cb_args->cb_fn(op->cb_args->channel, op->cb_args->cb_arg, -EBUSY);
+	free(op);
+}
+
+static void
+s3_bs_dev_copy(struct spdk_bs_dev *dev, struct spdk_io_channel *channel,
```
fslongjin, line 2502: **[Medium] `copy()` and prefetch skip the owner-thread bounce that read/write use here, leaving the ingest slots in a cross-thread data race.**

I found this during my first review round and failed to post it until now — apologies for the late addition. This code is unchanged since.

The comment at the top of this file (lines 282-285) documents that blobstore calls bs_dev entry points from whichever thread holds the channel (nvmf runs one thread per poll group), which is why read/write bounce to `owner_thread` (lines 1287-1293). But `s3_bs_dev_copy()` (here) and `s3_bs_dev_ingest_prefetch()` (line 2523) mutate `slot->state` / `slot->waiter` / `slot->ctx` and `in->inflight` directly on the calling thread, with the lock-free `ingest_find_slot()` scan (line 2181).

`spdk_blob_allow_esnap_copy(blob, true)` (`vbdev_s3lvol_xfer.c:3629`) opens the gate for the whole blob and the whole decouple. During that window, a guest first write to an unallocated cluster of the same volume from any poll-group thread goes through patch 0006's `blob_can_copy()` into this `copy()`, concurrently with the owner/decouple thread's `decouple_prefetch` (`vbdev_s3lvol_xfer.c:3402`) and the S3 completions (bounced back to the submitting thread, `s3_client_aws.c:355-367`) mutating the same slots. Under C11 this is UB; the outcomes include double-starting the same `src_chunk`, losing a waiter (blobstore waits forever), and miscounting `inflight` so `free(in)` runs while a callback still holds a slot (UAF).

Suggested fix (same shape as read/write): bounce `copy` and prefetch to `owner_thread` via `spdk_thread_send_msg`; or protect the ingest state with a lock only taken on the owner thread; or have `copy()` refuse non-materialise callers during the gate so guest CoW falls back to GET+write. A deterministic unit test over the ingest lifecycle permutations (copy completion × bind completion × ingest_end) would be the ideal companion.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -0,0 +1,1270 @@
...
+		}
+	}
+	pending_poller_sync();
+
+	{
+		struct pending_persist *p, *ptmp;
+
+		TAILQ_FOREACH_SAFE(p, &g_pending_persist, link, ptmp) {
+			if (spdk_uuid_compare(&p->lvs_uuid, lvs_uuid) == 0) {
+				TAILQ_REMOVE(&g_pending_persist, p, link);
```
fslongjin: **[Medium] Freeing the persist entry here does not wait for an in-flight registry PUT; after a same-uuid re-attach, a late PUT can write a cancelled mark back.**

Attach goes through `spdk_lvs_load_ext` (`vbdev_s3lvol_lvstore.c:1710`) and re-hangs the **same blobstore, same uuid** — so "unload then same-uuid revival" is the normal restart/re-attach path, not an exotic one. Two consequences, same root cause (the entry is freed while a PUT is in flight):

- **(a) a coalesced cancel is lost**: mark(A)'s PUT#1 is in flight, cancel(A) arrives (dirty is memory-only) → unload frees the entry here and nobody writes the registry again → PUT#1 lands late, registry = {A} → the re-attach's load reads A back; the user's cancellation is gone and the poller re-evaluates the delete.
- **(b) a same-uuid new entry gets cross-talked by the old completion**: after re-attach, a new mutation creates a fresh entry with the same uuid and submits PUT#2; the old PUT#1 then completes, and `pending_save_done` (line 759) finds the **new** entry by uuid and runs `end()` on its ctl — clearing `in_flight` while PUT#2 is still flying, breaking the one-in-flight invariant and allowing two concurrent PUTs to land out of order.

Suggestion: when this loop sees `p->ctl.in_flight`, don't free — mark the entry dead, clear the in-memory marks and set dirty, and let the completion callback write an **empty registry** before deleting itself; or have the unload path wait for the in-flight PUT before dropping the entry. Natural to fix together with the `pending_persist_ctl.h` comment.

CubeS3lvol/README.md
```diff
@@ -215,12 +215,17 @@ copying them:
...
+  target; snapshots are read-only. If the source is still decoupling from an
+  import, the snapshot keeps that external parent and the reply includes
+  `decouple_cancelled: true`.
 - `create_clone` takes `snapshot_name` (which **must be a read-only snapshot**)
   as the source and `clone_name` as the target.
 - `resize_lvol` can only grow, never shrink.
 - `delete_lvol` also clears the volume's activation record, so the next restart
-  does not try to restore it.
+  does not try to restore it. A snapshot that is still pinned is accepted as an
+  intent rather than carried out: the usual success envelope plus
```
fslongjin: **[Medium] This sentence unconditionally promises success + `deferred: true` for pinned snapshots, but a lease-less export actually returns EBUSY.**

`rcow_delete_lvol` only takes the success+deferred path when `s3lvol_snapshot_pending_deferred()` is true (`vbdev_s3lvol_rpc.c:1324-1344`), and that requires `pending_reason_auto(reason)` (`vbdev_s3lvol_pending.c:166-173`) — `export_legacy` is not in that set. Deleting a snapshot pinned by a lease-less export returns an **EBUSY error** (the mark is still recorded and waits for `--retry-pending`). README lines 454-457 describe the export_legacy behavior correctly; this paragraph just doesn't carve it out of "still succeeds".

The same sentence appears in `docs/guide/cross-node-snapshot.md:319` and `docs/zh/guide/cross-node-snapshot.md:369`. Please update all three to: "auto-completable reasons → success + `deferred: true`; a lease-less export still returns EBUSY (the mark is recorded all the same)". EN/ZH need to stay in sync.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol.h
```diff
@@ -744,27 +785,148 @@ int s3lvol_snapshot_query_lvol(struct spdk_lvol *lvol,
...
- * shows up in rcow_get_lvstores, and the caller (today
- * test/tools/s3lvol_rpc.py --retry-pending) reissues rcow_delete_lvol once the
- * blocker (export pin, clones, decouple) clears. The list lives in memory only
- * and is gone on restart.
+ * not complete" mark.
+ *
+ * The mark is reported by rcow_get_lvstores (the PEND column) and by
+ * rcow_get_pending_deletes, and a poller completes the delete once the blocker
+ * clears -- for the reasons that clear on their own. The list lives in memory
+ * only and is gone on restart.
```
fslongjin: **[Low] This sentence contradicts the persistence this very PR adds** (`vbdev_s3lvol_pending.c:605-608`, `:779-838`; the `s3lvol_pending_load` doc comment in this same file describes the restore). Please delete or rewrite it.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_rpc.c
```diff
@@ -2407,6 +2736,228 @@ rpc_rcow_get_imports(struct spdk_jsonrpc_request *request,
...
+};
+
+static const struct spdk_json_object_decoder rpc_pending_load_hold_decoders[] = {
+	{"stage",   offsetof(struct rpc_pending_load_hold, stage),   spdk_json_decode_string, true},
+	{"release", offsetof(struct rpc_pending_load_hold, release), spdk_json_decode_bool,   true},
+};
+
+static void
+rpc_rcow_pending_load_hold(struct spdk_jsonrpc_request *request,
+			    const struct spdk_json_val *params)
```
fslongjin, line 2913: **[Low] This is an ungated RPC left in the production binary**: any client can send `stage=head` and permanently park every pending load on that target (no timeout) — only a "Tests only" comment constrains it today. Please at least mark it test-only in the README RPC table; an existing mechanism (config switch / build option) would be better.

CubeS3lvol/test/README.md
```diff
@@ -12,14 +12,14 @@ them in this order -- the earlier ones are faster and need less environment.
 ## Running everything with one command
 
 ```sh
-make check           # 23 suites, 928 assertions, about 6 minutes (= test/run_all.sh)
-make check-offline   # only the suites needing no credentials and no root: 11, 491 assertions, about 40 seconds
+make check           # 24 suites (= test/run_all.sh; dataplane needs root + S3)
+make check-offline   # only the suites needing no credentials and no root: 12, 519 assertions, about 40 seconds
```
fslongjin: **[Low] The counts are stale**: the offline set is now 14 suites (11 + the three new unit suites) with roughly 552 assertions; `run_all.sh --list` and the file header comment (lines 6-7) don't list the new suites either. Consider replacing these numbers with "see `run_all.sh --list`" so they can't drift again.

CubeS3lvol/module/bdev/s3lvol/export_lease_term.h
```diff
@@ -0,0 +1,31 @@
...
+export_lease_miss_confirms_gone(bool lease_absent, uint64_t absent_at,
+				uint64_t expires_at, uint64_t now)
+{
+	if (!lease_absent || expires_at == 0) {
+		return false;
+	}
+	if (now < expires_at) {
+		return false;
+	}
+	return absent_at >= expires_at;
```
fslongjin, line 28: This predicate is right, and taking the miss time as the HEAD **submission** moment rather than its completion is the more conservative correct choice (completion lags submission, which would shrink the window around the deadline). Reaper, pin verdict and poller stop all use the same predicate — no path is left out.

CubeS3lvol/lib/s3bsdev/s3_copy_xml.h
```diff
@@ -0,0 +1,101 @@
...
+	if (!xml || !name) {
+		return false;
+	}
+	nlen = strlen(name);
+	if (nlen == 0 || len < nlen + 2) {
+		return false;
+	}
+	/* The character after the name is the delimiter. It must be inside the
+	 * buffer: i + 1 + nlen == len would read xml[len]. */
+	for (i = 0; i + 1 + nlen < len; i++) {
```
fslongjin, line 37: Clean fix on the bound (strict `<`, and the comment spells out the original bug). Two small test gaps to close opportunistically: a tag that is a prefix of the name followed by a non-delimiter character (e.g. `<Errorish>`), and `CopyObjectResult` and `Error` both present (lines 91-98 are currently only tested individually).

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -0,0 +1,1270 @@
...
+
+	if (g_pending_load_hold == PENDING_LOAD_HOLD_HEAD) {
+		pending_load_park(ctx, status, 0, false);
+		return;
+	}
+	pending_load_continue_head(ctx, status);
+}
+
+int
+s3lvol_pending_load_hold(const char *stage)
```
fslongjin, line 1225: The park/hold hooks are a good investment: they let the e2e suite construct "unload while the load is in flight" **deterministically** (step [11] triggers the HEAD and GET stages separately) instead of racing on timing. Together with uuid re-resolution and the client reference, this is a complete fix.

fslongjin (changes requested):
Thanks for this round of fixes. Every issue from my last review got a response, and most were fixed properly **with** deterministic regression tests — the lease `submitted_at` predicate, the extracted-and-tested XML parser, the pending-load park/hold hooks, and the `resolved` bitmap are the right fixes, not review-driven patchwork.

That said, re-reviewing the delta turned up two new HIGH issues **inside the fixes themselves**, plus one issue I should have raised earlier and one residual window. Not mergeable yet:

1. **[High] Non-atomic manifest refcnt** — a side effect of the manifest-snapshot fix: the new per-I/O `ref()` exposes the refcount to the multi-threaded read path for the first time. A lost increment frees the old manifest while reads are still in flight (data-plane UAF). ~2-line fix.
2. **[High] Pending-registry restart self-lock** — in the single-PUT serialization fix: once "a second mutation arrives while a PUT is in flight" happens once, that lvstore permanently stops persisting; every later mark/cancel is memory-only and lost on restart. Reproduced by driving the real header through the production call sequence. ~1-line fix.
3. **[Medium] Ingest slot cross-thread race** — found during my first review round but never posted to the PR until now (apologies for the late addition); unchanged in this push.
4. **[Medium] Unload does not wait for an in-flight registry PUT** — re-attach loads the same blobstore uuid via `spdk_lvs_load_ext` (the normal restart path), so a late PUT can write a cancelled mark back into the registry.
5. **[Medium] Docs over-promise `deferred: true`** — a delete pinned by a lease-less export (export_legacy) actually returns EBUSY, in three documents.

One observation for the next round: both new HIGHs sit at **integration seams of the fixes**, and both were masked by unit tests that exercise the extracted header in isolation rather than the wiring — the persist unit test models the resubmission as `puts++` and never goes through the `begin()` gate. Please pair each fix with a test that follows the production call path.

Once 1-5 are fixed I'm happy to approve — two of them are one-to-two-line changes. The remaining items (lease-key invariants, doc counts, test-only RPC labelling, cleanups) can be follow-ups.

### What's done well

- **Lease predicate**: taking the miss time at the HEAD **submission** moment rather than completion is the conservative, correct choice; the predicate lives in a unit-testable header and is applied uniformly by the reaper, the pin verdict, and the poller stop.
- **XML parsing**: extracted to a header-only module whose tests reproduce the original ASan case exactly (an 8192-byte body ending in `<Error`, byte-by-byte chunked assembly); truncation is now explicit.
- **Pending load**: no more raw `lvs` pointer — re-resolution is by **uuid** (not name), which also fixes "a same-named fresh store getting populated by a late GET"; the client reference keeps CRT alive; park/hold hooks make the unload race deterministic in e2e.
- **`resolved` bitmap**: only marks "allocated but reads as zeroes"; true holes still inherit, preserving REF-layout sparsity; not serialized, not in the CRC.
- **Manifest authority**: the `on_swap` callback lets the control plane follow the data plane and persists the new generation; swap refuses chunk_size changes.
- **get_bdev**: uuid/wwid + major:minor + udev-settled validation fixes the stale-node hazard after a same-nsid reactivate; `wait_ms=0` and the steady-state fast path keep their semantics.

### Checklist for the author

- [ ] Make `s3_export_manifest_ref`/`unref` atomic (~2 lines; the `dev->reads++`-style counters too)
- [ ] Drop the `in_flight` preset in `pending_persist_end`, covering both sync-failure restart paths (~1 line), plus a unit test that `begin()` right after a `true`-returning `end()` submits
- [ ] Bounce `s3_bs_dev_copy`/`s3_bs_dev_ingest_prefetch` to `owner_thread` (or an equivalent serialization)
- [ ] Defer the persist-entry free in `s3lvol_snapshot_pending_clear_lvs` while a PUT is in flight (dead-mark + empty-registry write, or wait for completion)
- [ ] Fix the `deferred` over-promise in README + EN/ZH guides (keep them in sync)
- [ ] (suggested) Delete the stale "lives in memory only" sentence in `vbdev_s3lvol.h`; refresh the test/README and `run_all.sh --list` counts
- [ ] (suggested) Mark `rcow_pending_load_hold` as test-only in the README RPC table
- [ ] (non-blocking) A two-thread concurrent-read + swap test; consider enabling ASan in `make check` by default

---

*Method: static review of `5faecf4d`; the three new unit suites were compiled and run independently under ASan/UBSan (all passing); the persist self-lock was reproduced by driving the real header through the production call sequence; the dataplane suites were not re-run here.*

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -2116,6 +2120,538 @@ s3_bs_dev_destroy(struct spdk_bs_dev *dev)
...
+ingest_prefetch_on_owner(void *arg)
+{
+	struct ingest_prefetch_msg *m = arg;
+
+	ingest_prefetch_work(m->ctx, m->src_lba, m->lba_count);
+	free(m);
+}
+
+void
+s3_bs_dev_ingest_prefetch(struct spdk_bs_dev *bs_dev, uint64_t src_lba,
```
fslongjin, line 2607: **[Low · non-blocking] One latent gap in the bounce: the prefetch message holds a raw `ctx` with no inflight accounting.**

The `ingest_prefetch_msg` holds a raw `ctx` across the hop, and nothing counts it in `ctx->inflight`. If a future caller invoked prefetch off the owner thread and the device were destroyed between the `send_msg` and the callback, this would be a use-after-free. Today the only caller (`decouple_prefetch`) runs on the owner thread, so the bounce branch is unreachable and this is theoretical. Suggest either counting the message in `ctx->inflight` (dropping it in the callback) or a comment/assert stating the caller-side constraint. (The copy bounce doesn't have this problem — the outstanding blobstore I/O keeps the device alive across the hop.)

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -0,0 +1,1324 @@
...
+
+		TAILQ_FOREACH_SAFE(p, &g_pending_persist, link, ptmp) {
+			if (spdk_uuid_compare(&p->lvs_uuid, lvs_uuid) != 0) {
+				continue;
+			}
+			if (p->ctl.in_flight || p->ctl.dirty) {
+				/* A PUT is still using this entry. Keep it so the
+				 * completion cannot run against a later attach, and
+				 * ask for one more write of the now-empty queue. */
+				p->dying = true;
```
fslongjin, line 350: **[Low · non-blocking] A narrow window in the dying design, fine to leave but worth writing down.**

If the in-flight PUT completes after a same-uuid re-attach but before the new load's GET has landed, this follow-up writes an *empty* registry while the GET then restores the old marks into memory — memory and S3 diverge until the next mutation, and a crash inside that window loses the restored intents (conservative direction: a pending delete is forgotten, not resurrected).

Also noting that the `dying`/bind/drop wiring here has no test coverage — the ctl unit test can't reach it. An e2e (unload with a PUT in flight → re-attach → assert the registry ends up empty) would lock the whole design in.

fslongjin (approved):
All five blockers from my last review are fixed in this push, and I verified each fix rather than taking the diff at face value: the refcnt atomics (correct memory orders, free gated on the pre-decrement value), the persist self-lock (preset dropped; the unit test now follows the production sequence — 18 assertions green under ASan/UBSan here), the ingest owner-thread bounce (every slot-mutation path traced: copy entry, prefetch, CopyObject completions, journal bind completions — all owner-confined now), the unload-vs-in-flight-PUT window (the `dying` entry survives unload and re-attach, which removes the same-uuid cross-talk at the root), and the `deferred: true` docs (all three, EN/ZH in sync).

Approving. What follows is everything I'd still want addressed — all non-blocking.

**New observations from re-reviewing the fixes (details inline):**

1. The prefetch bounce message holds a raw `ctx` with no inflight accounting. Unreachable today (the only caller runs on the owner thread), but it's a latent UAF if prefetch is ever called off-owner.
2. A narrow window in the dying design: an in-flight PUT completing after a same-uuid re-attach but before the new load's GET lands writes an empty registry while the load then restores the old marks into memory. Self-heals on the next mutation; a crash inside the window loses restored *delete intents* only (the conservative direction).

**Carried-over items, still open (all fine as follow-ups):**

- The lease-key invariant comment on `import_on_swap` (round 3): lease keys are only derived once from `imp->m` and not recomputed on swap. Safe today because both generation-bump paths produce a subset of the old source set, but that invariant is unwritten.
- The SPDK patch nits: `spdk_blob_allow_esnap_copy()` still only `assert`s non-NULL (release build would write through NULL), and `esnap_copy_refcnt` is a 0/1 flag, not a refcount.
- Dead code from the F6 rework: `dev->chunk_shift` / `dev->blocks_per_chunk` are now write-only (reads derive geometry from the per-I/O manifest), `s3_copy_xml_status()`'s `truncated` parameter is unused, and the `io->m ? ... : dev->m->...` fallback in `export_io_complete` is unreachable.

**Test asks:**

- The `dying`/bind/drop wiring in `vbdev_s3lvol_pending.c` has no coverage — the ctl unit test can't reach it. An e2e (unload with a PUT in flight → re-attach → assert the registry ends up empty) would lock the design in.
- The two-thread concurrent read + swap test from round 3 is still open.

---

*Method: static review of `9f7ef711`; the two modified unit suites were compiled and run independently under ASan/UBSan (all passing); the dying/self-lock state machines were traced path by path; the dataplane suites were not re-run here.*

fslongjin (approved):

PR #1664, opened 2026-09-07
refactor: move CubeDB module into pkgs/cubedb

fslongjin:
## Summary

Relocates the shared database module from `CubeDB/` to `pkgs/cubedb/` and
renames its Go module path to `github.com/tencentcloud/CubeSandbox/pkgs/cubedb`,
bringing it in line with `pkgs/CubeLog` and `pkgs/proto`.

## Changes

- **Module relocation**: `CubeDB/` → `pkgs/cubedb/`, with the module path
  updated accordingly; all staged migration files move with identical content.
- **Consumers**: CubeMaster and CubeOps imports now use the new path, and their
  local `replace` directives point at `../pkgs/cubedb`.
- **Tooling / CI**: the `Makefile` `cubedb-test` target and the path filters in
  the `build-check`, `unit-test-check`, and `migration-check` workflows are
  updated.
- **Migration immutability guard**: `migration-check` explicitly permits this
  same-content relocation (which appears as rename/delete pairs) while still
  rejecting any edit or delete of an already-applied migration.
- **Docs & ownership**: CODEOWNERS, the image build script, and affected docs
  are refreshed to the new location.

## Validation

- The move is detected as pure file renames; no applied migration content
  changed.
- No added lines introduce internal endpoints, IPs, or local paths.


ls-ggg:
LGTM

PR #1667, opened 2026-09-07
feat(clm): add active-standby failover

FakeLearne:
Run two warm CLM replicas in Kubernetes and use a Redis lease to gate singleton sweep and prune work while both replicas serve resume requests.

Use broadcast XREAD consumption, promotion catch-up, fencing epochs, and versioned state CAS to prevent stale replicas from overwriting newer state. Add Helm configuration, readiness observability, Redis transaction tests, and bilingual documentation. Keep one-click deployment single-replica.


cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -255,11 +367,166 @@ func replayRegistryTo(ctx context.Context, push *proxypush.Client,
...
+	defer eventApplyMu.Unlock()
+	if !lease.IsLeader() || generation != lease.Generation() {
+		return false
+	}
+	target, err := stream.LatestID(ctx)
+	if err != nil {
+		log.Warn("promotion high-water read failed", zap.Error(err))
+		return false
+	}
+	if err := catchUpStreamTo(ctx, target, stream, push, reg, ssDeps, progress, log); err != nil {
```
fslongjin, line 490: **Blocking (P0)** — `catchUpGeneration` passes the **same** `streamProgress` used by `consumeStream` into `catchUpStreamTo`. When the warm standby has already advanced the cursor to the stream high-water mark, the `cmp >= 0` check below makes promotion catch-up a no-op.

This conflicts with the design where standbys only warm-cache and external writes are replayed on promotion: if the current leader crashes before `WriteStateCAS` / `DeleteMeta`, the new leader cannot replay those events.

**Suggestion:** Maintain a separate `leaderAppliedCursor`, or on promotion force-flush from registry state + pending deletes — do not reuse the hot consumption cursor.

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -255,11 +367,166 @@ func replayRegistryTo(ctx context.Context, push *proxypush.Client,
...
+	push *proxypush.Client, reg *registry.Registry, ssDeps statesync.Deps,
+	progress *streamProgress, log *zap.Logger) error {
+
+	for {
+		cursor := progress.Cursor()
+		cmp, err := redisstream.CompareStreamIDs(cursor, target)
+		if err != nil {
+			return err
+		}
+		if cmp >= 0 {
```
fslongjin, line 595: **Blocking (P0)** — Returning immediately when `cursor >= target` means no `handleEvent` loop runs in the common failover case where the standby is already aligned with the stream. This is the direct trigger for P0-1.

Please add a test: `progress` already at `target`, registry reflects delete/state, but Redis/Proxy are stale → promotion must still push.

cube-lifecycle-manager/internal/statesync/handler.go
```diff
@@ -99,6 +120,19 @@ func Handle(ctx context.Context, d Deps, ev redisstream.Event) {
 		return
 	}
 
+	recordWarmState := func() {
+		if newState == lifecycle.StateRunning {
+			d.Registry.MergeLastActive(ev.SandboxID, now().UnixMilli())
+		}
+		d.Registry.SetRuntimeState(ev.SandboxID, newState)
+	}
+	if !persistEnabled(d) {
```
fslongjin: **Blocking (P0)** — Standby path only calls `recordWarmState()` and returns: no Redis write, no CubeProxy push. If the old leader dies before completing external writes, standby memory is updated but the outside world is not; promotion catch-up then skips replay because of the shared cursor.

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -255,11 +367,166 @@ func replayRegistryTo(ctx context.Context, push *proxypush.Client,
...
+	}
+	return true
+}
+
+func resolvePromotionState(
+	ctx context.Context, stream *redisstream.Client, entry registry.Entry,
+) (string, error) {
+	// After promotion catch-up the lease holder has CAS'd terminal state
+	// events into this key, so it is at least as fresh as RuntimeState.
+	state, ok, err := stream.GetState(ctx, entry.Meta.SandboxID)
```
fslongjin: **Blocking (P0)** — `resolvePromotionState` always returns `GetState` when the Redis key exists, ignoring potentially fresher `entry.RuntimeState`.

In the P0-1 crash window: standby has `SetRuntimeState(paused)` but Redis may still hold a TTL'd `running` → `hydrateFleet` pushes `SetState("running")` fleet-wide and CubeProxy stops auto-resume.

**Suggestion:** Hydration should reconcile registry vs Redis (e.g. trust registry when fresher), or only trust Redis after catch-up actually CAS'd the key.

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -231,12 +345,10 @@ func (s fleetSizer) Snapshot() int {
...
-// by the static-fleet initialization path. Errors are logged but not
-// escalated: reconciliation eventually converges via the stream consumer.
-func replayRegistryTo(ctx context.Context, push *proxypush.Client,
-	reg *registry.Registry, ep discovery.Endpoint, log *zap.Logger) {
+// replayRegistryTo pushes metadata and terminal state to a single admin
+// endpoint. Used by discovery.OnJoin and best-effort fleet hydration.
+func replayRegistryTo(ctx context.Context, push *proxypush.Client, stream *redisstream.Client,
+	reg *registry.Registry, ep discovery.Endpoint, log *zap.Logger) bool {
 
 	entries := reg.Snapshot()
```
fslongjin, line 365: **Blocking (P0)** — `replayRegistryTo` / `hydrateFleet` only walks entries still in `reg.Snapshot()`. Delete events remove sandboxes from the registry, so CubeProxy `DeleteMeta` **must** come from catch-up — which is a no-op on an already-aligned standby (see `main_test.go:166-170`).

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main_test.go
```diff
@@ -0,0 +1,554 @@
...
+
+func (standbyStatus) IsLeader() bool { return false }
+func (standbyStatus) Enabled() bool  { return true }
+
+type persistStatus struct{}
+
+func (persistStatus) IsLeader() bool { return true }
+func (persistStatus) Enabled() bool  { return true }
+
+func TestCatchUpStreamToDrainsPromotionHighWater(t *testing.T) {
```
fslongjin, line 34: **Blocking (P0) — test gap** — `TestCatchUpStreamToDrainsPromotionHighWater` starts from `newStreamProgress("1-0")`, covering only a lagging cursor. Missing: progress already at high-water, standby registry updated, Redis/Proxy not synced → promotion must still flush external state.

cube-lifecycle-manager/internal/redisstream/stream.go
```diff
@@ -210,13 +416,89 @@ func (c *Client) GetState(ctx context.Context, sandboxID string) (string, bool,
...
+			c.publishNotify(sandboxID)
+		}
+		return updated, nil
+	}
+	return false, fmt.Errorf("write state cas %s: transaction conflicted after %d attempts", key, maxAttempts)
+}
+
+const stateVersionPrefix = "v1|"
+
+func encodeStateValue(state, version string) string {
```
fslongjin: **Blocking (P0)** — `encodeStateValue` writes `v1|<streamID>|<state>`. On `master`, old `GetState` does **not** decode; old `classifyState` / `acquireResumeOwnership` only accept exact `"paused"`/`"running"`.

With Helm default RollingUpdate overlap, resume on an old Pod waits on versioned values → **503**. This triggers once any `OpState` is written via `WriteStateCAS` during the upgrade window.

**Suggestion:** Keep plain-text state values (version elsewhere), **or** switch CLM Deployment to Recreate and document the upgrade path.

deploy/kubernetes/chart/README.md
```diff
@@ -438,6 +438,19 @@ The default TLS mode is `selfSigned`, matching the one-click mkcert-style test e
 
 `cube-proxy` depends on chart-managed `cube-lifecycle-manager` for sandbox auto-pause / auto-resume. The chart wires nginx `$cube_sidecar_addr` to the lifecycle-manager Service, opens the proxy admin listener for in-cluster discovery, and registers each proxy replica in Redis. Do not deploy a separate cube-proxy-sidecar.
 
+The lifecycle manager runs two warm replicas by default. Both replicas consume
```
fslongjin, line 441: **Blocking (P0) — upgrade path** — This HA section is accurate for steady state, but the CLM Deployment still uses the chart default `controlPlane.deploymentStrategy` (RollingUpdate, `maxSurge: 1`). Upgrading from a single old CLM image with `replicas` 1→2 leaves old and new Pods on the Service together while new code writes versioned Redis state keys old binaries cannot parse → resume **503** on old Pods.

**Suggestion:** Document Recreate / scale-to-0 upgrade, or override lifecycle-manager strategy; pair with plain-text state values or a single-version rollout.

cube-lifecycle-manager/internal/httpapi/server.go
```diff
@@ -144,8 +153,16 @@ func (s *Server) handleHealthz(w http.ResponseWriter, _ *http.Request) {
 func (s *Server) handleReadyz(w http.ResponseWriter, _ *http.Request) {
 	w.Header().Set("Content-Type", "application/json")
 	resp := map[string]any{
```
fslongjin, line 155: **Blocking (P0) — related** — `/readyz` returns `"ok": true` for standbys (intentional, so the Service keeps warm standbys). During RollingUpdate overlap with old binaries, **old Pods are also Ready**, so kube-proxy may route resume traffic to processes that cannot parse versioned state.

If RollingUpdate is kept, you need another isolation mechanism or accept Recreate.

deploy/kubernetes/chart/templates/validate.yaml
```diff
@@ -198,6 +198,15 @@
 {{- fail "cubeProxy.enabled=true requires lifecycleManager.enabled=true (cube-lifecycle-manager replaces the obsolete cube-proxy-sidecar)" }}
 {{- end }}
 {{- if eq (include "cube.lifecycleManagerEnabled" .) "true" }}
+{{- $leaderElection := .Values.lifecycleManager.leaderElection | default dict -}}
+{{- if $leaderElection.enabled }}
+{{- if lt (int .Values.lifecycleManager.replicas) 2 }}
```
fslongjin: **Should fix (upgrade path)** — Only validates `leaderElection.enabled=true ⇒ replicas>=2`, which blocks the safer path of upgrading the image at `replicas=1` then scaling to 2.

**Suggestion:** Allow `replicas=1 && election=true`; document the upgrade sequence.

deploy/kubernetes/chart/templates/validate.yaml
```diff
@@ -198,6 +198,15 @@
...
+{{- fail "lifecycleManager.leaderElection.enabled=true requires lifecycleManager.replicas>=2" }}
+{{- end }}
+{{- if or (empty $leaderElection.leaseTTL) (empty $leaderElection.renewInterval) (empty $leaderElection.retryInterval) }}
+{{- fail "lifecycleManager.leaderElection.enabled=true requires leaseTTL, renewInterval, and retryInterval" }}
+{{- end }}
+{{- end }}
```
fslongjin, line 227: **Should fix (P1)** — Missing reverse check: when `replicas > 1`, `leaderElection.enabled` must be `true`. With election off, `leader.New` makes every replica `IsLeader()==true` → dual sweep / dual prune.

deploy/kubernetes/chart/values.yaml
```diff
@@ -411,7 +411,15 @@ lifecycleManager:
   # is enabled; CubeProxy points $cube_sidecar_addr at this Service and
   # registers itself in Redis for CLM discovery.
   enabled: true
-  replicas: 1
+  # Two replicas run active-standby when leader election is enabled. The
+  # preferred anti-affinity in the Deployment spreads them across nodes when
+  # the cluster has enough eligible control-plane nodes.
+  replicas: 2
```
fslongjin, line 417: **Nit** — Default `replicas: 2` + `leaderElection.enabled: true`. Upgrades that do not pin `replicas: 1` silently become dual-replica HA. Please document in upgrade docs how to disable HA (`replicas: 1`, `leaderElection.enabled: false`) and the safe upgrade sequence.

docs/guide/lifecycle.md
```diff
@@ -220,7 +220,7 @@ The repository ships with **no cluster-wide idle timeout** (`default_timeout_ins
...
 
 - **Pause fidelity**: CPU registers, process memory, TCP state (with no external peer), and filesystem mutations all survive the snapshot. Outbound sockets the sandbox itself opened are dropped on pause and must be reopened by the application after resume.
-- **Cluster coordination**: auto-pause is driven by the `cube-lifecycle-manager` service that runs on the control node. It consumes lifecycle events CubeMaster publishes via Redis stream, discovers every live CubeProxy replica through a Redis-backed registration table, and broadcasts state to each of them. Cross-replica races are resolved by Redis `SETNX` state locks so the same sandbox is never paused or resumed twice concurrently.
+- **Cluster coordination**: auto-pause is driven by `cube-lifecycle-manager`. Kubernetes runs two warm replicas: both consume lifecycle events, discover CubeProxy replicas, and serve resume callbacks, while a Redis lease elects one replica for idle sweep/kill and stale-registry pruning. Per-sandbox Redis state transitions and CubeMaster lifecycle locks serialize effective pause/resume work across replicas.
```
fslongjin: **Should fix (docs)** — "Kubernetes runs two warm replicas" reads as all Kubernetes deployments. Only the **Helm chart** defaults to 2; Terraform / one-click remain single-replica. The Chinese text correctly says "默认" — please align English and distinguish deployment paths.

docs/guide/lifecycle.md
```diff
@@ -220,7 +220,7 @@ The repository ships with **no cluster-wide idle timeout** (`default_timeout_ins
...
 - **Failure mode**: when an auto-resume RPC fails, CubeProxy returns `503 + Retry-After` to the client immediately rather than hanging on a long timeout. When the sandbox has already been killed (`killing` / `killed`) the proxy returns `410 Gone` instead, telling SDK clients to stop retrying.
 - **Diagnostics**: `docker logs cube-lifecycle-manager` (control node) is the runtime log for the auto-pause coordinator. Look for `create event applied`, `auto-paused sandbox`, `auto-resumed sandbox`, `timeout-killed sandbox`. Each CubeProxy replica additionally exposes `GET http://<node-ip>:8082/admin/healthz` reporting `heartbeat_last_pushed_ms` (the last time it announced itself to the manager). The admin port defaults to `8082`; override it with `CUBE_PROXY_ADMIN_PORT` when that port is already in use on the host (CubeProxy uses host networking).
```
fslongjin, line 225: **Should fix (docs)** — Troubleshooting uses `docker logs` and CubeProxy `node-ip:8082` (host networking). For Helm, use `kubectl logs` on the Deployment; CubeProxy admin is ClusterIP:8082 in-cluster.

cube-lifecycle-manager/internal/sweeper/sweeper.go
```diff
@@ -269,6 +322,7 @@ func (s *Sweeper) tryPause(ctx context.Context, e registry.Entry) error {
 		s.o.Log.Warn("write paused state failed",
 			zap.String("sandbox_id", sid), zap.Error(err))
 	}
+	s.o.Registry.SetRuntimeState(sid, lifecycle.StatePaused)
 	if err := s.o.ProxyPush.SetState(ctx, sid, "paused"); err != nil {
```
fslongjin, line 326: **Suggestion** — After `AcquireState("pausing")`, `tryPause` no longer checks `IsLeader()`; terminal `WriteState("paused")` + `ProxyPush.SetState` have no lease-generation fence. SETNX makes Redis "overwrite running" unlikely, but Proxy last-writer can still clobber `running` during failover.

**Suggestion:** Use expected-state CAS or check `reconciledLeader` generation on terminal writes.

cube-lifecycle-manager/internal/redisstream/stream.go
```diff
@@ -101,6 +107,90 @@ func (c *Client) EnsureGroup(ctx context.Context, group string) error {
 	return fmt.Errorf("xgroup create mkstream: %w", err)
```
fslongjin: **Suggestion** — `EnsureGroup` / `ReadGroup` / `Ack` are no longer called from `main.go` (switched to broadcast XREAD) but remain in the API. `Config.ConsumerGroup` comments still describe a PEL upgrade path — misleading.

**Suggestion:** Remove dead code or mark deprecated and fix comments.

cube-lifecycle-manager/internal/redisstream/stream.go
```diff
@@ -157,6 +247,47 @@ func (c *Client) ReadGroup(ctx context.Context, group, consumer string, block ti
 	return out, nil
 }
 
+// Read broadcasts lifecycle events to one CLM replica using a caller-owned
+// cursor. Unlike XREADGROUP, every replica receives every event and can keep
+// its in-memory registry warm. The returned cursor advances over malformed
+// entries too, preventing one bad message from wedging the loop.
+func (c *Client) Read(ctx context.Context, cursor string, block time.Duration, count int) ([]Event, string, error) {
```
fslongjin, line 197: **Suggestion (perf)** — Every `Read()` calls `CursorValid` (extra Redis RTT). On idle polls each replica pays 2 commands per block period. Consider checking only on empty batches or periodically.

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go
```diff
@@ -297,36 +562,117 @@ func bootstrapRegistry(ctx context.Context, stream *redisstream.Client,
...
-			case <-time.After(time.Second):
 			}
 			continue
 		}
 		for _, ev := range events {
-			handleEvent(ctx, ev, push, reg, ssDeps, log)
-			if err := stream.Ack(ctx, cfg.ConsumerGroup, ev.StreamID); err != nil {
-				log.Warn("ack failed",
-					zap.String("id", ev.StreamID), zap.Error(err))
+			eventApplyMu.Lock()
```
fslongjin, line 687: **Suggestion** — `eventApplyMu` is held across `handleEvent` (including CubeProxy HTTP) and shared with `catchUpGeneration`. A slow Proxy can block promotion catch-up and subsequent XREAD application.

**Suggestion:** Hold the lock only for registry + cursor; move HTTP I/O outside the lock.

cube-lifecycle-manager/cmd/cube-lifecycle-manager/main_test.go
```diff
@@ -0,0 +1,554 @@
...
+	}))
+	t.Cleanup(ts.Close)
+	return ts.URL, func() []string {
+		mu.Lock()
+		defer mu.Unlock()
+		return append([]string(nil), seen...)
+	}
+}
+
+// A delete event always drops the sandbox from the local registry, but only the
```
fslongjin: **Suggestion** — This comment correctly states deletes can only be pushed during catch-up and hydration cannot replay them — which is exactly why P0-1 is severe. Consider linking to a promotion test once fixed, or updating the invariant after the fix.

fslongjin (changes requested):
## Verdict: Request changes

Thanks for the CLM active-standby work — Redis lease + broadcast XREAD + warm standbys serving resume is a sound direction. The lease implementation, `WriteStateCAS`, standby `/readyz` staying ready, and keeping one-click single-replica are all reasonable. Unit tests on the lease/CAS/promotion paths (miniredis) are solid.

I cannot approve yet. Two production-facing issues need to be fixed first:

### Blocking

**1. Promotion catch-up shares the hot XREAD cursor (P0)**

`consumeStream` and `catchUpStreamTo` share the same `streamProgress`. A warm standby that already caught up to the stream high-water mark makes promotion catch-up a no-op (`cursor >= target`). Meanwhile, standbys apply `OpState` / `OpDelete` only in memory (`statesync.Handle` returns early when `!persistEnabled`) — no Redis write, no CubeProxy push.

The design relies on promotion catch-up to replay CAS writes and delete pushes (see `main_test.go` comment), but in the **most common HA topology** (both replicas already aligned on the stream) that path never runs. If the old leader crashes before `WriteStateCAS` / `DeleteMeta`:

- **deletes** cannot be replayed via `hydrateFleet` (it only iterates entries still in the registry);
- **pause/state** may be overwritten by `resolvePromotionState` preferring a stale Redis `running` over local `RuntimeState=paused`, so CubeProxy stops auto-resume.

Existing tests start from a lagging cursor (`0-0` / `1-0`); there is no "already-aligned standby promotion" case.

**Suggestion:** Use a separate leader-applied cursor, or on promotion force-flush terminal state / deletes the standby has already seen into Redis and CubeProxy; add regression tests.

---

**2. Rolling upgrade: RollingUpdate + versioned state breaks old CLM resume (P0)**

New code writes state keys as `v1|<streamID>|<state>` (`encodeStateValue`). On `master`, old `GetState` returns the raw Redis string and old `classifyState` only matches exact `"paused"` / `"running"`. Once any sandbox is written with `WriteStateCAS`, resume hitting an **old Pod** spins until timeout → CubeProxy **503**.

The CLM Deployment uses the chart default `controlPlane.deploymentStrategy` (RollingUpdate, `maxUnavailable: 0`, `maxSurge: 1`). When `replicas` goes 1→2 with an image change, old Pods (no election code) and new Pods are both in Service endpoints; `/readyz` returns `ok: true` for standbys too, so traffic is not isolated from old binaries. Old and new Pods share the same Redis DB.

**Suggestion (either one blocks resume-503):**

- Use **Recreate** for this CLM upgrade (or document scaling to 0 first); **or**
- Keep writing plain `paused`/`running` in Redis and store version elsewhere so old readers can still string-compare.

Also consider allowing `replicas=1 && leaderElection.enabled=true` as an intermediate upgrade step (current `validate.yaml` fails that).

---

### Should fix (not blocking alone)

- **P1 — dual-leader misconfig:** Helm does not reject `replicas>=2` with `leaderElection.enabled=false`. Terraform TKE (out of this PR) does not inject `CUBE_LCM_LEADER_*` — if operators set `replicas=2` there without election env, they get dual sweepers.
- **Docs:** Official `docs/guide/kubernetes/*` omits HA (not in this PR); `lifecycle.md` implies all Kubernetes deployments run two warm replicas (should say Helm chart default). Missing upgrade/disable-HA values examples.
- **Tests:** Missing already-aligned standby promotion, dual-client broadcast XREAD, discovery prune gating, Helm HA guard script.
- **Follow-up:** Stale leader terminal writes without generation fencing; remove dead `XREADGROUP` code; do not hold `eventApplyMu` across CubeProxy HTTP.

### What looks good

- Redis lease (token-safe renew/release, local deadline)
- Standbys keep serving resume; `/readyz` does not fail on standby
- One-click keeps election disabled; single-container behavior preserved
- `proxy_registry.lua` re-HSET every heartbeat is the correct fix for cross-slot prune
- Chart README is honest about single-replica Redis SPOF and lack of strict fencing on async replication

---

**Minimum to merge:** Fix P0-1 (cursor decoupling or forced flush + tests) + P0-2 (Recreate or plain-text state + upgrade docs). Happy to re-review after that.

fslongjin (approved):

PR #1674, opened 2026-09-08
fix(k8s): derive the sandbox guest MTU from the node uplink

### LINKED ISSUE #1673: [Bug Report] Sandbox guest MTU is fixed at 1500, stalling traffic on any overlay CNI
### Summary
The sandbox guest is given MTU 1500 regardless of the node uplink, so on a cluster whose Pod network
carries less (Flannel VXLAN 1450) the guest emits frames the uplink drops and connections stall
instead of failing.

### Environment
- CubeSandbox version / commit: measured on a v0.6.0-based build; the same packaged default is
  present on master @ `d38cb6e` (`Cubelet/config/config.toml:87`, `mvm_mtu = 1500`)
- Host OS and kernel version: Ubuntu 24.04.4 LTS, `Linux 6.8.0-136-generic #136-Ubuntu SMP
  PREEMPT_DYNAMIC x86_64` (k3s v1.35.1+k3s1)
- KVM info (`modinfo kvm`): in-tree `kvm` and `kvm_intel` for 6.8.0-136-generic (Intel VT-x); built
  with the kernel, so no `version` field is exported
- Deployment mode: cluster (Helm chart, k3s)
- Relevant component: CubeNet / Cubelet network runtime

### Steps to Reproduce
1. Install the Helm chart on a stock k3s cluster (Flannel, default VXLAN backend), no custom values.
2. Create a sandbox and exec into it.
3. `cat /sys/class/net/eth0/mtu` -> `1500`, while the cube-node Pod's eth0 is `1450`.
4. From the sandbox, request more than the path MTU from a host that does not clamp its MSS, e.g.
   `curl -sS https://google.com/ -o /dev/null` or any POST larger than ~1.2 KB.

### Expected Behavior
The guest MTU matches what the node uplink can carry, or the traffic fails fast with an error.

### Actual Behavior
The request hangs until it times out, with nothing logged in the guest, in cubelet, or on the node.
TLS handshakes to hosts whose certificate flight exceeds one segment never complete:
`_ssl.c:993: The handshake operation timed out`.

### Additional Context
The outcome tracks the peer's advertised MSS, not the port or the host. Measured on one host, one
port, one IP (`api.telegram.org:443`, anycast, which hands out both a fitting and an overflowing MSS
across repeated connections):

| wire frame (MSS + 12 timestamps + 40 header) | small request | large request |
|---|---|---|
| 1280 (fits the 1450 uplink) | 5/5 ok | 4/4 ok |
| 1500 (overflows) | 2/2 ok | **0/3 ok** |

Setting `TCP_MAXSEG` to 1360 on the socket before `connect()` makes every failing case pass, and
turns 20 s timeouts into 63-105 ms:

| target | unclamped | MSS clamped to 1360 |
|---|---|---|
| `google.com:443` | 20 s timeout | ok, 76 ms |
| `smtp.gmail.com:587` STARTTLS | 20 s timeout | ok, 105 ms |
| `imap.gmail.com:993` | 20 s timeout | ok, 96 ms |
| `1.1.1.1:853` | ok, 3215 ms | ok, 63 ms |

Why it cannot self-heal: there is no MSS clamp in the datapath, the egress path drops fragments, and
inbound ICMP handling matches only Echo Reply, so a fragmentation-needed message has no path back
into the guest.

Note `docs/guide/kubernetes/index.md` lists Flannel as the CNI for two of the three tested
environments; TKE's VPC-CNI carries 1500, which is presumably why this is not seen there.

Scope: this report and the patch cover the Kubernetes deployment path. The one-click/systemd
installer stages the same `config.toml` and is affected in the same way, but is not addressed here.

Related: #1270 reported the same defect and proposed exposing the knob. #1151 asks for post-install
functional coverage naming CNI variance and HTTPS egress, which is the case that would catch this.

Happy to send the patch: I have one ready that resolves `mvm_mtu` from the interface the node
entrypoint already auto-detects for `eth_name`, confined to `deploy/kubernetes/` with no Cubelet or
hypervisor changes, defaulting to a value that only ever lowers the packaged 1500.

rogelioRuiz:
Fixes #1673.

Same defect reported in #1270 (thanks @try-agaaain, whose analysis of the
tap -> VIRTIO_NET_F_MTU -> guest path this builds on). That PR exposed `mvm_mtu` as a chart knob and
was approved; it was closed unmerged after going stale on a rebase request. This keeps the knob and
additionally makes the default correct, so a stock install on an overlay CNI is not broken out of
the box.

The change is confined to the chart and the node entrypoint. It does not touch Cubelet or the
hypervisor, and adds no work to sandbox creation: the value is resolved once, at container start,
from the interface the entrypoint already auto-detects for `eth_name`.

`cube-node-entrypoint.sh` is deliberately left alone: the chart's `cubelet` container runs
`component-entrypoint.sh` (`CUBE_COMPONENT=cubelet`, `CUBE_ROLE=run`), and the `cube-node` image is a
separate delivery path the chart does not use. Happy to mirror the patch there if you would prefer
parity across both entrypoints.

`cubeNode.network.mtu` defaults to `auto`, which only ever lowers the packaged value. On any node
whose uplink is already >= the packaged 1500 the rendered config is byte-identical to today, so
TKE/VPC-CNI and bare-metal deployments see no behaviour change at all. An integer in `1280`..`65535`
pins `mvm_mtu`, and `0` keeps the packaged value.

Input handling follows the review findings on #1270: leading zeros are stripped before any arithmetic
(bash reads `010000` as octal, and TOML rejects a leading zero, so writing one back yields a
`config.toml` cubelet cannot parse), the digit count is gated before `(( ))` sees the value so a
`2^64 + n` input cannot wrap into the accepted range, and the write-back `sed` is anchored so a
commented example line is never rewritten. Pinning above the detected uplink is honoured but warned
about, since that silently recreates the blackhole the default exists to prevent.

It also closes the two findings left open on that thread. The write-back only rewrites a bare decimal
literal, because TOML accepts `0x5DC` and `1_500` too and the digit match would match on their
leading digits alone, splicing the remainder onto the new value; and a key that is missing, commented
out, or not a plain decimal is reported as skipped rather than logged as patched, so a no-op cannot
read as success. The write-back is a single `patch_mvm_mtu` helper the test sources directly, so the
checks cannot drift from the shipped expression.

If you would rather not change the default, flipping `mtu: "auto"` to `mtu: 0` in `values.yaml` is a
one-line change and preserves #1270's approved semantics. I chose `auto` because two of the three
environments in `docs/guide/kubernetes/index.md` ship Flannel at 1450, where the current default is
broken with no diagnostic — a knob only helps an operator who has already diagnosed it, and this
failure presents as a hang with nothing logged anywhere.

### Deployment note (belongs in the PR body)
Sandboxes are restored from a per-template VM memory snapshot, and the guest applies its interface
MTU in the agent's `create_sandbox` path, which a restore does not run. So a guest's MTU is whatever
it was when its template was built: **existing templates keep the old value and must be rebuilt for
this change to reach guests.** Verified on a live cluster — after the fix the tap, `cube-dev` and
`mvm_mtu` were all 1450 while a guest restored from a template built in July still reported 1500.
The intermediate state is safe (no VM died, no tap-write EIO), it is simply a no-op until the
template is rebuilt.

### Test plan
- `bash deploy/kubernetes/images/scripts/test-mvm-mtu-resolve.sh` — 40 checks: auto lowers, auto
  never raises, auto is a no-op at parity, the 1280 floor and the 65535 ceiling, leading-zero
  normalization, the `2^64 + n` wrap vector, each rejection asserted against the guard that produced
  it, the commented-example line left untouched, the resolved value written back into `config.toml`
  without touching neighbouring keys, and the values the rewrite refuses to touch (`0x5DC`, `1_500`,
  `0o2734`, `1.5e3`, and a config with no live `mvm_mtu` key) left byte-identical and never reported
  as patched.
- The four existing `deploy/kubernetes/images/scripts` tests still pass.
- `helm lint`, and `helm template` with `mtu` unset / `auto` / `1450` / `0` / `65535` / `null`.
- Live cluster (k3s + Flannel, Pod uplink 1450), after deploying the patched image and rebuilding
  the templates: guest `eth0` 1500 -> **1450**; the worst frame induced by an external peer 1500 ->
  **1450**; `google.com:443`, `smtp.gmail.com:587` STARTTLS, `:465` and `imap.gmail.com:993` all went
  from 20 s timeouts to completing; across 16 hosts the platform calls, TLS handshake failures went
  2/16 -> **0/16** and ~1.5 KB request failures 8/16 -> **0/16**. No VM died and no tap-write EIO was
  observed at any point.


rogelioRuiz:
Thanks — both inline comments are addressed in `e19f975`, along with two of the minor notes.

**Fail-open swallowing an explicit pin.** You are right that this reproduces the failure the PR
removes, so the two cases are now separated. A derived value stays fail-open, matching the
`eth_name`/`cidr` patches beside it, but a pinned value on a config without a plain-decimal
`mvm_mtu` key now fails the container: the operator named an MTU and would otherwise run with the
packaged one while believing it was set.

This is a deliberate difference from #1270, which downgraded the missing-key case to warn-and-skip
for all values. That is still right for the derived default — an unexpected config should not
crash-loop a node over a value nobody asked for — but a pin is an explicit request, and silently not
honouring it is the mismatch this change exists to remove. Happy to make it uniform in either
direction if you would rather it match #1270 exactly.

**Double error on invalid input.** Dropped the wrapper; `resolve_mvm_mtu` already reports the
specific reason, so only that line is printed now.

**Also applied:** leading zeros are stripped on the `packaged_mtu` read as well, since that value is
compared with `(( ))` where `01500` would read as octal; and the doc comment no longer claims the
"empty keeps the packaged value" path is reachable from the entrypoint, which defaults to `auto`.

**Left alone deliberately:** chart-level validation for `cubeNode.network.mtu`, to stay consistent
with `ethName`/`cidr`/`adminPort`, which also validate at container start. Glad to add it to
`validate.yaml` if you would prefer earlier feedback.

The suite is 51 checks; the new ones cover the pin/derive split and the setting-to-flag mapping at
the call site, since the hard-fail would otherwise be unreachable in the entrypoint even with the
function correct.

chenhengqi (commented):
Looks good overall.

fslongjin (approved):

PR #1675, opened 2026-09-08
fix(cubelet): decouple template artifact storage from kernel location

zyl1121:
## Motivation

Each template uses a read-only ext4 root filesystem artifact built from a Docker/OCI image. Nodes cache these large files locally, so disk usage grows as more templates are cached. By default, the cache lives under `/usr/local/services/cubetoolbox`, typically on the system disk.

Currently, the cache directory and shared `vmlinux` source path are derived from the same `cubetool_base_dir` setting. Changing that setting to place artifacts on `/data` also makes Cubelet look for the shared kernel under `/data`, even though it remains installed under `/usr/local/services/cubetoolbox`. This coupling prevents operators from configuring the artifact cache and shared kernel location independently.

This PR fixes the Cubelet path boundary so operators can configure the template artifact cache and shared kernel source independently, with one resolved artifact path for preparation, mounting, and cleanup. The historical toolbox paths remain the defaults. Deployment-specific relocation is handled separately by companion PR #1652, which keeps Cubelet's historical path and moves its backing directory through deployment-level softlinking.

## What Changed

- Add independent `image_base_path` and `shared_kernel_path` settings to the images plugin.
- Resolve those paths in `services/server.New` before plugin initialization, publish the resulting `pmem.Paths` for all consumers, and use the resolved image path for artifact preparation, VM mounting, and cleanup. The images plugin republishes the same resolved values in its own initialization, preserving its existing configuration ownership for standalone initialization paths.
- Keep `/usr/local/services/cubetoolbox/cubebox_os_image` as the default artifact cache for compatibility, while preserving explicitly configured legacy bases.
- Use `shared_kernel_path` for template artifacts and sandboxes without a template, so all kernel consumers use the same configured source.
- Keep deployment-layer relocation out of this change; the one-click and Helm data-disk softlink and migration behavior are handled by companion PR #1652.
- Remove the ineffective `cubetool_base_dir` entry misplaced under the runc runtime table in the shipped config. The images plugin still accepts this legacy setting when placed in its own table.
- Treat the images plugin table as the only user-facing source of artifact paths; legacy cbri `image_base_path` and `kernel_base_path` fields are ignored as artifact path sources. The early `server.New` bootstrap publishes the same resolved values before cbri initializes so all consumers use one path set.

The new `image_base_path` and `shared_kernel_path` settings are documented in the service management guide next to the Cubelet restart instructions.

## Upgrade Behavior

Upgrades with no explicit artifact path continue to use `/usr/local/services/cubetoolbox/cubebox_os_image`, so existing cached artifacts remain addressable. Operators that configure `image_base_path` explicitly are responsible for making the selected cache available before restarting Cubelet.

Cubelet does not move existing artifacts automatically. Before switching to a new path, make the affected cache or kernel files available there. Existing snapshots additionally require their historical artifact paths to remain resolvable (for example through a deployment-level softlink), or the snapshots must be recreated, because snapshot metadata stores absolute kernel and ext4 artifact paths. Deployment-specific data-disk migration, including preserving the historical Cubelet path through a softlink, is handled by the companion deployment change.

Legacy cbri artifact path fields may remain in an existing configuration, but they have no effect on artifact paths after this change. Configure artifact paths under the images plugin instead.

If a legacy cbri `base_path` was customized, set `shared_kernel_path` explicitly to the matching `vmlinux` under that installation. The cbri `base_path` still controls no-template guest and agent images, but no longer selects the kernel; the configured kernel must remain the node's selected bm/pvm variant.

## Validation

- Relevant Cubelet Go package tests and the Cubelet build passed, including default and explicit path resolution, legacy fallback, ignored cbri artifact fields, config dump/migrate output, shared-kernel selection, artifact lifecycle, and cleanup.
- Linux runtime validation covered separate artifact and kernel paths, template creation/restore/cleanup, and the historical `/usr/local` artifact path with the unchanged configuration.
- Full deployment migration, unavailable artifact sources, historical snapshot path variants, and application-specific readiness were not tested here; deployment migration belongs to companion PR #1652.

## Scope

This PR changes Cubelet template artifact path resolution and the related service-management documentation. It does not change one-click or Helm deployment migration behavior; that belongs to companion PR #1652.

Snapshot directory configuration and explicit snapshot component paths remain unchanged, as do artifact formats, kernel refresh semantics, and no-template kernel selection.

### DIFF
--- a/Cubelet/cmd/cubelet/config.go
+++ b/Cubelet/cmd/cubelet/config.go
@@ -34,6 +34,7 @@ import (
 	"github.com/containerd/plugin/registry"
 	ocispec "github.com/opencontainers/image-spec/specs-go/v1"
 	"github.com/pelletier/go-toml"
+	imageplugin "github.com/tencentcloud/CubeSandbox/Cubelet/services/images"
 	srvconfig "github.com/tencentcloud/CubeSandbox/Cubelet/services/server/config"
 	"github.com/urfave/cli/v2"
 	"golang.org/x/net/context"
@@ -52,6 +53,9 @@ func (c *Config) WriteTo(w io.Writer) (int64, error) {
 }
 
 func outputConfig(ctx context.Context, cfg *srvconfig.Config) error {
+	if _, err := imageplugin.ResolveConfiguredPaths(ctx, cfg); err != nil {
+		return err
+	}
 	config := &Config{
 		Config: cfg,
 	}
@@ -72,6 +76,11 @@ func outputConfig(ctx context.Context, cfg *srvconfig.Config) error {
 				return err
 			}
 
+			if imageConfig, ok := pc.(*imageplugin.Config); ok {
+				if _, err := imageConfig.ResolvePaths(); err != nil {
+					return err
+				}
+			}
 			config.Plugins[p.URI()] = pc
 		}
 	}
--- a/Cubelet/cmd/cubelet/config_paths_test.go
+++ b/Cubelet/cmd/cubelet/config_paths_test.go
@@ -1,0 +1,55 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package main
+
+import (
+	"context"
+	"os"
+	"path/filepath"
+	"testing"
+
+	"github.com/pelletier/go-toml/v2"
+	"github.com/stretchr/testify/require"
+	srvconfig "github.com/tencentcloud/CubeSandbox/Cubelet/services/server/config"
+)
+
+func TestConfigDumpArtifactPaths(t *testing.T) {
+	shipped, err := os.ReadFile(filepath.Join("..", "..", "config", "config.toml"))
+	require.NoError(t, err)
+	for _, tc := range []struct {
+		name, extra string
+		fail        bool
+	}{
+		{name: "shipped"},
+		{name: "conflicting legacy path", extra: "\n[plugins.\"io.cubelet.cbri.v1.cubebox\"]\nimage_base_path = \"/other/images\"\n", fail: true},
+	} {
+		t.Run(tc.name, func(t *testing.T) {
+			path := filepath.Join(t.TempDir(), "config.toml")
+			require.NoError(t, os.WriteFile(path, append(append([]byte{}, shipped...), []byte(tc.extra)...), 0600))
+			output, err := os.CreateTemp(t.TempDir(), "dump")
+			require.NoError(t, err)
+			original := os.Stdout
+			os.Stdout = output
+			defer func() { os.Stdout = original; _ = output.Close() }()
+			cfg := defaultConfig()
+			require.NoError(t, srvconfig.LoadConfig(context.Background(), path, cfg))
+			err = outputConfig(context.Background(), cfg)
+			os.Stdout = original
+			if tc.fail {
+				require.ErrorContains(t, err, "conflicts with images.image_base_path")
+				return
+			}
+			require.NoError(t, err)
+			data, err := os.ReadFile(output.Name())
+			require.NoError(t, err)
+			var decoded struct {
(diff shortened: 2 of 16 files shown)


luzhixing12345:
@zyl1121 Hi, with the current PR, if upgrading from an older version without an explicit path configuration, the system switches to `/data/cubelet/cubebox_os_image`. While the old files remain, the program looks for files in the new directory and attempts to re-download them; if the source is unavailable, the creation or restoration of the corresponding template may fail.

After comparing the two PRs, I personally lean towards the changes in #1652. It allows existing cache files to be accessed via their original paths, thereby reducing risks associated with re-downloads and dependencies on legacy paths. It also enables the file storage location to be adjusted through deployment configuration without requiring changes to Cubelet's internal path management. cc @fslongjin

zyl1121:
> @zyl1121 Hi, with the current PR, if upgrading from an older version without an explicit path configuration, the system switches to `/data/cubelet/cubebox_os_image`. While the old files remain, the program looks for files in the new directory and attempts to re-download them; if the source is unavailable, the creation or restoration of the corresponding template may fail.
> 
> After comparing the two PRs, I personally lean towards the changes in #1652. It allows existing cache files to be accessed via their original paths, thereby reducing risks associated with re-downloads and dependencies on legacy paths. It also enables the file storage location to be adjusted through deployment configuration without requiring changes to Cubelet's internal path management. cc @fslongjin

Thanks, that makes sense. The upgrade compatibility risk is real.

How about changing only the default path in this PR back to `/usr/local/services/cubetoolbox/cubebox_os_image`, while keeping the Cubelet-side path/configuration fixes? #1652 can then handle the relocation to `/data` through the symlink, preserving compatibility for existing installations.

The Cubelet-side fix in this PR is still useful independently: today the path configuration does not actually control where image artifacts are stored and used, so deployments outside this symlink-based setup cannot configure the location correctly. This PR fixes that configuration boundary and decouples artifact storage from the shared kernel path. The symlink can remain the compatibility path for existing installations, while deployments can use `image_base_path` directly once that compatibility layer is no longer needed.

luzhixing12345:
> Thanks, that makes sense. The upgrade compatibility risk is real.
>
> How about changing only the default path in this PR back to `/usr/local/services/cubetoolbox/cubebox_os_image`, while keeping the Cubelet-side path/configuration fixes? #1652 can then handle the relocation to `/data` through the symlink, preserving compatibility for existing installations.
>
> The Cubelet-side fix in this PR is still useful independently: today the path configuration does not actually control where image artifacts are stored and used, so deployments outside this symlink-based setup cannot configure the location correctly. This PR fixes that configuration boundary and decouples artifact storage from the shared kernel path. The symlink can remain the compatibility path for existing installations, while deployments can use `image_base_path` directly once that compatibility layer is no longer needed.

I think this is reasonable. We can combine the two pull requests and retain the new configuration capabilities added in the current pull request. When no path is configured, we can still use /usr/local/services/cubetoolbox/cubebox_os_image.

Please modify the code. By the way, please briefly supplement the user configuration documentation to explain the newly added configuration items.

zyl1121:
@luzhixing12345 Updated as discussed. I restored the historical default artifact path, kept the independent `image_base_path` / `shared_kernel_path` configuration, and added the corresponding user documentation.

Deployment-level relocation remains handled by #1652.

zyl1121:
@luzhixing12345 Updated as discussed. I restored the historical default artifact path, kept the independent `image_base_path` / `shared_kernel_path` configuration, and added the corresponding user documentation.

Deployment-level relocation remains handled by #1652.

Also, the current Cubelet unit-test failure looks like a flaky test and doesn't appear to be related to this change. The same `Cubelet/storage` tests have failed intermittently in previous runs of this PR, while other runs pass. Could you please rerun the failed job? This flaky test may be worth fixing separately.

luzhixing12345:
LTGM, all tests passed.

Issue #1151, opened 2026-07-24
[Feature Request] Add Post-Install Functional E2E Tests for Kubernetes Deployments


zyl1121:
## Problem / Motivation

A CubeSandbox Helm release can be installed successfully, and all Kubernetes workloads may become `Ready`, without proving that the installed system can actually create and run a sandbox through the full control-plane and data-plane path.

This matters because Kubernetes environments differ significantly across CNI plugins, dataplanes, kernels, node capabilities, storage classes, image delivery paths, and routing behavior. Some failures only appear when a real sandbox is created and asked to execute workloads, access the network, enforce policy, and expose services.

For example, CubeVS attaches TC programs in the `cube-node` networking path, and its interaction with host-side or workload-side CNI dataplanes may differ across Flannel, Calico, and Cilium configurations. But CNI compatibility is only one part of the broader problem. Similar post-install failures may also come from node virtualization prerequisites, runtime assets, image pulls, PVCs, node registration, Template readiness, scheduling capacity, DNS, CubeProxy routing, or the sandbox runtime itself.

Kubernetes readiness checks alone do not distinguish these cases.

The goal of this issue is to provide a reproducible post-install functional validation that answers:

> After CubeSandbox is installed on this Kubernetes cluster, can a real sandbox be created, run basic workloads, use the expected network policy, and expose services correctly?

## Proposed Scope

Add a Kubernetes post-install compatibility validation to the existing E2E framework.

The validation should focus on a small set of real sandbox scenarios that are most likely to expose Kubernetes environment compatibility problems:

* Sandbox creation and cleanup
* Basic command execution
* Basic file operations
* Public DNS and HTTPS egress
* Kubernetes Service access through DNS
* Kubernetes Service access through ClusterIP
* CubeProxy routing to a declared custom port

Before running sandbox-dependent cases, the validation should also perform lightweight observable preflight checks for common deployment blockers, such as:

* Helm release status
* Workload readiness
* PVC readiness
* Image pull failures
* Compute-node placement
* CubeMaster node registration
* Schedulable healthy capacity
* Runtime assets
* Template readiness

These preflight checks are not intended to become a full Kubernetes certification suite. They should provide enough context to avoid reporting every failure as a generic sandbox timeout.

## Kubernetes Networking Compatibility Cases

Networking should be one of the core validation areas because it is highly dependent on the cluster’s CNI, dataplane, routing mode, kernel features, and CubeVS policy enforcement.

[...]

PR #1677, opened 2026-09-08
test(e2e): add Kubernetes post-install functional validation

saurlax:
## Summary

Add `--k8s-post-install` to the existing SDK E2E runner to validate an installed CubeSandbox Helm release through real sandbox operations.

- Check Helm status, workloads, PVCs, compute-node registration, reported capacity, and runtime prerequisites before the existing SDK/template preflight.
- Reuse lifecycle, command, and file tests; add Service ClusterIP/FQDN, public DNS/HTTPS, and CubeProxy custom-port coverage.
- Create temporary Kubernetes Services with explicit network allow-outs, record same-node/cross-node placement, and verify cleanup.
- Add regression tests and English/Chinese usage documentation.

Refs #1151.

## Usage

Configure kubeconfig, `SDK_E2E_K8S_CONTEXT`, `SDK_E2E_K8S_NAMESPACE`, `SDK_E2E_K8S_RELEASE`, and the existing CubeAPI/template/proxy settings, then run:

```bash
pytest --run-e2e --k8s-post-install
```

The template must provide Python 3, trusted TLS roots, cluster DNS configuration, and the declared custom port (8088 by default).

## Validation

Local checks on the current revision:
- 132 tests passed; 172 opt-in live tests skipped.
- Ruff passed for changed Python files, excluding existing RUF100/SIM102 findings in the root conftest.
- `git diff --check` passed.

Live validation was performed on a Tencent Cloud bare-metal server:
- High I/O BMI5, 96 cores, 384 GB RAM, 5 Mbps bandwidth.
- TencentOS Server 4, kernel 6.6.117-45.11.6.tl4.x86_64.
- Single-node K3s v1.36.4+k3s1 with Flannel VXLAN.
- CubeSandbox v0.7.0 using native KVM.

All 9 selected checks passed in 43.97 seconds. No test sandboxes or temporary namespaces remained, and all CubeSandbox Pods were Ready.

This successful live run preceded the final scope reduction, which removed the standalone preflight-only option and reduced documentation/formatting changes. Local checks passed again afterward; a live rerun of the final revision is pending because SSH connectivity to the validation server timed out.

## Limitations

- Both Service paths were same-node; cross-node routing and other CNIs were not validated.
- An earlier run encountered `snapshot storage unavailable` during sandbox creation. Subsequent runs passed; the transient storage condition was not fully diagnosed.

Autonomously-by: Codex:GPT-6


saurlax:
I checked the findings and ran focused local reproductions.

- **Fixture resolution:** The same-name dependency is supported by pytest: an overriding fixture can request the fixture it overrides. Three checks using the current fixture definitions passed, covering default options, inherited marker options, and Service-specific options. I don't think a separate base fixture is necessary.
- **Missing `cubelet` container:** Confirmed. A Ready Pod carrying the `cube-node` label but lacking a container named `cubelet` raises `StopIteration`. The current chart uses that name, so this does not affect the normal deployment, but it is a real diagnostic gap. A small guard and regression test would address it.
- **Cleanup diagnostics:** Confirmed that successful REST cleanup clears earlier cleanup errors. This does not suppress a failure from the test body's explicit SDK `kill()` call, but retaining fallback diagnostics could improve troubleshooting.
- **Flags and backends:** The `--run-e2e` requirement follows the existing runner's opt-in behavior. The E2B adapter also targets the configured CubeAPI endpoint; it exercises CubeSandbox through another SDK. Service-specific cases already skip unsupported backends.

lkml-likexu:
@saurlax Wonderful contributions. Any test report demo to share ?

saurlax:
**Run date:** 8 September 2026  
**Tested revision stage:** Before final scope reduction  
**Environment:** CubeSandbox v0.7.0 · K3s v1.36.4+k3s1 · Flannel VXLAN · native KVM · single node · CubeSandbox SDK

**Execution summary:** `9 passed, 291 deselected, 8 warnings in 43.97s`. All nine teardown phases completed successfully.

| Test | Concrete execution result | Evidence detail / limit |
| --- | --- | --- |
| Deployment preflight | The preflight completed without a blocking error, and its scheduler-node list was non-empty. The historical environment report records successful Helm, workload, PVC, node-registration, and runtime checks. | The retained JSON records the test outcome, but does not contain the full preflight payload or numerical capacity readings. |
| Sandbox creation | The created sandbox had a non-empty ID. The ID returned by `info()` matched that ID. `uname -s` completed successfully and its normalized output equaled `linux`. | These assertions were satisfied. The sandbox ID and complete command response were not retained. |
| Command execution | The command result carried **exit code `7`**. Captured stdout contained **`hello-out`** and captured stderr contained **`hello-err`**. | Verified non-zero exit-code propagation and separate stdout/stderr capture. Assertions checked containment, not exact equality of the entire output streams. |
| File write/read | Reading `/tmp/sdk-compat-file.txt` after writing through the SDK returned exactly **`hello file`**. | Exact content equality was asserted; this was a small-text round trip. |
| Service via ClusterIP | The sandbox request to the temporary Service on port **`8080`** returned **HTTP `200`**. The body exactly matched the fixture’s namespace-derived marker; the probe emitted **`HTTP_OK`**. General internet access was disabled, with explicit outbound permissions for DNS and the Service IP. | Recorded compute node: **`vm-0-8-tencentos`**. Recorded endpoint node: **`vm-0-8-tencentos`**. Path: **`same-node`**. Actual Service IP and marker were not retained. |
| Service via FQDN | Access through `http.<namespace>.svc.<cluster-domain>:8080` completed using the guest’s existing resolver. The request returned **HTTP `200`**, its body exactly matched the fixture marker, and the probe emitted **`HTTP_OK`**. | Compute and endpoint nodes were both **`vm-0-8-tencentos`**; path: **`same-node`**. The actual namespace/FQDN and DNS response were not retained. |
| Public DNS / HTTPS | The configured HTTPS request returned **HTTP `200`** and the probe emitted **`HTTP_OK`**, with inherited proxies disabled and TLS certificate verification enabled. | Source default: `https://example.com/`. The effective configured URL was not retained, so the default is not asserted to be the actual target. No exact public response-body check was performed. |
| CubeProxy custom port | The historical report identifies port **`8088`**. A request routed through CubeProxy returned **HTTP `200`**, with a body exactly matching the unique **`cube-k8s-<UUID>`** marker generated for that sandbox, within the test’s 30-second retry window. | This checked the intended sandbox server, not just an HTTP health response. The actual UUID, proxy address, retry count, and per-test latency were not retained. |
| Sandbox deletion | After SDK `kill()`, the reconnect helper confirmed connection failure. API lookup for the same sandbox ID returned **`{}`**. Teardown completed without cleanup errors. | Confirms reconnect rejection and API absence. The exact reconnect exception and deleted sandbox ID were not retained. |

## Resource Cleanup and Deployment State

| Measurement | Recorded value |
| --- | --- |
| Sandboxes remaining after execution | **0** |
| Temporary test namespaces remaining | **0** (`remaining_mock_namespaces=[]`) |
| Sandbox cleanup records | **8**, each with **`errors=[]`** |
| Teardown results | **9 passed**, no failed teardown in the retained record |
| Deployment Pod state | **13/13 Running; 13/13 Ready** |
| Observed Service topology | **2 same-node paths; 0 cross-node paths** |

## Additional Execution Results

| Check / attempt | Recorded result |
| --- | --- |
| Local suite after scope reduction | **132 passed; 172 skipped**, including **129 framework tests**. The skipped tests were opt-in live cases. |
| Added Kubernetes/cleanup regression instances | **21**, included in the passing local total. |
| Ruff | No remaining findings under the executed check configuration; existing root-conftest **RUF100/SIM102 were excluded**. |
| `git diff --check` | Completed without reported whitespace errors. |
| Earlier live attempt | Both Service cases failed during sandbox creation with **error `130597`**. Master logged **`snapshot storage unavailable`**. The subsequent nine-test execution succeeded; the transient storage failure’s root cause remained unresolved. |
| Warnings in successful live run | **8**. The historical report attributes them to the internal HTTP CubeAPI endpoint. |

tests/e2e/sdk_compat/conftest.py
```diff
@@ -64,6 +64,8 @@ def _strip_env_value(value: str) -> str:
 
 def pytest_addoption(parser: pytest.Parser) -> None:
     group = parser.getgroup("sdk compat e2e")
+    group.addoption("--k8s-post-install", action="store_true", default=False,
```
fslongjin, line 67: The help text documents "requires --run-e2e", but nothing enforces the combination. As-is, `pytest --k8s-post-install` alone collects the 9 tests, skips them all with "live SDK E2E disabled", and exits 0:

```
$ pytest --k8s-post-install -q
9 skipped, 295 deselected in 0.12s        # exit code 0
```

For an acceptance gate this is the worst failure mode: a misconfigured CI job or operator command goes green while validating nothing.

Since every post-install test is inherently live, the flag can simply imply the gate (in `pytest_configure`, before collection reads the option):

```python
if config.getoption("--k8s-post-install"):
    config.option.run_e2e = True
```

This also simplifies the documented invocation to a single flag. If you'd rather keep them independent, please raise `pytest.UsageError("--k8s-post-install requires --run-e2e")` instead — either fix works for me.

tests/e2e/sdk_compat/framework/kubernetes_suite.py
```diff
@@ -0,0 +1,22 @@
...
+
+SHARED_CASES = {
+    "lifecycle/test_create.py::test_create_returns_usable_sandbox",
+    "lifecycle/test_kill.py::test_kill_prevents_reconnect",
+    "commands/test_run.py::test_command_stdout_stderr_and_exit_code",
+    "filesystem/test_read_write.py::test_file_write_read_roundtrip",
+}
+
+
+def select_post_install_tests(items):
```
fslongjin, line 14: With `--sdk-e2e-backends=cubesandbox,e2b`, the selector admits every backend's parametrization — collection grows from 9 to 17 items (verified via `--collect-only`), including `test_public_dns_and_https[e2b]` and `test_cubeproxy_custom_port[e2b]`. Probing e2b sandboxes doesn't validate this Helm release, and the only items that skip are the two e2b Service cases — via the `k8s_service` fixture, by incident rather than by design.

Suggested shape:

```python
for item in items:
    spec = getattr(item, "callspec", None)
    backend = spec.params.get("sdk_backend") if spec else None
    include = include and backend in (None, "cubesandbox")
```

(`test_kubernetes_preflight` has no `callspec`, hence the `None`.)

tests/e2e/sdk_compat/framework/kubernetes.py
```diff
@@ -0,0 +1,589 @@
...
+            )
+        except (RuntimeError, ValueError):
+            # The API may have committed the namespace before the client lost
+            # its response. Recover ownership, not a blind name-only delete.
+            observed = kube.run(
+                "get", "namespace", name, "--ignore-not-found", "-o", "json"
+            )
+            if observed:
+                metadata = json.loads(observed)["metadata"]
+                if all(
```
fslongjin, line 456: Robustness of the lost-response recovery branch: when the create call failed and the namespace genuinely doesn't exist, newer kubectl prints `null` (rather than empty output) for `get --ignore-not-found -o json`. `json.loads("null")` is `None`, so `["metadata"]` raises `TypeError: 'NoneType' object is not subscriptable` — which replaces the original create error with a confusing traceback. Suggested:

```python
if observed:
    metadata = (json.loads(observed) or {}).get("metadata") or {}
```

This branch also has no unit coverage today — `LostResponseKube` only exercises the "namespace exists" path, so a parametrized "not found" case would be a natural addition.

tests/e2e/sdk_compat/framework/kubernetes.py
```diff
@@ -0,0 +1,589 @@
...
+    ops = [
+        s
+        for s in services
+        if s["metadata"].get("labels", {}).get("app.kubernetes.io/component")
+        == "ops"
+    ]
+    registry = (
+        observe(
+            "CubeOps scheduler node view",
+            lambda: kube.service_get(ops[0], "/internal/v1/nodes"),
```
fslongjin, line 302: `ops[0]` takes whichever `component=ops` Service the API happens to list first. Correct today (the chart ships exactly one), but discovery becomes order-dependent the moment a second ops Service appears. A name match, a deterministic sort, or a one-line comment noting the single-Service assumption would all work.

tests/e2e/sdk_compat/conftest.py
```diff
@@ -150,6 +152,8 @@ def pytest_configure(config: pytest.Config):
     # the ``dsession`` controller when distribution mode activates). That lets
     # ``_verify_xdist_activated`` observe the real activation state below.
     for marker in (
+        "k8s_post_install: Kubernetes post-install functional validation",
```
fslongjin, line 155: The two new markers are registered here via `addinivalue_line` while the existing ~20 live in `pytest.ini`. Mind moving them there for consistency? It also keeps `--strict-markers` viable later.

tests/e2e/sdk_compat/cases/kubernetes/conftest.py
```diff
@@ -0,0 +1,52 @@
...
+            endpoint_node=service["endpoint_node"],
+            path="same-node"
+            if target["node"] == service["endpoint_node"]
+            else "cross-node",
+        )
+        yield service
+
+
+@pytest.fixture()
+def sdk_create_options(request, sdk_create_options):
```
fslongjin, line 43: The same-name override (`sdk_create_options` requesting `sdk_create_options`) is valid pytest — the parameter resolves to the root conftest's definition — but it's subtle enough that the next reader will pause on it. A one-line comment ("same-name override; the argument resolves to the root conftest fixture") would save them the detour.

tests/e2e/sdk_compat/framework/cleanup.py
```diff
@@ -55,16 +59,35 @@ def safe_kill(adapter: SandboxAdapter, config: SdkE2EConfig) -> list[str]:
...
+        try:
+            api = ApiClient(config)
+            if api.get_sandbox(adapter.sandbox_id):
+                errors.append(
+                    f"sandbox {adapter.sandbox_id} still exists after cleanup"
+                )
+            else:
+                # Kill tests already removed the sandbox; a confirmed 404 also
+                # validates a successful REST fallback after an SDK error.
+                errors.clear()
```
fslongjin, line 82: `errors.clear()` discards *all* earlier errors once a 404 is confirmed — including `.info failed` and kill failures, not just the REST-fallback case the comment describes. I believe that's the right semantics for this profile (what matters post-install is that the sandbox is actually gone), but the breadth looks easy to trip over later, so confirming: intended?

tests/e2e/sdk_compat/framework/kubernetes.py
```diff
@@ -0,0 +1,589 @@
...
+                    "namespace",
+                    name,
+                    "--wait=true",
+                    "--timeout=45s",
+                    timeout=55,
+                )
+                reporter.record(
+                    "kubernetes_mock_cleanup", namespace=name, outcome="passed"
+                )
+            except RuntimeError as exc:
```
fslongjin, line 582: If the test body fails and namespace cleanup also fails, the `RuntimeError` raised from this `finally` is what pytest surfaces; the original failure survives only in the JSONL events and `__context__`. Fine for a validation profile (cleanup failures must stay visible), but appending the in-flight exception to the cleanup error message would keep both visible in the terminal as well.

tests/e2e/sdk_compat/cases/kubernetes/test_post_install.py
```diff
@@ -0,0 +1,122 @@
...
+import pytest
+import requests
+from framework.assertions import assert_command_ok
+from framework.capabilities import NETWORK_ALLOW_DENY, NETWORK_PUBLIC_ACCESS
+from framework.network_probe import TRAFFIC_ACCESS_TOKEN_HEADERS
+
+pytestmark = [pytest.mark.e2e, pytest.mark.sdk_compat, pytest.mark.k8s_post_install]
+
+
+def test_kubernetes_preflight(k8s_environment):
```
fslongjin, line 20: Since a failed preflight aborts the session before this test can run, the assert can only pass — its real job is to give the preflight result a test-level record. Consider asserting the `kubernetes_preflight_passed` report event instead, which makes that intent explicit and gives the assert something to actually check.

fslongjin (changes requested):
**Request Changes — but scoped to a single blocking item; everything else is nits/questions.**

I cross-checked the PR's assumptions against the current tree and they all hold: the chart labels (`cube-node`, `ops`), the `data-cubelet` mount path via `cube.nodeToolboxVolumeMounts`, the unauthenticated CubeOps `/internal/v1/nodes` route and `SchedulerNode` JSON fields, SDK `distribution_scope` / `allow_internet_access` / `network.allow_out`, `MaxMvmLimit <= 0` meaning unlimited, `get_sandbox` 404 → `{}`, and the same-name fixture override (verified with a minimal local repro). Local numbers reproduce yours exactly: 132 passed / 172 skipped, the 21 new framework tests green, ruff clean on the changed files, and the selector picks exactly the advertised 9 tests.

**Blocking (1):** `pytest --k8s-post-install` without `--run-e2e` silently skips all 9 tests and exits 0 — a misconfigured gate goes green while validating nothing. Please make the flag imply `--run-e2e` (or raise `UsageError`). Details inline.

**Non-blocking, in priority order:**
- Multi-backend runs admit e2b parametrizations (9 → 17 collected items).
- The namespace-recovery branch can hit `TypeError` on kubectl's `null` output, masking the original create error.
- Nits: marker registration location, `ops[0]` determinism, a comment for the same-name fixture override, cleanup error masking the original failure, the near-tautological preflight test.

**One ask before merge:** the 9/9 live run in the description predates the final scope reduction (the standalone preflight-only option was removed and selection logic changed with it), and the re-run is noted as pending. Please re-run the final revision on the validation host once SSH is back and post the numbers.

lkml-likexu:
@saurlax This one is wonderfull. Would you help to move on ?

saurlax:
Thanks for the review. I pushed `20f595f` to address the blocking flag issue: `--k8s-post-install` without `--run-e2e` now raises a usage error instead of reporting a successful skipped run.

The post-install profile now selects only the native CubeSandbox backend. I also handled `kubectl` returning `null` during namespace recovery, made a missing `cubelet` container an actionable preflight error, and addressed the marker, Service ordering, fixture comment, and cleanup diagnostic suggestions. The live run exposed one more gap: public HTTPS needed an explicit allow-out for the cluster DNS IP; that is now covered by a regression test.

I reran the updated revision on a single-node K3s v1.36.4+k3s1 deployment with CubeSandbox v0.7.0 and native KVM: **9 passed, 305 deselected in 49.44s**. Local Linux checks were **142 passed, 172 skipped**. After the run, all 13 CubeSandbox Pods were Ready, with no test sandboxes or temporary namespaces remaining.

Both Service paths were same-node. Cross-node routing and other CNIs remain untested. A separate hostNetwork experiment returned CubeProxy upstream 502s, so this successful run covers the chart's default Pod network configuration.

PR #1684, opened 2026-09-09
fix(cubelet-volume): keep detach safe when the refcount store fails

dushulin:
## Summary

`Manager.Detach` in `Cubelet/plugins/volume/manager.go` misuses the
`ReleaseResult` when the underlying refcount store returns an error.
Because Go zero-initializes the result on error, the code ends up
setting `req.RefCount = 0` and `req.NodeRefLastDetach = true`, i.e. it
tells the plugin *and* CubeMaster that this was the last detach when in
reality the refcount was never decremented.

The unknown-driver branch inside the same function has the same shape
of bug (`rr, _ := rc.Release(...); req.NodeRefLastDetach = rr.After == 0
&& !rr.NotHeld`).

## Impact

For any plugin that treats `RefCount == 0` as the trigger for
host-level teardown (unmounting a shared FUSE / cosfs / NFS mount,
releasing a lease, etc.), a transient store error on one sandbox's
detach can tear down resources that other sandboxes on the same node
still depend on. On the control-plane side, CubeMaster observes a
spurious 1→0 transition and removes the node from the volume's holder
list, which then lets a subsequent `Volume.destroy` bypass the
"volume in use" protection.

## When it fires

`refcount.Store.Release` is a `bbolt` `db.Update` transaction and can
return an error on:
- data disk full / read-only remount / inode exhaustion,
- bbolt txn abort (mmap failure, concurrent write timeout),
- record unmarshal error (corrupted or version-mismatched value),
- `db.Close`d before the call reaches it during shutdown races.

## Fix

Both branches:
- on `Release` error, do **not** propagate the zero-value result to the
  plugin or to `NodeRefLastDetach`;
- set `RefCount = 1` on the main path so plugins that key on
  `RefCount == 0` will conservatively skip host-level cleanup;
- keep `NodeRefLastDetach = false` so CubeMaster is not told 1→0;
- log a `WARN` (was previously silent — `_ = err`) so operators can see
  the store degraded;
- still return `nil` from `Detach` so the caller can reclaim the
  per-sandbox bind, matching the original "non-fatal; don't block
  Destroy" intent.

Once the store recovers a later successful `Detach` will emit the
correct transition; the current node's record is left as-is and
requires operator cleanup, but that's already the expected posture when
the store has thrown an I/O error.

## Tests

New `Cubelet/plugins/volume/manager_test.go` covering:
- `TestDetachRefcountReleaseError` — main path with a closed bbolt DB:
  plugin sees `RefCount=1` and `NodeRefLastDetach=false`, `Detach`
  returns `nil`.
- `TestDetachRefcountReleaseSuccessLastHolder` — healthy last-holder
  path unchanged: `RefCount=0`, `NodeRefLastDetach=true`.
- `TestDetachRefcountReleaseSuccessMultiHolder` — healthy path with
  another holder still present: `RefCount=1`, `NodeRefLastDetach=false`.
- `TestDetachUnknownDriverRefcountReleaseError` — unknown-driver +
  `Release` error: `NodeRefLastDetach` stays `false`.

All four pass; `go test ./plugins/volume/...` and `go vet` clean.

### DIFF
--- a/Cubelet/plugins/volume/manager.go
+++ b/Cubelet/plugins/volume/manager.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/containerd/log"
 	"github.com/tencentcloud/CubeSandbox/Cubelet/plugins/volume/refcount"
 )
 
@@ -170,8 +171,19 @@ func (m *Manager) Detach(ctx context.Context, req *DetachRequest) error {
 	if err != nil {
 		// Unknown driver during Detach: release ref-count and continue.
 		if rc := m.rcStore; rc != nil {
-			rr, _ := rc.Release(req.Namespace, req.VolumeID, req.SandboxID)
-			req.NodeRefLastDetach = rr.After == 0 && !rr.NotHeld
+			rr, rErr := rc.Release(req.Namespace, req.VolumeID, req.SandboxID)
+			if rErr == nil {
+				req.NodeRefLastDetach = rr.After == 0 && !rr.NotHeld
+			} else {
+				// Refcount store failure: the count is unknown; suppress
+				// the 1→0 event so CubeMaster does not incorrectly remove
+				// this node from the volume's holder list.
+				log.G(ctx).WithError(rErr).
+					WithField("driver", req.Driver).
+					WithField("volume_id", req.VolumeID).
+					WithField("sandbox_id", req.SandboxID).
+					Warn("volume: refcount release failed during unknown-driver detach; suppressing last-detach event")
+			}
 		}
 		return nil
 	}
@@ -187,13 +199,30 @@ func (m *Manager) Detach(ctx context.Context, req *DetachRequest) error {
 	if rc := m.rcStore; rc != nil {
 		rr, err := rc.Release(req.Namespace, volID, req.SandboxID)
 		if err != nil {
-			_ = err // non-fatal; don't block Destroy
-		} else if !rr.NotHeld {
-			released = true
+			// Refcount store I/O failed — the true count is unknown.
+			// Conservatively tell the plugin the volume is still held so
+			// it will NOT tear down shared host resources that other
+			// sandboxes on this node may still be using, and clear
+			// NodeRefLastDetach so CubeMaster does NOT observe a spurious
+			// 1→0 transition. Detach itself still succeeds so the caller
+			// can reclaim the per-sandbox bind. Operators must clean up
+			// the refcount store manually once the underlying I/O issue
+			// is resolved.
+			log.G(ctx).WithError(err).
+				WithField("driver", req.Driver).
+				WithField("volume_id", volID).
+				WithField("sandbox_id", req.SandboxID).
+				Warn("volume: refcount release failed; degrading to safe mode (RefCount=1)")
+			req.RefCount = 1
+			req.NodeRefLastDetach = false
+		} else {
+			if !rr.NotHeld {
+				released = true
+			}
+			req.RefCount = rr.After // 0 = last detach
+			// 1 → 0 on this node: last sandbox here stopped referencing the volume.
+			req.NodeRefLastDetach = rr.After == 0 && !rr.NotHeld
 		}
-		req.RefCount = rr.After // 0 = last detach
-		// 1 → 0 on this node: last sandbox here stopped referencing the volume.
-		req.NodeRefLastDetach = rr.After == 0 && !rr.NotHeld
 	}
 
 	if pluginErr := p.Detach(ctx, req); pluginErr != nil {
--- a/Cubelet/plugins/volume/manager_test.go
+++ b/Cubelet/plugins/volume/manager_test.go
@@ -1,0 +1,204 @@
+// Copyright (c) 2026 Tencent Inc.
+// SPDX-License-Identifier: Apache-2.0
+
+package volume
+
+import (
+	"context"
+	"path/filepath"
+	"testing"
+
+	bolt "go.etcd.io/bbolt"
+
+	"github.com/tencentcloud/CubeSandbox/Cubelet/plugins/volume/refcount"
+)
+
+// fakePlugin records the RefCount and NodeRefLastDetach it observes on Detach,
+// so tests can assert the Manager's ref-count bookkeeping is correct.
+type fakePlugin struct {
+	name          string
+	attachErr     error
+	detachErr     error
+	gotRefCount   int64
+	gotLastDetach bool
+	detachCalls   int
+}
+
+func (p *fakePlugin) Name() string                                 { return p.name }
+func (p *fakePlugin) PluginType() PluginType                       { return PluginTypeBuiltin }
+func (p *fakePlugin) Init(_ context.Context, _ PluginConfig) error { return nil }
+
+func (p *fakePlugin) Attach(_ context.Context, req *AttachRequest) (*AttachResult, error) {
+	if p.attachErr != nil {
+		return nil, p.attachErr
+	}
+	return &AttachResult{VolumeID: req.VolumeID, HostPath: "/tmp/" + req.VolumeID}, nil
+}
+
+func (p *fakePlugin) Detach(_ context.Context, req *DetachRequest) error {
+	p.detachCalls++
+	p.gotRefCount = req.RefCount
+	p.gotLastDetach = req.NodeRefLastDetach
+	return p.detachErr
+}
+
+func (p *fakePlugin) Close() error { return nil }
+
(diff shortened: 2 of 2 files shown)


Cubelet/plugins/volume/manager.go
```diff
@@ -187,13 +199,30 @@ func (m *Manager) Detach(ctx context.Context, req *DetachRequest) error {
...
+			// 1→0 transition. Detach itself still succeeds so the caller
+			// can reclaim the per-sandbox bind. Operators must clean up
+			// the refcount store manually once the underlying I/O issue
+			// is resolved.
+			log.G(ctx).WithError(err).
+				WithField("driver", req.Driver).
+				WithField("volume_id", volID).
+				WithField("sandbox_id", req.SandboxID).
+				Warn("volume: refcount release failed; degrading to safe mode (RefCount=1)")
+			req.RefCount = 1
```
fslongjin, line 216: **Docs: the `RefCount=1` sentinel isn't documented on any plugin-author-facing surface**

On this path `RefCount` becomes a fabricated "still held" sentinel rather than the real post-detach count, but every surface an external plugin author reads still documents the strict contract: `context.go` ("number of sandboxes still attached AFTER this call"), `interface.go`, `volumeplugin.proto` (`ref_count`), and both volume-plugin guides (en/zh). Binary/RPC plugins receive the value verbatim (`--ref-count` / gRPC field), so a plugin author has no way to learn that a nonzero value can mean "count unknown" rather than "exactly one holder remains".

Not behavior-blocking (in-repo and example plugins only branch on `== 0`, where the sentinel is exactly equivalent), but the divergence is introduced here, so it's cheapest to fix now — one caveat sentence on `DetachRequest.RefCount` (plus the guide), e.g.: *"If the node-local refcount store fails, Cubelet passes refCount=1 as a conservative sentinel, not a real count — treat any value > 0 as 'do not tear down'."*

Cubelet/plugins/volume/manager.go
```diff
@@ -187,13 +199,30 @@ func (m *Manager) Detach(ctx context.Context, req *DetachRequest) error {
...
-			_ = err // non-fatal; don't block Destroy
-		} else if !rr.NotHeld {
-			released = true
+			// Refcount store I/O failed — the true count is unknown.
+			// Conservatively tell the plugin the volume is still held so
+			// it will NOT tear down shared host resources that other
+			// sandboxes on this node may still be using, and clear
+			// NodeRefLastDetach so CubeMaster does NOT observe a spurious
+			// 1→0 transition. Detach itself still succeeds so the caller
+			// can reclaim the per-sandbox bind. Operators must clean up
```
fslongjin, line 208: **Comment: "Operators must clean up the refcount store manually" is inaccurate — a restart self-heals**

`RecoverRefCounts` (`refcount/store.go:316`) runs at every cubelet startup (`storage/plugin.go:559-570`) and removes refs of sandboxes that are no longer live. The sandbox whose Release failed here was just destroyed, so the stale ref is reconciled automatically on the next restart once the store is healthy — no manual store surgery needed (the store's own package doc, "Recovery after restart", describes exactly this path).

The comment also inverts the residual risk: the store record is the part that self-heals; what actually leaks is the host mount the plugin was told not to tear down (benign — the cosfs example's `Mount` checks `isMountPoint` first and reuses it). Suggested wording: "the stale ref is reconciled by RecoverRefCounts at the next cubelet restart; the host mount is intentionally left up and reused by the next Attach".

(Same-area nit: the unknown-driver comment says CubeMaster would "remove this node from the volume's holder list" — master keeps no per-node list; a 1→0 event decrements the aggregate `t_cube_volume.refcount`, which gates `Volume.destroy`. Substance right, mechanism imprecise.)

Cubelet/plugins/volume/manager.go
```diff
@@ -170,8 +171,19 @@ func (m *Manager) Detach(ctx context.Context, req *DetachRequest) error {
 	if err != nil {
 		// Unknown driver during Detach: release ref-count and continue.
 		if rc := m.rcStore; rc != nil {
-			rr, _ := rc.Release(req.Namespace, req.VolumeID, req.SandboxID)
-			req.NodeRefLastDetach = rr.After == 0 && !rr.NotHeld
+			rr, rErr := rc.Release(req.Namespace, req.VolumeID, req.SandboxID)
```
fslongjin, line 174: **Asymmetric with the main error path below; suppression here relies on the caller's zero-value request**

Two small things, both fixable in this `else`:

1. On `rErr != nil` this branch leaves `req.RefCount` at the caller's value (0), while the main path sets the `1` sentinel. `Detach` returns nil, so the caller's success log prints `refcount_after=0` (`pluginvolume.go:306`) — indistinguishable from a genuine last-detach, exactly when an operator is triaging a degraded store. (The success arm never sets `RefCount` either — pre-existing; writing `rr.After` there and the sentinel on error would unify both arms.)
2. The 1→0 suppression works only by *not touching* `NodeRefLastDetach`, relying on callers passing a fresh zero-valued request (both in-tree callers do). Since `context.go` documents Manager as owning these fields, an explicit `req.RefCount = 1; req.NodeRefLastDetach = false` here makes the branch self-contained — the main path already uses the explicit form.

Cubelet/plugins/volume/manager_test.go
```diff
@@ -0,0 +1,204 @@
...
+		t.Errorf("plugin got RefCount=%d, want 1 (one holder still present)", p.gotRefCount)
+	}
+	if p.gotLastDetach {
+		t.Error("plugin got NodeRefLastDetach=true, want false")
+	}
+}
+
+// Unknown-driver Detach must also suppress NodeRefLastDetach when the refcount
+// store fails, so CubeMaster is never told 1→0 on an unrecovered error.
+func TestDetachUnknownDriverRefcountReleaseError(t *testing.T) {
```
fslongjin, line 180: **Test gap: the success arm of the unknown-driver branch (`manager.go:175-176`) is untested**

This PR split that branch into err/non-err arms but only tests the error arm. Regressions to the success arm — inverted condition, an "always suppress" simplification, dropping the `!rr.NotHeld` clause — pass the current suite. Two cheap cases would pin it:

- unknown driver + single holder, Release OK → `NodeRefLastDetach == true` (this arm emits a real 1→0 to CubeMaster, so a regression here is control-plane visible);
- unknown driver + never-acquired sandbox (`NotHeld`) → `false`.

Cubelet/plugins/volume/manager_test.go
```diff
@@ -0,0 +1,204 @@
...
+
+	"github.com/tencentcloud/CubeSandbox/Cubelet/plugins/volume/refcount"
+)
+
+// fakePlugin records the RefCount and NodeRefLastDetach it observes on Detach,
+// so tests can assert the Manager's ref-count bookkeeping is correct.
+type fakePlugin struct {
+	name          string
+	attachErr     error
+	detachErr     error
```
fslongjin, line 21: **Test gap: the plugin-failure rollback path (`manager.go:228-239`) has no coverage — the scaffolding for it is right here, unused**

`detachErr` (and `attachErr`, for the Attach-side rollback paths at `manager.go:146-157`) are declared but never exercised. The re-Acquire rollback + `NodeRefLastDetach=false` reset is the trickiest invariant in `Detach`, and this PR's `released=false`-on-store-error change interacts with it. ~15 lines with the existing helpers: single holder + `detachErr` → assert the error is returned, `store.Get("ns", "vol1") == 1` (rollback took effect), `NodeRefLastDetach == false`.

(Optional: the two healthy-path tests could also assert `store.Get(...)` — 0 after last-holder detach, 1 after multi-holder — to pin the Release side-effect and record deletion at zero.)

fslongjin (commented):
Verified the fix from several angles (correctness, tests, docs, security, project conventions): the bug is real — a store error made both the plugin and CubeMaster act on a fabricated "count reached zero" — and the fix is correct. On this head: `go test ./plugins/volume/...` and `go vet` pass locally; success-path semantics are unchanged field-by-field; branching on `err != nil` also discards the partially-populated result case (callback ran, commit failed), which is one more way the old code was unsafe. The 4 new tests pin both fixed branches with deterministic error injection (closed bbolt → `ErrDatabaseNotOpen`).

Findings below — none blocks merge. R1–R3 are small pre-merge polish; R4–R5 are test additions that can also land as follow-ups.

Aside: the `Test Cubelet (amd64)` failure is `pkg/queueworker`'s `TestQueueWorker` — a timing-sensitive test unrelated to this PR (it asserts a full queue while the QueueWorker's own background workers are concurrently draining it; it flakes under CI scheduling jitter and passes locally). A re-run should go green.

ls-ggg:
Please return the failure instead, and keep sandbox deletion reentrant.

PR #1688, opened 2026-09-09
fix(s3): keep pause package after Resume so Snapshot stays incremental

ls-ggg:
## Summary

- After S3 Resume, Cubelet used to delete the pause catalog immediately. `CommitSandbox` only resolves a memory base from that catalog, so the next Snapshot fell through to a full 2GiB dump (same-node and cross-node).
- Resume-time `CleanupTemplate` now no-ops on S3 the same way it already does on XFS while a live sandbox still holds the pause id. The next Pause or Destroy still GCs the old package. Cross-node origin leftover (tombstone + package) is still dropped.
- Cross-node Resume never writes a local pause catalog. `prepareCommitMemoryArtifact` now clones the imported `sb-*-memory` volume (the same base Pause already uses) instead of full-dumping.

XFS Resume keep / mmap behavior is unchanged.

## Test plan

- [ ] Unit: `TestKeepLiveXFSPausePackage` — live sandbox keeps the package; PAUSED / forged annotation still allow delete
- [ ] Same-node S3: Create → Pause → Resume → Snapshot — memory is incremental / soft-dirty, not full; pause package dir still present until next Pause or Destroy
- [ ] Same-node S3: Resume → Pause again — second pause is incremental; previous package is cleaned
- [ ] Cross-node S3: Pause@A → Resume@B → Snapshot@B — Snapshot is incremental (own volume), not Device busy
- [ ] Destroy / DelPaused still delete the pause package

### DIFF
--- a/CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
+++ b/CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go
@@ -449,11 +449,12 @@ func resumeFromPauseSnapshot(ctx context.Context, req *types.UpdateRequest, host
 	// After Create the sandbox runs on private disks. Ask Cubelet to drop
 	// the pause package on the node that still holds it: origin when that
 	// IP is known and not the target (also the PAUSED tombstone), otherwise
-	// this node. Cubelet decides whether the catalog actually goes away
-	// (S3 already cloned onto sb-*-memory; XFS Resume still mmaps the
-	// package and no-ops until the next Pause or Destroy). Do not key this
-	// off CrossNode — a cross-node placement with a blank origin would
-	// otherwise skip both cleanup paths and leak the package.
+	// this node. Cubelet no-ops while a live sandbox still restores from
+	// that catalog (XFS mmap; S3 Snapshot last-restore) and GCs on the
+	// next Pause or Destroy. Cross-node origin is not live, so the origin
+	// package still goes away. Do not key this off CrossNode — a
+	// cross-node placement with a blank origin would otherwise skip both
+	// cleanup paths and leak the package.
 	origin := strings.TrimSpace(rec.NodeIP)
 	if origin != "" && origin != targetIP {
 		if err := pausesnap.DropOriginTombstone(ctx, req.RequestID, req.SandboxID, origin, rec.SnapshotID, rec.Backend); err != nil {
--- a/CubeMaster/pkg/service/sandbox/util.go
+++ b/CubeMaster/pkg/service/sandbox/util.go
@@ -281,7 +281,7 @@ func ConstructCubeletReq(ctx context.Context, req *types.CreateCubeSandboxReq) (
 
 	out := &cubebox.RunCubeSandboxRequest{
 		RequestID:         req.RequestID,
-		Labels:            req.Labels,
+		Labels:            stripUserCubeMasterLabels(req.Labels),
 		InstanceType:      req.InstanceType,
 		NetworkType:       req.NetworkType,
 		Annotations:       make(map[string]string),
@@ -325,8 +325,12 @@ func ConstructCubeletReq(ctx context.Context, req *types.CreateCubeSandboxReq) (
 		delete(out.Annotations, AnnotationPluginVolumeSources)
 	}
 	// Sync any annotations added by checkAndGetVolumes (e.g. plugin-volume-sources)
-	// into the outgoing RunCubeSandboxRequest.
+	// into the outgoing RunCubeSandboxRequest. Do not re-add keys
+	// checkAndGetAnnotation already dropped (pause / restore-base).
 	for k, v := range req.Annotations {
+		if rejectUserCubeMasterAnnotation(k) {
+			continue
+		}
 		if _, exists := out.Annotations[k]; !exists {
 			out.Annotations[k] = v
 		}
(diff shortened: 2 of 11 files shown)


fslongjin:
## Missing Fix regression tests

This PR fixes a user-visible bug: after **S3 Resume**, the next **Snapshot** was falling through to a **full ~2GiB memory dump** because the pause catalog was deleted too early.

The only test change is `TestKeepLiveXFSPausePackage`, which now asserts the keep predicate after the S3 early-return was removed. That locks in a helper, but **not the behavior users actually hit**.

Please add unit tests that would **fail on `master` and pass with this patch**, covering at least:

1. **`CleanupTemplate` with `honorLiveXFSPause=true`** — while a live S3 sandbox holds the pause snapshot id and the catalog entry is `Kind=pause_snapshot`, cleanup is a **no-op**; with `honorLiveXFSPause=false` (Destroy / next Pause GC), the package is still **deleted**.
2. **`prepareCommitMemoryArtifact` Tier 2b** — when both catalog bases are missing but `ImportedMemoryVol` is set (typical cross-node Resume), the commit path returns **incremental** (clone of `sb-*-memory`), not full.
3. **Full fallback** — when catalog bases are missing **and** `ImportedMemoryVol` is empty, the path still falls back to **full**.

`snapshot_base_memory_fallback_test.go` already documents the commit decision tree but was not extended here. SDK e2e is unlikely to stably assert incremental vs full dump, so this fix really needs a Cubelet unit regression before merge.

---

## Security (Medium): user `Labels` can pin another tenant’s S3 pause package

Removing `if storage.IsS3Backend(backend) { return false }` from `keepLiveXFSPausePackage` aligns S3 with XFS for same-node keep — but it also **removes a safety net** that used to let origin-node cleanup proceed even when something looked “live.”

**Issue:** Master already strips forged **`Annotations`** via `rejectUserCubeMasterAnnotation`, but **`req.Labels` are forwarded as-is** in `ConstructCubeletReq` and written into the CubeBox at Create. `keepLiveXFSPausePackage` only trusts **Labels** (not Annotations — correctly tested), and `CleanupTemplate` with `honorLiveXFSPause=true` no-ops when any RUNNING/CREATED/PAUSING sandbox’s Labels match the pause snap id.

**Cross-node Resume attack path (origin node):**
1. Victim resumes on node B; Master calls `DropOriginTombstone` on origin node A.
2. Cleanup goes through `CleanupTemplate(honorLive=true)`.
3. Attacker on node A creates a sandbox with `Labels["cube.master.pause.snapshot.id"] = <victim's snap-id>`.
4. Cleanup no-ops → victim’s **full VM memory pause package** stays on A (disk pressure / snapshot retention), even though no one is actually restoring from it on A.

This was mostly theoretical for S3 before because S3 always returned `false` from the keep predicate. This PR makes it real.

**Suggested mitigations (any one helps; combining is better):**
- Strip `cube.master.*` keys from user **`Labels`** at Create (mirror the Annotation filter), and/or
- On **cross-node origin cleanup**, use `honorLiveXFSPause=false` (origin has no live restore), and/or
- Tie the keep predicate to sandbox identity / Cubelet-stamped resume flow, not arbitrary Label values.

Please add a test that a forged **Label** (not just Annotation) cannot block GC of another sandbox’s pause package on the origin path.

ls-ggg:
Addressed in `a32ded9e`.

**Security.** Create now drops the same `cube.master.*` keys from **Labels** that were already rejected on Annotations (Master + Cubelet). Keep is keyed off the Cubelet-stamped restore-base, not a pause-id Label, so a forged Label cannot pin another tenant's pause package. A delayed Cleanup of the previous package still keeps it because restore-base still names that package while Pause restamps the pause id.

**Tier 2b.** If cloning the imported sandbox memory volume fails, CommitSandbox now Warnf-falls through to a full dump (same as Pause) instead of failing the RPC.

**Regression tests** (fail on `master`, pass here):

- `cleanupTemplate` + live pause catalog + honorLive=true → no-op; honorLive=false (next Pause / Destroy) still deletes
- forged pause-id Label does not block origin GC
- `prepareCommitMemoryArtifact`: catalog miss + imported volume → incremental; empty import → full; 2b clone failure → full
- Master `ConstructCubeletReq` strips the forged Labels and does not re-add the stripped Annotations

**Cluster e2e** after deploying this commit (one export at a time; s3lvol left running):

| Case | Result |
|---|---|
| Same-node Pause → Resume → Destroy | Pass. Package dir kept after Resume; live volumes only sandbox rootfs/memory/metadata; Destroy drops the dir and live volumes |
| Pause → Resume → Pause → Resume → Destroy | Pass. Second Pause GCs the first package dir and keeps the new one; Destroy clears both |
| Pause → DelPaused (no Resume) | Pass. PAUSED is not live, so keep does not apply; dir and live volumes gone |
| Resume → Snapshot | Pass. Package still present; Snapshot ready in ~4s (incremental, not a full memory dump); pause id not listed as a customer snapshot |
| Cross-node Resume | Pass on retry. Origin package/tombstone gone; target has no pause catalog; live volumes present; Destroy leaves no this-run sandbox volumes on the target |

Leftovers were only export-busy package objects (existing S3 delete-busy behavior). No unexpected lvols, no leftover package dirs or sandbox mounts, empty sandbox/snapshot lists, both nodes healthy and unisolated.

No new leak or lifecycle regression from the keep / Label-strip / 2b-fallthrough changes.

PR #1694, opened 2026-09-09
kernel: disable Guest RAID6 PQ startup benchmark

anti-entropy123:
Disable the optional RAID6 PQ implementation benchmark in Guest kernels while retaining RAID6 parity and recovery support. This avoids running an unrelated benchmark during each cube sandbox cold start.

## Benefit
In our test environment, this benchmark adds approximately 476 ms to Guest kernel startup. Disabling it is expected to avoid this overhead while keeping RAID6 support enabled.


lkml-likexu:
@anti-entropy123 Thanks for your contribution. Could you look into whether there are more config options like this that improve performance?

You may update guest cmdline to tune more details:
```
earlyprintk=serial,ttyS0,115200 console=ttyS0,115200 loglevel=8 ignore_loglevel initcall_debug rd.debug rd.shell rd.break=pre-pivot log_buf_len=1M
```

PR #1695, opened 2026-09-09
Validate original destinations before domain-based egress allow

UgOrange:
CubeEgress previously matched only the presented Host/SNI before forwarding to the original destination. For traffic reaching the L7 proxy, client hosts or `curl --resolve` overrides could therefore authorize an unrelated IP. Host-only HTTPS rules could also match one Host while forwarding another SNI on a shared IP.

Bind domain-constrained allow rules to trusted proxy-side DNS A records before connecting upstream or injecting credentials. Validate answer owner/class/section and CNAME chains; require HTTPS Host constraints to agree with SNI. Preserve the original IP/port and behavior for rules without domain constraints.

Asynchronous lookups have a five-second whole-operation deadline, five-alias limit, 16 active lookups per worker, and up to 256 coalesced same-name waiters. Shared positive caching uses the shortest A/CNAME TTL with a 300-second cap and elapsed lookup time deducted; zero and sub-millisecond TTLs are not cached. Timer-owned work and independent deadline cleanup handle client disconnects and callbacks dropped under resource pressure. Cache keys distinguish resolver configurations.

## Validation

- Lua port/scheme and DNS suites, Lua bytecode compilation, and `git diff --check` passed.
- Real curl HTTP/HTTPS with container `/etc/hosts` and `--resolve` overrides: all four forged-destination cases returned 200 on baseline `62da0187`, and 403 with the fix. Normal HTTP/HTTPS requests returned 200.
- Real DNS/OpenResty/TLS fixture tests cover injection, redacted audit, Host/SNI mismatch, CNAME ownership and expiry, TTL zero, NXDOMAIN, timeouts, IP literals, and explicit deny precedence. Denied requests did not increase upstream hit counts.
- 32 overlapping TTL-zero requests used one DNS query; 32 NXDOMAIN requests also used one query. Forty distinct concurrent misses used at most 16 active names. A slow CNAME chain was denied at approximately 5.02 seconds, with subsequent recovery. Initiator disconnection, dropped timers, late callbacks, and sub-millisecond TTLs have regression coverage.

Run `make -C CubeEgress test-lua` and `make -C CubeEgress test-dns-integration`. Reproduce the baseline with `python3 CubeEgress/tests/dns_auth_integration.py --baseline-ref 62da0187`.

## Scope and limitations

The container harness uses production access/audit modules and proxy locations, but fixture policy storage, static test certificates, direct sockets instead of TPROXY, and separate backend ports. It does not verify a deployed CubeSandbox guest, CubeVS/eBPF interception, or the patched production image. Test OpenResty is 1.31.1.1; the production base is 1.29.2.5 and was not rebuilt here.

CDN/split-DNS views and rotation may reject legitimate destinations. DNS membership does not establish tenant ownership or replace private-IP policy and TLS verification. IPv4 only; positive cache is shared across workers, while in-flight coalescing is per worker. English and Chinese guides document these boundaries.

### DIFF
--- a/CubeEgress/Makefile
+++ b/CubeEgress/Makefile
@@ -32,7 +32,8 @@ test-lua:
 	$(Q)LUA_BIN=$$(command -v luajit || command -v lua); \
 		test -n "$$LUA_BIN" || { echo "error: lua or luajit not installed"; exit 1; }; \
 		$$LUA_BIN tests/port_scheme_test.lua && \
-		$$LUA_BIN tests/port_scheme_extra_test.lua
+		$$LUA_BIN tests/port_scheme_extra_test.lua && \
+		$$LUA_BIN tests/dns_auth_test.lua
 
 .PHONY: build
 build:
--- a/CubeEgress/Makefile
+++ b/CubeEgress/Makefile
@@ -35,6 +35,10 @@ test-lua:
 		$$LUA_BIN tests/port_scheme_extra_test.lua && \
 		$$LUA_BIN tests/dns_auth_test.lua
 
+.PHONY: test-dns-integration
+test-dns-integration:
+	$(Q)python3 tests/dns_auth_integration.py
+
 .PHONY: build
 build:
 	$(call msg,BUILD IMAGE $(IMAGE_LOCAL):$(IMAGE_TAG)-$(ARCH))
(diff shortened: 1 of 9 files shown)


chenhengqi:
@UgOrange Could you please open an issue and let's gather some consensus before the fix? Thanks.

UgOrange:
@chenhengqi Opened #1697 with the reproduction, scope, and DNS/CDN trade-offs for discussion. We can use that issue to agree on the expected authorization behavior before proceeding with this PR.

PR #1701, opened 2026-09-10
fix(s3lvol): keep snapshot REF exports until snapshot delete

### LINKED ISSUE #1692: [S3 backend] Pause snapshot volumes stay allocated for ~1h (export TTL) after sandbox destroy/resume, and accumulate across repeated pause/resume cycles
**Components:** CubeS3lvol (pending-delete / export pin), Cubelet (S3 cleanup), CubeMaster (pause lifecycle)
**Environment:** v0.7.x, all-in-one single-node deployment, S3 storage backend

## Summary

After a sandbox on the S3 backend is paused, the three temporary snapshot volumes created by that pause (`tpl-snap-<id>-memory-snap`, `tpl-snap-<id>-rootfs`, `s3-meta-snap-<id>-snap`) are **not released immediately** when the sandbox is resumed or destroyed. Their deletion is deferred until the S3 export created during that pause expires (default TTL 3600s), at which point the s3lvol background poller finally completes it. Until then, the S3 objects and lvstore clusters stay allocated.

When the same sandbox goes through several pause/resume cycles within one export TTL window, every cycle leaves one such deferred group behind, so the **peak storage usage grows linearly with the number of cycles**.

## Steps to reproduce

Scenario A — single destroy/resume:

1. Create a sandbox with the S3 backend and pause it (pause completes, snapshot export finishes).
2. Resume it (or destroy the paused sandbox directly).
3. The sandbox's own live volumes (`sb-<id>-*`) are deleted immediately, but the three pause snapshot volumes show up as pending-delete (reason=export).
4. For at least the next hour (until the export TTL expires), the S3 object count/bytes do not drop.
5. Only after the export expires (plus up to one 60s poller tick) are the three volumes actually deleted and the space reclaimed.

Scenario B — repeated pause/resume:

1. Cycle the same sandbox pause → resume → pause → resume … with intervals shorter than the export TTL.
2. Every resume's cleanup request gets parked in the pending-delete queue, so several snapshot groups coexist in the lvstore.
3. Usage ≈ (number of groups) × (size per group), until each group's export expires in turn.

## Expected behavior

Once a sandbox is resumed or destroyed, the temporary snapshots of that pause should be deleted and their storage released promptly (seconds to minutes), not deferred by an export TTL that is unrelated to the sandbox lifecycle at that point.

## Actual behavior

- The delete request is refused by s3lvol and parked in the pending-delete queue until the export TTL lapses.
- The cleanup caller (Cubelet) treats the refusal as success, so the control plane has no way to notice the delay.

## Root cause analysis

Three layers stack up:

1. **Pause snapshots and their exports share one pin lifecycle.**
   When a pause completes (S3 remote-ready), s3lvol creates a ref-layout export for each snapshot volume to enable cross-node restore. Later, the cleanup path issues `rcow_delete_lvol`, and `s3lvol_lvol_destroy` sees the snapshot is behind an export pin (`export_pin_verdict` returns LEASE: no lease object exists yet, TTL not expired), refuses the delete, and records a pending-delete intent (reason=export) in `vbdev_s3lvol_pending.c`.

2. **Cubelet has no way to release an export, and treats "busy" as success.**
   The cubecow engine does not expose a release_export call. `S3Cow.DeleteByKind` (`Cubelet/storage/s3_volume_manager.go`) treats a "device busy" refusal for a snapshot as already deleted (it logs "recording it as deleted and leaking the object"), then continues removing the local directory and catalog and returns success. From the control plane's point of view the cleanup is done, while the objects actually sit in s3lvol's pending queue.

3. **The pending queue's auto-complete condition is exactly the export TTL.**
   The pending poller (60s period) only completes the delete once the export pin turns STALE (TTL expired and the lease miss confirmed). In other words, **the lower bound of the reclaim delay is the export TTL (default 3600s, `S3LVOL_EXPORT_DEFAULT_TTL_SEC`)**.

Representative target-side log lines during a destroy/resume cleanup:

```
ERROR: lvol 'tpl-snap-<id>-memory-snap' is the snapshot behind export <uuid>,
       which another node may be reading through. Release that export, or wait
       for it to expire, before deleting this.
NOTICE: delete of 'tpl-snap-<id>-memory-snap' is pending: export
        (will complete once the blocker clears)
```

`rcow_get_pending_deletes` shows the three volumes with reason=export; `rcow_get_exports` shows the matching exports with `lease_absent=true`, `pin=lease`, `expires_at ≈ pause time + 1h`.

## Impact

- Every pause/resume or destroy of a paused sandbox leaves the temporary snapshot space allocated for up to one hour.
- Under frequent pause/resume (e.g. auto-pause scheduling), the peak storage footprint grows with the number of cycles and is not bounded.
- Cleanup reports success while the objects remain, so storage monitoring and capacity planning disagree with reality.

## Suggested fixes (for discussion)

1. **Root fix: release the export before deleting the snapshots.**
   `rcow_release_export` already exists on the s3lvol side but is not exposed or called by Cubelet/cubecow. If the pause-package cleanup (resume/destroy → `CleanupPauseSnapshot` → Cubelet `CleanupTemplate`) released the corresponding exports before deleting the snapshots, the pin would clear immediately and the pending poller would finish the delete on its next tick (≤60s), with no dependence on the TTL.
2. **Alternative: deterministic sweep of the previous pause group when the next pause starts.**
   A successful resume means the previous pause package is no longer needed; reclaiming it at the start of the next pause would bound the accumulation to one group.
3. **Semantics: stop treating "busy" as success in Cubelet.**
   `DeleteByKind` should surface export-pinned refusals (or wait for / trigger the pending completion) so the upper layer can observe the real reclaim progress instead of silently recording "deleted + leaked".

## Relevant code paths

- `CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_lvstore.c` — `s3lvol_lvol_destroy` (refuses deletes behind an export pin)
- `CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c` — pending-delete recording + 60s poller (`pending_lvol_deletable`)
- `CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c` — `export_pin_verdict` / lease semantics
- `CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol.h` — `S3LVOL_EXPORT_DEFAULT_TTL_SEC = 3600`
- `Cubelet/storage/s3_volume_manager.go` — `DeleteByKind` / `isS3DeleteBusy` (busy → recorded deleted + leaked)
- `Cubelet/services/cubebox/template_ops.go` — `cleanupTemplate` (pause package cleanup entry)
- `CubeMaster/pkg/pausesnap/delete_paused.go` — destroy paused / `CleanupPauseSnapshot`
- `CubeMaster/pkg/service/sandbox/sandbox_resume_pause.go` — cleanup invocation after a successful resume

changpe1:
## Summary
- Snapshot-backed REF exports stay valid for the snapshot lifetime: `ttl_sec` is accepted but ignored, and new manifests use `expires_at=0`. Cubelet does not need `rcow_release_export`.
- Deleting the snapshot releases remaining REF exports internally, including pre-lease registry entries. A live importer lease, local esnap reader, clone_count, decouple, or in-flight export still defers the delete.
- A missing manifest is treated as already released (drop the registry entry and finish the destroy). Export lookup prefers `snapshot_uuid` so a recreated snapshot does not inherit a stale uuid; older registries fall back to `blob_id`.
- Re-exporting the same snapshot returns the existing uuid (in flight or after DONE, including a materialised rewrite of that uuid). A later export is allowed only after that uuid is released.
- A queued decouple is removed before `decouple_start()` so a synchronous completion cannot double-free the queue node; `-EBUSY` requeues at the head. Lease grace is measured from the first miss of this absence; `deletable` uses the cached lease poll.

Fixes #1692.

## Test plan
- [x] `run_pending_delete_test.sh` (131 passed) — TTL ignored, snapshot delete releases the export, late importer still pins, pre-lease delete, missing-manifest 404
- [x] `run_selfimport_test.sh` (28 passed) — snapshot delete revokes the export; reused snapshot name does not revive it
- [x] `run_export_test.sh` (197 passed) — idle vs leased deletable, one export per snapshot, `expires_at=0`
- [x] `make check` in `CubeS3lvol/` (37/38; `run_cubecow_client_test.sh` NVMe node flake, 27/27 on retry)

Assisted-by: Cursor Grok 4.6


CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -895,39 +741,16 @@ export_pin_verdict(const struct s3lvol_export *exp)
...
-				return export_lease_miss_confirms_gone(
-					       true, exp->lease_absent_at,
-					       exp->expires_at,
-					       (uint64_t)time(NULL))
-				       ? S3LVOL_EXPORT_PIN_STALE
-				       : S3LVOL_EXPORT_PIN_LEASE;
-			}
-			return s3lvol_export_is_expired(exp)
-			       ? S3LVOL_EXPORT_PIN_STALE
+			return exp->lease_absent ? S3LVOL_EXPORT_PIN_STALE
```
fslongjin: A confirmed lease miss returns STALE with no time condition — the age check at :759 only runs when `lease_updated_at != 0`. A miss is not proof: an older importer reading a manifest with `expires_at == 0` skips the lease by design (`master xfer.c:336`), and any importer whose first lease PUT has not landed yet also has none. Master gated this on `export_lease_miss_confirms_gone()`, which required the miss to be observed at or after the deadline; this PR removes that helper and its test. Could the `lease_absent` branch get a time condition back — a miss only counts once it is older than N lease periods — or keep publishing a non-zero `expires_at` so pre-lease importers still write a lease?

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -998,15 +842,11 @@ s3lvol_export_pinning(struct s3lvol_lvstore *lvs, const char *snapshot_name)
 			/* An importer is reading, or may be. */
 			return exp;
 		case S3LVOL_EXPORT_PIN_LEGACY:
-			/* No lease: the TTL decides, as it always has. */
-			if (!s3lvol_export_is_expired(exp)) {
-				return exp;
-			}
+			/* Status still says liveness is unknown. An explicit snapshot
```
fslongjin, line 893: This used to be `if (!s3lvol_export_is_expired(exp)) return exp;` — a pre-lease export was pinned until its TTL ran out, which is exactly the window before a pre-lease reader's first lease PUT lands. The comment above says *"Status still says liveness is unknown"* and the next line treats that as grounds to proceed. If "an explicit snapshot delete is the revocation" is the intended semantics, it needs to be in the user docs.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -449,23 +419,18 @@ pending_lvol_deletable(struct s3lvol_lvstore *lvs, struct spdk_lvol *lvol,
...
-	 * So LEGACY is skipped whether or not the TTL has passed, which also
-	 * covers an entry that was recorded before the first lease GET had
-	 * answered. */
+	/* A stale or missing lease, and a pre-lease export, let the pending
+	 * delete proceed: the recorded intent is the revocation. */
 	switch (s3lvol_export_pin_state(lvs, lvol->name)) {
 	case S3LVOL_EXPORT_PIN_LEASE:
 		*why = "an importer may still be reading an export of it";
 		return false;
 	case S3LVOL_EXPORT_PIN_LEGACY:
```
fslongjin, line 435: With this, the poller completes a pre-lease export unattended — no human in the loop — and `pending_reason_auto()` includes EXPORT_LEGACY. `destroy_mark_pending()` now always records EXPORT, so the only way to reach a LEGACY verdict here is a registry entry restored from before leases existed, i.e. the rolling-upgrade case.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -405,33 +402,10 @@ import_lease_start(struct s3lvol_import *imp)
...
-	now = (uint64_t)time(NULL);
-	remaining = (m->expires_at > now) ? (m->expires_at - now) : 0;
-	ttl = remaining / 3;
-	if (ttl < S3LVOL_LEASE_RENEW_MIN_SEC) {
-		ttl = S3LVOL_LEASE_RENEW_MIN_SEC;
-	}
-	imp->lease_interval_us = ttl * SPDK_SEC_TO_USEC;
+	/* A fixed cadence makes leases independent of export age. It is reported as
+	 * renew_s, and the source applies a grace of at least three intervals. */
+	interval = S3LVOL_LEASE_RENEW_MIN_SEC;
```
fslongjin, line 407: Dropping `remaining_ttl / 3` makes `renew_s` a constant 20 s; master defaulted to 1200 s. The source derives `grace = 3 * renew_s`, floored at `S3LVOL_LEASE_MIN_GRACE_SEC` (60 s), so every source's protection window shrinks from 3600 s to 60 s — including sources still running the old build, which never got the memo. That window is the only thing absorbing a stalled lease PUT.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -583,31 +489,15 @@ s3lvol_export_lease_stop(struct s3lvol_export *exp)
...
-	 * allows. */
-	remaining = (exp->expires_at > (uint64_t)time(NULL))
-		    ? (exp->expires_at - (uint64_t)time(NULL)) : 0;
-	interval = remaining / 3;
-	if (interval < S3LVOL_LEASE_RENEW_MIN_SEC) {
-		interval = S3LVOL_LEASE_RENEW_MIN_SEC;
-	}
+	/* Poll at the same fixed cadence importers renew. The grace period is still
+	 * derived from the importer's reported renew_s, not this local timer. */
+	interval = S3LVOL_LEASE_RENEW_MIN_SEC;
```
fslongjin, line 513: Same cadence change here: master computed `remaining / 3` (~20 minutes for a default export), so this is now one HEAD per export per 20 s — 60× the request rate against the bucket for exports nobody is importing. Not wrong, but worth being deliberate about.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -1265,16 +1125,10 @@ exports_reap(void *arg)
...
-				/* Even if this tick's batch is full, the watch
-				 * has nothing left to learn. */
-				export_lease_stop_if_terminal(exp);
-			}
 			if (n >= S3LVOL_EXPORT_REAP_BATCH) {
 				continue;
 			}
-			if (export_snapshot_alive(exp) &&
-			    !export_lease_is_terminal(exp)) {
+			if (export_snapshot_alive(exp)) {
```
fslongjin, line 1162: With the `!export_lease_is_terminal(exp)` half gone, the reaper only reclaims an entry whose snapshot is already gone, so nothing bounds the registry while snapshots keep existing. `S3LVOL_MAX_EXPORTS` (256) is a fixed decode table with no cap check on publish; once the count passes it, `exports.json` still writes but the next attach fails to decode and takes the lvstore down with it.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -3761,6 +3765,10 @@ decouple_start(struct s3lvol_lvstore *lvs, struct spdk_lvol *lvol,
 	struct s3lvol_decouple *d;
 	uint64_t i;
 
+	if (lvol->action_in_progress) {
```
fslongjin, line 3773: The guard is right, but the only other caller is `decouple_start_next_queued()`, which treats any non-zero return as terminal — it removes and frees the entry, then reports the error. So a queued decouple that meets another `action_in_progress` holder (a `rcow_resize_lvol`, say) at drain time is dropped for good rather than left queued. For a queue, "try again later" is closer to the intent.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_lvstore.c
```diff
@@ -3008,47 +3132,10 @@ s3lvol_lvol_destroy(struct spdk_lvol *lvol,
...
-			    "be deleted yet\n", lvol->name);
-		/* The decouple that is running will clear, after which the delete can
-		 * succeed; record the intent for --retry-pending, same contract as
-		 * the export-pin refusals above. */
-		destroy_mark_pending(lvol, S3LVOL_PENDING_DECOUPLE);
-		return -EBUSY;
-	}
+	/* snapshot_destroy_check_clones() runs before any export is released, so a
+	 * synchronous clone-count refusal leaves every manifest intact. Once the
+	 * release chain has started there is no rollback: an asynchronous blob
```
fslongjin, line 3142: Accurate comment, and good to have — but it documents a state this PR creates: the export is revoked, the snapshot survives, and nobody can import it (the manifest is gone). The operator gets a `PENDING_FAILED` with no hint that re-exporting is the way out.

CubeS3lvol/test/dataplane/run_pending_delete_test.sh
```diff
@@ -1153,38 +1191,28 @@ for _ in $(seq 30); do
...
+sleep "$((LATE_TTL + 3))"
+if rpc rcow_get_snapshot_status \
 		"$(printf '{"export_uuid":"%s"}' "${LATE_UU}")" >/dev/null 2>&1; then
-	pass "the export was not reaped after a pre-deadline 404 plus expiry"
+	pass "the export remains valid after the old ttl_sec"
 else
-	fail "the export was reaped despite a lease PUT after the 404"
+	fail "the export was reaped while its snapshot still exists"
 fi
 lvol_exists lateimp \
```
fslongjin, line 1238: Step [12] still never issues a snapshot delete — it releases the export at :1227 — so `lvol_exists lateimp` is true whether or not the late lease was treated as a live reader. Adding the delete is what would make the step test the thing its header describes.

CubeS3lvol/test/dataplane/run_selfimport_test.sh
```diff
@@ -508,20 +480,12 @@ rpc rcow_create_snapshot '{"lvol_name":"tmp3","snapshot_name":"snap_dup"}' >/dev
...
 rpc rcow_deactive_bdev '{"device_name":"tmp3"}' >/dev/null 2>&1
 
-OUT="$(rpc rcow_import_lvol \
-	"$(printf '{"lvol_name":"c_replaced","export_uuid":"%s"}' "${U3}")" 2>&1)"
-MODE="$(import_mode "${OUT}")"
-if [ "${MODE}" = "esnap" ]; then
-	pass "mode=esnap: the blob id did not match, so the name was not trusted"
-else
-	fail "mode='${MODE}': it cloned the replacement, which is not what was exported"
+if OUT="$(rpc rcow_import_lvol \
```
fslongjin, line 503: The old version asserted `mode=esnap` and grepped the log for the blob-id mismatch; this version only asserts the import fails. But the delete above already released the export, so the import fails before `import_local_parent()`'s blob-id guard is reached — the guard this step exists to exercise.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -1872,6 +1839,32 @@ s3lvol_lvol_export(struct s3lvol_lvstore *lvs, struct spdk_lvol *snapshot,
...
+
+	published = s3lvol_export_find_for_snapshot(lvs, snapshot->name);
+	if (published) {
+		s3lvol_export_get(published, &info);
+		if (export_uuid && export_uuid[0] != '\0' &&
+		    strcmp(info.export_uuid, export_uuid) != 0) {
+			SPDK_ERRLOG("snapshot '%s' is already exported as %s, "
+				    "not %s\n", snapshot->name,
+				    info.export_uuid, export_uuid);
+			return -EEXIST;
```
fslongjin, line 1859: The `-EEXIST` branch for a mismatched explicit `export_id` has no test — nothing under `CubeS3lvol/test/` passes an `export_id`. Given it is the one path where the idempotency rule can reject a caller, worth a case in `run_export_test.sh`.

fslongjin (commented):
Reviewed at e691ced0.

**One issue blocks: `export_pin_verdict()` treats "no lease observed" as proof that nobody is reading, with no time condition on that branch — and this revision applies the same rule to pre-lease exports by removing the TTL pin they used to have.**

Three parts of the old rule are gone in this PR:

- `export_lease_miss_confirms_gone()` is deleted, along with its test (`test/integration/s3_lease_term_test.c`) and `export_lease_term.h`. In master a confirmed miss only counted once it was observed at or after the export's deadline. Now any miss is final.
- `s3lvol_export_pinning()` no longer pins a legacy export: master had `if (!s3lvol_export_is_expired(exp)) return exp;` (`exports.c:845`, removed here).
- `pending_lvol_deletable()` now treats LEGACY as deletable, so the poller completes it unattended (`pending.c:427`).

"No lease yet" is ambiguous, and the code says so itself — the new comment on the LEGACY branch reads *"Status still says liveness is unknown"* and the next line falls through to deletable. Two readers leave no lease at the moment the check runs:

- **An importer reading a manifest with `expires_at == 0`.** Publishing a zero is now deliberate (`xfer.c:2194`), but the older build skips writing a lease for exactly that shape (`master xfer.c:336`). On a source-first rolling upgrade the source 404s the lease, marks the export absent, and reads it as STALE forever — for a reader that is actively reading.
- **A reader whose first lease PUT has not landed yet.** Master covered this with the TTL pin above. Nothing covers it now.

The delete then goes through, and it is silent (`rcow_get_snapshot_status` reports `deletable: YES` throughout) and replayable unattended (`pending_reason_auto()` includes both EXPORT and EXPORT_LEGACY).

No redesign needed — the rule just needs a time dimension, the way master had one. Either gate the `lease_absent` branch on age (a miss counts only once it is older than N lease periods), or keep writing a non-zero `expires_at` so pre-lease importers still write a lease.

**Not blocking:**

- `renew_s` is now a constant 20 s (`xfer.c:407`), where master derived it from the remaining TTL (1200 s by default). The source computes `grace = 3 * renew_s` floored at 60 s, so every source's protection window drops from 3600 s to 60 s — the window that absorbs a stalled lease PUT, consumed by the pending poller with nobody watching.
- The source-side poller picks up the same 20 s (`exports.c:500`), so an export nobody is importing now costs one HEAD per 20 s instead of one per 20 minutes.
- `exports_reap()` no longer reclaims a terminal-lease entry (`exports.c:1131`), so nothing bounds the registry while snapshots keep existing. `S3LVOL_MAX_EXPORTS` is a fixed decode table with no cap check on publish; past it, `exports.json` writes fine but the next attach fails to decode. The table and the missing check predate this PR, but the reap rule you removed was what kept the count away from it.
- `decouple_start()`'s `-EBUSY` (`xfer.c:3768`) makes `decouple_start_next_queued()` drop a queued entry for good — it treats any non-zero return as terminal. A queued decouple that meets an `rcow_resize_lvol` at drain time is lost, not retried.
- Docs: `docs/guide/cross-node-snapshot.md:322` (and the zh twin) still says a snapshot that "has been exported" is refused with `EBUSY`; `CubeS3lvol/README.md:533`'s `deletable: YES` summary omits the local-esnap-reader blocker that the field description above it lists; and `README.md:516` still says a failing bdev unregister is not recorded, although `vbdev_s3lvol_lvstore.c:2889` records it.

**Tests.** Step [14] is a good addition and does cover the missing-manifest fix. Two things I would still like:

- a snapshot with a queued decouple, deleted → assert the decouple is not started and the run is ASAN-clean;
- a delete while a reader that writes no lease is importing → assert it is refused or deferred.

Also: step [12] (`run_pending_delete_test.sh:1217`) still never issues a snapshot delete, so `lvol_exists lateimp` is true regardless; and `run_selfimport_test.sh`'s steps [4]/[5] (`:483`) now delete the snapshot before asserting the import fails, so `import_local_parent()`'s blob-id guard is no longer exercised.

Happy to re-review as soon as the pin verdict has a time condition.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -3895,26 +3911,35 @@ decouple_start_next_queued(void)
 			continue;
 		}
 
-		TAILQ_REMOVE(&g_decouple_queue, q, link);
-		cb_fn  = q->cb_fn;
-		cb_arg = q->cb_arg;
-
-		rc = decouple_start(q->lvs, q->lvol, q->uuid_str, cb_fn, cb_arg);
+		rc = decouple_start(q->lvs, q->lvol, q->uuid_str, q->cb_fn, q->cb_arg);
```
fslongjin, line 3918: **Should fix (memory safety).** This ordering is a regression from the previous revision, which removed `q` from the queue *before* calling `decouple_start()`. Keeping `q` linked is what makes `-EBUSY` retryable, but it also means that if `decouple_start()` returns 0 and completes synchronously, the completion re-enters `decouple_start_next_queued()` while `q` is still linked — the drain picks `q` up again, removes and frees it, and then the outer frame's `TAILQ_REMOVE`/`free` at :3916 hits freed memory. Synchronous completion is real: the SPDK patches call their callback inline on `-ENOMEM`/`-EINVAL`, and `decouple_cluster_done()` calls `decouple_finish()` inline on error. Remove before the call, and re-insert with `TAILQ_INSERT_HEAD` on `-EBUSY` — that keeps both "don't lose the entry" and "re-entry safe".

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -292,50 +264,28 @@ struct export_lease_get_ctx {
...
-
+/* Record that the last completed lease lookup found no current reader.
+ *
+ * lease_absent_at is the first miss in this absence, not the most recent one.
+ * Refreshing it on every 20 s poll would keep miss_age under one interval and
+ * the grace would never elapse. */
 static void
 export_lease_note_absent(struct s3lvol_export *exp, uint64_t submitted_at)
 {
+	if (!exp->lease_absent || exp->lease_absent_at == 0) {
```
fslongjin, line 283: Pinning the anchor to the *first* miss is necessary (refreshing it each poll would keep `miss_age` under one interval and the grace would never elapse), but it has a consequence worth deciding on: for an export that has been idle for a while, `miss_age` is already past `S3LVOL_LEASE_MIN_GRACE_SEC`, so `export_pin_verdict` returns `STALE`. A newly admitted importer's first lease PUT is therefore covered only until the next poll — and the delete path never forces a fresh GET, it reads this cached verdict (`s3lvol_export_pinning` from `xfer.c:2264`, and `s3lvol_export_pin_state` from the pending poller). Master's TTL pin covered that window. Could the delete path re-check the lease before releasing, or is the window meant to be accepted?

CubeS3lvol/README.md
```diff
@@ -228,34 +228,46 @@ copying them:
...
 scripts/s3lvol_rpc.py rcow_export_snapshot '{"snapshot_name":"snap0"}'
 ```
 
+`ttl_sec` is accepted for compatibility but ignored. New reference manifests use
+`expires_at: 0`; their lifetime follows the source snapshot.
+
+Rolling upgrade: bring **importers** to this build before sources that publish
+`expires_at: 0`. An older importer skips writing a lease for that shape, so a
+source-first upgrade would treat a still-reading node as absent after the
+grace window. Same-version importers renew every 20 s; the source pins a miss
```
fslongjin: "the source pins a miss for at least 60 s so the first lease PUT can land" only holds when the reader arrives during the first 60 s of the absence. Once an export has been idle past that (see `exports.c:279`), the miss is already older than the grace, so the first lease PUT of a new importer is covered only up to the source's next poll. Worth describing what actually happens. The same goes for `:265`'s "never cached": the verdict is recomputed per call, but from the cached lease state.

fslongjin (commented):
Re-reviewed at 8110754e. The blocker from the previous revision is fixed: the `lease_absent` branch now has a 60 s grace (`export_pin_verdict`, `exports.c:773`), and an importer of this build writes a lease for an `expires_at == 0` manifest (`import_lease_start`, `xfer.c:325`). The decouple queue also no longer drops an entry on `-EBUSY`. Good direction.

Two things left before merge.

**1. A regression introduced by the `-EBUSY` fix — heap double-free.**
`decouple_start_next_queued()` (`xfer.c:3914`) now calls `decouple_start()` first and `TAILQ_REMOVE`s after:

```c
rc = decouple_start(q->lvs, q->lvol, q->uuid_str, q->cb_fn, q->cb_arg);
if (rc == 0) { TAILQ_REMOVE(&g_decouple_queue, q, link); free(q); return; }
```

The previous revision removed `q` *before* the call, which is what made it safe against re-entry. Keeping `q` linked across the call is what makes `-EBUSY` retryable, but it also means that if `decouple_start()` completes synchronously and re-enters the drain, `q` is removed and freed twice. The synchronous path exists: the SPDK patches invoke their callback inline on `-ENOMEM`/`-EINVAL` (`patches/0002`, `patches/0003`), and `decouple_cluster_done()` finishes inline on error — so this is reachable under memory pressure. A fix keeps both properties: remove before the call, re-insert on `-EBUSY`.

**2. The grace window does not do what the README says it does.**
`lease_absent_at` is the *first* miss of an absence (`exports.c:279`) and is only cleared by a successful GET (`exports.c:387`). For an export that has been idle for a while, `miss_age` is already past 60 s, so the verdict is `STALE`. When a new importer is then admitted and writes its first lease, the source keeps reading its cached verdict until the next poll — so the reader is unprotected for up to one poll interval, and for longer if its first lease PUT is slow. Master covered exactly that with the TTL pin. `README.md:247`'s "the source pins a miss for at least 60 s so the first lease PUT can land" therefore does not hold in the steady state, and `:265`'s "never cached" is inaccurate — the verdict is recomputed, but from cached lease state. Either force a fresh lease GET on the delete path, or fix the wording.

**Smaller, no need to block:** `renew_s` is now a constant 20 s (`xfer.c:407`), so every source's protection window drops from 3600 s to 60 s; the source poller picks up the same 20 s (`exports.c:509`), i.e. one HEAD per export per 20 s for exports nobody imports; and `exports_reap()` no longer reclaims a terminal-lease entry, so nothing bounds the registry while snapshots keep existing — `S3LVOL_MAX_EXPORTS` (256) is a fixed decode table with no cap check on publish, and the entry that overflows it makes the next attach fail to decode.

**Tests.** Step [14] is a good addition. Still missing: (a) a snapshot with a queued decouple, deleted — assert the decouple is not started and the run is ASAN-clean; that would have caught the double-free above. (b) `run_pending_delete_test.sh` step [12] still never issues the snapshot delete it is named for, so `lvol_exists lateimp` is true regardless. (c) `run_selfimport_test.sh` steps [4]/[5] now delete the snapshot before asserting the import fails, so `import_local_parent()`'s blob-id guard is no longer reached.

Fix the 3 lines in (1) plus the wording in (2) and this is good to go.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
fslongjin, line 407: Dropping `remaining_ttl / 3` makes `renew_s` a constant 20 s; master defaulted to 1200 s. The source derives `grace = 3 * renew_s`, floored at `S3LVOL_LEASE_MIN_GRACE_SEC` (60 s), so every source's protection window shrinks from 3600 s to 60 s — including sources still on the old build, which never got the memo. The number of missed renewals tolerated is unchanged (3); only the wall-clock allowance is. Fine if deliberate, worth saying so.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -583,31 +498,15 @@ s3lvol_export_lease_stop(struct s3lvol_export *exp)
...
-	 * allows. */
-	remaining = (exp->expires_at > (uint64_t)time(NULL))
-		    ? (exp->expires_at - (uint64_t)time(NULL)) : 0;
-	interval = remaining / 3;
-	if (interval < S3LVOL_LEASE_RENEW_MIN_SEC) {
-		interval = S3LVOL_LEASE_RENEW_MIN_SEC;
-	}
+	/* Poll at the same fixed cadence importers renew. The grace period is still
+	 * derived from the importer's reported renew_s, not this local timer. */
+	interval = S3LVOL_LEASE_RENEW_MIN_SEC;
```
fslongjin, line 513: Same cadence change: master computed `remaining / 3` (~20 minutes for a default export), so this is now one HEAD per export per 20 s — 60x the request rate against the bucket for exports nobody is importing.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_pending.c
```diff
@@ -449,23 +421,18 @@ pending_lvol_deletable(struct s3lvol_lvstore *lvs, struct spdk_lvol *lvol,
...
 
-	/* Asked as a state rather than a yes/no, because the two ways an export
-	 * stops pinning are not equally good. A stale lease is evidence: an
-	 * importer wrote it and stopped renewing, so nobody is reading. An export
-	 * with no lease is not -- its TTL lapses on its own, and completing a
-	 * delete on that basis is exactly the hazard that refuses to automate.
-	 * So LEGACY is skipped whether or not the TTL has passed, which also
-	 * covers an entry that was recorded before the first lease GET had
-	 * answered. */
+	/* A stale or missing lease, and a pre-lease export, let the pending
```
fslongjin, line 429: **Polish.** With this, a pre-lease export is completed by the poller unattended — and `pending_reason_auto()` includes `EXPORT_LEGACY`. "An explicit snapshot delete is the revocation" is defensible, but it is a user-visible behaviour change; please make sure the user-facing docs state it, not only the code comment.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -998,15 +865,11 @@ s3lvol_export_pinning(struct s3lvol_lvstore *lvs, const char *snapshot_name)
 			/* An importer is reading, or may be. */
 			return exp;
 		case S3LVOL_EXPORT_PIN_LEGACY:
-			/* No lease: the TTL decides, as it always has. */
-			if (!s3lvol_export_is_expired(exp)) {
-				return exp;
-			}
+			/* Status still says liveness is unknown. An explicit snapshot
+			 * delete is the revocation, same as STALE. */
```
fslongjin: **Polish.** The comment is accurate and the decision is documented here, but this is the place the "delete is the revocation" semantics actually lives, and it is silent: `rcow_get_snapshot_status` reports `deletable: YES` throughout, so an operator deleting an export a legacy reader is still reading has no signal. A note in the user docs would close the loop.

changpe1:
All dataplane regression tests are passed.

fslongjin (commented):
Re-reviewed at ec63d52b. The blocker from the last revision is gone: `decouple_start_next_queued()` removes the entry before the call again and keeps `-EBUSY` retryable with `TAILQ_INSERT_HEAD` (`xfer.c:3917-3925`) — the base ordering plus the retry, so the re-entrancy double-free is fixed. The `snapshot_uuid` registry identity is a good hardening against blob_id reuse; I went through the memory and serialization paths and they are clean. Test step [12] in `run_pending_delete_test.sh` and step [4] in `run_selfimport_test.sh` now do what they say.

One wording fix, plus a few decisions to record.

**The protection window is stated in reverse.** `README.md:251` (with `docs/guide/cross-node-snapshot.md:322`, `docs/zh/guide/cross-node-snapshot.md:363`) says a newly admitted importer of an already-idle export is "protected only until the source's next lease poll (20 s)". It is the other way round. Until the poll observes the PUT, `export_pin_verdict()` still sees `lease_updated_at == 0` with a miss older than the grace, so the verdict is `STALE` and the importer is **unprotected**; protection begins at that poll, once `lease_updated_at != 0`. Suggested wording: "unprotected until the source's next lease poll, when the source first sees the PUT." The identical inverted phrasing is in the code comment at `exports.c:276-277`, which is where the docs took it from — worth fixing there too.

**Decisions to record (not blockers):**

- the delete path decides from the cached poll, so a reader of an already-idle export has an unprotected window of up to one poll interval (<= 20 s; longer if its first PUT is slow). Master's TTL pin covered that. It is now documented; is that the intended trade-off, or should the delete path force a fresh lease GET before releasing?
- `renew_s` is a constant 20 s (`xfer.c:407`), so with `grace = 3 * renew_s` floored at 60 s every source's protection window is 60 s where master allowed 3600 s — including sources still on the old build.
- the source poller is 20 s (`exports.c:513`): one HEAD per export per 20 s for exports nobody is importing (master ~20 min).
- `exports_reap()` only reclaims an entry whose snapshot is gone (`exports.c:1162`), so nothing bounds the registry while snapshots exist; `S3LVOL_MAX_EXPORTS` (256) is a fixed decode table with no cap check on publish, and the entry that overflows it makes the next attach fail to decode.

**Tests still missing:** a queued decouple meeting a snapshot delete (ASAN), and a direct test of the registry-side `snapshot_uuid` round-trip.

Fix the wording above and this is good to go.

CubeS3lvol/README.md
```diff
@@ -228,34 +228,54 @@ copying them:
...
+`expires_at: 0`; their lifetime follows the source snapshot.
+
+Rolling upgrade: bring **importers** to this build before sources that publish
+`expires_at: 0`. An older importer skips writing a lease for that shape, so a
+source-first upgrade would treat a still-reading node as absent after the
+grace window. Same-version importers renew every 20 s. A confirmed miss younger
+than 60 s still pins — that covers a first lease PUT that arrives in the first
+minute after publish (or after a live lease disappears). Once the export has
+been idle longer than that, the miss is already older than the grace: a newly
+admitted importer is protected only until the source's next lease poll (20 s),
```
fslongjin: **Should fix.** This states the window backwards. For an export that has been idle longer than the grace, `export_pin_verdict()` returns `STALE` while `lease_updated_at == 0` (`exports.c:766-781`), so a newly admitted importer is **unprotected** until the poll observes its PUT — protection *begins* at that poll, once `lease_updated_at != 0` (`exports.c:786-798`). "protected only until the next lease poll" reads as if the reader is safe for up to 20 s after admission, which is exactly the exposed interval. The same inverted sentence is in `docs/guide/cross-node-snapshot.md:322`, `docs/zh/...:363`, and the code comment at `exports.c:276-277`. Suggested: "unprotected until the source's next lease poll, when the source first sees the PUT."

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -895,40 +754,31 @@ export_pin_verdict(const struct s3lvol_export *exp)
...
+
+			if (!exp->lease_absent) {
+				return S3LVOL_EXPORT_PIN_LEASE;
+			}
+			now = (uint64_t)time(NULL);
+			if (exp->lease_absent_at == 0 || now < exp->lease_absent_at) {
+				return S3LVOL_EXPORT_PIN_LEASE;
+			}
+			miss_age = now - exp->lease_absent_at;
+			if (miss_age < S3LVOL_LEASE_MIN_GRACE_SEC) {
```
fslongjin, line 778: The grace itself is right. The consequence worth deciding on is upstream of here: nothing forces a fresh lease GET on the delete path, so this verdict is always the last poll's. A reader admitted into an export that has already been idle past the grace therefore has an unprotected window of up to one poll interval (<= 20 s, longer if its first PUT is slow) — the window master's TTL pin covered. It is now documented; do you want the delete path (`s3lvol_lvol_destroy_impl` -> `s3lvol_export_pinning`) to force a fresh GET before releasing, or is the <= 20 s window the accepted trade-off?

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
fslongjin, line 407: Dropping `remaining_ttl / 3` makes `renew_s` a constant 20 s; master defaulted to 1200 s. The source derives `grace = 3 * renew_s`, floored at `S3LVOL_LEASE_MIN_GRACE_SEC` (60 s), so every source's protection window shrinks from 3600 s to 60 s — including sources still on the old build. The number of missed renewals tolerated is unchanged (3); only the wall-clock allowance is. Fine if deliberate, worth saying so in the description.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -583,31 +502,15 @@ s3lvol_export_lease_stop(struct s3lvol_export *exp)
...
-	 * allows. */
-	remaining = (exp->expires_at > (uint64_t)time(NULL))
-		    ? (exp->expires_at - (uint64_t)time(NULL)) : 0;
-	interval = remaining / 3;
-	if (interval < S3LVOL_LEASE_RENEW_MIN_SEC) {
-		interval = S3LVOL_LEASE_RENEW_MIN_SEC;
-	}
+	/* Poll at the same fixed cadence importers renew. The grace period is still
+	 * derived from the importer's reported renew_s, not this local timer. */
+	interval = S3LVOL_LEASE_RENEW_MIN_SEC;
```
fslongjin, line 513: Same cadence change: master computed `remaining / 3` (~20 minutes for a default export), so this is now one HEAD per export per 20 s — 60x the request rate against the bucket for exports nobody is importing.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_exports.c
```diff
@@ -1265,16 +1156,10 @@ exports_reap(void *arg)
...
-				/* Even if this tick's batch is full, the watch
-				 * has nothing left to learn. */
-				export_lease_stop_if_terminal(exp);
-			}
 			if (n >= S3LVOL_EXPORT_REAP_BATCH) {
 				continue;
 			}
-			if (export_snapshot_alive(exp) &&
-			    !export_lease_is_terminal(exp)) {
+			if (export_snapshot_alive(exp)) {
```
fslongjin, line 1162: With this as the only reclaim condition, nothing bounds the registry while snapshots keep existing. `S3LVOL_MAX_EXPORTS` (256) is a fixed decode table and `spdk_json_decode_array` fails past it, so an `exports.json` with more than 256 entries makes the next attach fail to decode and takes the lvstore down with it. The table and the missing cap check predate this PR, but the reap rule you removed was what kept the count away from them.

fslongjin:
LGTM.

The protection-window wording now matches `export_pin_verdict()`: `STALE` (unprotected) while `lease_updated_at == 0` past the grace, `LEASE` only once a GET observes the PUT. Nothing else changed in this revision; memory-safety paths are clean.

The remaining items (cached-verdict delete path, 60 s grace, 20 s poller, registry cap) are worth an issue, not a blocker.

fslongjin (approved):

fslongjin (approved):

PR #1705, opened 2026-09-10
fix(one-click): align Redis DB with Helm #1638

xiaojunxiang2023:
## Summary

- Fix the one-click equivalent of #1638 by deriving Master, TemplateCenter, CubeOps, both CubeProxy Redis consumers, and LCM from one validated `CUBE_EXTERNAL_REDIS_DB` value.
- Remove stale `REDIS_URL` and per-component DB keys from the managed runtime environment so CubeOps cannot bypass `REDIS_DB` and write metrics to another logical database.
- Patch TemplateCenter's external Redis host/Sentinel/password together with Master, and make `redis.db_no` updates block-scoped, atomic, and verified.
- Preserve a consistent non-zero legacy DB during upgrade; when legacy values conflict, prefer Master `redis.db_no`, which owns scheduler reads.
- Fail runtime consumers clearly on invalid DB values and document that cross-DB moves require Redis `COPY ... DB` / `MIGRATE` or a cold start. The key migration script only renames keys within one DB.

## Test plan

- [x] `bash -n deploy/one-click/lib/common.sh deploy/one-click/install.sh deploy/one-click/scripts/common/validation.sh deploy/one-click/scripts/systemd/cubeops-start.sh deploy/one-click/scripts/one-click/up-cube-proxy.sh deploy/one-click/scripts/one-click/up-cube-lifecycle-manager.sh deploy/one-click/tests/test_redis_runtime_env.sh`
- [x] `bash deploy/one-click/tests/test_redis_runtime_env.sh`
- [x] `bash deploy/one-click/tests/test_env_merge.sh`
- [x] Linux Bash run covers legacy non-zero derivation, conflicting legacy values, standalone/Sentinel TemplateCenter endpoint patching, `REDIS_URL` removal, scoped YAML patch success/failure, and runtime DB validation.
- [ ] Fresh one-click install against a live external Redis with a non-zero DB.

## Scope note

The packaged CubeOps image must include #1607 so the binary honors `REDIS_DB`; rebuilding or republishing that image is intentionally outside this PR.


deploy/one-click/install.sh
```diff
@@ -280,6 +280,18 @@ apply_one_click_toggles
 # opposite-engine CUBE_EXTERNAL_* marker from .one-click.env.
 apply_one_click_database_intent
 
+# Old packages had no unified knob. Preserve a previous non-zero Master /
+# per-component DB instead of silently resetting all consumers to DB 0.
+if [[ "${INSTALL_MODE}" == "upgrade" ]]; then
+  derive_one_click_redis_db_from_legacy \
```
fslongjin, line 286: **[R1] Multi-node upgrades: derivation diverges between control and compute nodes — split-brain silently survives** (🔴 please address before merge)

Scenario: a multi-node user once hand-edited the control node's Master conf `db_no` to 5 (everything else sat on 0 — exactly the split-brain they wanted to fix). On upgrade:

- **Control node**: `derive_one_click_redis_db_from_legacy` reads its own Master conf → derives 5 → the whole control plane moves to DB 5. ✅
- **Compute nodes**: they run the same upgrade command, but the compute role never deploys CubeMaster (`install.sh:1802-1822`), and there are no legacy keys in the compute node's `.one-click.env` → derivation yields **0** → compute proxies stay on DB 0.

Result: control plane on DB 5, compute proxies on DB 0 — the exact split-brain this PR set out to fix, still present after the upgrade, with no log line or doc mentioning it. The env.example comment only says "set it in .env" and never states the multi-node distribution requirement.

Strictly speaking this is not a regression (compute was on 0 before the upgrade too), but the single-source-of-truth promise silently fails for this deployment shape, and nobody is told.

Suggestion — either one is cheap, both is best:

```bash
# Option A: warn in install.sh when a non-zero DB is derived
if [[ "${CUBE_EXTERNAL_REDIS_DB:-0}" != "0" ]]; then
  log "WARNING: derived non-zero Redis DB ${CUBE_EXTERNAL_REDIS_DB}; on multi-node"
  log "deployments set CUBE_EXTERNAL_REDIS_DB in .env on EVERY node, or compute"
  log "nodes stay on DB 0 (split-brain)."
fi
```

Option B: one sentence in the env.example comment block stating the multi-node requirement.

deploy/one-click/lib/common.sh
```diff
@@ -1293,10 +1560,24 @@ persist_one_click_database_runtime_env() {
...
+# cannot leave Master/Ops/TC on different databases.
 persist_one_click_redis_runtime_env() {
   local env_file="$1"
   [[ -n "${env_file}" ]] || die "persist_one_click_redis_runtime_env: env file path required"
 
+  local redis_db
+  redis_db="$(one_click_redis_db)"
+  upsert_env_kv "${env_file}" "CUBE_EXTERNAL_REDIS_DB" "${redis_db}"
+  _remove_env_keys "${env_file}" \
+    REDIS_URL \
```
fslongjin, line 1594: **[R2] Stripping `REDIS_URL` is completely silent** (🟡 please address — one-line log)

Scenario: a user once hand-added `REDIS_URL=redis://other-host:6379/9` to `.one-click.env` to point CubeOps at a specific Redis. After this upgrade:

- the line is stripped — correct by design, since CubeOps gives `REDIS_URL` top priority (`CubeOps/internal/config/config.go:413`; `redis_metric.go:84-85`) and it would bypass `REDIS_DB`;
- but CubeOps from now on **silently** derives its endpoint from `CUBE_EXTERNAL_REDIS_*`. If that `REDIS_URL` pointed at a different instance, the metrics data source just changed with zero log output.

Anyone later debugging "where did my node metrics go" has no trail pointing at this removal.

Suggestion:

```bash
if grep -q '^REDIS_URL=..*' "${env_file}"; then
  log "removing legacy REDIS_URL; CubeOps now derives its Redis endpoint from CUBE_EXTERNAL_REDIS_*"
fi
```

One line of cost, an hour of troubleshooting saved. The `unset REDIS_URL` in `cubeops-start.sh` could get the same treatment.

deploy/one-click/scripts/systemd/cubeops-start.sh
```diff
@@ -83,11 +83,18 @@ fi
...
 export REDIS_HOST="${CUBE_EXTERNAL_REDIS_HOST:-${CUBE_SANDBOX_REDIS_HOST:-127.0.0.1}}"
 export REDIS_PORT="${CUBE_EXTERNAL_REDIS_PORT:-${CUBE_SANDBOX_REDIS_PORT:-6379}}"
 export REDIS_PASSWORD="${CUBE_EXTERNAL_REDIS_PASSWORD:-${CUBE_SANDBOX_REDIS_PASSWORD:-ceuhvu123}}"
 export REDIS_MASTER_NAME="${CUBE_EXTERNAL_REDIS_MASTER_NAME:-}"
 export REDIS_SENTINEL_NODES="${CUBE_EXTERNAL_REDIS_SENTINEL_NODES:-}"
 export REDIS_SENTINEL_PASSWORD="${CUBE_EXTERNAL_REDIS_SENTINEL_PASSWORD:-}"
+# Prefer CUBE_EXTERNAL_REDIS_DB; keep legacy REDIS_DB as fallback. Must match
+# CubeMaster conf.yaml db_no (patched from the same env at install).
+export REDIS_DB
+REDIS_DB="$(normalize_redis_db "${CUBE_EXTERNAL_REDIS_DB:-${REDIS_DB:-0}}" "CUBE_EXTERNAL_REDIS_DB/REDIS_DB")"
```
fslongjin, line 101: **[R3] No runtime guard for the "edit env + restart only" footgun** (🟡 suggested — fine as an immediate follow-up)

Root cause: Master/TC `db_no` lives in conf.yaml, which **only install.sh writes**; Ops/Proxy/LCM re-derive from `.one-click.env` on every service start (`scripts/one-click/common.sh:6`). The two update paths move at different speeds:

- **Correct procedure** (set in `.env` → re-run `install.sh --mode=upgrade`): both sides move together; covered by tests. ✅
- **Wrong procedure** (hand-edit `.one-click.env` → restart services): Ops/Proxy/LCM move to the new DB, Master/TC stay on the old value → a **fresh split-brain**.

The env.example comment says "restart alone is not enough for those two" — but that is only a doc convention; nothing at runtime detects or blocks the wrong procedure. `cubeops-start.sh` validates the DB range, but never cross-checks against the Master conf.

Suggestion — cross-check at start (control nodes only, where the Master conf exists):

```bash
master_db_no="$(<awk-extract redis.db_no from CubeMaster/conf.yaml>"
[[ "${REDIS_DB}" == "${master_db_no}" ]] \
  || die "REDIS_DB=${REDIS_DB} != CubeMaster conf redis.db_no=${master_db_no}; re-run install.sh --mode=upgrade"
```

Die with an actionable message — turn the doc convention into a machine-enforced one.

deploy/one-click/scripts/one-click/up-cube-proxy.sh
```diff
@@ -108,7 +108,21 @@ CUBE_PROXY_HEARTBEAT_INTERVAL_MS="${CUBE_PROXY_HEARTBEAT_INTERVAL_MS:-5000}"
...
+if [[ -z "${CUBE_EXTERNAL_REDIS_DB:-}" \
+   && -n "${CUBE_PROXY_REDIS_DB:-}" \
+   && -n "${CUBE_PROXY_REGISTRY_REDIS_DB:-}" ]]; then
+  CUBE_PROXY_REDIS_DB="$(normalize_redis_db "${CUBE_PROXY_REDIS_DB}" "CUBE_PROXY_REDIS_DB")"
+  CUBE_PROXY_REGISTRY_REDIS_DB="$(normalize_redis_db \
+    "${CUBE_PROXY_REGISTRY_REDIS_DB}" "CUBE_PROXY_REGISTRY_REDIS_DB")"
+  [[ "${CUBE_PROXY_REDIS_DB}" == "${CUBE_PROXY_REGISTRY_REDIS_DB}" ]] \
+    || die "CUBE_PROXY_REDIS_DB (${CUBE_PROXY_REDIS_DB}) and CUBE_PROXY_REGISTRY_REDIS_DB (${CUBE_PROXY_REGISTRY_REDIS_DB}) conflict; set CUBE_EXTERNAL_REDIS_DB or make them equal"
+fi
+CUBE_PROXY_REDIS_DB="$(normalize_redis_db \
```
fslongjin, line 122: **[R4] Asymmetric conflict check; legacy values silently overridden** (🟢 non-blocking)

Three silent paths in this block:

1. only one of the two legacy keys is set (e.g. a lone `CUBE_PROXY_REGISTRY_REDIS_DB=3`) → no conflict check runs, and the value is silently promoted to both DBs;
2. `CUBE_EXTERNAL_REDIS_DB=5` coexists with a differing legacy `CUBE_PROXY_REDIS_DB=3` → 5 silently wins, nothing notes that 3 was ignored;
3. the `die` only fires when **both** legacy keys are set **and** they differ (that part is right).

Suggestion: log a WARNING wherever a legacy value is ignored or overridden — same rationale as R2: one-click troubleshooting is log-only, and users should be told which of their knobs got taken over.

deploy/one-click/scripts/common/validation.sh
```diff
@@ -47,3 +47,18 @@ validate_host_port() {
   (( 10#${port} >= 1 && 10#${port} <= 65535 )) \
     || die "invalid ${name}: ${value} (port out of range)"
 }
+
+normalize_redis_db() {
```
fslongjin, line 51: **[R5] The 0–15 range check is asymmetric with the Helm chart** (🟢 non-blocking)

The one-click side strictly validates 0–15 (leading zeros and oversized numerics are both rejected — the implementation itself is good). But the chart's `cube.redisDB` (`deploy/kubernetes/chart/templates/_helpers.tpl:1052-1054`) is just `.Values.redis.db | default 0 | int` — **no range check at all**; `redis.db=99` renders through fine.

So the two deployment paths disagree on what "valid" means: one-click rejects values Helm accepts. (Also note 0–15 encodes the default `databases 16` assumption; Redis can be configured with more.)

Keeping the stricter one-click check is fine — but either leave a one-line comment here stating the asymmetry is intentional, or add the same range check to the chart for parity with the spirit of #1638, so future readers don't assume the two paths behave identically.

deploy/one-click/tests/test_redis_runtime_env.sh
```diff
@@ -87,25 +106,435 @@ test_external_redis_persists_external_password() {
...
 
+test_cubeops_start_derives_redis_db() {
+  local start_sh="${ONE_CLICK_DIR}/scripts/systemd/cubeops-start.sh"
+  grep -Fq 'unset REDIS_URL' "${start_sh}" \
+    || fail "cubeops-start.sh must unset REDIS_URL before selecting REDIS_DB"
+  grep -Fq 'normalize_redis_db "${CUBE_EXTERNAL_REDIS_DB:-${REDIS_DB:-0}}"' "${start_sh}" \
+    || fail "cubeops-start.sh must validate CUBE_EXTERNAL_REDIS_DB with REDIS_DB fallback"
+}
+
+test_proxy_lcm_derive_unified_db() {
```
fslongjin, line 154: **[R6] Grep-the-source tests are brittle** (🟢 non-blocking)

This batch of assertions checks for fixed strings inside script sources:

```bash
grep -Fq '"${CUBE_EXTERNAL_REDIS_DB:-${CUBE_PROXY_REDIS_DB:-...}}"' up-cube-proxy.sh \
  || fail "up-cube-proxy.sh must derive ..."
```

Rename a variable → the test goes red (false positive); behavior breaks while the string survives → the test stays green (false negative). The behavioral tests added in this same PR (`test_one_click_patch_conf_redis_db_*`: run the real function, check the real file) are much better.

Non-blocking — this matches the existing style of this test file. Suggestion for a later iteration: convert the grep assertions to execute-and-check-outputs; the `persist_clean` subshell pattern already in this file is a ready-made template.

deploy/one-click/install.sh
```diff
@@ -601,63 +626,13 @@ patch_cubemaster_external_deps() {
...
-      -e "s|nodes: \".*\"|nodes: \"${redis_nodes_esc}\"|" \
-      -e "s|password: \".*\"|password: \"${redis_pwd_esc}\"|" \
-      "${cfg}"
+  one_click_patch_conf_redis_endpoint "${cfg}" "CubeMaster" "${restore_bundled_redis}"
+
+  # TemplateCenter writes progress snapshots through the same Redis cache and
+  # has no environment override for its endpoint. Patch both endpoint and DB.
+  local tc_cfg="${PKG_ROOT}/CubeTemplateCenter/conf.yaml"
+  if [[ -f "${tc_cfg}" ]]; then
+    one_click_patch_conf_redis_endpoint "${tc_cfg}" "CubeTemplateCenter"
```
fslongjin: **[R7] nit: the TC endpoint patch doesn't pass `restore_bundled` — asymmetric with the Master call** (🟢 non-blocking)

The Master call at `install.sh:629` passes `"${restore_bundled_redis}"` as the third argument; this one doesn't (defaults to 0).

Currently harmless: `PKG_ROOT` is a fresh unpack every run (the comment at `install.sh:588` says so explicitly), so the TC template already carries bundled values and the restore branch is effectively dead code. But this plants an asymmetry: if conf handling ever switches to preserve-on-upgrade instead of regenerate, it becomes "Master restored to bundled, TC still pointing at the external Redis".

Pass the same argument for symmetry (one-word change), or leave a one-line comment explaining why TC doesn't need it — leave the reasoning behind instead of making the next reader rediscover it.

deploy/one-click/env.example
```diff
@@ -112,6 +112,26 @@ WEB_UI_IMAGE=cube-sandbox-image.tencentcloudcr.com/opensource/openresty:1.21.4.1
 # Example IPv6: CUBE_EXTERNAL_REDIS_SENTINEL_NODES=[2001:db8::10]:26379
 # Leave SENTINEL_PASSWORD empty when Sentinel has no requirepass (do not reuse master password).
 # CUBE_EXTERNAL_REDIS_SENTINEL_PASSWORD=
+# Single logical Redis DB for the whole one-click stack (Master/TC/Ops/Proxy/LCM).
```
fslongjin, line 115: **[R8] nit: the comment block reads like implementation narrative** (🟢 non-blocking, optional)

The content itself is good — upgrade semantics, cross-DB data consequences, and the `redis_key_migrate.sh` limits are all spelled out honestly. But it mixes in installer internals, e.g. "install.sh snapshots this key like the DB-engine intent keys so the new value wins over a preserved .one-click.env line" — that describes a mechanism operators can neither act on nor verify.

Per this repo's comment conventions (rules and the durable why only), the block could compress to ~10 lines: keep *what it is / default / non-zero use / must re-run upgrade / data consequences*; drop the snapshot mechanics and the stripping details. Optional, but it would make the block much easier to scan.

fslongjin (commented):
## Review: fix(one-click): align Redis DB with Helm #1638

**Verdict: ✅ Mergeable — please address R1 and R2 first; R3 can be an immediate follow-up.**

This PR converges the one-click Redis logical DB from four mutually-unaware sources into a single knob (`CUBE_EXTERNAL_REDIS_DB`), strictly mirroring the merged Helm #1638. The design direction and implementation are correct, and it fixes a real latent bug along the way: TemplateCenter's conf was never re-pointed at external Redis, and the bundled Redis container isn't deployed when external Redis is configured (`up-support.sh:141-145`) — so on every external-Redis install, TC's progress snapshots have been silently failing (the Redis client is lazily connected, so TC starts fine and only fails at write time).

I ran the three test suites locally (`test_redis_runtime_env.sh`, `test_env_merge.sh`, `test_runtime_file_safety.sh`) — all pass. Every factual claim in the inline comments below was verified against the code; `file:line` references included.

The issues below concentrate on upgrade-path gaps where users are not told what happened. The mechanism itself won't break most upgrades (the majority of existing installs upgrade as a no-op), but specific user profiles hit silent surprises. The fixes are cheap — one-line logs and one warning — which is why R1/R2 are marked "address before merge".

| # | Severity | One-liner | Suggested handling |
|---|---|---|---|
| R1 | 🔴 address first | Multi-node: control node derives a non-zero DB, compute nodes derive 0 — split-brain silently survives the upgrade | warning + doc |
| R2 | 🟡 address first | Legacy `REDIS_URL` is stripped with zero logging; the data source silently switches | one-line log |
| R3 | 🟡 suggested | Editing the env and only restarting (without re-running install.sh) creates a fresh split-brain; no runtime guard | startup cross-check; follow-up OK |
| R4 | 🟢 non-blocking | Legacy values are silently overridden; the conflict check only covers the both-set case | add WARNING |
| R5 | 🟢 non-blocking | 0–15 validation is asymmetric with the chart (which validates nothing) | comment, or chart-side check |
| R6 | 🟢 non-blocking | Grep-the-source tests are brittle (assert text, not behavior) | convert over time |
| R7 | 🟢 nit | TC endpoint patch doesn't pass `restore_bundled` (harmless today, asymmetric with the Master call) | pass the arg, or comment why |
| R8 | 🟢 nit | The env.example block mixes operator rules with installer internals | optional trim |

**Worth preserving in future changes** (these design choices carry real reasons):

- **Early validation before the destructive phase** — a typo like `CUBE_EXTERNAL_REDIS_DB=16` fails before anything is stopped or wiped.
- **The awk block-scoped YAML patch** — the conf contains two `db_no` keys, and `instance_db_config.db_no` is the MySQL database index; a naive `sed s/db_no/` would corrupt it. Atomic tmp+mv plus read-back verification, with failure-path behavior locked by tests.
- **The intent-key split** (engine keys vs. full intent keys) — prevents a redis-DB-only upgrade from scrubbing preserved MySQL/PostgreSQL markers, locked by `test_redis_db_intent_survives_upgrade_merge`.

**Pre-merge checklist**

- [ ] R1: warning/doc for the multi-node case
- [ ] R2: removal log for `REDIS_URL`
- [ ] Confirm the released CubeOps binary/image includes #1607 (honors `REDIS_DB`) — declared out of scope in the PR description; just don't let it slip past the release notes.

xiaojunxiang2023:
@fslongjin Thanks for the thorough review!

Addressed R1/R2 (and the one-word R7):

- R1: on upgrade with a non-zero Redis DB, install.sh now warns that multi-node deployments must set CUBE_EXTERNAL_REDIS_DB in .env on every node; env.example states the same requirement.
- R2: persist_one_click_redis_runtime_env logs before stripping REDIS_URL, and cubeops-start.sh logs when ignoring a preserved REDIS_URL at runtime.
- R7: TemplateCenter's endpoint patch now passes restore_bundled_redis, matching Master.

Leaving R3 as an immediate follow-up (runtime Master-conf cross-check on restart-only edits), and holding R4–R6/R8 as non-blocking for now.

fslongjin (approved):

PR #1709, opened 2026-09-10
fix(vmm): classify incremental CoW pages from pagemap bit 61

ls-ggg:
## Summary
- Incremental Pause decided which guest pages to write by following pagemap PFNs into `/proc/kpageflags`. That walk is not atomic with the pagemap snapshot, so compaction can make `KPF_ANON` miss a dirty page and leave template bytes on the dest.
- Classify from the same pagemap word only: swapped (bit 62) or present exclusive anon (bit 61 clear). File-backed / shared-anon pages stay on the reflink base.
- Concurrent XFS Pause/Resume with a dirty guest (4-way, write load, two pause/resume cycles) passed on a freshly built template and on an older one that previously reproduced guest death after resume.

## Test plan
- [x] `rustfmt --edition 2021` on the changed file
- [x] Build `cube-runtime` / shim from current `master` plus this change
- [x] 4-way create → dirty guest → Pause → Resume → Pause → Resume on a new template
- [x] Same load on the older template that used to fail
- [ ] CI format / unit checks

### DIFF
--- a/hypervisor/vmm/src/pagemap_anon.rs
+++ b/hypervisor/vmm/src/pagemap_anon.rs
@@ -4,17 +4,16 @@
 
 //! PagemapAnon snapshot support
 //!
-//! This module provides functionality for creating pagemap-anon-based snapshots
-//! that only save CoW anonymous pages (pages actually written by the Guest)
-//! by inspecting `/proc/self/pagemap` and `/proc/kpageflags`.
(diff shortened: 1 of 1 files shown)


fslongjin:
This PR is similar with  https://github.com/TencentCloud/CubeSandbox/pull/1343 , but it lacks a check for the kernel version, which would cause kernels prior to 6.44 to be unable to correctly perform incremental snapshots.

Issue #1227, opened 2026-07-30
[Feature Request] Introduce a CubeSandbox-maintained minimal envd implementation


aptacc2421:
## Problem / Motivation

Every sandbox currently runs envd as an in-guest data-plane service. It is the boundary used by SDKs for readiness checks, command execution, file access, and initialization. It is therefore a critical compatibility boundary between E2B SDK clients and the CubeSandbox runtime.

The current base image builds the Go envd implementation from a pinned [`e2b-dev/infra@ENVD_REF`](https://github.com/TencentCloud/CubeSandbox/blob/master/docker/Dockerfile.cube-base) and records its ref, version, and commit. Pinning controls which revision CubeSandbox adopts, but the implementation roadmap, fix cadence, release schedule, and future behavior evolution are still owned primarily by the E2B project.

The upstream envd is designed for E2B's own runtime environment and includes integration paths for Firecracker MMDS, access-token initialization, instance log exporting, NFS/Volume initialization, and Hyperloop configuration. The current CubeSandbox integration is narrower: its [`/init` request](https://github.com/TencentCloud/CubeSandbox/blob/master/Cubelet/services/cubebox/probe.go) only supplies `envVars`. Meanwhile, the upstream [`main.go`](https://github.com/e2b-dev/infra/blob/2026.16/packages/envd/main.go) and [`init.go`](https://github.com/e2b-dev/infra/blob/2026.16/packages/envd/internal/api/init.go) contain MMDS polling paths that static analysis suggests may still run in the current setup.

A configuration change or small upstream patch may be sufficient for this individual MMDS mismatch, but it would not address the broader dependency on another project's implementation and release priorities. As CubeSandbox evolves as an independent self-hosted platform, it needs an explicit, independently versioned compatibility contract for the command and file behavior exposed to SDK users, rather than coupling that contract permanently to E2B-specific runtime implementation details.

## Proposed Solution

Introduce a CubeSandbox-maintained Rust `cube-envd` that preserves the E2B SDK-observable behavior required by CubeSandbox, implements the agreed minimal subset, and owns its tests, versioning, fixes, and release process in this community. It will be developed behind the existing `ENVD_BIN` switch, with the Go envd retained as a rollback.

### Why Rust

[...]

PR #1723, opened 2026-09-11
feat(cube-envd): CubeSandbox-maintained in-guest data-plane daemon 

chaojixinren:
**Closes:** #1227

---

## Summary

This PR replaces the upstream Go `envd` in the base image with `cube-envd`, a CubeSandbox-maintained Rust implementation of the same in-guest data plane, behavior-matched fixture-by-fixture against a recorded Go envd 0.5.13 baseline.

The Go binary is still built and shipped as `/usr/bin/envd-go`, so rollback is a single environment variable at sandbox startup and needs no image rebuild.

| | |
|---|---|
| Change set | `cube-envd/` implementation, conformance harness, image integration, and SDK compatibility changes |
| Build | Rust 1.89, static musl binary; Go 0.5.13 retained for rollback |
| New test asset | `tests/e2e/envd_conformance/` — black-box conformance harness |
| Default changed | `docker/Dockerfile.cube-base` now installs cube-envd as `/usr/bin/envd` |
| Rollback | `ENVD_BIN=/usr/bin/envd-go`, or rebuild with `--build-arg ENVD_IMPL=go` |

Everything lands behind the existing `ENVD_BIN` switch that `docker/cube-entrypoint.sh` already reads, so the guest contract (`envd -port 49983 -isnotfc` on `:49983`) is unchanged.

## Motivation

envd runs inside every sandbox and is the compatibility boundary between the E2B SDKs and the CubeSandbox runtime. Consuming it from `e2b-dev/infra` means the implementation roadmap, fix cadence, and release schedule are owned elsewhere, and the shipped binary carries integration paths CubeSandbox never exercises: Firecracker MMDS polling, Hyperloop configuration, and NFS volume initialization.

Issue #1227 laid out the proposal, the Rust-vs-Go reasoning, and the conditions for adopting cube-envd on the default path. This PR delivers the full implementation rather than the MVP subset #1227 scoped as a first milestone, so it closes that issue rather than reporting against it: declared conformance differences are explicit and allowlisted, SDK end-to-end covers command and file behavior, a template built with cube-envd creates a usable sandbox, performance measurements are published per workload, and the `ENVD_BIN` rollback to Go envd is verified.

The merged performance work is deliberately reported per workload instead of collapsed into one headline. Rust has lower idle RSS and faster direct startup; the optimized command/PTY spawn paths and process-output path compare well with Go in their measured shapes; `/files` remains workload-dependent, with near-parity single-stream behavior and a remaining advantage for Go's `sendfile` path once high-concurrency downloads become CPU-bound.

Rust is not adopted on the promise of a universal benchmark win; the numbers are published here so they can be checked.

## Scope

This is a full envd implementation, not an MVP subset. Every RPC the E2B envd service declares is implemented and conformance-tested, including the parts issue #1227 originally deferred: PTY, interactive stdin, the whole watch family, and cgroup-specific behavior. `cube-envd/src/app/routes.rs` is the complete list — 8 of 8 `process.Process` RPCs, 9 of 9 `filesystem.Filesystem` RPCs, and the REST surface used by the platform.

**In scope and implemented.**

- REST: `GET /health` (204), `POST /init` (envVars merge, optional accessToken, timestamp gate), `GET /envs`, `GET /metrics`, `GET/POST /files` (octet-stream + multipart, relative paths, ownership, Go errno vocabulary).
- `process.Process`: `Start`, `Connect`, `List`, `SendSignal`, `SendInput`, `StreamInput`, `CloseStdin`, `Update` — including PTY allocation with window resize and CRLF line discipline, pipe and PTY stdin, injected input, and `CloseStdin` EOF.
- `filesystem.Filesystem`: `Stat`, `ListDir`, `MakeDir`, `Move`, `Remove`, `WatchDir`, `CreateWatcher`, `GetWatcherEvents`, `RemoveWatcher` — including the streaming watch family and pull watchers.
- CLI: Go `flag`-compatible parsing for `-port`, `-isnotfc`, `-version`, `-commit`, and `-h`.
- Auth: `Authorization: Basic base64("<user>:")` / `username` query, `/etc/passwd` resolution, privilege drop, and `X-Access-Token` gating.
- Guest integration: cgroup v2 per-command leaf placement, OOM attribution, process-group cleanup, and a deployment-reachable cgroup memory cap.

**Residual protocol gaps.** These are explicit rather than silent:

| Gap | Behavior | Why |
|---|---|---|
| `/files/compose` | HTTP 501 | The endpoint stays in the route table so callers get an explicit 501 rather than a misleading 404. |
| gzip download encoding | Serves identity only; refuses unacceptable identity with upstream-compatible 406 behavior | gzip is applied by CubeProxy rather than the daemon. |
| Multi-range `GET /files` | Serves the whole file with 200 instead of upstream's `206 multipart/byteranges` | A multipart range writer is not implemented; single-range and conditional requests are implemented. |
| `HEAD /files` | Axum serves 200 with the GET headers, while upstream has no HEAD route and returns 405 | Framework-level route behavior; recorded as a declared conformance difference. |
| Connect binary-protobuf codec | Rejected up front with a stable `unimplemented` naming the JSON codec | Every known client in this repository and the official E2B Python/JS SDKs uses the JSON codec. |

`/files` signature verification is **not** a gap: cube-envd implements the endpoint surface upstream exposes, and upstream's signing path is not part of the SDK-observable contract this daemon reproduces. An earlier draft of this PR description listed it as unimplemented; that was incorrect, and it is not in the code or the conformance suite.

Unrecognized **input shapes** on implemented RPCs — an empty process selector, a `StreamInput` event with zero or multiple arms, or an empty `ProcessInput` oneof — answer `unimplemented` with upstream-compatible text. That is alignment with Go envd's default branch, not a missing feature, and unit tests pin the behavior. Nothing is signalled, attached to, or otherwise side-effected on those paths.

## Design

`cube-envd/` is organized as a contract index and a dependency graph: a path states what is promised, an edge states who may call whom. Six layers, one allowed direction.

```mermaid
graph TD
app["<b>app/</b><br/>HTTP surface + startup wiring<br/>routes · handlers · lifecycle · metrics<br/>cli · pool · middleware"]
fs["<b>filesystem/</b><br/>Stat · ListDir · MakeDir · Move · Remove<br/>download · upload · watch<br/>http/ — ranges · preconditions · httpdate"]
proc["<b>process/</b><br/>Start · Connect · List · SendSignal<br/>SendInput · StreamInput · CloseStdin · Update<br/>supervisor · table · engine · cgroup"]
plat["<b>platform/</b><br/>config · identity · lock"]
proto["<b>protocol/</b><br/>frames · stream · error · keepalive · timeout"]
compat["<b>compat/</b><br/>vocab — Go errno text table"]

app --> fs
app --> proc
app --> plat
fs --> plat
fs --> proto
fs --> compat
proc --> plat
proc --> proto
proc --> compat
plat --> proto
plat --> compat
proto --> compat

fs -. "never" .-> proc
proc -. "never" .-> fs
fs -. "never" .-> app
proc -. "never" .-> app
plat -. "never" .-> app
proto -. "never" .-> plat
compat -. "never" .-> proto

classDef allowed fill:#1f6feb22,stroke:#1f6feb,stroke-width:2px
classDef sibling fill:#8957e522,stroke:#8957e5,stroke-width:2px
class app,plat,proto,compat allowed
class fs,proc sibling
```

Solid edges are the allowed direction; dotted edges are the ones the layer rule forbids. `filesystem/` and `process/` are siblings and never reference each other, and nothing below `app/` reaches back into it.

This is enforced, not merely intended: `cube-envd/tests/layer_rule.rs` reads `src/` and fails the test suite on a forbidden module path, so an inverted edge cannot land while `cargo test`, `clippy`, and `fmt` stay green.

| Layer | Responsibility |
|---|---|
| `app/` | HTTP surface and startup wiring: `routes.rs`, `handlers.rs`, `lifecycle.rs`, `metrics.rs`, Go-flag-compatible `cli.rs`, blocking-thread `pool.rs`, `middleware/{cors,legacy}.rs` |
| `filesystem/` | Filesystem domain; `http/` owns content disposition, encoding, HTTP date, preconditions, and ranges; `watch/` owns inotify, tree, and watch pumping |
| `process/` | Process domain; command lifecycle, output fan-out, PTY, cgroup, supervision, table, and engine |
| `platform/` | Host-facing services shared by both domains: config, identity, user/group lookup, path anchoring, and locking |
| `protocol/` | Connect wire layer: envelope codec, error model, keepalive, timeout, and stream primitives |
| `compat/` | Go compatibility vocabulary and stable error text |

`spec/` holds the protocol snapshots the `wire.rs` files mirror — the upstream OpenAPI document plus `filesystem.proto` and `process.proto` copied from `e2b-dev/infra@2026.16`. It is a conformance reference, not a build input.

Notably, `/init` semantics were re-derived from the upstream source rather than assumed. An **equal** timestamp is accepted (`utils.AtomicMax.SetToGreater` only rejects strictly older values), and the timestamp gate runs **before** token validation, so a stale-timestamp `/init` returns 204 rather than 401. `/init` is also on the upstream auth allowlist: it does not check `X-Access-Token`, and token semantics come entirely from the request body.

## Integration and rollback

`docker/Dockerfile.cube-base` compiles both implementations in parallel stages and selects one. The build context changes from `docker/` to the repository root because `cube-envd/` must be inside it.

```mermaid
flowchart TD
subgraph build["Build time — docker/Dockerfile.cube-base, context = repo root"]
direction TB
src_go["e2b-dev/infra@ENVD_REF<br/>pin: 2026.16"]
src_rust["cube-envd/<br/>this repo"]

subgraph b1["Stage 1a"]
go_b["envd-go-builder<br/>Go builder"]
end

subgraph b2["Stage 1b"]
cube_b["envd-cube-builder<br/>Rust 1.89 + musl-tools"]
end

src_go --> go_b
src_rust --> cube_b
go_b --> go_bin["/out/envd<br/>Go binary"]
cube_b --> cube_bin["/out/envd<br/>musl static"]
go_bin --> sel{"envd-default<br/>ENVD_IMPL"}
cube_bin --> sel
end

subgraph img["Runtime image — Ubuntu 22.04"]
direction LR
e1["/usr/bin/envd<br/>selected impl"]
e2["/usr/bin/envd-go<br/>always present"]
e3["/etc/cubesandbox-envd-impl<br/>cube | go"]
end

sel -->|"ENVD_IMPL=cube (default)"| e1
sel -->|"ENVD_IMPL=go"| e1
sel --> e3
e1 --> entry["cube-entrypoint.sh<br/>envd -port 49983 -isnotfc"]
e2 -. "rollback target — ENVD_BIN" .-> entry
```

Rollback is available at both runtime and build time:

```mermaid
flowchart LR
start{"cube-envd is<br/>the live data plane"} --> q1{"rollback scope?"}
q1 -->|"one deployment,<br/>image, or test template"| r1["set<br/>ENVD_BIN=/usr/bin/envd-go"]
q1 -->|"whole image,<br/>at build time"| r2["docker build<br/>--build-arg ENVD_IMPL=go"]
q1 -->|"verify side by side"| r3["run the two conformance<br/>containers, one per impl"]
r1 --> out["previous Go envd behavior restored"]
r2 --> out
r3 --> out
```

Installing cube-envd as the literal `/usr/bin/envd` rather than only selecting it through `ENVD_BIN` matters because Cubelet version collection executes `envd --version`.

`.dockerignore` admits `cube-envd/` while still excluding `cube-envd/target/`. CI triggers include `cube-envd/**`, the component matrix runs Rust tests/clippy, and `CUBE_ENVD_COMMIT` is threaded from the source revision into the image so `envd -commit` reports the build's source commit.

## SDK changes

The Rust daemon reports richer termination information than upstream's EndEvent. This is a **CubeSandbox extension**. `signal`, `oomKilled`, and `killedBy` are additive and optional, so clients that ignore them keep the previous behavior.

| SDK | Change |
|---|---|
| Go | `CommandResult` gains `Signal *int`, `OOMKilled bool`, `KilledBy string`; `PtyHandle` gains matching accessors |
| Python | `CommandResult` gains `signal`, `oom_killed`, `killed_by`; `PtyHandle` gains matching properties |
| Node | `CommandResult` gains optional `signal`, `oomKilled`, `killedBy`; `PtyHandle` gains matching getters |

Two behavioral corrections ride along:

1. **Python `Commands.run` always uses the in-house Connect-JSON decoder.** It previously preferred the generated E2B client when the optional `e2b` package happened to be installed, which dropped the termination extension and made behavior depend on the user's environment.
2. **Go `Pty.Connect` no longer sends `Connect-Timeout-Ms`.** Attachments are long-lived, so a client-side idle timeout must not become a hard server-side deadline.

## Verification

Historical functional evidence is recorded under `tests/e2e/envd_conformance/`; the conformance README contains the commands. Performance is reported below from the merged optimization work, with each workload kept separate.

### Protocol conformance

Two containers from the same image, one running each implementation, are compared message by message. Normalization removes volatile values, header case, and transfer framing. Declared differences are allowlisted with reasons.

The current allowlist contains six declared differences:

- `rest_files_gzip_accept` — upstream supports gzip; cube-envd serves identity only.
- `rest_files_compose_probe` — upstream implements `/files/compose`; cube-envd returns 501.
- `fs_bad_json` — parser-specific JSON error wording; status/error class still match.
- `rest_init_timestamp_out_of_range` — upstream `UnixNano()` wraps and treats the request as stale; cube-envd rejects the caller input with 400.
- `rest_files_range_multi` — upstream returns `206 multipart/byteranges`; cube-envd serves the whole file with 200.
- `rest_files_head` — upstream has no HEAD route and returns 405; Axum supplies HEAD with 200 and the GET headers.

A fixture that starts matching must be removed from the allowlist; otherwise a future regression could be silently downgraded from FAIL to DECLARED-DIFF.

### Unit tests

At the current performance-integrated head, the latest recorded test run is **376 passed / 3 ignored**. `cargo clippy --all-targets` and `cargo fmt --check` are clean. The ignored tests require privileges / a writable cgroup v2 environment that normal CI does not provide.

The suite includes guards for the layer rule, poison-recovering locks, `ProcHandle` process-table indirection, constant-time access-token comparison, byte-exact Go error text, output-bus backpressure, download resource budgets, and the process-wide read-buffer pool.

### SDK end-to-end

The latest recorded candidate-binary SDK run contains **21 checks, PTY included**, against a real sandbox. It covers readiness, command round trips, stdout/stderr and terminal events, stdin/PTY paths, exit status, env/user/cwd handling, timeout behavior, and text/binary file operations.

### Performance — current merged measurements

The old `3c326264` direct/CubeProxy table and its temporary runners applied only to that pinned revision and are no longer used as the current performance conclusion. The sections below report the merged optimization paths separately. Topologies differ; absolute numbers from one subsection must not be mixed with another.

#### Command spawn path — [chaojixinren/CubeSandbox#25](https://github.com/chaojixinren/CubeSandbox/pull/25)

The fork-free path uses `clone(CLONE_VM|CLONE_VFORK|SIGCHLD)` while keeping the same child-side credential, cwd, process-group, and cgroup setup.

| Measurement | Go | cube-envd |
|---|---:|---:|
| Start → `/health`, P50 | 5.45 ms | **1.55 ms** |
| First command after readiness | 3.94 ms | **3.43 ms** |
| Idle RSS | 15.00 MiB | **2.98 MiB** |
| Command latency, C=64 | 86.32 ms | **77.63 ms** |
| Command throughput, C=64 | 732 rps | **804 rps** |
| Failed requests | 0 | 0 |

This is a command-spawn measurement, not a file-transfer or full CubeProxy result.

#### PTY spawn path — [chaojixinren/CubeSandbox#27](https://github.com/chaojixinren/CubeSandbox/pull/27)

The PTY path now uses the same fork-free spawn engine instead of maintaining a second `fork`-based path. With 2 vCPU and 2,000 PTY sessions at C=64, repeated for five rounds:

| Measurement | Go | cube-envd |
|---|---:|---:|
| Session throughput | 684 rps | **798 rps** |
| minor faults / command | **1.25** | 6.97 |
| sys ms / command | 0.290 | **0.260** |

The throughput result measures session creation. Existing-session PTY input/output round trips do not spawn a process and remain around parity with Go.

#### Process output and slow-consumer behavior — [chaojixinren/CubeSandbox#28](https://github.com/chaojixinren/CubeSandbox/pull/28)

PR #28 replaced the two bounded broadcast rings with an explicit per-process `OutputBus`. A slow live consumer now applies backpressure instead of making the daemon truncate the stream. A subscriber that is actually blocked and makes no progress for 300 seconds is evicted independently; idle subscribers are not evicted, and a reserved terminal slot guarantees that `End` still fits after data.

| Process-stream check | Go 0.5.11 | cube-envd |
|---|---:|---:|
| 32 MiB stream, median of five | 81.2 MB/s | **128.6 MB/s** |
| Delivery | complete | complete |
| PTY keystroke round trip, median | 0.44 ms | **0.34 ms** |
| PTY keystroke round trip, P95 | 4.00 ms | **0.56 ms** |

For a 200 MiB process stream inside the guest, cube-envd delivered all **209,715,200 bytes** with a matching SHA-256 in **1.50–1.74 s**, while the daemon peaked at **2.1 MB RSS**. The Go run completed the wire transfer in **3.8–7.0 s**, but left roughly **1.0–1.5 GB** resident in the 1.9 GB guest, leaving about 139 MB available and causing a normal client to fail while writing what it received.

A three-subscriber 200 MiB run (fast, ~10 MB/s, and one subscriber that does not read) delivered the full 200 MiB with matching SHA-256 to both progressing subscribers; only the stalled subscription was evicted after the 300-second window. Attachments are bounded at 8 per process and 64 per envd.

#### `/files` data plane — [chaojixinren/CubeSandbox#29](https://github.com/chaojixinren/CubeSandbox/pull/29) + [#34](https://github.com/chaojixinren/CubeSandbox/pull/34)

PR #29 moved large downloads to body-sized pooled/prefetched reads, added `TCP_NODELAY`, and introduced explicit thread, memory, and concurrency budgets. PR #34 then moved free read buffers into one process-wide `ReadPool`, eliminating most per-request page-fault churn at high concurrency.

Single-stream, in-guest PVM measurement:

| 200 MiB body | Median | vs Go |
|---|---:|---:|
| Go | **0.040 s** | 1.00x |
| cube-envd | 0.041 s | 1.03x |

Small-body keep-alive latency in the same guest:

| 4 KiB body | Median |
|---|---:|
| Go | **0.11 ms** |
| cube-envd | 0.27 ms |

Latest high-concurrency read-pool result, 2 vCPU guest, in-guest C client, 32 concurrent 4 MiB downloads:

| 4 MiB × 32 | Go 0.5.x | cube-envd |
|---|---:|---:|
| Throughput | 4855–6244 MiB/s | **3867–4133 MiB/s** |
| daemon CPU | **0.01 ms/MiB** | 0.20–0.23 ms/MiB |
| minor faults | **0.1 / MiB** | 1.6 / MiB |
| peak RSS | **6 MiB** | 20–21 MiB |

The shared pool makes cube-envd roughly **4.7–6.7× faster** than its per-body-pool predecessor in this shape and brings it to about **65–85% of Go**. Go still wins at this concurrency because `sendfile` avoids the two userspace copies that remain in the Rust path.

The pool is bounded rather than free. It starts empty at about **2.5 MiB idle RSS**, retains at most 32 MiB at the default configuration, and measured about **19.5 MiB settled RSS** after repeated 32×4 MiB bursts and **22.7 MiB** after one 200 MiB body. Go stayed around 6.1–7.2 MiB because its `sendfile` path does not retain userspace read buffers.

Stalled clients are bounded separately from the free-buffer cache. With 200 stalled large downloads, the configured global cap serves **128** and rejects **72 with 503**; held RSS remains about **277.2–277.9 MiB** across the tested buffered variants. The important property is that memory and blocking-thread use follow configured budgets instead of the number of connections a client can open.

### Sampling and limits

Percentiles, throughput, and RSS are meaningful only inside the topology that produced them. Command-spawn, PTY, process-stream, and `/files` measurements use different harnesses and should not be cross-divided. Where a result is reported as a range, the range is kept instead of converting a small number of runs into a fake stable percentile.

Historical measurements remain useful for regression archaeology, but they are not the current headline numbers. A new retest should record the source revision, image digest, binary hash, resource limits, client placement, and raw output together.

### cgroup and termination — real machine

cgroup behavior is not exposed over RPC, so the conformance harness cannot compare it. `termination_e2e.py` covers it on an isolated Linux/QEMU guest: real main-process OOM, descendant-only OOM, timeout and user attribution, pipe and PTY placement, escaped-descendant cleanup, and fail-closed leaf allocation.

The deployment-reachable cgroup memory cap is also verified on a real PVM guest: `cgroup_memory_max_e2e.py` records **25/25** checks for flag/env/derived precedence, clamping to the enclosing limit, invalid-value handling, `memory.high`, and the real `ENVD_EXTRA_ARGS` entrypoint path.

## Deliberate behavioral differences

Each difference is documented in `cube-envd/README.md` and, where a fixture exists, allowlisted with a reason in `conformance.py`.

- **Process-group cleanup.** cube-envd starts each command in its own process group and signals the whole group, so a timeout or `SendSignal` reaps the shell **and** its descendants. Go envd 0.5.13 signals only the direct child and can leave grandchildren behind. cube-envd preserves the normal exit/deadline contract and additionally publishes richer terminal metadata.
- **Bounded watch buffers.** The pull-watch event buffer is capped at 10,000 events. Upstream accumulates without bound. At the cap, cube-envd fails the watcher through `GetWatcherEvents` rather than silently dropping or growing forever.
- **Keepalive cadence.** 30 seconds rather than upstream's 90 seconds, chosen to stay below the common 60-second idle timeout in front of CubeProxy. The `Keepalive-Ping-Interval` header still tunes the cadence.
- **CLI validation is stricter.** Invalid flags, positional arguments, and malformed values terminate startup with exit code 2 instead of being silently ignored.
- **Out-of-range `Keepalive-Ping-Interval` degrades instead of panicking.** Upstream can panic when it feeds an invalid value into `time.NewTicker`; cube-envd falls back to 30 seconds.
- **Slow output consumers are back-pressured, with bounded eviction.** Progressing subscribers are not truncated. A subscriber that is blocked and makes no progress for 300 seconds is evicted with `resource_exhausted`; the child and other subscribers continue.
- **Uploads match upstream's non-atomic write.** Both upload paths stream into the target with `O_WRONLY|O_CREATE|O_TRUNC`. An interrupted upload leaves partial content, concurrent readers can observe growth, and a failed upload is not rolled back.

## Risk and rollback

**Blast radius:** every sandbox's data plane.

The mitigation is that rollback is a runtime switch, not a rebuild: `ENVD_BIN=/usr/bin/envd-go` restores the previous Go daemon for a deployment, image, or test template. `ENVD_IMPL=go` moves the same decision to image build time.

**Not covered by this PR:** promoting the new default to production templates is a separate operational decision. This PR provides the compatibility, SDK, resource, and rollback evidence needed for that decision; it does not perform the rollout.

**Known residual limitations:**

- Multipart part filenames that are not valid UTF-8 return 400 where Go would write the file.
- A non-UTF-8 basename falls back to `download` in `Content-Disposition`; the current query path is UTF-8-constrained, so this is effectively unreachable.
- OOM attribution is command-leaf scoped, not PID scoped; an unrelated external SIGKILL concurrent with a descendant OOM is inherently ambiguous.
- `-cmd` and `-cgroup-root` are recognized but not implemented; they warn and skip.

## Reviewer guide

Suggested reading order:

1. `cube-envd/README.md` — design, compatibility scope, and the declared differences table.
2. `cube-envd/src/app/routes.rs` — the entire URL surface.
3. `cube-envd/tests/layer_rule.rs` — the architectural dependency rule.
4. `cube-envd/src/process/supervisor.rs` and `process/table.rs` — process lifecycle, `ProcHandle`, output fan-out, backpressure, and attachment limits.
5. `cube-envd/src/filesystem/http/` — conditional requests, Range handling, and HTTP file semantics.
6. `tests/e2e/envd_conformance/conformance.py` — especially `DECLARED_DIFFERENT`.
7. `docker/Dockerfile.cube-base` — the dual-build, default switch, and rollback binary.


ls-ggg:
Thanks for the work on this.
Please get CI green and squash the commit history before we merge.

cc @wbzdssm for a look at the SDK side, and @lkml-likexu for the tests.

aptacc2421:
Hi, @ls-ggg ,thanks for review. We squashed. 
CI has two red: one is cubelet, another is our pr bigger then 20000 lines so bot can't review. 
Other pr also red. 
All not our problem.
And I've made some fix about the CI problem below:
https://github.com/TencentCloud/CubeSandbox/pull/1738
https://github.com/TencentCloud/CubeSandbox/pull/1737
https://github.com/TencentCloud/CubeSandbox/pull/1736

ls-ggg:
Thanks for the cube-envd work. We reproduced **compile + `cargo test`** on a Linux amd64 host with the repo-pinned Rust 1.89 toolchain, at `56162df`.

**Build**
- `cargo build --locked` (debug): pass
- `cargo build --release --locked --target x86_64-unknown-linux-musl`: pass (static-pie, ~3.4M)

**Unit tests** (`cargo test --locked`, matching `cube-envd/Makefile`)
- **376 passed / 3 ignored / 0 failed**
- Ignored cases match the write-up (process-wide attachment budget; two tests need root + a writable cgroup v2 mount)

The PR text currently says 347 passed / 2 ignored. The suite on this head is green and slightly larger than that snapshot — worth updating the numbers so reviewers do not chase a stale count.

We have not run the conformance harness or SDK e2e in this pass.

chaojixinren:
@ls-ggg Thanks for running the independent repro at 56162df — that's exactly the check we wanted.

Updated the PR text accordingly. The Unit tests section now reads 376 passed / 3 ignored, matching your numbers; the 347 / 2 figure was a stale snapshot from before the performance work landed, so it was wrong to leave it in.

Issue #1690, opened 2026-09-09
[Bug Report] Cross-node resume of a paused sandbox fails after the export TTL (1 h): s3lvol auto-reaps the ref export while the sandbox is still paused


fslongjin:
## Summary

A sandbox that is paused can no longer be resumed on a **different node** if the cross-node resume happens after the pause export's TTL (default **1 hour**) expires.

When a sandbox is paused, the `s3lvol` side publishes a *ref*-layout export (lease-aware, `pin=lease`) so that a later cross-node resume can import it. If nothing imports it while the sandbox stays paused, the export becomes "terminal" once its lease is confirmed absent past the TTL deadline, and the reaper removes it **automatically** — manifest deleted, registry entry cleared. From that point on, any cross-node resume attempt fails because the manifest it must import returns 404. Same-node resume keeps working because it is served from the local pause catalog / lvstore snapshot volume and never depends on the export.

## Environment

- Affected version: current `master` — the automatic reaper behavior that causes this was introduced in #1663 (the fix for #1634).
- Deployment mode: cluster (cross-node resume).
- Relevant components: CubeS3lvol (export/lease) / CubeMaster (restoreplace cross-node placement) / Cubelet (resume).

## Description / mechanism

1. Pause completes. The `s3lvol` side publishes a *ref*-layout export for the pause package (manifest under `<lvstore>/meta/exports/<uuid>.json`) with `lease_aware=true` and the default TTL `S3LVOL_EXPORT_DEFAULT_TTL_SEC = 3600` (1 h).
2. Cross-node resume depends on that export: `restoreplace.Decide` only allows a resume to leave the origin node while the pause record reports `remote_status=ready` (i.e. the export exists), and Cubelet on the destination node imports it through the export UUIDs (`remoteUUIDs`).
3. If the export expires with no lease taken (`lease_absent` — nothing imported it during the pause), `export_lease_is_terminal()` becomes true and the periodic `exports_reap` poller reaps it **automatically**: the manifest is deleted and the registry entry cleared. No explicit delete / release request is involved.
4. Any later cross-node resume reads the already-removed manifest → 404 → the resume fails. Same-node resume is unaffected: it goes through the local pause catalog + lvstore snapshot volume and does not depend on the export.

## Expected Behavior

A sandbox that is still paused should remain resumable on another node (cross-node) without being silently limited to the ~1 h export TTL. Reaching the export deadline should not invalidate a pause that is still owed to a possible cross-node recovery.

[...]

Issue #1683, opened 2026-09-08
[Bug Report] Explicit connect can race with the lifecycle sweeper and immediately re-pause a resumed sandbox


wdmzsslj:
## Summary

An explicit `Sandbox.connect()` can race with the lifecycle sweeper, causing a successfully resumed sandbox to be immediately paused again because CLM still uses a stale `LastActiveMs`.

## Environment

- CubeSandbox version: `v0.7.0`
- Host OS: Ubuntu 20.04.6 LTS
- Kernel: Linux `5.15.0-139-generic`, x86_64
- KVM: in-tree KVM module for `5.15.0-139-generic`
- Deployment mode: single-node
- Relevant components: CubeAPI, CubeMaster, CubeProxy, and `cube-lifecycle-manager`
- Client SDK: `e2b@2.46.1`

## Steps to Reproduce

1. Create a sandbox with automatic pause and resume enabled:

   ```ts
   const sandbox = await Sandbox.create(templateId, {
     timeoutMs: 300_000,
     lifecycle: {
       onTimeout: "pause",
       autoResume: true,
     },
   });
   ```

2. Pause the sandbox and leave it paused long enough that CLM's `LastActiveMs` is older than the configured idle timeout and the Redis lifecycle state key has expired.

3. Explicitly resume the sandbox through the E2B-compatible API:

   ```ts
   await Sandbox.connect(sandbox.sandboxId, {
     timeoutMs: 300_000,
   });
   ```

   The equivalent direct request is:

   ```http
   POST /sandboxes/{sandboxID}/connect
   Content-Type: application/json

   {"timeout": 300}
   ```

4. Immediately access the sandbox envd health endpoint or execute a command.

5. Inspect the sandbox state and `cube-lifecycle-manager` logs.

This is a scheduling race with the five-second sweeper interval. A deterministic regression test can enforce the problematic ordering with synchronization barriers.

## Expected Behavior

After `Sandbox.connect()` returns successfully:

- The sandbox remains in the `running` state.
- CLM refreshes `LastActiveMs` to the current time.
- Redis and CubeProxy are synchronized to `running`.
- The sweeper does not pause the sandbox until the configured idle timeout expires.
- The envd health endpoint is reachable.

## Actual Behavior

`POST /sandboxes/{id}/connect` returns HTTP 200 and CubeMaster reports the sandbox as `running`, but CLM still observes the old activity timestamp.

[...]

PR #1726, opened 2026-09-13
fix(network,lifecycle): per-request MTU, cross-node lease renewal, sweeper grace

rollroyces:
# Fix three sandbox lifecycle issues: per-request MTU, cross-node lease renewal, sweeper-after-resume race

Addresses #1673, #1690, #1692, #1683.

Includes a follow-up commit addressing feedback from the
in-house PR review bot (lease-renewer stop-on-resume gap, pool
tap MTU re-write, mtu_interface default-clobber fix, per-call
MTU resolution consolidation).

## Summary

Three independent fixes in the sandbox provisioning and lifecycle stack:

1. **#1673** — Guest MTU is no longer hardcoded to 1500. Operators can
   configure a runtime default via `mvm_mtu` or have the runtime
   mirror a host link's live MTU via `mtu_interface` (typically the
   Kubernetes parent bridge `cbr0`). Required for overlay networks
   whose transport MTU is below the Ethernet default.

2. **#1690 / #1692** — Cross-node resume no longer races the S3 lease
   reaper. The source Cubelet keeps the paused snapshot's export
   alive via a process-local lease renewer until Resume / Destroy /
   a Pause failure explicitly stops it.

3. **#1683** — The auto-pause sweeper no longer immediately re-pauses
   a freshly-resumed sandbox. A new `ResumeGrace` window (default 60 s)
   exempts resumed sandboxes from the idle check, gated on a
   `ResumedAtMs` timestamp the resumer stamps on success.

### Cubelet — network runtime (Task 2 / #1673)

* `Cubelet/network/runtime/config.go` — new `MTUInterface` config
  field, `DefaultGuestMTU` constant (1500), `EffectiveMTU()` resolver
  in three tiers (per-request → `MvmMtu` → live host-link MTU →
  default), `resolveMTUOnce()` helper for the single-shot
  EnsureNetwork path.
* `Cubelet/network/runtime/controller.go` — `acquireTap` now takes an
  MTU; the pool-hit branch calls `ensureTapMTU` so a pooled tap
  whose previous owner had a different MTU is re-MTU'd.
  `actualInterfaces` takes a pre-resolved MTU rather than calling
  `EffectiveMTU()` independently (transient netlink errors can
  no longer desync host TAP and guest eth0). cube-dev / cube-router
  use `EffectiveMTU()`.
* `Cubelet/network/runtime/{tap_lifecycle,startup_recover,tap_cleaner}.go` —
  every tap create / restore site now calls `EffectiveMTU()`
  instead of `cfg.MvmMtu` directly (no more `LinkSetMTU(tap, 0)`
  when `mtu_interface` mode is active and `mvm_mtu` is unset).
* `Cubelet/network/types/network.go` — new `NetRequest.MTU` field
  (omitempty, backward compatible). Plumbed through the
  EnsureNetwork path; the host TAP and the guest eth0 share the
  resolved value via a single `resolveMTUOnce` call so a transient
  netlink error cannot desync them. (`NetRequest.MTU` is set by
  the caller's `cube.master.net` annotation today; wiring it
  from a Master API is a follow-up.)
* `Cubelet/network/plugin_shim.go` — `buildEnsureNetworkRequestFromIntent`
  resolves the requested MTU in three tiers matching the runtime.
* `Cubelet/network/plugin_tap.go` — new `MTUInterface` TOML field,
  plumbed into the runtime Config. The MvmMtu assignment is now
  conditional so an unset `mvm_mtu` in TOML no longer clobbers
  the runtime default with zero.

No Rust changes — `CubeShim/shim/src/sandbox/net.rs` already passes
`Mtu` from `ShimNetReq` straight through to the guest kernel.

### Cubelet — pause lease renewer (Task 3 / #1690, #1692)

* `Cubelet/storage/lease_renewer.go` (new) — `Renewer` type with
  per-snapID renewal goroutines (30 s interval, lazy-init
  singleton on `localStorage`).
* `Cubelet/storage/lease_renewer.go` — `Unregister` is deterministic:
  the per-call context in `renew` derives from `entry.cancel`,
  so closing it aborts the in-flight `UploadSnapshot` immediately
  rather than racing a 5 s wall-clock against a 60 s per-call
  timeout. Aborted-by-Unregister errors log at debug so the
  normal log filter no longer shows the path-out as a failure.
* `Cubelet/storage/local.go` — `RegisterPauseLease` /
  `UnregisterPauseLease` / `ActivePauseLeases` accessors.
* `Cubelet/services/cubebox/pause_cow.go` — Register on Pause
  success when remote UUIDs were returned. Unregister when
  `destroyLiveAfterPause` fails (the snapshot stays on disk but
  Master will not drive a Resume).
* `Cubelet/services/cubebox/service.go` — Unregister on Destroy
  immediately before `bestEffortCleanupPauseSnapshot` runs.
  Unregister on Resume (Create-from-pause) when the resume-side
  cubebox is still findable in the manager.

### cube-lifecycle-manager — sweeper race (Task 4 / #1683)

* `cube-lifecycle-manager/internal/registry/registry.go` — new
  `Entry.ResumedAtMs` field, new `Registry.MarkResumed()` setter.
* `cube-lifecycle-manager/internal/config/config.go` — new
  `Config.ResumeGrace` field, default 60 s, env var
  `CUBE_LCM_RESUME_GRACE`.
* `cube-lifecycle-manager/internal/sweeper/sweeper.go` — new gate in
  `sweepOnce`; new `Options.ResumeGrace`.
* `cube-lifecycle-manager/internal/resumer/resumer.go` — stamp
  `ResumedAtMs` after success-bookkeeping writes.
* `cube-lifecycle-manager/cmd/cube-lifecycle-manager/main.go` — wire
  `cfg.ResumeGrace` into the sweeper Options.

## Tests

* `Cubelet/network/runtime/config_test.go` (new) — locks the
  `EffectiveMTU` resolution order, the live-link branch on Linux,
  and `resolveMTUOnce`.
* `Cubelet/storage/lease_renewer_test.go` (new) — covers
  idempotent registration, XFS / unknown-backend no-ops,
  per-snapID concurrent uniqueness, and the new
  deterministic Unregister path (returns within 2 s even
  mid-tick).
* `cube-lifecycle-manager/internal/registry/registry_resumed_test.go`
  (new) — covers `MarkResumed` forward-only semantics and Upsert
  preservation.
* `cube-lifecycle-manager/internal/sweeper/sweeper_test.go` — three
  new tests for the `ResumeGrace` gate (in-window skip,
  post-window act, zero-grace disable).

Verified locally on macOS via cross-compiled Go test binaries:
`cube-lifecycle-manager/internal/{registry,sweeper,resumer,config}`
all pass (no regressions, 30 + new cases). Rust build is
Linux-only — full test run belongs in the project's
`cube-sandbox-builder:ubuntu2004` Docker CI.

## Risk / compatibility

* **MTU change is additive.** Existing deployments that left
  `mvm_mtu` at 1500 see no change. Setting `mtu_interface = "cbr0"`
  is opt-in.
* **Lease renewer is idempotent and bounded.** A Renewer goroutine
  re-exports via `UploadSnapshot` every 30 s with a 60 s per-call
  timeout; Pause retries and process restart are safe. Unregister
  is deterministic — closing `entry.cancel` aborts the in-flight
  call instead of racing a wall-clock timeout.
* **ResumeGrace defaults to 60 s.** Set `CUBE_LCM_RESUME_GRACE=0`
  to disable the gate entirely (legacy behaviour). Operationally
  silent — a sandbox paused because it was genuinely idle for > 60 s
  after resume still pauses normally.

Autonomously-by: Hermes Agent: MiniMax-M3


chenhengqi (changes requested):
Please split this PR into smaller changesets focused on each specific issues.
BTW, most of issues you mentioned are already closed. Please check.

fslongjin:
This is a wide-ranging PR that includes multiple fixes. Some of the bugs have already been resolved by other PRs (for example, the MTU issue) and have been merged. We recommend that each independent feature or fix be submitted as a separate PR. Could you please split this PR into multiple sub-PRs and resubmit them? Thank you.

PR #1733, opened 2026-09-14
feat: add GAUGE perf metrics and replace --enable-ivshmem

aichitudou123:
Add host-side performance collection for sandboxes. The guest kernel module writes counters into a shared-memory device; the host maps that same region and reads them. The device has to exist in the template snapshot, so collection is opted in at template build with --enable-metric.

Drop the unused open-source --enable-ivshmem /
cube.master.enable_ivshmem path. The old demo ivshmem helpers now implement GAUGE only: CubeMaster, CubeShim, CubeAPI, and the Python SDK expose a single switch (--enable-metric / enableMetric / cube.master.perf.metric). A VM still has one ivshmem slot, so the two flags could never coexist.

How to collect metrics:

1. Build the template with --enable-metric:

   cubemastercli tpl create-from-image \ --image <sandbox-code:latest> \ --writable-layer-size 1G \ --expose-port 49999 --expose-port 49983 \ --probe 49999 \ --enable-metric

2. After the sandbox is running, insmod cube_gauge.ko in the guest.

3. On the host, read the sandbox gauge.shmem with open-cube-gauge-sdk:

   CGO_ENABLED=1 go run ./demo <sid>


lkml-likexu:
@aichitudou123 Any output demo and performance data ? Furthermore, it is unacceptable to merge changes to different components into a single commit.

kinwin-ustc:
> After the sandbox is running, insmod cube_gauge.ko in the guest.

Is `insmod` an automatic behavior of the kernel on the guest machine, or does it require manual triggering? For user-defined kernels, is some documentation required?

aichitudou123:
> > After the sandbox is running, insmod cube_gauge.ko in the guest.
> 
> Is `insmod` an automatic behavior of the kernel on the guest machine, or does it require manual triggering? For user-defined kernels, is some documentation required?

insmod is not a kernel auto-load, and it is not a manual guest-side step either.
cube_gauge.ext4 lives next to the guest kernel at
/usr/local/services/cubetoolbox/cube-kernel-scf/cube_gauge.ext4.
It contains cube_gauge.ko, which must match the vermagic of that
directory's vmlinux.
When the template was built with --enable-metric and that ext4 is present:
1. Shim attaches the host file as virtio-pmem; the guest sees it as /dev/pmem2.
2. cube-agent mounts /dev/pmem2 at /run/cube-gauge.
3. After restore from the template, Shim asks the agent to
   insmod /run/cube-gauge/cube_gauge.ko.
If the flag is off or the ext4 is missing, the sandbox still starts;
we skip attach and skip insmod. The kernel never loads this module by itself.

**For a custom guest kernel:** yes, a short note in the docs is needed.
cube_gauge.ko is out-of-tree, so a module built for the stock guest kernel
will not insmod on a different vmlinux. Rebuild the module against that
guest kernel, pack it into cube_gauge.ext4, and place it beside the matching
vmlinux under cube-kernel-scf/. We will add this to the PR description / user guide.

zyl1121:
Thanks for working on this. My main concern is whether it is worth giving up generic ivshmem for GAUGE.

CubeSandbox already has a per-sandbox resource metrics path through `Task.Stats` and `/v1/metrics/resource`. CPU and memory are implemented today, and the existing stats structures already have the basis for disk and network metrics as well, even though those fields are not populated yet.

The existing collector defaults to 5s sampling, while GAUGE defaults to 15s and supports intervals down to 1s. So both paths are currently targeting seconds-scale resource monitoring. If the main benefit of GAUGE is significantly lower collection overhead at that rate, it would be useful to see the performance difference between the two paths.

At the same time, this PR removes the generic `enableIvshmem` interface and makes the ivshmem slot effectively dedicated to GAUGE. Generic ivshmem has useful cases beyond metrics, such as streaming framebuffer data or using shared-memory ring buffers for high-throughput host/guest data exchange. With this change, users who need those kinds of shared-memory channels can no longer request an ivshmem device at all.

Given that the existing metrics path can be extended to cover these metrics, is this tradeoff necessary? Even if GAUGE is added, I think it should at least coexist with the generic ivshmem capability rather than replace it.

kinwin-ustc:
> Given that the existing metrics path can be extended to cover these metrics, is this tradeoff necessary? Even if GAUGE is added, I think it should at least coexist with the generic ivshmem capability rather than replace it.

@zyl1121 I agree with your opinion. @aichitudou123 In the kernel of the sandbox, `insmod` is not a very user-friendly operation for open source users. If the only purpose is monitoring and data collection, I feel there are solutions with better compatibility and stability. This choice should be re-evaluated.

aichitudou123:
> > Given that the existing metrics path can be extended to cover these metrics, is this tradeoff necessary? Even if GAUGE is added, I think it should at least coexist with the generic ivshmem capability rather than replace it.
> 
> @zyl1121 I agree with your opinion. @aichitudou123 In the kernel of the sandbox, `insmod` is not a very user-friendly operation for open source users. If the only purpose is monitoring and data collection, I feel there are solutions with better compatibility and stability. This choice should be re-evaluated.

Thanks @zyl1121 @kinwin-ustc.
1.We should compare this approach with the existing one, and decide whether we should use it.
2.We should not overwrite generic ivshmem. The hypervisor currently only supports one ivshmem device, so we may need to add support for multiple ivshmem devices.
3.If insmod is not appropriate, we may need to consider building it into the kernel.

PR #1736, opened 2026-09-14
feat(s3lvol): hot upgrade — sandbox I/O pauses instead of breaking

fslongjin:
## The problem

Upgrading the s3lvol target process currently goes through a "planned stop": `nvme disconnect`
first, then unload the lvstore. Both steps remove the block device from the host, so a running
sandbox immediately sees I/O errors (`vda` reports I/O errors, ext4 goes read-only). Upgrading
therefore means interrupting the workload.

In short: **an upgrade today is an outage.**

## The goal

When the target is upgraded, sandbox I/O should **stall for a moment and then continue in
place**, and the host should still see **the same `/dev/nvmeXnY`** — the block device is never
deleted or recreated.

## Approach: make the upgrade a controlled crash recovery

No disconnect, no unload. The sequence is:

1. **Online**: `flush` (push everything acknowledged to S3) + `checkpoint` (snapshot the chunk
   map, truncate the journal)
2. Record a layout snapshot, then **SIGKILL** the target outright (a crash, not a graceful stop)
3. The replacement rebuilds **exactly the same NQN / (subsys, nsid) / UUID layout**
4. The host's nvme_tcp reconnects on its own to the same device, and I/O resumes

This reuses the recovery logic that **crash self-healing** already uses (`Restart=on-failure`
takes this path); it is not a new mechanism.

### New

| Item | Purpose |
|---|---|
| `scripts/rcow_upgrade.sh` | The hot stop: online flush + checkpoint + layout snapshot + SIGKILL + residue cleanup. Deliberately does **not** disconnect, unload, or touch `active_lvols` |
| `rcow_get_build_info` RPC + `s3lvol_tgt --print-build-info` | **Version gate**: compares the on-disk format versions, the subsystem grid and the NQN prefix between the running and the candidate binary *before* anything is swapped. Incompatible sides are refused; there is no in-place conversion. The candidate is required on a real hot stop, and the systemd path gets it from the marker |
| `rcow_active_bdev_batch` RPC | Restores a batch of volumes in a single round-trip. One `rcow_active_bdev` per volume is one process startup per volume, all inside the pause window |

### Fixes

1. **Namespaces were lost after a hot start** (the critical one)

   Replay answered from a registry entry as if it meant "this process has already exposed it",
   so every volume came back as `already_active` and the real attach was skipped — after the
   restart the namespace count was 0 and the host removed the gendisks.

   Fixed by separating **"recorded"** (the file says what a restart should restore) from
   **"already attached by this process"**; only the latter may answer `already_active`.

2. **`rcow_flush_lvstore` hung indefinitely under write load**

   The 30 s deadline check sat *after* two early returns (`in_flight > 0` and
   `super_sync_active`), and `kick` refills `in_flight` before checking — so as long as writes
   keep arriving the deadline is never reached and the callback never fires. The deadline is
   now read *before* those gates and carried in a separate `drain_expired` flag; once set, no
   new round starts and only the uploads already in flight are waited for.

   This also removes a side effect: the old timeout path set the flusher `stopping` permanently,
   so a `--dry-run` — which must be side-effect free — silently stopped the uploads of a live
   volume.

3. **Hot start deadlocked on a host-wide `sync`**

   The `sync` inside `rcow_replay_registry()` goes through `sync_bdevs()`, which waits for
   writeback on **every** block device; the sandbox's nvme_tcp device is mid-reconnect, so its
   writeback cannot complete — and the listeners that would let the controller reconnect are
   added *after* that step. Observed: a 9-minute deadlock, until the kernel deleted the
   controllers and I/O turned from "paused" into "failed".

   Fixed by `fsync`-ing only the one file (and its directory entry) that actually has to survive.

### Control plane

`cubecow` gains `rpc_connect_budget_ms` (default 60 s): during the hot-upgrade window the socket
is unlinked, so **a connect failure is the expected case, not an error**. Only connection
establishment is retried (100 ms doubling, 2 s ceiling); once connected, a JSON-RPC error
response is an answer and returns immediately — retrying a real error would only turn a fast,
correct failure into a slow one. There is deliberately no watchdog that deactivates volumes or
marks the node NotReady: a control-plane outage must not become a data-plane action.

### Other

- `--keep-connected` on `rcow_stop.sh` is now **refused**: it reads as "leave the host
  connected" but still unloads the lvstore, which hands the host a namespace-removal AEN — a
  silent hot-removal, worse than the plain stop it was meant to avoid.
- `cube-s3lvol-stop.sh` / `cube-sandbox-s3lvol.service`: the stop script's three branches, and
  the rationale for `KillMode=process` (`control-group` would let systemd's post-ExecStop sweep
  reach a target that a stop script deliberately left running).

### Follow-ups from review

- **The target is now identified by its RPC socket as well as its binary name**, and a
  hot-restart marker that names a live target survives a failed identity check instead of being
  discarded. The strict path comparison was the one check that cannot hold during an upgrade --
  the candidate has already moved into place -- so a marker could be consumed and thrown away,
  dropping the stop into the full-teardown branch that this PR exists to avoid. Only a marker
  that is stale by construction (another boot, or malformed) is now removed outright.
- **`rpc_connect_budget_ms` bounds the whole call, queueing included.** It used to start its
  clock after acquiring the connection mutex, which the retry loop holds across its sleeps, so
  concurrent callers each got a fresh allowance and the last one waited about N x the configured
  budget.
- **`connect_backoff` clamps the shift amount** rather than relying on `checked_shl`, which
  rejects only a shift past the bit width: at two values it returned a zero-length sleep instead
  of the ceiling.
- **The build stamps are evaluated once per build**, not once per translation unit, so the three
  `git` calls no longer re-run for every compile and objects cannot be stamped differently if
  the tree goes dirty part-way through.
- **CI now exercises the Rust half.** The crate's own tests are run by `cubecow-test-native`, and
  `cubecow/**` triggers `unit-test-check` (which owns them) instead of `cube-s3lvol-test`, whose
  job never reads the crate.

### Follow-ups from the second review

- **The version gate now runs on the path that ships.** The systemd stop called the script with no
  arguments, so the gate was skipped on the only entry point that is actually wired up, while the
  README described `--candidate` as an operator option. A candidate is now required on a real
  run, and it is deliberately *not* defaulted to `RCOW_TGT_BIN`: an upgrade switches the versioned
  directory in only after the old process is confirmed dead, so that path still names the outgoing
  binary while the stop is running, and a gate handed it would compare the running build with
  itself. The candidate travels in the hot-restart marker instead -- the writer is the only side
  that knows it -- and the stop path forwards what it finds there.
- **The hot stop pins the initiator timeouts.** The connect flags only reach a fresh connection
  and an upgrade runs against controllers that already exist, so their rollback budget was never
  pinned, while the unit's `TimeoutStopSec` was sized against the value it was supposed to have.
  A `fast_io_fail_tmo` that is already enabled is now a refusal rather than a warning: that
  setting turns every pause into a fast failure and nothing else undoes it.
- **Two states that used to succeed by breaking something now refuse.** A target that is alive but
  does not answer RPC had its residue cleared, which unlinks the socket the live process is
  listening on and leaves it with no clean exit; and a marker for this boot naming a pid that is
  not a running target fell through to the full teardown, deleting a live target's gendisks.
  Both refuse, leave everything in place, and exit non-zero. Recovery from a refusal is
  `systemctl reset-failed` and then the stop by hand, which is the trade taken deliberately.
- **`rcow_hot_stop.sh` is renamed `rcow_upgrade.sh`.** It stopped being only a stop once the
  version gate and the layout snapshot went into it.

### Unit / regression

| Suite | Result |
|---|---|
| `s3_flush_test`, new case `[10]`: under load a drain must report a timeout rather than hang | **102 passed / 0 failed** (the same case was **100 passed / 2 failed** before the fix — the old code hung and never called back) |
| `run_control_test.sh`, new case `[7b]`: activating a "recorded but not exposed by this process" volume must not answer `already_active`, and must really create the namespace at the recorded position | failed at the first step before the fix |
| `cubecow`, new `connect_backoff` unit test: the backoff doubles and holds at a 2 s ceiling | passes |

### On a real machine, with a real workload

On a machine running this build, fio runs inside a sandbox (**4k randrw 70/30, direct,
iodepth 4, 900 s runtime**, data on the s3lvol volume) while the target is hot-upgraded **in the
same time window**.

| Criterion | Result |
|---|---|
| Business I/O does not error | fio `error=0`, ran the **full 900 s**; **zero I/O errors** in the sandbox's dmesg, no ext4 errors |
| Same block device | **`diskseq` and `uuid` unchanged for all 18 nvme device nodes** — no gendisk was ever deleted |
| Layout restored exactly | `rcow_verify_active --expect` **rc=0** (same subsystem, nsid and device) |
| Controllers reconnect in place | **all 32 back**; `new ctrl`=0, `removing namespace`=0 |
| Sandbox survives | alive throughout, fio never broke, finished after the upgrade |

**Performance impact**:

| | mean | p99 | max |
|---|---|---|---|
| read | 0.46 ms | 2.87 ms | 435 ms |
| write | 0.62 ms | 2.87 ms | 39.9 s |

- **A p99 of 2.87 ms** means performance **outside** the window is identical to normal — no
  regression.
- **The 39.9 s write maximum** is the one I/O that happened to straddle the window. Its latency
  **equals the length of the pause window itself**: the completion-latency log shows a single
  **39.93 s gap** in which no I/O completed, spanning exactly from the SIGKILL to the
  replacement reporting ready. It **waited; it did not fail** — which is the semantics this work
  is about.

The **≈40 s pause window** breaks down as follows (each row is verifiable in the host logs):

| Time | What |
|---|---|
| 8 s | SIGKILL + systemd restarting the target |
| 5 s | creating 32 subsystems, the WAL bdev, the S3 namespace |
| **30 s** | **the target attaching the lvstore itself (blobstore metadata load + journal/WAL replay)** ← dominant |
| 2 s | restoring volumes and adding listeners |

> The dominant part is the target's own attach cost, which is **not in this PR's code**. It is
> the one place worth optimizing, and the lever for shortening the window later.

## Compatibility and risks

- **Version gate**: `super_version / wal_version / ckpt_version / journal_op_max / num_subsys /
  ns_per_subsys / nqn_prefix` must match on both sides, otherwise the upgrade is **refused**;
  there is no in-place conversion.
- **The journal has no version field**, so `journal_op_max` is what guards it: a log written by
  a newer build is indistinguishable from a torn tail to an older reader, and replay **stops
  there by design** — which would be a silent truncation. The gate is the mitigation.
- **The hard ceiling on the pause** is the kernel's `ctrl_loss_tmo`; past it the kernel deletes
  the controllers and I/O turns from "paused" into "failed". This build connects with
  `reconnect_delay=1` / `ctrl_loss_tmo=1800` (30 minutes); every controller must be checked
  before an upgrade.

## Not included yet

- **The production orchestration script is not landed**, so one-click upgrades still take the
  planned (disconnect) path. This PR delivers the mechanism, reachable by hand via
  `rcow_upgrade.sh`.
- Cross-SPDK layout compatibility has no end-to-end verification.
- The `TimeoutStopSec` edge (a stop taking longer than 600 s) has not been stressed.


changpe1 (commented):
## Review notes (items 1–3)

The attach / flush / recorded-vs-attached core still looks right. These comments are the remaining stop-path and version-gate issues. Marker identity + delete-after-honour was already tightened on HEAD; the leftover problem is what ExecStop does when consume returns false, plus production still taking the planned stop, plus the optional gate.

I did not re-run dataplane / fio.

CubeS3lvol/scripts/rcow_common.sh
```diff
@@ -586,6 +707,231 @@ rcow_target_alive()
...
+		[ "${boot}" != "${now}" ]; then
+		rm -f "${RCOW_HOT_MARKER}"
+		return 1
+	fi
+
+	if ! printf '%s\n' "$(rcow_target_instances)" | grep -qxF -- "${pid}"; then
+		rcow_warn "${RCOW_HOT_MARKER} names pid ${pid}, which is not a running \
+target; leaving the marker in place rather than discarding the intent"
+		return 1
+	fi
```
changpe1, line 791: **Hot-intent consume still maps to a planned stop on failure (finding 1, residual).**

Identity is now `rcow_target_instances` and the file is only unlinked after a successful honour (or a stale boot_id). That fixes the renamed-binary false negative that used `rcow_pid_is_target`.

The contract below still undoes it:

> A marker left in place still reads as absent, so the stop falls through the same way it did before.

`cube-s3lvol-stop.sh` treats **any** consume failure as “nobody asked for a hot restart” and, if a target is still visible, runs `rcow_stop.sh` (disconnect + unload). Leaving the marker on a pid mismatch is therefore only half a fix: the intent survives on disk, but this stop still removes the gendisks.

If consume returns 1 because the named pid is not (yet) recognised as the live target, ExecStop should **fail closed** — refuse, leave the marker, do not fall through to `rcow_stop.sh`. Same for a missing `rcow_hot_stop.sh` (already refused). A missing/stale marker is the only case that should take the planned path.

deploy/one-click/scripts/systemd/cube-s3lvol-stop.sh
```diff
@@ -41,9 +45,29 @@ fi
...
+    # Refusing rather than falling back: the full teardown drops the nvme
+    # controllers, and the upgrade that wrote the marker is not expecting it.
+    log "CubeS3lvol: hot restart was requested but ${RCOW_HOT_STOP} is missing or not executable"
+    exit 1
+  fi
+  log "CubeS3lvol: hot restart requested; stopping via rcow_hot_stop.sh, initiator untouched"
+  "${RCOW_HOT_STOP}"
+elif rcow_target_alive; then
   log "CubeS3lvol: target alive, full teardown via rcow_stop.sh"
   "${RCOW_STOP}"
```
changpe1: **Consume failure falls through to a full teardown (finding 1 + 2).**

```bash
if rcow_hot_marker_consume; then
  "${RCOW_HOT_STOP}"          # no --candidate
elif rcow_target_alive; then
  "${RCOW_STOP}"              # disconnect + unload
```

Two independent problems:

1. After the marker rewrite, a *live* intent that consume cannot honour (pid not in `rcow_target_instances`) still hits the `elif` and runs `rcow_stop.sh`. That is the outage this PR removes, taken by the branch that exists to distinguish hot vs planned stop. Fail closed instead: if a marker file exists and is this boot, do not disconnect.

2. Even the success path never passes `--candidate`. systemd ExecStop is the production entry for a marker-driven stop, so the version gate (journal_op_max / on-disk formats) is skipped unless a human invoked `rcow_hot_stop.sh` by hand.

Also: nothing in this PR writes the marker (`install.sh`, k8s `cube-s3lvol-entrypoint.sh`, terraform still `systemctl stop` / SIGTERM). One-click and DaemonSet rolls keep the planned path after merge. Either wire `rcow_hot_marker_write` + `--candidate` there, or do not describe fleet upgrades as pause-not-break.

deploy/one-click/systemd/cube-sandbox-s3lvol.service
```diff
@@ -7,24 +7,28 @@ RequiresMountsFor=/data/cubelet /data/log
...
 Type=simple
 EnvironmentFile=-/usr/local/services/cubetoolbox/.one-click.env
 ExecStart=/usr/bin/bash /usr/local/services/cubetoolbox/scripts/systemd/cube-s3lvol-supervise.sh
-# Conditional stop: full teardown (disconnect -> flush/unload -> SIGTERM)
-# when the target is alive; when it has crashed, only clean target-side
-# state and never disconnect the NVMf initiator, so nvme_tcp reconnects
-# to the rebuilt NQN/NSID grid without an I/O interruption.
+# See the header of cube-s3lvol-stop.sh: this stop never disconnects the NVMf
+# initiator, so nvme_tcp reconnects to the rebuilt NQN/NSID grid.
 ExecStop=/usr/bin/bash /usr/local/services/cubetoolbox/scripts/systemd/cube-s3lvol-stop.sh
```
changpe1: **This comment is not true of the default ExecStop path (finding 2).**

`cube-s3lvol-stop.sh` has three branches. The planned-stop branch (`target alive`, no honourable marker) still runs `rcow_stop.sh`, which **does** disconnect the initiator and unload the lvstore.

As written, an operator reading the unit will think `systemctl stop cube-sandbox-s3lvol` keeps namespaces in place. Today that is only true with a valid hot-intent marker (not written by install/k8s in this PR) or when the target is already dead.

Please restore the previous wording (conditional: disconnect only on a planned alive-target stop) until the orchestrator actually writes the marker and ExecStop always takes `rcow_hot_stop.sh` for upgrades.

CubeS3lvol/scripts/rcow_upgrade.sh
```diff
@@ -0,0 +1,300 @@
...
+fi
+
+# ==========================================================================
+# Before the target is touched. The running side describes itself only while it
+# is alive, and the two descriptions can only be compared while both exist, so
+# this is the last moment the decision can be made at all.
+if [ -n "${CANDIDATE}" ]; then
+	rcow_step "version gate against ${CANDIDATE}"
+	if ! rcow_version_gate_check "${CANDIDATE}"; then
+		fail_live "the version gate refused the hot upgrade"
```
changpe1, line 223: **Version gate is optional on the path that actually kills the target (finding 3).**

The journal has no version field; an unknown op is treated as a torn tail and silently truncates acknowledged writes. The comment above this script is explicit that the gate must run *before* the kill, while both binaries can still describe themselves.

`if [ -n "${CANDIDATE}" ]` means:

- `cube-s3lvol-stop.sh` → `rcow_hot_stop.sh` with no args skips the gate entirely
- a hand invocation without `--candidate` will flush, checkpoint, and SIGKILL, then `rcow_start.sh` execs whatever `RCOW_TGT_BIN` is now

The mechanism cannot bootstrap itself (`rcow_get_build_info` missing → refuse), which is good — but only if the gate is actually invoked.

Please require `--candidate` on any non-`--dry-run` hot stop (or have ExecStop pass the staged binary). A missing candidate should `fail_live`, not skip the comparison.

CubeS3lvol/scripts/rcow_common.sh
```diff
@@ -586,6 +707,231 @@ rcow_target_alive()
...
+
+	python3 - "${running}" "${candidate}" <<'PY'
+import json, sys
+
+# journal_op_max is here because the journal has no version field: see the note
+# above this function.
+EQUAL = ("super_version", "wal_version", "ckpt_version", "journal_op_max",
+         "num_subsys", "ns_per_subsys", "nqn_prefix")
```
changpe1, line 854: **`rpc_api_version` is emitted but never compared (finding 3).**

The PR description says the gate compares on-disk formats **and the RPC API version**. `vbdev_s3lvol_buildinfo.c` does include `rpc_api_version` in the document; this `EQUAL` tuple does not.

A missing field on either side is correctly a refusal for the fields you *do* list. `rpc_api_version` can drift with no refusal. Additive RPCs (`rcow_active_bdev_batch`) may not need a bump, but a real API break would not be caught here — and replay would then depend on PARSE_FAIL treating the whole batch as pending, which is a worse failure mode than refusing before SIGKILL.

Either add `rpc_api_version` to `EQUAL` (and test it next to `journal_op_max`), or drop the claim from the PR text / `--print-build-info` comments.

CubeS3lvol/scripts/rcow_upgrade.sh
```diff
@@ -0,0 +1,300 @@
+#!/usr/bin/env bash
```
changpe1, line 1: Can we rename this script with `rcow_upgrade.sh` ?

deploy/one-click/scripts/systemd/cube-s3lvol-hot-upgrade.sh
```diff
@@ -0,0 +1,282 @@
...
+  # against as if it described the layout being swapped now.
+  rm -f "${RCOW_HOT_SNAPSHOT}"
+  # `reset-failed` first, as the cold path does: a stop on a failed unit is a
+  # no-op that reports success, and a unit left failed by an earlier refusal or
+  # crash-loop is a state this can arrive in. Without it the stop does nothing
+  # while claiming to have done something, and the target it was asked to
+  # replace is still there for the guard below to refuse.
+  systemctl reset-failed "${SERVICE}" >/dev/null 2>&1 || true
+  if ! systemctl stop "${SERVICE}"; then
+    warn "the in-place stop failed; falling back to a cold stop"
```
changpe1: **A failed in-place stop must not fall back to a cold teardown of a target that is still serving (finding 1).**

`systemctl stop` returning non-zero here means `rcow_upgrade.sh` refused. Every failure path in it before the kill leaves the target unsignalled and running (`fail_live` says so explicitly). The only non-zero path after the kill is "pid survived SIGKILL". So this branch runs in the state "the target is alive and serving; the in-place route did not work this time" — and the response is `MODE=cold` → `stop_and_confirm` → a second stop with no marker → `rcow_stop.sh` → `nvme disconnect` + unload. That is the outage this mechanism exists to prevent, taken automatically because an attempt failed, on a target that was still fine.

The branch a few lines below already has the right posture for the same situation: "the target is still running after the stop; leaving everything alone", exit 1. This fallback should do the same: refuse, leave the initiator connected, exit non-zero. Do not disconnect a live target just because the hot path could not complete.

A concrete way to reach it: the pre-check only asks whether a target process exists, not `rcow_target_ready`. A target that is up but has not attached its lvstore yet still answers `rcow_get_build_info`, so the gate can pass, the flush/checkpoint then fails, and we land here with a healthy process and a cold stop queued.

This also composes with the marker being unlinked on consume (see the comment on `rcow_hot_marker_consume`): after this refusal the marker is already gone, so the prescribed recovery (`systemctl reset-failed` then stop again) takes the planned path as well.

changpe1 (commented):
## Review notes (items 1–2)

The attach / flush / recorded-vs-attached core still looks right. These two remaining holes are on the stop / marker path and they compose: a refused hot stop still tears down a live target.

I did not re-run dataplane / fio.

CubeS3lvol/scripts/rcow_common.sh
```diff
@@ -586,6 +714,275 @@ rcow_target_alive()
...
+	fi
+
+	if ! printf '%s\n' "$(rcow_target_instances)" | grep -qxF -- "${pid}"; then
+		rcow_warn "${RCOW_HOT_MARKER} names pid ${pid}, which is not a running \
+target; leaving the marker in place rather than discarding the intent"
+		return 2
+	fi
+
+	RCOW_HOT_CANDIDATE="${candidate}"
+	rm -f "${RCOW_HOT_MARKER}"
```
changpe1: **The marker is unlinked as soon as it is recognised, not once the upgrade has actually happened (finding 2).**

The comment above says the file is removed only once it has been honoured. This `rm -f` runs before the caller does anything with it: `cube-s3lvol-stop.sh` calls `rcow_hot_marker_consume`, gets 0, and only then runs `rcow_upgrade.sh`. But `rcow_upgrade.sh` can still refuse *after* that point and leave the target running, by design:

- the version gate (`rcow_version_gate_check` → `fail_live`)
- `hot_online_op` flush/checkpoint failing with anything but the tolerated `"code": -110`
- an initiator with `fast_io_fail_tmo` enabled

Each of those exits 1 via `fail_live` — "the target is still running and was not signalled" — and the marker is already gone.

The recovery both headers prescribe for a non-zero ExecStop (`systemctl reset-failed` and then the stop by hand) then does exactly the outage this path exists to avoid: consume returns 1 (no marker), so the stop falls through to `elif rcow_target_alive` → `rcow_stop.sh` → `rcow_disconnect_all` + `rcow_unload_lvstore`. Combined with the orchestrator's cold fallback on a failed `systemctl stop`, a single refused hot attempt deletes the gendisks.

Leave the marker until the kill has succeeded (or the target is confirmed gone). A failed attempt must still look like a hot intent on retry, not like a planned stop.

deploy/one-click/scripts/systemd/cube-s3lvol-hot-upgrade.sh
fslongjin: Fixed, with one correction to the premise.

`systemctl stop` returning non-zero is not what a plain refusal looks like here. The ExecStop failure shows up as a failed unit, and the stop still returns 0 -- so what answered was the branch below it, "the target is still running after the stop", which was already refusing. The failure branch is reached through a job-level failure instead (a stop timeout, for one), and it refuses now, for the reason you gave: every failure before the kill leaves the target running and unsignalled. A marker that cannot be written refuses too, same reasoning -- without it the stop takes the planned route on its own.

You were also right that this composes with the marker, and that turned out to be the sharper half. With the marker now left in place, the retry has to be able to reach the in-place stop -- and it could not. A refusal leaves the unit stopped while the target it would not kill keeps running, so a stop has no unit left to drive ExecStop, and a start is refused by rcow_start.sh's instance guard. The upgrade now drives that stop itself when the unit is not running, with the same script ExecStop would have run. That is what makes the refusal a retryable state rather than the end of the road.

Verified on a node: a forced refusal (a controller with fast_io_fail_tmo set) left the target running, the initiator connected, the diskseq unchanged and the marker in place. The next run drove the stop itself, completed the upgrade and spent the marker, with the diskseq still unchanged and the controllers live again afterwards.

CubeS3lvol/scripts/rcow_common.sh
fslongjin: Fixed: the marker is no longer unlinked where it is merely recognised. `rcow_hot_marker_consume` leaves it and hands back the candidate, and `rcow_upgrade.sh` spends it once the target it names is gone -- or, in the branch that found nothing to kill, once that is confirmed. Each of the paths you listed (the gate, a flush or checkpoint that is not the tolerated -110, an initiator with fast_io_fail_tmo) now leaves the intent behind, so a retry is still an in-place stop rather than the planned teardown.

The tripwire in the negative suite moved with it: the marker was "never touched", and is now pinned as "never written, and removed only after the kill is confirmed", which is the property that was missing.

changpe1:
LGTM.

PR #1741, opened 2026-09-14
feat(kubernetes): add operator DNS resolver opt-in

purebluesong:
## Background

CubeVS deliberately denies private, link-local, and loopback destination ranges by default. This protects host-local services, cluster internals, and cloud metadata from sandbox traffic.

In Kubernetes, the platform DNS resolver is commonly exposed through a private Service address. A guest can receive that resolver through Cubelet's administrator-managed DNS configuration, but an IP-only sandbox policy cannot reach it unless the resolver is also admitted through CubeVS. Explicitly placing that address in every sandbox request works, but leaks environment-specific infrastructure into application callers.

This PR does not change the default policy contract. It adds an explicit operator opt-in for Kubernetes deployments that need the platform-configured resolver reachable independently of a sandbox's domain/L7 egress policy.

## Design

```yaml
cubeNode:
  dns:
    sandbox:
      autoAllowDefaultResolvers: false # default
```

When `false`, the existing behavior is unchanged: resolver CIDRs are appended only for domain/L7 policies.

When `true`, Cubelet admits a resolver for an IP-only policy only when it is both:

1. Configured by the operator as a Cubelet default resolver.
2. Actually used by the guest's effective DNS configuration.

Per-sandbox `dns_config.servers` cannot enable this exception. The admitted operator resolver CIDRs are persisted separately so an IP-only policy update preserves only the operator-approved subset, rather than arbitrary resolver CIDRs recorded by a domain-policy create.

The Chart exposes the setting as an environment variable and writes it into Cubelet dynamic configuration. It defaults to `false`, preserving existing installations and security semantics.

## Security boundary

The current CubeVS `allow_out` data plane is CIDR-wide, not DNS-port-aware. Enabling this option admits all ports to the configured resolver IP. The feature is therefore intentionally operator-controlled and disabled by default. A future port-aware policy map could narrow the exception to UDP/TCP 53.

## Validation

- Full Linux tests pass for `Cubelet/pkg/config`, `Cubelet/network`, and `Cubelet/network/runtime`.
- Linux/amd64 Cubelet build passes.
- Helm render confirms the opt-in produces `CUBE_SANDBOX_AUTO_ALLOW_DEFAULT_DNS_SERVERS=true` for `cube-node`.
- Both Kubernetes entrypoint scripts pass `bash -n`.
- Tests cover default IP-only behavior, operator opt-in admission, exclusion of caller-provided resolvers, domain/L7 compatibility, policy updates, and durable state round-tripping.

Assisted-by: Codex:GPT-5


chenhengqi (commented):
The PR summary does not explain the `why`.

purebluesong:
## Sanitized integration validation

The policy behavior was additionally validated in a private Kubernetes environment using a v0.6 backport for the legacy external network-agent architecture. This backport carries the same resolver-admission behavior as this PR while preserving the existing public-resolver guard.

Test case:

```text
allow_internet_access=false
allow_out=[]
guest DNS=private Kubernetes Service resolver
```

Observed result:

```text
MicroVM sandbox created successfully
Guest resolved a private application hostname successfully
No explicit resolver /32 was supplied by the caller
Temporary sandbox cleanup completed successfully
```

The same code path was also covered by Linux unit tests for private resolver admission, mixed private/public resolver lists, and the pre-existing public-resolver and domain/L7 behavior.

This validates the intended policy semantics and real DNS data path. It does not claim that the current PR head was deployed as-is in that environment; the deployed validation artifact was the compatibility backport required by the older network-agent architecture.

purebluesong:
## Follow-up: resolver trust boundary tightened

The automated review correctly identified that an effective guest resolver can originate from caller-controlled `dns_config.servers`. The prior revision used its private-address classification as an implicit admission signal, which was too broad because the current policy representation is CIDR-only.

This revision changes the admission source:

- Cubelet separately resolves the trusted default resolver list without a sandbox request.
- Automatic admission is limited to the intersection of that list and the guest effective resolver list.
- In IP-only policies, the intersection is additionally limited to private/link-local/loopback IPv4 resolvers.
- Caller-provided resolvers, including private addresses, require explicit `allow_out`.
- Only resolver CIDRs actually admitted at create are persisted, preventing a later policy update from promoting an untrusted resolver.

New Linux tests cover a caller-provided link-local resolver in both IP-only and domain-policy cases, plus the create-path persistence behavior. Full `Cubelet/network` tests pass in the Linux builder.

chenhengqi:
The current behavior is by design. You can add the DNS Server IP to allow_out when creating sandbox.

purebluesong:
## Follow-up on DNS policy intent

Thanks for clarifying that explicitly putting the DNS server IP in `allow_out` is the current intended behavior. We confirmed that this works in our Kubernetes environment.

The revised PR is therefore no longer a default-behavior bug fix. It proposes an optional, operator-controlled Kubernetes integration:

```yaml
cubeNode.dns.sandbox.autoAllowDefaultResolvers: false # default
```

When explicitly enabled, only resolvers configured by the operator as Cubelet defaults and actually used by the guest are admitted for IP-only policies. Per-sandbox `dns_config.servers` cannot trigger the exception. Default behavior remains unchanged, and the Chart wires the flag into Cubelet dynamic configuration.

This lets a platform owner keep resolver topology out of application sandbox requests while retaining an explicit deployment-level security decision. We also document the current CIDR-wide limitation (not DNS-port-only). Full Linux config/network/runtime tests, Cubelet build, Helm render, and entrypoint syntax checks pass; repository CI has been retriggered.

chenhengqi:
@purebluesong Could you please open an issue for this feature? Let's discuss to make sure we are on the same page. Thanks.

purebluesong:
Opened #1776 to discuss the Kubernetes operator-level design and confirm the recommended integration path. I will keep this PR focused on the agreed design after that discussion.

PR #1748, opened 2026-09-15
Virtiofs: speed up migration preserialization

lisongqian:
This patch replaces the default directory-tree DFS (PathReconstructor) with a new StoreOnlyReconstructor that only iterates the inode store and uses readlinkat on /proc/self/fd to recover each inode's absolute path, then reverse-looks up parents through an incrementally-built path -> inode hash table. Cost drops from O(shared-tree) to O(inode-store).


ls-ggg:
LGTM.

Internal CICD sdk_compat e2e on this PR (merge commit of the branch into master at the time of the run): **273 passed, 0 failed, 14 skipped, 1 xfailed**. About 5m19s. No errors or reruns.

Skips are environment gates, not regressions: platform AutoPause/lifecycle (flag off), CubeAPI simple-key auth (not enabled), and one L7 HTTPS case (image does not trust the intercept CA). The xfail is the existing host-mount symlink-escape case (`test_symlink_escapes_allowed_prefix`); it is unrelated to this change.

Same-node Pause/Resume, Snapshot/Rollback/Clone, and host-mount paths in that suite stayed green after deploying the store-only virtiofsd build. That is a useful no-regression signal for everyday sandbox lifecycle.

It is **not** coverage of the new preserialization path itself. The changed virtiofsd files do not map to a dedicated live-migration / large-tree / `drop_caches` e2e, and Cube Pause is typically a VMM memory snapshot plus cubecow rather than QEMU virtiofs device migration. So this run does not measure the O(shared-tree) → O(inode-store) claim or the filter-pin fix. Those still need a targeted virtiofs live-migration check (including the DFS fallback flag) if we want them proven in CI.

Issue #1569, opened 2026-08-29
[Bug Report] One-click / cubetoolbox upgrade silently drops third-party VolumePlugins (binary + volume_plugins registration)


HuChundong:
## Summary

**中文摘要：** 一键安装树（`/usr/local/services/cubetoolbox/.../plugin/`）以及同树下的 `conf.yaml` / `config.toml` 在平台升级时会被覆盖或重置。官方示例与文档把第三方 VolumePlugin 引导安装到该目录，导致升级后第三方 driver 静默消失，所有依赖该 volume 的 sandbox 创建失败（例如 `unknown driver: juicefs`），租户持久卷不可用。协议本身允许 `binary_path` 指向任意绝对路径，但文档未警示「装进 toolbox 会被升级清掉」，也未给出升级安全的安装布局与配置保留策略。

Upgrading CubeSandbox (one-click / cubetoolbox package replace) can **silently remove third-party VolumePlugins** that were installed under the documented `CubeMaster/plugin/` and `Cubelet/plugin/` paths, and can also **reset `volume_plugins` registration** in `conf.yaml` / `config.toml` that live in the same tree.

After upgrade, create/attach with a custom `driver` fails with errors such as `unknown driver: <name>` / `no plugin registered for driver "<name>"`, while tenant data on the backend (object store + metadata DB) may still be intact. This is an **upgrade footgun**, not a plugin protocol bug: the framework already allows arbitrary `binary_path`, but docs and examples steer operators into a path that the installer owns and replaces.

## Environment

- CubeSandbox version / commit: observed when upgrading to **v0.7.0-amd64** (from a prior 0.6.x one-click install)
- Host OS: Linux (amd64), bare-metal / VM with one-click `cubetoolbox` layout
- Deployment mode: single-node (CubeMaster + Cubelet under `/usr/local/services/cubetoolbox/`)
- Relevant component: CubeMaster, Cubelet, Volume Plugin registration / packaging
- Custom driver: third-party JuiceFS VolumePlugin (binary type), previously registered and working

## Steps to Reproduce

1. Install CubeSandbox via the one-click / cubetoolbox layout.
2. Follow the same install pattern as the official COS/S3 examples: place a **third-party** VolumePlugin binary (and its credential/config file) under:
   - `/usr/local/services/cubetoolbox/CubeMaster/plugin/<driver>`
   - `/usr/local/services/cubetoolbox/Cubelet/plugin/<driver>`
[...]

PR #1750, opened 2026-09-15
feat(volume-juicefs): add JuiceFS volume plugin

WaylandYang:
Adds a binary VolumePlugin for [JuiceFS](https://github.com/juicedata/juicefs) under `examples/volume/juicefs/`, following the layout of the COS binary plugin.

## Motivation

The S3 / COS plugins mount object storage with s3fs / cosfs, where every file operation becomes an object request. That works for reading and writing output files, but agent workloads usually also run git, SQLite and create many small files in the mounted directory.

JuiceFS keeps data in object storage and metadata in Redis / PostgreSQL / MySQL / TiKV, so the mount behaves like a POSIX file system: directory listing, renames, file locks and git work at close to local speed, and sandboxes on different nodes see each other's writes immediately.

## Tested on AWS

We run CubeSandbox on AWS and tested this plugin there end to end:

- **Environment:** CubeSandbox v0.7.1 at 52e5e3d, one-click install on EC2 `m7i.xlarge` (PVM) in `ap-northeast-1`
- **Storage:** Amazon S3 general purpose bucket in the same region
- **Metadata:** PostgreSQL 16
- **Credentials:** the file system is formatted without access keys, so JuiceFS uses the EC2 instance role; no static keys on the host

Same operations inside a sandbox with the Python SDK:

| Operation | s3 plugin | juicefs | juicefs `--writeback` | sandbox local disk |
|---|---|---|---|---|
| 200 small files | 31.6 s | 6.6 s | 1.2 s | 37 ms |
| list a directory | 441 ms | 39 ms | 47 ms | 31 ms |
| 50 MB write + read | 1.06 s | 603 ms | 374 ms | 236 ms |
| git init + commit | 12.4 s | 1.4 s | 589 ms | 62 ms |
| SQLite write | fails (`disk I/O error`) | 301 ms | 118 ms | 61 ms |
| second sandbox sees a write | — | yes | yes | — |

Running the same 200-file loop directly on the host gives 6.4 s (default) and 0.83 s (`--writeback`). The default mode's cost is the synchronous upload on close, not virtiofs.

## Design

- **One file system, one directory per volume.** `create` makes `volumes/<volume_id>`; `destroy` removes it with `juicefs rmr`.
- **Per-volume mount on the node.** `attach` runs `juicefs mount --subdir volumes/<volume_id>` at `<volume-base-dir>/juicefs-<volume_id>`, so a sandbox only ever sees its own volume. Mounts are reused while `refCount > 0`, and `detach` unmounts with `--flush`.
- **Control-plane mount.** `create` / `destroy` need the file system root, so the plugin keeps one mount on the CubeMaster node. It also runs JuiceFS background jobs (trash cleanup, compaction); per-volume mounts pass `--no-bgjob`.
- **No object storage credentials in the plugin.** They are stored in metadata by `juicefs format`. If you format without keys, the node's cloud identity is used. The plugin config holds only `META_URL` and an optional `META_PASSWORD`, which is passed via env so it stays out of `ps`.
- **`volume_id` is validated** (`^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$`) because it becomes a path component on the node and inside the file system.

- **No changes to CubeMaster, Cubelet or one-click.** Registration is manual, the same as adding any third-party binary plugin to `conf.yaml` / `config.toml`.
- **`--writeback` is not on by default.** It makes small-file writes much faster, but data sits in the node's cache directory until it is uploaded. The README explains the trade-off and shows how to enable it through `MOUNT_OPTS`.
- **No rpc variant** and no link from `docs/guide/volume-plugin.md` yet. Happy to add either in a follow-up if you'd like this in the docs index.

## Related

- #1569: third-party plugins installed under the toolbox tree are removed on upgrade. The README installs this plugin under `/usr/local/lib/cube-volume-plugins/juicefs` instead.

## Tested

- `shellcheck -S warning` on both scripts — clean
- `install-deps.sh --all` in a fresh `ubuntu:24.04` container: downloads JuiceFS 1.4.1 and verifies its checksum; `--check-only` on the host
- Hooks called directly, following README §4: create / attach / write / detach / destroy; an invalid `volume_id` (`../x`) is rejected
- Python SDK on the AWS setup above (`Volume.create(driver="juicefs")`), 29/29 checks passed: read/write, 200 small files, 50 MB file, delete / overwrite / append / rename file and directory, git commit, SQLite, 2000-file directory, two sandboxes sharing one volume, data persists after sandbox destroy, read-only mount rejects writes


kinwin-ustc:
@WaylandYang Some of the comments on cubebot are reasonable; please help improve them. Also, when fetching the JuiceFS from GitHub, could you please fix it to a specific version to avoid incompatibility issues later?

WaylandYang:
@kinwin-ustc Thanks — pushed all of it.

**Pinned the JuiceFS release.** `install-deps.sh` no longer resolves "latest" from the GitHub API; it installs the version the plugin is written and tested against, and `JUICEFS_VERSION` overrides it when you want to move deliberately:

```sh
JUICEFS_DEFAULT_VERSION="1.4.1"
local version="${JUICEFS_VERSION:-$JUICEFS_DEFAULT_VERSION}"
```

The checksum verification against the release `checksums.txt` is unchanged.

**From the bot review, the three findings:**

1. **Volume ids no longer stricter than the platform.** `^[A-Za-z0-9._-]{1,128}$`, with `.` and `..` refused explicitly. `_scratch` and `-dash` now work, matching what CubeAPI and CubeMaster accept; the id is still only ever one path component under `volumes/`, never an argument. Verified:

   ```
   _scratch   -> passes validation (fails later on the missing config, as expected)
   -dash      -> passes validation
   ok.name    -> passes validation
   ..         -> ERROR: invalid volume id: ..
   .          -> ERROR: invalid volume id: .
   a/b        -> ERROR: invalid volume id: a/b
   ```

2. **Locks are bounded.** Both the per-volume lock and the admin-mount lock now take `flock -x -w "$LOCK_WAIT_SECONDS"` (60 s by default, overridable), so a lock left held by a killed invocation fails that attach with a clear error instead of blocking every later attach and detach for the volume forever. Measured: the call returns after the timeout rather than hanging.

3. **Dependency check covers what the scripts use.** `--check-only` now also checks `flock`, `mountpoint`, `curl`, `tar` and `sha256sum`, so it can no longer report success on a node that then fails at the first attach.

**Also took the optional follow-up:** the plugin is now listed in the Reference Implementations table in both `docs/guide/volume-plugin.md` and `docs/zh/guide/volume-plugin.md`, since that table is where people look for plugins.

Both READMEs are updated for the id pattern, the lock timeout and the pinned version. `shellcheck -S warning` is clean and `bash -n` passes on both scripts.

One note on the JuiceFS behaviour the review could not verify: it is all from 1.4.1, which is what the installer now pins, and the plugin has been running on our deployment since — several sandboxes sharing one volume, idle mounts detaching, and the trash behaving as documented.

WaylandYang:
Good catch on the READMEs — both were corrupted by my own edit, and they are rebuilt from a clean copy now (`README.md` 170 lines, `README.zh.md` 158, one H1 each, the ID rule stated once).

The cause is worth naming since it is easy to repeat: the replacement text contained `` `^[A-Za-z0-9._-]{1,128}$` ``, and in a JavaScript string replacement `` $` `` means "everything before the match" — so the whole document was spliced in mid-bullet. Redone with a replacer function.

On the three rules for the same id: they are one rule now.

| Source | Rule |
|---|---|
| `cube-volume-juicefs.sh:104` | `^[A-Za-z0-9._-]{1,128}$`, then `.` and `..` refused |
| `README.md` / `README.zh.md` | the same, stated once |
| CubeAPI / CubeMaster | `^[a-zA-Z0-9_-]+$`, ≤ 128 — a strict subset of the above |

So a volume CubeMaster accepted always attaches here, which is what the earlier first-character class broke. `_data` and `-x` pass now; `.`, `..` and anything with `/` are refused, and the id stays a single quoted path component (`volumes/<id>`, `<base>/juicefs-<id>`, `${LOG_DIR}/<id>.log`), never a bare argv element.

The other two inline points were already in the push you reviewed: both locks take `flock -x -w "$LOCK_WAIT_SECONDS"` (60 s, overridable) so a stuck lock fails the hook instead of hanging Cubelet's create/destroy, and `--check-only` now also checks `flock`, `mountpoint`, `curl`, `tar` and `sha256sum`.

On the control-plane mount not getting `--cache-dir` / `--cache-size`: that one is real and I will fix it in the next push — the root mount should use the configured cache directory like the per-volume mounts do, rather than writing 100 GiB of default cache to `/var/jfsCache` on the system disk.

WaylandYang:
Pushed the cache fix: the control-plane mount now takes `--cache-dir "$CACHE_DIR" --cache-size "$CACHE_SIZE"` like the per-volume mounts, so a deployment that points `CACHE_DIR` at a data disk no longer has the root mount quietly filling `/var/jfsCache` on the system disk with up to the 100 GiB default.

`shellcheck -S warning` and `bash -n` are clean.

Current state of this PR: ids match what the control plane accepts, both locks are bounded, `--check-only` covers the tools the scripts use, the JuiceFS release is pinned with checksum verification, both mounts share the configured cache, the READMEs are rebuilt, and the plugin is listed in the framework doc's reference table in both languages.

kinwin-ustc:
Can you squash all of these commits into one

WaylandYang:
@kinwin-ustc Squashed — this PR is one commit now (`feat(volume-juicefs): add JuiceFS volume plugin`, 9 files, +813). #1751 is squashed to a single commit as well.

Before squashing I fixed the three things the latest review found, one of which was a real bug I had just introduced:

1. **`detach` read unset variables.** The writeback check used `$MOUNT_OPTS` and `$CACHE_DIR`, but `do_detach` never called `load_config`, so under `set -u` the hook aborted instead of logging and falling back. It loads the config now — detach does not need `META_URL`, but it does need to know whether the mount was made with `--writeback` and where its cache is.
2. **The version check was a substring match**, so a node with 1.4.10 satisfied a request for 1.4.1. Now matched on a token boundary, which still accepts the build suffix:

   ```
   juicefs version 1.4.1+2026-07-30.0b90c7d   -> already the pinned version
   juicefs version 1.4.10+2026-01-01.abc      -> reinstalls
   juicefs version 1.3.9+2025                 -> reinstalls
   ```
3. **A duplicated comment block** in `README.md` — removed; the Chinese page had it once already.

Also confirmed against the pinned release that `juicefs umount --flush` exists in 1.4.1 (`--flush   wait for all staging chunks to be flushed`), which is what the detach path and the `--writeback` recommendation rely on.

`shellcheck -S warning` and `bash -n` are clean on both scripts.

ls-ggg:
The DCO check is failing — the commit is missing the Signed-off-by trailer.Please add it

PR #1751, opened 2026-09-15
feat(volume-s3): use the node's cloud identity when no access key is configured

WaylandYang:
## What it does

When both `ACCESS_KEY_ID` and `SECRET_ACCESS_KEY` are empty in `volume-s3.conf`, the S3 volume plugin uses the node's cloud identity instead of failing validation:

- the controller side (`s3api`) uses minio-go's `credentials.NewIAM("")`;
- the node side mounts with s3fs `-oiam_role=auto` and does not write a passwd file.

Setting only one of the two keys is still an error.

## Motivation

On AWS, the recommended way to give a host access to a bucket is an instance role. Today the plugin requires a long-lived access key in a file on every CubeMaster and Cubelet node. With this change, operators can scope bucket access with IAM and keep static keys off the hosts.

## Tested on AWS

We run CubeSandbox on AWS and tested this change there end to end:

- **Environment:** CubeSandbox v0.7.1 at 52e5e3d on EC2 `m7i.xlarge` (PVM) in `ap-northeast-1`
- **Credentials:** an EC2 instance role whose policy is scoped to one S3 bucket; `ACCESS_KEY_ID` / `SECRET_ACCESS_KEY` left empty in `volume-s3.conf` on both CubeMaster and Cubelet
- **Endpoint:** `https://s3.ap-northeast-1.amazonaws.com`
- **Result:** create / attach / detach / destroy through the Python SDK all work. Objects land under `volumes/<id>/` in the bucket. Inside the sandbox there are no credentials in the environment and IMDS is unreachable.
- **Checks passed:** read/write, 200 small files, 50 MB file, delete / overwrite / append / rename, git commit, data persists after sandbox destroy, read-only mount rejects writes

- **Static keys keep working exactly as before.** With keys set, the code path is unchanged.
- **Only the default credential chain is supported.** There are no role ARN or assume-role options; happy to add them in a follow-up if useful.
- **One-click is not touched.** We configured `volume-s3.conf` directly; this PR does not change how one-click handles empty `CUBE_S3_*` keys.

## Tested

- `gofmt -l` — clean
- `go vet ./...` — clean
- `go test ./...` — pass, including the new `TestLoadInstanceRole`, `TestMountArgsInstanceRole` and `TestEnsurePasswdFileSkippedForInstanceRole`
- End to end on AWS as described above


WaylandYang:
Answering the review's open questions, with the checks run rather than reasoned about — and adding the test it noted was missing.

**1. Does a stray `/etc/passwd-s3fs` shadow `iam_role=auto`? No.** Tested with s3fs 1.95 (Debian 13), a world-readable bucket name and no network route to IMDS, so the logs show which path it takes:

- `-o iam_role=auto`, no passwd file → goes straight to `http://169.254.169.254/latest/api/token`, then `GetIAMRoleFromMetaData`.
- `-o iam_role=auto` **with** `/etc/passwd-s3fs` present (`AKIAFAKE:secretfake`, mode 0600) → **identical path**: IMDS token, `GetIAMCredentialsURL`, `GetIAMRoleFromMetaData`. The file is not read.
- No `iam_role`, passwd file present → the normal static path, no IMDS request at all.

So `iam_role` wins over the default passwd files, and a node that happens to have one does not mount with the wrong identity.

**2. Is `credentials.NewIAM` still the right API in minio-go v7.3.0? Yes.** It is present and not deprecated at that tag (`pkg/credentials/iam_aws.go`). `NewChainCredentials` also exists, but a chain is the wrong shape here: it would also pick up `AWS_ACCESS_KEY_ID` from the environment and `~/.aws/credentials`, so a node with stray credentials would silently mount as the wrong identity — exactly what an operator who deliberately left the keys empty does not want. `NewIAM("")` says "use this node's identity" and nothing else.

**3. Coverage: the controller-side choice is now pinned by a test.** The decision moved out of `New` into a small `credentialsFor`, so it can be asserted without a network:

- keys configured → `Get()` returns exactly those keys, and the credentials do not start expired;
- no keys → the provider starts unresolved (it has asked nobody yet), which is what tells an IAM provider from a static one offline.

`gofmt -l`, `go vet ./...` and `go test ./...` are clean on the example module.

@kinwin-ustc when you have a moment — this one is small (the static-key path is unchanged byte for byte) and #1750 depends on the same idea: a node's cloud identity rather than keys in a config file.

WaylandYang:
Second round — all of it addressed.

**The stale credential file is now removed.** This was the best point in the review: a node migrating from static keys to the instance role kept `/etc/cube/.passwd-s3fs-volume-<bucket>` on disk, plaintext and mode 0600, which is exactly the long-lived secret this change exists to get rid of. `EnsurePasswdFile` now removes it in instance-role mode (ignoring "not there"), the doc comment says so, and a test covers both the removal and the second call being a no-op.

**The error message names the field again**, and says what the accepted state means:

```
config: SECRET_ACCESS_KEY is empty (set both keys, or leave both empty to use the EC2 instance role via IMDS)
```

**A silent fallback now leaves a trace.** Since unknown keys are dropped, a misspelled `ACCESS_KEY_ID` used to fail fast and would now select the node's identity instead. `Load` logs when it falls back:

```
volume-s3: no ACCESS_KEY_ID in /path/volume-s3.conf; using the EC2 instance role (IMDS)
```

I kept the XOR rule rather than gating on `USE_INSTANCE_ROLE=true`: the two-key rule is already the contract, and one more knob to get wrong seemed worse than a loud log line. Happy to switch if you prefer fail-closed.

**Doc comment fixed** — `// New builds an S3 client from plugin config.` sits above `New` again, and `credentialsFor` has its own comment, which now also states that this is IMDS specifically, not the AWS credential chain.

**Both READMEs say which deployments can reach this.** A note next to the footnote: only the manual install on that page (one-click `die`s with empty keys and re-renders `volume-s3.conf` on every upgrade; Helm needs `volumeS3.existingSecret`), that empty keys mean the EC2 instance role via IMDS and not the wider chain — so COS, R2 and MinIO still need a static pair — and that the role has to exist on every CubeMaster *and* Cubelet host, scoped to `BUCKET` rather than the account.

`gofmt -l`, `go vet ./...` and `go test ./...` are clean on the module.

ls-ggg:
Thanks for the change — the static-key path stays intact, and the empty-key path is consistent on both sides (minio-go `NewIAM` + s3fs `-oiam_role=auto`). A few things to tighten before merge:

1. **Do not treat empty keys as the only switch.** A missing/overwritten conf, a typo (`ACCESS_KEY` vs `ACCESS_KEY_ID`), or one-click rewriting the file will silently fall through to IMDS. The log line helps, but an explicit setting (e.g. `CREDENTIALS=iam` / `instance_role`) is harder to misfire. Empty keys should be valid only when that flag is on; COS / R2 / MinIO with empty keys should fail at load time instead of hanging on IMDS.

2. **`UseInstanceRole()` should require both fields empty.** It currently checks only `AccessKeyID == ""` and relies on `validate()`’s XOR. Callers that skip `Load` (or future test helpers) could treat a secret-only config as IAM. Use `AccessKeyID == "" && SecretAccessKey == ""`.

3. **Probe IMDS at `Load` / `New` when in instance-role mode.** If CubeMaster is not on EC2, IMDS is off, or the hop limit is 1, the first volume create is what times out today. A startup `Get()` that fails with “instance role requires IMDS” is much easier to debug than an s3fs mount failure.

**CI:** the DCO check is red — the commit is missing a `Signed-off-by: Name <email>` trailer. Please add it (`git commit --amend --signoff` / rebase `--signoff`) and force-push. The rest of the checks look green.

@fslongjin Could you CR the volume-s3 / IAM path when you have a chance?

WaylandYang:
Thanks — all three addressed, and DCO is fixed (both commits signed off).

1. **Explicit opt-in.** New `CREDENTIALS=static|instance_role` (default `static`). Empty keys without `instance_role` now fail at load; so do keys together with `instance_role`, an unknown value, and a non-AWS `ENDPOINT` (COS/R2/MinIO fail at load instead of hanging on IMDS).
2. **`UseInstanceRole()`** now requires the opt-in *and* both keys empty.
3. **IMDS probe.** In instance-role mode, create/destroy/attach fetch credentials once (bounded to 5s — minio-go's default client has no timeout) and fail with "no credentials from the EC2 instance metadata service (is this host on EC2, with IMDS enabled and an instance role attached?)". The fetched value is reused by the client. Covered by tests against a fake IMDS.

Also corrected the README: minio-go's IAM provider does honour `AWS_WEB_IDENTITY_TOKEN_FILE` / `AWS_CONTAINER_CREDENTIALS_*` if set, so I no longer claim those are ignored.

WaylandYang:
Follow-up to the automated review in d958f51: the IMDS probe now runs only when a mount is actually attempted (a repeat attach of a mounted volume no longer touches IMDS), and the stale passwd file is removed on every attach — including when the volume is already mounted — with a failure logged rather than failing the attach.

PR #1756, opened 2026-09-16
perf(s3lvol): LRU nsid reuse, dest cache, and faster restore reads

changpe1:
## Summary
- Reuse the longest-idle nsid instead of the lowest free slot so attach/detach churn does not keep growing namespace IDs.
- Restore reads now go through whole-object GET coalescing, a 256 GET budget, and dest fills that skip bounce copies when a sole waiter can donate its buffer.
- Dest cache is the local copy of immutable S3 objects: partial residency via a per-block bitmap, a 1 GiB RAM hot pool for whole-object hits, and per-chunk overlay bypass so a dirty overlay does not disable the rest of the dest cache.
- Same dest cache also keeps a lower-priority exact-object-key lane so a second import of the same S3 objects hits locally. CopyObject dest reads can alias that source object without copying bytes or evicting native write-cache slots.
- Dense imports (`present_chunks >= ceil(num_chunks/2)`) set host `read_ahead_kb` to 4 MiB; sparse stays at 1 MiB. Userspace dest/export sequential prefetch is gone — kernel readahead already covers a chunk — while CopyObject ingest prefetch remains.
- Close races on that path: refuse unload while decouple is in flight, drain in-flight writes before detaching the esnap parent, and never serve dest cache under a uuid that still has overlay data.

## Test plan
- [x] CubeS3lvol `S3LVOL_SKIP_HOT_UPGRADE=1 ./test/run_all.sh` (41 suites, 2029 assertions, 1 expected-fail; hot-upgrade pair skipped)
- [x] `s3_export_read_test` / `s3_cache_test` cover demand whole-object GET, READY LRU, object-key cache, and CopyObject aliases
- [ ] Confirm attach/detach still reuses nsids as LRU (`s3_active_nsid_test`)
- [ ] Optional: mmap restore probe on a real import after merge (not in `run_all.sh`)

Assisted-by: Cursor:Grok-4.6


CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -613,6 +684,394 @@ s3_chunk_read_should_reread(struct s3_chunk_io *cio)
...
+		}
+	}
+	pthread_mutex_unlock(&ctx->read_fill_lock);
+
+	if (bounce_owner && ctx->owner_thread != NULL &&
+	    ctx->owner_thread != spdk_get_thread()) {
+		rc = spdk_thread_send_msg(ctx->owner_thread,
+					  s3_dest_fill_finish, fill);
+		if (rc != 0) {
+			SPDK_ERRLOG("dest fill owner bounce failed: %s "
```
fslongjin: **[H1 Site A — merge blocker]** `dest_fills_inflight` is incremented in `s3_dest_fill_submit_waiter` (`:1029`) and decremented only at the end of `s3_dest_fill_finish` (`:892-896`). On bounce failure we `return` without running finish, so the counter never returns to zero. Teardown (`:2689-2709`) waits on this counter; if it cannot register the quiesce poller, the code leaks the entire device per its own comment.

This is not a theoretical race — the code itself says `leaking the fill`. The export path in the same PR already shows the correct rule (`s3_export_bs_dev.c:697-704`).

Fix note: the cio path calls `s3_read_apply_overlay()`, which is only legal on the owner thread. On bounce failure, please guarantee `dest_fills_inflight--`, finish all waiters, and `free(fill)`; if a cio cannot return to the owner, complete its waiter with an error instead of running overlay on the GET thread.

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -613,6 +684,394 @@ s3_chunk_read_should_reread(struct s3_chunk_io *cio)
...
+				if (waiter != direct_waiter) {
+					memcpy(bs_io->payload,
+					       (uint8_t *)fill->buf + waiter->offset,
+					       waiter->length);
+				}
+				rc = spdk_thread_send_msg(waiter->origin,
+							  s3_bs_submit_fill_finish,
+							  bs_io);
+				if (rc != 0) {
+					SPDK_ERRLOG("dest fill completion bounce failed: %s "
```
fslongjin: **[H1 Site B]** Here finish is already running, so `dest_fills_inflight` is not stuck. But `bs_io` never completes, so blobstore waits on this I/O forever. The error path in the same function (`:826-836`) already completes inline. Please deliver `s3_bs_submit_fill_finish` inline when the origin bounce fails.

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -1245,6 +1716,190 @@ s3_bs_io_submit(void *arg)
...
+	bs_io->cache_status = status;
+
+	/* A local bdev is allowed to complete during spdk_bdev_read(). Always
+	 * defer so blobstore never receives its callback from inside its own
+	 * bs_dev->read call. The delayed recheck in finish also widens protection
+	 * against a write entering the overlay behind this cache read. */
+	rc = spdk_thread_send_msg(bs_io->submit_thread,
+				  s3_bs_submit_cache_finish, bs_io);
+	if (rc != 0) {
+		SPDK_ERRLOG("Failed to defer submit-thread cache completion: %d "
```
fslongjin: **[H1 Site C]** This sends to **itself** to defer completion (avoiding synchronous completion on the `bs_dev->read` stack); it does not touch `dest_fills_inflight`. On failure a single I/O hangs. Inline completion may still be on the `spdk_bdev_read()` stack, so this needs to align with the same reason `s3_bs_io_complete` (`:600-609`) chose to leak. At minimum, please stop claiming this path can stick the inflight counter.

CubeS3lvol/test/integration/s3_active_nsid_test.c
```diff
@@ -0,0 +1,167 @@
...
+	expect_eq("never-used nsid beats the one just freed", nsid, 2);
+
+	rc = add_at("vol-b", 0, nsid);
+	if (rc != 0) {
+		fail("add vol-b");
+	}
+	nsid = s3lvol_active_alloc_nsid(0);
+	expect_eq("next auto-alloc still prefers a never-used nsid", nsid, 3);
+
+	s3lvol_active_note_nsid(0, 10);
```
fslongjin: **[M1 — false cover]** At this point slots 1 and 2 are occupied, 3 has been touched in-flight, and the lowest never-used slot is 4 by construction. **Removing the `note_nsid(0, 10)` call still passes this assertion.** Please note the *next* slot that would have been picked (e.g. 4) and assert alloc returns 5, so the test fails when the call is removed.

**[M1 — wiring]** The production `want_nsid` branch in `vbdev_s3lvol_rpc.c` (around `nsid = want_nsid`) only does a conflict check + assignment, never touches the generation, and `alloc_nsid`'s `taken[]` only sees entries already in `g_active`. With two attaches racing and `want_nsid` happening to be the free slot the current algorithm would pick, NVMf will contend for the same namespace. Please call `s3lvol_active_note_nsid(subsys, nsid)` right after `nsid = want_nsid`.

fslongjin: **[M1 — false cover]** At this point slots 1 and 2 are occupied, 3 has been touched in-flight, and the lowest never-used slot is 4 by construction. **Removing the `note_nsid(0, 10)` call still passes this assertion.** Please note the *next* slot that would have been picked (e.g. 4) and assert alloc returns 5, so the test fails when the call is removed.

CubeS3lvol/include/s3lvol/s3_cache.h
```diff
@@ -109,23 +124,46 @@
...
-#define S3_CACHE_STAGING_BUFS   4
+#define S3_CACHE_STAGING_BUFS   16
+/* Export-parent population is opportunistic and must not consume the staging
+ * pool needed by native lvstore reads and flushes. */
+#define S3_CACHE_OBJECT_FILLS_MAX 4
+
+/* Whole-object DRAM hot set. RPC callers use DEFAULT when the parameter is
+ * omitted and may pass zero to retain the disk-only behaviour. The limit keeps
+ * a single malformed request from prefaulting an unbounded anonymous mapping. */
+#define S3_CACHE_HOT_BUFS_DEFAULT 1024
```
fslongjin, line 135: **[H2 default]** Default 1024 × 1 MiB with `mmap(..., MAP_POPULATE)` reserves ~1 GiB RSS per lvstore at create/attach time, outside the DPDK `tgtMemMB` budget. `rcow_start.sh` / Helm / terraform do not pass this knob, so old clients omitting the field get the new default.

It is not a format/protocol incompatibility, but it is operator-visible. Please either default to 0, or expose the knob in `rcow_start.sh` / Helm and update the EN+ZH capacity docs (still 18 GiB). Please confirm 16 + 1 + process overhead does not hit the 20 GiB limit.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_xfer.c
```diff
@@ -3512,6 +3605,72 @@ decouple_cleared(void *cb_arg, int bserrno)
...
+	 * transition have been queued by the bdev layer and resume normally. */
+	d->status = bserrno;
+	d->parent_cleared = bserrno == 0;
+	rc = spdk_bdev_unquiesce(d->lvol->bdev, vbdev_s3lvol_get_module(),
+				 decouple_unquiesced, d);
+	if (rc != 0) {
+		SPDK_ERRLOG("Could not unquiesce lvol '%s' after detaching export %s: %s\n",
+			    d->lvol_name, d->uuid_str, spdk_strerror(-rc));
+		decouple_remember_status(d, rc);
+		/* spdk_bdev_unquiesce() removes the public quiesce record before
```
fslongjin, line 3650: **[M2]** The comment assumes SPDK has already removed the public quiesce record on `rc != 0`; this tree has no SPDK implementation to verify that. If the failure happens before the record is removed, host I/O stays queued at the bdev layer while management thinks decouple is done. At minimum, log "bdev still quiesced, restart lvstore" instead of presenting it as a normal decouple failure.

CubeS3lvol/module/bdev/s3lvol/vbdev_s3lvol_active.c
```diff
@@ -139,14 +168,26 @@ s3lvol_active_alloc_nsid(uint32_t subsys)
 		}
```
fslongjin, line 169: **[M3 — file-level, posted here since `s3lvol_active_load` was not modified in this PR]** In `s3lvol_active_load`, on `active_insert` failure, already-inserted entries stay in `g_active` and `g_loaded` remains false. The next RPC re-loads the file: `find` hits the first copy, `remove` only deletes the first copy, ghost entries keep occupying nsids, and `alloc_nsid`'s `taken[]` treats those slots as used. Please roll back inserted entries on failure.

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -613,6 +684,394 @@ s3_chunk_read_should_reread(struct s3_chunk_io *cio)
...
+			   uint64_t chunk_index, const struct spdk_uuid *uuid,
+			   uint32_t valid_bytes, const char *key)
+{
+	struct s3_dest_fill *fill;
+	int rc;
+
+	pthread_mutex_lock(&ctx->read_fill_lock);
+	TAILQ_FOREACH(fill, &ctx->read_fills, link) {
+		if (fill->chunk_index == chunk_index &&
+		    spdk_uuid_compare(&fill->uuid, uuid) == 0) {
```
fslongjin, line 1050: **[L9, downgraded from M4]** The coalescing key is only `chunk_index + uuid`, not `valid_bytes`. The bs_io `memcpy` (`:805-811`) uses `waiter->length` with an `assert`, while the cio path uses `spdk_min`. `s3_bs_try_submit_cache` already requires `offset + length <= valid_bytes` before enqueue, and dest objects are immutable by design, so overflow is not realistic. Please align the `memcpy` with cio using a runtime `min` as a defensive measure.

CubeS3lvol/lib/s3bsdev/s3_bs_dev.c
```diff
@@ -613,6 +684,394 @@ s3_chunk_read_should_reread(struct s3_chunk_io *cio)
...
+static int
+s3_dest_fill_submit_waiter(struct s3_ctx *ctx,
+			   struct s3_dest_fill_waiter *waiter,
+			   uint64_t chunk_index, const struct spdk_uuid *uuid,
+			   uint32_t valid_bytes, const char *key)
+{
+	struct s3_dest_fill *fill;
+	int rc;
+
+	pthread_mutex_lock(&ctx->read_fill_lock);
```
fslongjin, line 1043: **[L10]** The teardown comment (`:2684-2686`) says dest fill re-checks `destroying` under `read_fill_lock`, but `s3_dest_fill_submit_waiter`'s join/create does not actually do this. The blobstore contract is that all `read()` completes before destroy, so this looks like a stale comment rather than a stable UAF. Please reject join/create under `destroying` to match the comment.

fslongjin (changes requested):
The direction is right: the NSID LRU itself is clean (only a static generational array, no heap nodes, so there is no "caller holds a pointer to an evicted entry" UAF), and the dest cache layering plus per-chunk overlay bypass are sound. Main-path tests (NSID LRU, dest cache hit/miss, whole-GET coalescing, 256 budget, overlay dirty-chunk isolation) are wired into `run_all.sh` and are good quality.

**But this revision is not mergeable yet.** The single hard blocker is the dest fill completion path dropping I/O when `spdk_thread_send_msg` fails:

- `s3_dest_fill_done` only logs `leaking the fill` and returns when the owner bounce fails, so `s3_dest_fill_finish` never runs and `dest_fills_inflight` never decrements → teardown spins; if the quiesce poller cannot be registered, the code leaks the entire device per its own comment.
- The export path in the same PR already shows the correct rule (`s3_export_bs_dev.c:697-704`); the dest success path did not align with it.
- The trigger window (unload / hot-remove / target thread stops accepting messages) is exactly the `io_device` lifetime / delayed unregister / decouple quiesce surface this PR changes.

Fix constraint: **do not** blindly call the full `s3_dest_fill_finish` on the GET thread — the cio path calls `s3_read_apply_overlay()`, and overlay apply is only legal on the owner thread. On bounce failure you must guarantee `dest_fills_inflight--`, finish all waiters, and `free(fill)`; if a cio cannot be bounced back to the owner, complete its waiter with an error rather than running overlay on the GET thread. A component test that mocks `send_msg` failure would be welcome.

Several items should be addressed in the same window (none of them alone blocks merge):

1. **`s3lvol_active_note_nsid` is not wired into the production path** (file-level comment on `vbdev_s3lvol_rpc.c`). The RPC `want_nsid` branch only does `nsid = want_nsid` and never touches the generation; `alloc_nsid`'s `taken[]` only sees entries already in `g_active`. With two attaches racing and `want_nsid` happening to be the free slot the current algorithm would pick, NVMf will contend for the same namespace. The existing test asserts `nsid == 4` after `note_nsid(0, 10)` — removing that call still passes, so it is a false cover. Please call `note_nsid` right after `nsid = want_nsid`, and have the test note the *next* slot that would have been picked (e.g. 4) so it fails when the call is removed.
2. **Default `cache_hot_bufs=1024` makes create/attach `MAP_POPULATE` ~1 GiB RSS per lvstore**, and `rcow_start.sh` / Helm / terraform do not pass this knob. It is not a format/protocol incompatibility, but it is operator-visible. Please either default to 0, or expose the knob in `rcow_start.sh` / Helm and update the EN+ZH capacity docs (they still say 18 GiB).
3. **`spdk_bdev_unquiesce` submission failure clears `bdev_quiesced=false` and treats decouple as finished.** The comment assumes SPDK has already removed the public quiesce record on `rc != 0`; this tree has no SPDK implementation to verify that. If that is not always true, host I/O stays queued at the bdev layer while management thinks decouple is done. At minimum, log "bdev still quiesced, restart lvstore" instead of presenting it as a normal decouple failure.
4. **README is missing user-visible behavior**: `cache_hot_bufs`, NSID LRU, `RCOW_READ_AHEAD_KB`, unload returning `-EBUSY` on in-flight decouple, and the new `rcow_get_lvstores` cache stat fields. `test/README.md` suite counts are also stale.
5. **`s3lvol_active_load` OOM mid-load leaves a half-built table** (file-level comment on `vbdev_s3lvol_active.c`): inserted entries stay in `g_active`, `g_loaded` stays false, the next load re-inserts the file, `find`/`remove` only touch the first copy, and ghost entries occupy nsids. Please roll back inserted entries on failure.

Before merging, please run `run_all.sh` without `S3LVOL_SKIP_HOT_UPGRADE=1` (both hot-upgrade suites) — it is the only end-to-end net for the lifetime changes in this PR, and the author's own test plan skipped it.

The NSID LRU portion could be merged on its own; it is the dest cache bulk that carries the risk.

---
_Review assisted by Cursor (GLM-5.2). Per `AGENTS.md`, no `Signed-off-by` is added by the AI; the human submitter certifies the DCO._

fslongjin (approved):
LGTM

PR #1758, opened 2026-09-16
feat(hypervisor): reclaim guest-free memory with balloon reporting

zyl1121:
## Motivation

Long-running sandboxes can have bursty memory usage. A workload may grow close to the guest memory limit, release that temporary allocation, and then keep running with only a small steady-state footprint.

Before this change, the guest kernel could see the released memory as free again, but the VMM RSS stayed near the peak. That high-water RSS still consumed host memory capacity, reducing effective sandbox density and leaving less headroom for other sandboxes on the same node.

Pausing an unused sandbox can avoid that cost, but it does not help a sandbox that must remain running between bursts. This is common for long-running agent workloads, where the sandbox may stay alive for session state, background work, or future tool calls even after a memory-heavy step has completed.

This PR lets the host reclaim pages that the guest kernel already considers free while the VM remains running. The goal is to improve host memory utilization for long-running sandboxes without requiring pause/resume, reducing guest-visible memory, or forcing the guest to reclaim application-owned pages.

## Approach

Enable virtio-balloon free page reporting so the guest can report pages that its kernel already considers free and the VMM can return those pages to the host.

The balloon target remains zero. This does not resize guest-visible memory, trigger traditional balloon inflation, or force the guest to reclaim application-owned pages.

## What Changed

- Add a zero-target virtio-balloon device with `free_page_reporting=true` to newly created VMs.
- Enable `CONFIG_MEMORY_BALLOON`, `CONFIG_BALLOON_COMPACTION`, `CONFIG_PAGE_REPORTING`, and `CONFIG_VIRTIO_BALLOON` in the regular x86_64 guest kernel config. The aarch64 and PVM guest configs already enable these options.
- Restore reporting queue topology from the feature state saved in each snapshot, preserving legacy snapshots without a balloon.
- Backport Cloud Hypervisor's balloon descriptor hardening from `cfec1307` and `39264eea`, together with the checked-descriptor prerequisite from `4c6d60f`: cap inflate/deflate descriptor length, iterate descriptor chains through validated checked descriptors, and isolate malformed descriptors and PFNs so later queue entries continue.
- Harden reported ranges by clamping them to their guest memory region, handling zero-length ranges, and avoiding descriptor-length overflow.
- Release reported pages according to their memory mapping: discard `MAP_PRIVATE` CoW pages with `MADV_DONTNEED` without modifying their backing file, and punch holes only for file-backed `MAP_SHARED` regions. A shared-backing punch failure is reported but does not prevent the local mapping from being discarded.
- Keep process-wide and strict per-thread seccomp policies compatible with the balloon worker's existing `fallocate` call for shared file-backed memory.
- Fix stacked seccomp behavior for device threads by permitting the same constrained vsock `getsockopt(SOL_SOCKET, SO_PEERCRED)` query in the inherited VMM filter and the vsock filter.
- Add focused balloon and stacked seccomp tests, direct hypervisor integration coverage for cold boot and full snapshot restore, and a Cube API E2E for create and Cube pause/resume. The E2E rejects mismatched VMM processes and fails unless sandbox and temporary-template cleanup complete.

## Validation

- **Running-VM reclamation:** Before this change, after the guest released a touched mapping of approximately 1.8 GiB, VMM RSS stayed near the 1.9 GiB peak. With free page reporting enabled, VMM RSS returned close to the pre-workload baseline, reclaiming approximately 1.8 GiB while the VM remained running.
- **Snapshot restore:** After full restore, another approximately 1.8 GiB high/low cycle returned VMM RSS close to baseline without a lifecycle transition, confirming that reporting continued through the restored queue topology and private-mapping discard path.
- **Cube integration:** A legacy template was redone under upgraded components and passed Cube API create, a complete 1.5 GiB high/low cycle, Cube pause-to-snapshot/restore, and another complete high/low cycle in the restored VMM. Sandbox process, task-state, and API cleanup completed. The run reclaimed more than 1.2 GiB in each phase; the runner waits for the guest's ready marker before sampling the peak.
- **Stability:** An eight-hour soak completed 474 high/low cycles while interleaving 47 raw pause/resume operations, 23 full snapshot/restore cycles, and 11 two-clone concurrent restores. No reclamation failure, snapshot corruption, or upward drift in released RSS, VMM file descriptors, or worker threads was observed.
- **Compatibility:** Read-only snapshot integrity, incremental snapshot/rollback, successive soft-dirty snapshots, concurrent clones, and restoration of a legacy no-balloon snapshot passed. Memory still held by a guest process remained resident until the process exited.
- **Balloon queue and backing safety:** Regression tests reject oversized and malformed descriptor chains and invalid PFNs without terminating the balloon worker, verify that later valid queue entries still complete, and confirm that reclaiming a writable `MAP_PRIVATE` mapping does not modify its backing file.
- **Seccomp:** Strict trap mode passed cold boot, running-VM reclaim, full snapshot/restore with a second reclaim cycle, and two concurrent clone restores. Regression tests cover the vsock filter alone and stacked after the inherited VMM filter; the latter reproduces the original `SIGSYS` without the fix.
- **Automated and build coverage:** Focused balloon and seccomp tests, the complete `virtio-devices` unit suite, CubeShim workspace tests, and hypervisor integration checks passed. Hypervisor/CubeShim release builds passed, and the regular x86_64 and aarch64 guest kernel builds contained the required reporting options.

Default-policy runtime logs contained no reported-range release failure, balloon worker panic, guest kernel panic, Oops, or lockup.

## Maintainer Validation Required

**Runtime functionality has been verified on x86_64 PVM only. I do not currently have access to a standard x86_64 KVM bare-metal host or an Arm KVM host, so those runtime paths still need functional validation before merge.**

The standard x86_64 KVM and Arm KVM paths have build/config coverage, but build coverage does not prove runtime free page reporting behavior.

Please run the functional smoke test on:

- a standard x86_64 KVM bare-metal host using the rebuilt regular x86_64 guest kernel
- an Arm KVM host using the aarch64 guest kernel

For each path, confirm that the guest negotiates free page reporting, host VMM RSS eventually falls after the guest returns a touched mapping to its kernel, and reporting still works after full snapshot/restore under the default runtime security mode.

## Scope / Compatibility

This affects new cold boots and snapshots or templates rebuilt from them. Restoring an existing snapshot preserves its saved device topology, so legacy snapshots do not gain a balloon implicitly. The new VMM restores legacy no-balloon snapshots, while snapshots containing reporting topology must continue to use compatible new components through the existing multi-version component inventory.

Free page reporting does not reclaim memory still held by an application, resize guest-visible memory, or introduce a forced balloon target. This change does not modify CubeMaster APIs, annotations, component pinning, or the control-plane data model.


up2wing:
Thanks for the work! However, some of the code is from cloud hypervisor upstream, it's better to put them into standalone commits rather than being folded into the feature commit, also with upstream attribution in the commit message.

There is an example here:
https://github.com/TencentCloud/CubeSandbox/commit/29c20149ff4651a09d6bcd8ef0c2e6e211f8c305

zyl1121:
> Thanks for the work! However, some of the code is from cloud hypervisor upstream, it's better to put them into standalone commits rather than being folded into the feature commit, also with upstream attribution in the commit message.
> 
> There is an example here: [29c2014](https://github.com/TencentCloud/CubeSandbox/commit/29c20149ff4651a09d6bcd8ef0c2e6e211f8c305)

Thanks for the suggestion. I'll split these into standalone commits with proper upstream attribution.

up2wing:
Great. 

Once this PR is merged, the balloon feature will be enabled by default, so security is a critical factor. We noticed that the upstream community introduced these two commits cfec1307 / 39264eea to harden balloon security; could you backport them?

zyl1121:
> Great.
> 
> Once this PR is merged, the balloon feature will be enabled by default, so security is a critical factor. We noticed that the upstream community introduced these two commits cfec1307 / 39264eea to harden balloon security; could you backport them?

There are still a few test issues I'm working through. once those are fixed, I'll finish the upstream backports and attribution.

zyl1121:
@up2wing I have rebased against latest master, updated the PR description, and split the changes into 4 standalone commits following the upstream attribution conventions:

1. `virtio-devices: balloon: Clamp range to region before PUNCH_HOLE` (backported from upstream `326cd8807448`, authored by Dylan Reid)
2. `virtio-devices: balloon: Handle guest queue errors gracefully` (backported upstream queue/descriptor hardening from `c1b5b18`, `ce74caa`, `4c6d60f`, `cfec130`, and `39264ee`)
3. `feat(hypervisor): support balloon reporting across snapshot restore` (CubeSandbox adaptation: respect mapping semantics across snapshot restore, and fix stacked seccomp)
4. `feat(cube): enable virtio-balloon free page reporting` (kernel config, CubeShim enablement, integration tests, and E2E validation)

Along with the split, I also refactored the release path in Commit 3 to follow memory mapping semantics directly: `MAP_PRIVATE` mappings (such as snapshot bases and CoW backings) skip hole punching entirely and discard local pages via `MADV_DONTNEED`, completely removing the previous `EBADF`/`fcntl` workaround.

The full details and test validation have been updated in the PR description. All unit tests and E2E suites have passed.

up2wing (approved):
lgtm

configs/kernel-oc9.x86_64.config
lkml-likexu, line 1: No plan for arm64 or pvm  guest ?

lkml-likexu (dismissed):

configs/kernel-oc9.x86_64.config
zyl1121: > No plan for arm64 or pvm guest ?

They are both supported and planned.

It looks like there was some config drift across the guest kernels in the repo: `configs/kernel-oc9.aarch64.config` and `deploy/pvm/configs/pvm_guest` already have the balloon reporting options enabled, while regular x86_64 was left behind. So in this PR I only needed to update `kernel-oc9.x86_64.config` to align them.

For runtime verification, all my current tests were run on **x86_64 PVM**. Since I don't have access to Arm KVM or bare-metal standard x86 KVM hardware, could you help run a quick smoke test on those two environments?

lkml-likexu: You may update deploy/pvm/configs/pvm_guest for pvm-guest.

zyl1121: > You may update deploy/pvm/configs/pvm_guest for pvm-guest.

I checked `deploy/pvm/configs/pvm_guest` earlier, and it already has all the required balloon options enabled:

- `CONFIG_MEMORY_BALLOON=y` (line 1016)
- `CONFIG_BALLOON_COMPACTION=y` (line 1017)
- `CONFIG_PAGE_REPORTING=y` (line 1020)
- `CONFIG_VIRTIO_BALLOON=y` (line 3574)

(Similarly, `configs/kernel-oc9.aarch64.config` already has them enabled as well.)

So only `kernel-oc9.x86_64.config` needed updating in this PR.

PR #1781, opened 2026-09-18
fix(hypervisor): detect PVM through supported MSRs

bysui:
### Problem ###

The physical address bits of a PVM guest are only allowed below 44 bits for 4-level paging guest; see https://github.com/virt-pvm/linux/issues/1 for more information.

Recently, the upstream PVM repository changed the PVM ABI related to the CPUID signature. As a result, the out-of-tree module built from the upstream repository no longer works with CubeSandbox, because it is recognized as a KVM platform instead.

### Problem ###

Detect the PVM PV MSR instead, as it is stable in the PVM ABI (as promised).


lkml-likexu:
@bysui Cool. On your pvm-host, could you run cubesandbox using this https://github.com/TencentCloud/CubeSandbox/releases/download/kernel-release-260812-1/vmlinux-pvm-amd64 ?

bysui:
> @bysui Cool. On your pvm-host, could you run cubesandbox using this https://github.com/TencentCloud/CubeSandbox/releases/download/kernel-release-260812-1/vmlinux-pvm-amd64 ?

@lkml-likexu, the provided guest kernel cannot boot with our OOT module because it does not include our event delivery ABI changes. As a result, your guest kernel is not compatible with our OOT module. I also could not find the event ABI changes in the [Tencent Cube PVM branch](https://gitee.com/OpenCloudOS/OpenCloudOS-Kernel/commits/linux-6.6/released/cube-pvm).

However, this change will not break the version you maintain, because the PVM virtual MSR index is fixed.

PR #1786, opened 2026-09-20
sdk/python: remove e2b command fallback and add cubesandbox-only user/path e2e coverage

wbzdssm:
Background:
The Python SDK still kept an e2b connect fallback in the commands path, which conflicts with the goal of fully removing e2b dependencies. sdk_compat also lacked direct e2e coverage for user/cwd/path arguments.

Changes:
- Remove Commands._run_with_e2b_connect and its helpers; commands.run now always uses envd Connect RPC.
- Clean up unit tests that patched _run_with_e2b_connect.
- Add sdk_compat e2e cases for commands.run cwd/env/user and filesystem extended APIs with the user argument.
- Add non-root (nobody) + /tmp path scenarios to validate "other user + specified path" behavior.

Verification:
- Unit tests: pytest sdk/python/tests/test_sandbox.py -> 19 passed.
- e2e (cubesandbox backend):
  pytest --run-e2e --sdk-e2e-backends=cubesandbox cases/commands cases/filesystem
  -> 11 passed in 2.96s.
- Dual-backend attempt (SDK_E2E_BACKENDS=e2b,cubesandbox): the e2b backend preflight stopped the session because the e2b SDK is not installed in that environment ("e2b backend requires e2b-code-interpreter or e2b"). After installing it, rerun:
  pip install e2b-code-interpreter
  export E2B_API_KEY=<key>
  pytest --run-e2e -m "smoke or p0" cases/commands cases/filesystem

Notes:
- Raw assertions run only on the cubesandbox backend; e2b skips by design.
- Non-root cases skip automatically when the target user does not exist in the template.
- Earlier python3 failures came from a template image without python3 and are unrelated to this change; with a python3-capable template, the commands cases now pass (11 passed).


sdk/python/cubesandbox/_commands.py
```diff
@@ -61,62 +58,13 @@ def run(
...
-            events = rpc.start(
-                request,
-                headers=_user_headers(user),
-                timeout=timeout,
-                request_timeout=self._sandbox._config.request_timeout,
-            )
-            return _collect_process_events(events)
-        finally:
-            pool.close()
+        return self._run_with_connect_fallback(
```
luzhixing12345: rename the function? this is not fallback

luzhixing12345:
LGTM, please squash into one commit

PR #1798, opened 2026-09-21
docs: agent platform freeze/resume integration guide

AkatQuas:
## Summary

- Add English and Chinese guides for agent orchestrators integrating manual `sandbox_freeze` / `connect_by_id` on CubeSandbox.
- Cover the three independent S3 layers (Volume, s3lvol, template backend), the fact that `pause()` does not stop idle kill timers, and the recommended control-plane flow (`pause` → `timeout` → later `connect` before envd).
- Link the new pages from lifecycle docs.

## Test plan

- [x] Markdown renders in docs site preview
- [x] Links from `lifecycle.md` resolve to `agent-platform-freeze.md` (en + zh)


ls-ggg:
Two corrections:

1. The table row for template `--backend s3` reads as if pause/snapshot only exists for that template family. XFS templates pause too. S3 is what cross-node pause/snapshot needs.
2. The Chinese guide has corrupted text: `必���先经`. Please fix that sentence.

PR #1799, opened 2026-09-21
feat(api): optional retention timeout on pause

AkatQuas:
## Summary

- Add optional `PauseSandbox { timeout?: i32 }` body to `POST /sandboxes/{id}/pause`.
- After pause succeeds, call `set_timeout` when `timeout` is provided so agent freeze flows can extend paused retention (e.g. 24h) in one round-trip.
- Empty/missing request body remains backward compatible with existing clients.

## Test plan

- [x] `cargo test pause_sandbox_applies` passes
- [x] `cargo test resume_and_connect_accept` passes (PauseSandbox validation)
- [ ] Manual: `POST /sandboxes/{id}/pause` with `{"timeout": 86400}` then verify `endAt` extended via `GET /sandboxes/{id}`


ls-ggg:
There is already a dedicated timeout API, `POST /sandboxes/{id}/timeout`. Please keep the pause API to pausing only.

PR #1804, opened 2026-09-21
fix(shim): make the agent ttrpc call timeout configurable (was a hard…

oiovoyo:
…coded 3s)

Creating a sandbox larger than ~768MiB fails reproducibly with

    Create sandbox failed:reset guest time failed:ttrpc err: Receive packet timeout

The SandBox context used for every agent RPC (set_guest_date_time, reseed_random_dev, stats_container, wait_notify) carried a hardcoded 3s timeout. A bigger MicroVM restoring from a snapshot does not always answer set_guest_date_time within that window, so creation aborts; 512MiB and 768MiB sandboxes stay under the limit and start fine.

Raise the default to 15s and allow CUBE_AGENT_CTX_TIMEOUT_SECS to override it (values outside 1..=600 fall back to the default), so operators can tune it without a rebuild.


oiovoyo:
Thanks — all three points were correct, including the factual error in my doc comment.

**`wait_notify` is not an agent RPC.** Confirmed: `CubeShim/protoc/src/agent_ttrpc.rs` has
no such method; the only one is `CubeHypervisor::wait_notify`. It reads `self.ctx.timeout_nano`
in two places (`sb.rs:994` and `sb.rs:1428`), so the original patch silently changed the
pause-to-snapshot wait as well. That was not intended and the doc comment was wrong.

**Reworked as the narrower fix you suggested.** `SandBox::ctx` is back to 3s; the widened
deadline now applies only inside `reset_guest`, following the local-override pattern
`create_sandbox` already uses at `sb.rs:516-518`:

```rust
let mut ctx = self.ctx.clone();
ctx.timeout_nano = (secs as i64) * 1_000_000_000;
```

stats_container and the per-container control RPCs that share the client keep their 3s
bound, and both wait_notify waits are untouched.

Silent fallback fixed. Parsing moved to parse_agent_ctx_timeout_secs(Option<&str>) -> CResult<u64>,
which returns the offending text; reset_guest logs warnf! when a value is ignored. The
doc now states explicitly that 0 is not a spelling for "unlimited".

Tests added in the existing mod tests, covering unset / in-range / boundary / zero /
out-of-range / non-numeric:

test sandbox::sb::tests::agent_ctx_timeout_defaults_when_unset ... ok
test sandbox::sb::tests::agent_ctx_timeout_accepts_in_range_values ... ok
test sandbox::sb::tests::agent_ctx_timeout_reports_unusable_values ... ok
test result: ok. 3 passed; 0 failed

cargo build --release --locked is clean.

One thing worth flagging: this changes the default behaviour of the original patch, not of
upstream. Anyone who had already deployed the first version was getting a 15s bound on
stats_container too; after this rework they get 3s there again, which is the pre-existing
upstream behaviour.

oiovoyo:
shim: make the agent ttrpc call timeout configurable (was a hardcoded 3s)

---

## Problem

`SandBox::new()` builds the ttrpc `Context` used for **every** agent RPC with a hardcoded
3-second deadline:

```rust
ctx: context::with_timeout(1000 * 1000 * 1000 * 3),
```

That single `Context` is reused by `set_guest_date_time`, `reseed_random_dev`, the snapshot
path, `stats_container` and `wait_notify`. `set_guest_date_time` is the **first** call the
shim makes after a restore, so when the deadline is too tight, sandbox creation fails
outright:

```
Create sandbox failed: reset guest time failed: ttrpc err: Receive packet timeout
```

The SDK client sees the generic form:

```
Create sandbox failed: Recv len invalid:0
```

## Evidence that the deadline — not the VM — is what fires

From `/data/log/CubeVmm/vmm.log` for a failing create:

```
  1ms   API request event: VmRestore(RestoreConfig { source_url: "file:///…/2C1024M/snapshot", … })
  1ms   restore non-shared map, speed up restore by share map memory file
 46ms   Starting vCPUs: desired = 2, allocated = 2
 46ms   vm has been restored
 46ms   vm has been resumed
 46ms   vsock: local-init connection: local_port=1073741826, peer_port=1024
 47ms   vsock unix accept cost 37 us, established 791 us
25049ms ERROR ttrpc-0.5.8/src/asynchronous/client.rs:72 -- write_message got error:
        Os { code: 32, kind: BrokenPipe, … }
```

and the matching shim log:

```
create req start
start vm start
agent is ready                                                  <- 47ms
Create sandbox failed: reset guest time failed: ttrpc err: …    <- at the deadline
```

The VM restores and resumes in ~47 ms and the vsock channel is established, but the guest
agent does not answer the first RPC within the budget. The error timestamp tracks the
configured deadline exactly — 25 049 ms with the timeout set to 25 s, ~15 066 ms with 15 s —
which is what identifies the deadline as the trigger rather than a stuck VM.

## When it shows up

On our host the failure was **memory-size dependent**, with the same image, the same node
and the same time window:

| template memory | result |
|---|---|
| 256 MiB | OK |
| 512 MiB | OK |
| 768 MiB | OK |
| 1024 MiB | fails, every attempt |

It is not an absolute wall at 1 GiB. After we freed host memory and reaped leaked
`containerd-shim-cube-rs` processes, 1 GiB sandboxes created fine again. The point is that
3 s is a very tight budget for a snapshot restore on a loaded node, and today there is no
way to adjust it without rebuilding the shim.

Each failed create also leaks a `containerd-shim-cube-rs` process (~92 MB RSS) that keeps
holding node quota, so repeated failures degrade the node until those processes are reaped
manually.

## Change

- Extract the deadline into `agent_ctx_timeout_nano()`.
- Read `CUBE_AGENT_CTX_TIMEOUT_SECS`; accept `1..=600`, otherwise fall back to the default.
- Raise the default from 3 s to **15 s**.

Behaviour is unchanged when the variable is unset, other than the larger default. No new
dependencies. The env var name follows the existing `CUBE_*` convention already used in the
shim (`CUBE_BIND_SHARE_TYPE`, `CUBE_SYS_PATH`, …).

## A note for operators choosing a value

`POST /sandboxes` is served on CubeAPI's default 30 s route budget
(`CubeAPI/src/routes.rs:26`, `DEFAULT_ROUTE_TIMEOUT`). Setting
`CUBE_AGENT_CTX_TIMEOUT_SECS` above ~25 s is therefore not useful for the create path: the
API returns `408` before the shim's own deadline expires and the underlying error is lost.
We run 25 s.

## Testing

- `cargo build` for the shim; deployed the resulting binary to a single-node cubelet.
- 1 GiB template (2 vCPU / 1024 MiB): 3/3 successful creates at 0.24–0.48 s after the
  change, versus 0/3 before.
- 256 MiB template unaffected: 3/3 at 0.18–0.19 s.
- Full functional suite (lifecycle, commands, files, artifact generation, network policy,
  TTL, concurrency): 36/36.
- Sustained load: 120 rounds of create → run → file I/O → kill at concurrency 8, 120/120
  with no failures.

## Open question for maintainers

Would you prefer this as a field in `conf.yaml` (alongside `common.sandbox_exec_cmd_time_out`)
rather than an environment variable? An env var was the smallest change that could be rolled
out without touching the config plumbing, but a config field would be more consistent with
the rest of the shim's tunables. Happy to rework.

ls-ggg:
Widening the shim timeout is fine.

The better approach is to take the timeout from the RPC context or the request that Cubelet passes down. Configuring it through an environment variable is a bit odd — please avoid that.

charliecgxu:
I’m curious why reset_guest’s runtime scales with VM memory size. set_guest_date_time and reseed_random_dev appear to be constant-cost operations to me. What actually scales with memory within that execution window?

oiovoyo:
@charliecgxu 
> I'm curious why reset_guest's runtime scales with VM memory size. set_guest_date_time and reseed_random_dev appear to be constant-cost operations to me. What actually scales with memory within that execution window?

You were right to be suspicious. It doesn't scale with memory — I re-ran it with instrumentation and the correlation in my description is a confound. **What actually scales is the number of orphaned `containerd-shim-cube-rs` processes left behind by earlier failed creates.** Every one of them slows down every subsequent create, regardless of guest size.

Here is the measurement.

## Setup

Single node, Linux 6.6, 16 cores / 12 GiB quota, host ~97% idle (loadavg 0.18) throughout. No root required: `/proc/<pid>/stat` and `/proc/vmstat` are world-readable, and each shim is locatable by its `-id <sandbox_id>` argument.

Two templates from the **same image**, differing only in configured RAM:

| alias | spec |
|---|---|
| `harness-py-256` | cpu=2000m, **mem=256Mi** |
| `harness-py` | cpu=2000m, **mem=1024Mi** |

## Result: the variable is orphan count, not guest memory

Same test, run at three different orphan counts on the node:

| orphans | 256Mi median | 1024Mi median | 256Mi range | 1024Mi range |
|---|---|---|---|---|
| **0** | **180 ms** | **175 ms** | 155–478 ms | 165–190 ms |
| **3** | 2457 ms | 2429 ms | 1169–4748 ms | 645–6212 ms |
| **6** | 6167 ms | 5188 ms | 2196–19716 ms | 162–22533 ms |

Degradation factor relative to a clean node:

```
256Mi    0->3 orphans  13.6x     0->6 orphans  34.2x
1024Mi   0->3 orphans  13.9x     0->6 orphans  29.6x
```

The two memory sizes degrade by the **same factor**. A 1 GiB guest on a clean node creates in 175 ms; a 256 MiB guest on a node with six orphans takes six seconds. Guest RAM is not the variable — orphan count is, and it is monotonic.

## Nothing memory-proportional happens in that window

Per-create fault profile (5 rounds each, clean node):

```
template          spec      latency   minflt  majflt  RSS_MB  cpu_s
harness-py-256    256Mi       478      2816       3      64   0.02
harness-py-256    256Mi       167      2746       0      62   0.02
harness-py-256    256Mi       180      2752       0      61   0.02
harness-py        1024Mi      165      2773       0      55   0.02
harness-py        1024Mi      187      2770       0      54   0.02
harness-py        1024Mi      175      2767       0      55   0.03
```

Minor faults (~2750), RSS (~55–62 MB) and CPU (~0.02 s) are indistinguishable between a 256 MiB and a 1024 MiB guest, and major faults are ~0. **A guest configured with 1 GiB leaves the shim resident at ~55 MB** — the restore never touches pages in proportion to configured RAM.

That matches the code: on the fast-restore path (`memory_manager.rs:1342`) guest RAM is an mmap of the snapshot file and `fill_saved_regions` is skipped, and `prefault` defaults to `false` (`update_ext.rs:67`), so no `MAP_POPULATE`. Pages arrive on demand, and only the working set is ever touched — and the working set is a function of what the guest runs, not of how much RAM it was given. So, as you said: `set_guest_date_time` and `reseed_random_dev` are constant-cost, and nothing else in that window is memory-proportional either.

Even under the degraded condition the shim is not *doing* anything. Timestamping when the shim process appears relative to the API call:

```
template          total     shim appears at   remaining after that
harness-py-256    5330ms         227 ms              5104ms
harness-py-256    9699ms          80 ms              9618ms
harness-py       26717ms          53 ms             26664ms
```

The shim is up within 53–227 ms; the remaining 5–26 s is spent with **zero major faults, 0.02–0.10 s of CPU, and the host ~97% idle**. It is a pure block waiting for the guest agent to answer.

## Where the orphans come from

One template in my environment (`base` — same artifact and same `cpu=2000m,mem=256Mi` spec as a template that works) fails deterministically, 3/3, with the API returning 408 at exactly 30.00 s. Each failure leaves a shim behind:

```
API /sandboxes                 0 sandboxes
host containerd-shim-cube-rs   6 processes, ~60 MB RSS and 24 threads each
```

They survive `SIGTERM` and needed `SIGKILL`. The control plane has no idea they exist. Trace of one failing create:

```
  t       RSS     minflt   majflt   cpu_ticks
  0.60s     7MB      333        0        0
  4.26s    49MB     2541        0        2
 11.62s    51MB     2856        0        4
 29.99s    53MB     3109        0        6     <- API already returned 408
 41.02s    53MB     3242        0        8     <- shim still alive
```

Which closes the loop: a failed create leaks a shim → each orphan adds seconds to every subsequent create → creates start exceeding the 3 s agent deadline → more failures → more orphans. That is self-reinforcing, and it is what I was actually looking at when I wrote the memory table. My own PR text already contained the giveaway — "after we freed host memory **and reaped leaked `containerd-shim-cube-rs` processes**, 1 GiB sandboxes created fine again" — I just attributed the recovery to the wrong half of that sentence.

## What I'm changing in the PR

1. Dropping the "template memory | result" table. It is a confound.
2. Rewriting the justification around the measurement above: the first agent RPC after a restore needs 5–26 s once a node has accumulated orphans, so a 3 s constant fails it. The knob is worth having, but as a mitigation for a degraded node, not as a fix for anything memory-related.
3. Saying plainly that the leak is the more serious bug. Raising the deadline only slows the feedback loop down; it does not stop orphans accumulating.

## Open

I could not determine *why* the agent takes seconds to answer when orphans are present — nothing on the host is saturated (CPU, memory, disk all idle), so my guess would be contention on a shared resource in the vsock/CubeVS path rather than anything scheduler-related, but I have not verified that and would rather not speculate in the PR. I also have not found why `base` fails deterministically when a template with an identical artifact and spec succeeds. If you want me to chase either one I'm happy to, though the leak looks like it deserves its own issue rather than riding on this PR.

oiovoyo:
@ls-ggg 

> Widening the shim timeout is fine. The better approach is to take the timeout from the RPC context or the request that Cubelet passes down. Configuring it through an environment variable is a bit odd — please avoid that.

Understood — I'll drop the environment variable. It was the smallest change that could be rolled out without touching the config plumbing, but "smallest to deploy" is not a good reason to add a tunable that lives outside the normal path, and a caller-supplied deadline is clearly the right shape for this.

Both routes you name already exist in the tree, so before I rewrite it I'd like to know which one you consider the intended channel.

**a) From the RPC context.** `TaskService::create` already receives the context and currently ignores it:

```rust
async fn create(
    &self,
    _ctx: &TtrpcContext,        // task_srv.rs:325
    req: api::CreateTaskRequest,
) -> TtrpcResult<api::CreateTaskResponse> {
```

Deriving the agent deadline from the caller's remaining budget is the most self-consistent option — the shim would stop outliving the deadline its caller is working to. What I can't confirm from this workspace is whether Cubelet's ttrpc client actually sets a deadline on create today, or whether that would need a change on the Cubelet side first. If you know it's already populated, this is my preference.

**b) From the request Cubelet passes down.** There's an established pattern for exactly this in `update_ext.rs` — Cubelet sends JSON config through `cube.shimapi.update.*` annotations, and `RollbackRestoreConfig` already carries knobs of this kind (`prefault`, `dirty_log`, `memory_vol_url`). A create-side equivalent would sit naturally alongside them and needs no ttrpc semantics.

One thing that argues for plumbing a deadline rather than just widening one constant: the shim currently has five independent hardcoded deadlines for agent/hypervisor calls —

```
container/mod.rs:245   10s   container ctx
sandbox/sb.rs:138       3s   sandbox ctx  (the one this PR is about)
sandbox/sb.rs:570      25s   create
sandbox/sb.rs:623       5s
sandbox/sb.rs:758     200ms  destroy
```

A caller-supplied budget would give all of these a single origin instead of five, though I'd keep this PR scoped to the create path unless you want the wider cleanup.

Separately, and it affects how this PR should be framed: @charliecgxu asked why `reset_guest` would scale with guest memory, and on re-measuring, it doesn't. The real variable is the number of orphaned `containerd-shim-cube-rs` processes left by earlier failed creates — median create latency on the same node went 180 ms (0 orphans) → 2457 ms (3) → 6167 ms (6), with a 256 MiB and a 1024 MiB guest degrading by the same factor. Measurements are in my reply to him.

That changes what this patch is: a mitigation that keeps a degraded node usable, not a fix. The underlying bug is that a failed create leaks a shim the control plane never learns about, and each orphan slows every subsequent create until they start breaching the deadline too — which leaks more. I'll say so explicitly in the description rather than leaving the impression that widening the deadline solves the problem. The leak probably deserves its own issue; let me know if you'd prefer I open one.

Let me know which channel you'd prefer for the timeout and I'll rework the patch accordingly.

ls-ggg:
You can take it from the RPC context. The create call already carries a deadline, and the shim receives it as `TtrpcContext.timeout_nano`. `Task.Create` currently ignores that context.

For a quick fix, raise the timeout configured in the shim. Reading it from the RPC context can wait.

PR #1807, opened 2026-09-21
webui: add an audit trail and template-scoped API key management (RFC, backend not yet upstream)

oiovoyo:
---

## Status: draft / RFC — not mergeable as-is

The two pages in this PR call backend endpoints that **do not exist upstream yet**. Merged
as-is, users would get two new nav items that return 404.

I am opening it as a draft to show a working frontend implementation and to ask maintainers
where the backend should live and what the API should look like. If the direction is
agreed, I am willing to implement the backend in CubeAPI / CubeOps and complete this PR.

If you feel these features are out of scope for Cube, please say so directly — I will keep
them in our own deployment and stop taking up review time.

## Motivation

We self-host Cube as the sandbox backend for an internal agent harness. Two gaps showed up
once it was in real use.

**1. No history at all.** Once a sandbox is destroyed, Cube keeps no record that it ever
existed — what commands ran, which files were read or written, who created it. For any
setup that has to audit after the fact (our agents execute model-generated code inside the
sandbox) this is a hard blocker.

**2. API keys are all-or-nothing.** `CUBE_API_KEY` is a binary switch: whoever holds it can
use any template and act on any sandbox. We need to hand out per-team / per-purpose keys and
restrict which templates each one may use — e.g. key A may only start `harness-py-256`,
key B only `sandbox-code`.

## What is in this PR

Frontend only, four commits:

| commit | content |
|---|---|
| `feat(webui): add Audit page backed by the gateway audit log` | `/audit` list page and `/audit/:id` detail page (timeline) |
| `feat(webui): API Keys page (template-scoped SDK keys)` | `/apikeys` page: create / delete / enable / disable keys, pick allowed templates |
| `feat(webui): audit states running/paused/complete/killed with endedBy` | state classification and filtering |
| `fix(webui): keep audit/API-keys tables from overflowing` | table layout fix |

Adds `web/src/pages/{Audit,AuditDetail,ApiKeys}.tsx`, `web/src/api/{audit,keys}.ts` and
English/Chinese locales; two entries in the nav rail.

**Why both features are in one PR:** originally I claimed they were coupled in the code.
That was wrong — `web/src/api/audit.ts` has no imports and neither audit page imports from
`keys.ts`; see the review discussion below. They are kept together only until a maintainer
weighs in on whether either belongs in Cube at all; I will split them on request.

## The backend we currently run (not included here)

We put an OpenResty gateway in front of CubeAPI and CubeProxy which does SDK entry, key
authorization and audit logging. The frontend in this PR was written against it.

**I am aware this shape is not suitable for upstream** — it adds an nginx sidecar and does
not fit Cube's existing Rust/Go component split. The contract is written out below to
discuss what it should look like if CubeAPI / CubeOps provided it instead.

### Audit query

```
GET /_audit/api/sandboxes?days=&q=&template=&key=&state=&page=&size=
    -> { days, files[], events, sandboxes, killed, running,
         templates[], keys[], total, page, size, items[] }

GET /_audit/api/sandboxes/:sandboxId?days=
    -> full timeline for one sandbox
```

List item fields:

```
id, createdAt, killedAt, durationSec, state, endedBy,
template, keyName, client, commands, files, requests, firstCommand, metadata
```

`state` semantics:

- `running` / `paused` — still alive
- `complete` — destroyed after its commands had finished (normal lifecycle)
- `killed` — destroyed **while a command was still executing**

`endedBy` distinguishes `kill` (explicit DELETE) from `ttl` (timeout reclaim).

> Implementation note, from a bug we hit: this classification needs **millisecond**
> timestamps. We first recorded only seconds, and since the typical SDK pattern is "command
> returns, then kill immediately", the two usually land in the same second — so every
> sandbox was classified as `killed`. The predicate has to be "command end time > time the
> kill request *arrived*", where the arrival time is `killTs - killLatency`.

### Key management

```
GET    /_gw/api/templates          list templates (for the picker)
GET    /_gw/api/keys               list keys
POST   /_gw/api/keys               create { name, templates[], note }
PUT    /_gw/api/keys/:key          update { name, templates[], enabled, note }
DELETE /_gw/api/keys/:key          delete
```

Key record: `{ key, name, templates[], admin, enabled, createdAt, note }`.
`templates[]` accepts a mix of aliases and template IDs.

Authorization semantics: a non-admin key may only create sandboxes from templates on its
list, **and** only sees those templates and their sandboxes in `GET /sandboxes` and
`GET /templates`. Creating out of scope returns `403`:

```
template 'base' is not allowed for API key 'harness'
```

The page authenticates with the WebUI's existing session JWT (forwarded to CubeOps
`/api/v1/auth/session` for validation) — no second login mechanism is introduced.

## Frontend wiring

Both pages use relative paths `/auditapi/*` and `/keysapi/*`, proxied by the WebUI nginx:

```nginx
location /auditapi/ { proxy_pass http://<backend>/_audit/api/; }
location /keysapi/  { proxy_pass http://<backend>/_gw/api/; }
```

**Those two proxy entries are deliberately not in this PR** (`deploy/one-click/webui/nginx.conf`
is untouched), because the target backend is undecided. If the endpoints end up on CubeAPI
or CubeOps, the frontend only needs the two base constants at the top of
`web/src/api/audit.ts` and `web/src/api/keys.ts` changed.

## Testing

- `npx tsc --noEmit` — 0 errors.
- English and Chinese locales complete.
- Running against our gateway backend in a live deployment for several days.
- The `complete` / `killed` classification verified with two controlled cases: command runs
  to completion then kill → `complete`; kill during `sleep 30` → `killed`.

## Questions for maintainers

1. **Are these features in scope for Cube**, or do you consider them something each
   deployment should build around it?
2. If in scope, **where should the backend live?** Audit capture is a natural fit for
   CubeAPI (every SDK request already goes through it); key management feels closer to
   CubeOps. Split, or together?
3. **How should audit data be stored?** We currently use per-day JSONL files, which is fine
   for a single node but not for a cluster. Would you prefer MySQL, ClickHouse, or a
   pluggable sink?
4. **Should this be two PRs?** (See "Why both features are in one PR" above.)

## What I can take on

If the direction is agreed, I can implement the backend and complete this PR. I would like
to settle questions 2 and 3 first so the work goes in the right place.


oiovoyo:
Thanks — this is a genuinely useful review. All three defects reproduce, the deployment
concern is real, and you are right that my stated reason for keeping the two features in one
PR does not hold.

## Fixed (new commit on the branch)

**Blocking: the pages landed in the default deployment.** Confirmed — neither
`deploy/one-click/webui/nginx.conf` nor `web/vite.config.ts` proxies `/auditapi/*` or
`/keysapi/*`, and the SPA fallback at `nginx.conf:133-135` turns a missing endpoint into
`200 + index.html`, which both API modules then fail to parse. The two nav items and the
three routes are now behind `VITE_ENABLE_GATEWAY_PAGES=1`, following the existing
`VITE_HIDE_AGENT_RECOVER` precedent (`pages/AgentHub.tsx:50`). A default build is unchanged.

**#1 — template stored by ID could not be revoked.** Confirmed. The chip rendered "on" for
either spelling but `toggle()` always operated on the alias, so clicking an ID-backed chip
*added* the alias instead of removing the ID. Now it toggles whichever spelling is actually
stored:

```ts
const stored = value.includes(name)
  ? name
  : value.includes(tpl.templateID)
    ? tpl.templateID
    : null;
const on = stored !== null;
// onClick={() => toggle(stored ?? name)}
```

**#2 — unguarded `timeline` deref.** Confirmed, and inconsistent with the memo one line
above it (`(data?.timeline ?? [])`). Now `data?.timeline?.length`.

**#3 — `ops('/auth/session')` as a refresh trigger.** Agreed that this leans on an
undocumented side effect. Left as-is for now only because it sits in the half of the diff
whose fate depends on Q1 below; if the API-keys feature survives that discussion I will
replace it with the shared session helper rather than a second token read.

`AGENTS.md`: the `Assisted-by:` trailer was missing — my oversight. It is on the new commit,
and I will add it to the PR description. Title convention noted; I will move to
`feat(webui): …` when the branch is next rewritten so the change is not buried in a force-push
mid-review.

## Retraction

> **4. Two PRs?** Yes, please split. The coupling argument in the description doesn't hold.

You are right and I was wrong. `web/src/api/audit.ts` has no imports at all, and neither
audit page imports from `keys.ts`. The only link is a URL string and one column. I will
correct that paragraph in the description.

On *when* to split: I would rather do it once a maintainer has weighed in on Q1, because the
split is only worth doing if at least one half is wanted. If you would prefer the split
first so the two can be triaged independently, say so and I will do it immediately — it is a
mechanical change, not a rework.

## On the design notes — I think you are right on both, and now is the cheap moment

**Keys are stored recoverably.** This is the one I would most like a maintainer opinion on
before any backend exists. Our current model returns the full secret from `GET /keys` and the
table has a reveal control, which does make any WebUI session a credential dump. Hash-on-write
with show-once-on-create is the better default, and changing it later means invalidating every
issued key. If the feature is accepted, I would implement it that way unless told otherwise.

**`stdin` bodies are recorded.** Confirmed and the wording is misleading: the detail subtitle
says "stdout/stderr contents are not recorded" while `case 'stdin'` renders `item.body`
(`AuditDetail.tsx:243-249`). Whatever a user pastes into a sandbox can contain secrets. I lean
towards dropping `body` from the timeline rather than documenting it — capturing input but not
output is an odd place to land, and the command line itself already gives the audit value. Open
to the opposite view if there is a use case for it.

**Key identity by display name.** Correct — the filter links use `name`, which is not unique.
Should be keyed on `key`, with the name shown as a label.

## On Q2, and a scope point worth keeping

> a gateway only sees SDK-routed traffic, so anything reaching a sandbox directly
> (`/sandbox/<id>/<port>/`) is invisible

This is the most important observation in the review and it is one I had not stated plainly.
A gateway-based audit trail is structurally incomplete: the path-based CubeProxy route bypasses
it entirely. That is an argument for lifecycle capture in CubeMaster/CubeAPI and
command/file capture in envd, as you suggest, rather than for polishing the gateway approach.
If the feature is accepted I would rather build it there than port our gateway upstream.

I will keep the "seen by the gateway" scoping visible in the feature name and subtitle for as
long as the implementation is gateway-based.

## Still open for a human maintainer

Q1 in particular — whether read-only lifecycle history and template-scoped keys belong in Cube
at all, and whether they should be decided as one question or two. Everything above is
mechanical; that one is not, and I would rather not guess at it.

PR #1810, opened 2026-09-21
fix(cubelet): do not reclaim sandbox resources while the shim may still be alive

fslongjin:
## Summary
- Identify the sandbox shim by pid and `/proc` start time so a recycled pid cannot be mistaken for our runtime.
- Destroy and resume-replace wait until that identity is gone, and refuse to release tap, IP, and volume resources when evidence is unresolved (including the spawn window before a pid is recorded).
- Quarantine repeatedly failing cleanups: alert once, keep retrying at a slower interval, and add `cubecli diag tap-holders` / `cubecli diag reclaim` so an operator can kill the live holder without guessing.

This is an alternative approach to the shim-reclaim gap described in #1765 (credit: @dushulin). Same problem, different mechanism: process identity and fail-closed reclaim instead of bundle-root PID scan plus startup orphan reaping.

## Test plan
- [ ] `go test` for Cubelet packages `pkg/utils`, `services/cubebox`, `services/gc`, and `cmd/cubecli/commands/diag` (including `-race` on the GC quarantine tests)
- [ ] Create and destroy a sandbox; confirm the shim exits and the address can be reused
- [ ] Pause and resume; confirm a new shim is not treated as gone because of a stale recorded pid
- [ ] Freeze a shim so destroy cannot complete; confirm cleanup stops after the retry bound, the alert fires once, and resources stay held
- [ ] Kill that holder with `cubecli diag reclaim` using pid and start time from `cubecli diag tap-holders`; confirm the next slow cleanup succeeds and the address returns to the pool


ls-ggg:
The `130459` leak is real: a missing pid was treated as “runtime gone,” so destroy/replace could return the tap/IP while an orphan shim still held the fds. That part needs to change. The rest of this PR is a larger redesign than that bug requires.

**What actually has to change**

1. **Missing / unreadable evidence must fail closed.** Empty `collectSandboxRuntimePIDs` and `WaitProcessGone` on a missing pid must not be success. If we cannot prove the shim has exited, do not run the network/volume release step.

2. **Resume-replace must wait before deleting the PAUSED tombstone.** That row is the last place the old shim identity lives. Fail with `Conflict` (so create failover/rollback can run). Do not use `PreConditionFailed` — that skips failover after network/volume have already been allocated. Wait only; do not SIGKILL from bookkeeping.

3. **When `l.shims.Get` returns nothing, still look at the on-disk bundle** (`shim.pid` / `vmm.pid` under the containerd task dir). That is the window that produced the leak (VMM dead, containerd dropped the task, shim still alive). `ShimSpawned` only protects sandboxes created after this binary. Upgrade-era tombstones and already-detached orphans stay fail-open without a disk fallback (and preferably a guarded startup scan). That is why this is not a full substitute for #1765.

Those three close the incident. Pid+starttime, GC quarantine, and `cubecli diag tap-holders`/`reclaim` are useful follow-ups, not prerequisites.

**If this larger design stays in the PR, these are merge blockers**

- `WaitIdentityGone` must not return on the first `LivenessUnknown`. Retry until the wait context is done, then fail closed. A single `/proc` read error should not burn the cleanup budget and quarantine the sandbox.
- A live pid with no recorded start time should be `unresolved`, not `addBare` (adopting whoever holds the number now and waiting on them).
- Restore the bundle-pid collection test; add coverage for `waitReplacedSandboxGone` and the pre-`NewTask` `ShimSpawned` persist. Do not delete bundle-file coverage unless the disk fallback is in.

Please do not close #1765 as superseded unless the disk fallback (and an explicit story for existing orphans) is in this change or a maintainer accepts ops-only reclaim for the in-fleet leftovers.

PR #1812, opened 2026-09-21
fix(cubelet): log routine requests at info level

### LINKED ISSUE #1809: [Bug Report] Normal RunCubeSandbox requests are logged at error level
## Summary

`Cubelet/services/cubebox/service.go` logs every normal `RunCubeSandboxRequest` at `Error` level when debug logging is disabled.

## Environment

- CubeSandbox: `0.7.0`
- Relevant component: Cubelet

## Steps to Reproduce

1. Start Cubelet with debug logging disabled.
2. Send a valid `RunCubeSandboxRequest`.
3. Check the Cubelet logs.

## Expected Behavior

The sanitized request should be logged at `Info` level because receiving a request does not indicate an error.

## Actual Behavior

The request is logged at `Error` level:

```go
if log.IsDebug() {
    log.G(ctx).Debugf("RunCubeSandboxRequest:%s", utils.InterfaceToString(req))
} else {
    log.G(ctx).Errorf("RunCubeSandboxRequest:%s", safePrint(req))
}
```

This may produce misleading error logs, trigger false alerts, and distort log-based error metrics.

## Suggested Fix

Use `Infof` for the sanitized non-debug log:

```go
log.G(ctx).Infof("RunCubeSandboxRequest:%s", safePrint(req))
```

Similar request logging in `DestroyCubeSandbox` and `CreateImage` may also need to be reviewed.

bernalalexis-try:
Closes #1809

`RunCubeSandbox` (outside debug mode), `DestroyCubeSandbox` and `CreateImage` logged every incoming request at `Error`, so routine calls raised false alerts and counted toward error-level metrics. They are now logged at `Info`. Failures are still reported at `Error` by each handler's deferred check.

Autonomously-by: Claude:claude-opus-5


kinwin-ustc:
I think simply replacing Errorf with Infof would suffice; there's too much unnecessary code added here, including the pointless unit tests.

bernalalexis-try:
Thanks, done: the PR is now just the three `Errorf` → `Infof` changes, with the helpers and tests removed.

One thing worth a separate look, raised by the bot review as well: some of these request logs contain registry credentials in clear text, independent of the level. `CreateImage` prints the whole request, and the spec annotations carry `cube.master.image.username` / `cube.master.image.token`. `safePrint` in `RunCubeSandbox` masks the top-level annotations but not the per-container image annotations, which carry the same keys. `SafePrintImageSpec`, already used by `DestroyImage`, masks them. Happy to open a separate issue or PR for that if you want it handled.

kinwin-ustc:
Yes，credentials should be masked,  welcome to solve it. By the way, add DCO in your commit

fslongjin:
This is not a bug, but rather a design trade-off. In many places in Cubelet, `Errorf` is actually used as "logs that must be output." By default, Cubelet uses `Warn` as the log level. If the logging calls are directly changed to `Infof`, then logs for creating/destroying sandboxes would not be printed, which is not expected and would make troubleshooting difficult. Additionally, if Cubelet's default log level is changed to `Infof`, the log volume (including within dependent components) would increase significantly. Therefore, this issue has a fairly broad scope and cannot be simply resolved by changing the log level of Create/Destroy Sandbox requests to `Infof`. Doing so would make faults and issues hard to locate.

Issue #1785, opened 2026-09-20
[Bug Report] v0.7.0 x86_64 guest ENTRY_TRAMPOLINE panic after template restore; connect succeeds after TaskExit


hanydd:
## Summary

On v0.7.0, an x86_64 guest created from a template snapshot panicked approximately 35 seconds after creation while a Python process was running. The stack shows a fault in `sync_regs+0x7`, repeated faults in `exc_page_fault+0x19`, and ultimately `ENTRY_TRAMPOLINE` stack exhaustion. Cubelet recorded the task's abnormal exit, but subsequent E2B-compatible `GET /sandboxes/<id>` and `POST /sandboxes/<id>/connect` still returned HTTP 200 while CubeProxy continued to return connection-timeout 504s for that sandbox.

This is a production incident report with correlated guest, compute-node, proxy, and client logs. We have not established a deterministic reproducer or the underlying kernel/VMM defect. Identifiers, internal addresses, and document paths below are anonymized consistently.

## Environment

- Deployed CubeSandbox version: **v0.7.0**, using the official `cube-proxy:v0.7.0` and `cube-lifecycle-manager:v0.7.0` images. Cubelet records also identify `v0.7.0`.
- Deployment: a three-node cluster, managed through systemd and Docker; the affected sandbox was on a compute node, with CubeProxy on the control node.
- Guest architecture and kernel: **x86_64, `6.6.1199-0009-03_2.0.1 #1`**, `PREEMPT SMP NOPTI`, not tainted at the first oops.
- Guest hardware banner: `Cube Hypervisor cube-hypervisor, BIOS 0`.
- Template restore path ends in `1C1024M/snapshot`; local XFS-backed template memory volume. The sandbox initially passed an envd command health check and executed commands successfully.
- Host OS/kernel, CPU model, `modinfo kvm`, and actual KVM/PVM mode: **not captured in the available incident bundle**, so we cannot yet attribute this to a particular host backend. The version above is the guest kernel, not the host kernel.
- Relevant components: guest kernel, CubeShim/hypervisor, Cubelet; CubeAPI/control-plane status handling is a separate recovery concern.
- Client: a Java client using the E2B-compatible control API and Connect protocol envd endpoints through `/sandbox/<id>/49983/...`; three application instances. Two concurrent sessions belonging to one user shared this sandbox. No concurrent sandbox pause/resume is present in the incident lifecycle log.
[...]

PR #1814, opened 2026-09-21
fix(cubeapi): stop reporting exited sandboxes as running

zyl1121:
## Motivation

When CubeAPI receives a backend sandbox status such as `Stopped`, it reports the sandbox as usable instead of terminal.

`CubeAPI/src/services/sandboxes.rs` masked the state in two places:

- `sandbox_state_from_status` fell through to `Running` for every status it did not handle explicitly. `CONTAINER_EXITED` is represented as `Stopped`, so `GET /sandboxes/<id>` reported `"state": "running"` for a sandbox whose container had exited.
- `connect_sandbox` only special-cased `Paused` to trigger a resume. Other statuses, including `Stopped`, reached the success path and returned HTTP 200 with connection information.

Clients therefore treated a dead sandbox as healthy: they kept calling `POST /sandboxes/<id>/connect`, received HTTP 200, and only failed once traffic reached the unreachable backend through the proxy.

This addresses the CubeAPI state-exposure part discussed in #1785.

## Reproduction

With an exited sandbox, CubeMaster reports status `2` (`CONTAINER_EXITED`) and the node lists the sandbox as `Exited` (`cubecli cubebox ls -a`; the default view shows only running sandboxes). Against that sandbox:

| Endpoint | Before | After |
| --- | --- | --- |
| `GET /sandboxes/<id>` | HTTP 200 with `"state": "running"` | HTTP 200 with `"state": "unknown"` |
| `POST /sandboxes/<id>/connect` | HTTP 200 with connection info | HTTP 409 with `sandbox <id> is not connectable in state 'stopped'` |

## What Changed

- Add `Unknown` to the public `SandboxState` enum. The published response schema and the generated web types already advertised `running | paused | pausing | unknown`; the Rust enum was missing the variant.
- Stop defaulting unhandled backend statuses to `Running`. `sandbox_state_from_status` now maps `Pausing` to `pausing` (it previously reported `running`) and maps the terminal and unrecognized statuses `Stopped`, `Error`, and `Unknown` to the public `unknown` state. `sandbox_state_from_str` maps only `running`, `paused`, and `pausing`, and treats every other value as `unknown`.
- Reject connect on terminal states. `connect_sandbox` returns HTTP 409 `Conflict` for `Stopped`, naming the backend state in the message; `Paused` still resumes and `Running` is unchanged. (`Error` is guarded defensively, but no CubeMaster status currently maps to it.)
- Accept `pausing` and `unknown` in the `/v2/sandboxes` `state` filter so clients can filter for the newly exposed states. The backend label `stopped`, which is the one the connect 409 quotes, and the raw status text `error` are accepted as aliases for `unknown`, so a client replaying the state named in the connect 409 narrows the list instead of falling through to the legacy behaviour of returning every sandbox. Because the alias resolves to the whole `unknown` bucket, it also matches sandboxes that are still being created.
- Extend the documented 409 on the connect endpoint, in the handler annotation and `openapi.yml`, to cover terminal states as well as lifecycle conflicts.
- Document the resulting vocabulary where clients look for it: the lifecycle guide (`docs/guide/lifecycle.md` and its Chinese mirror) now lists `unknown` and says that a connect 409 is not always retriable, the `/v2/sandboxes` `state` parameter documents its accepted values and the `stopped`/`error` aliases next to its definition, and the Python SDK `Sandbox.connect` docstring records the terminal case.
- Add regression tests for the state mappings, the state filter, the 409 response, and the list path, including the shape CubeMaster sends for a creating sandbox, where the `status` field is omitted rather than sent as `0`.

Terminal states are exposed as `unknown` rather than as a new public `stopped` state: `unknown` is already part of the published schema, and CubeOps has reported created, exited, stopped, and unrecognized values as `unknown` since #1350. Keeping both APIs on one vocabulary avoids two names for the same condition, at the cost that `unknown` is not by itself a terminal signal: a sandbox that is still being created, or one whose state could not be read from its node, also reports `unknown`, and connect still succeeds for those. The 409 from `connect` is the reliable terminal signal.

## Validation

- CubeAPI test suite: 158 passed, 0 failed, including the new regression tests. Re-run on the final revision, which changes only comments, documentation, and the list-path test's payload.
- `cube-api --export-openapi` output for the `/v2/sandboxes` `state` parameter is byte-identical to the committed `openapi.yml`, so the spec entry added here is what the generator produces.
- Release build of the change succeeded.
- Runtime before/after on a single-node deployment: created a sandbox, terminated its runtime shim so the task exited, then replayed that exited sandbox against both binaries. The old binary reported `"state": "running"` and returned HTTP 200 from connect; the rebuilt binary reported `"state": "unknown"` and returned HTTP 409 with `state 'stopped'`. CubeMaster reported the sandbox as `CONTAINER_EXITED` (status `2`) throughout, so the difference came from CubeAPI's mapping rather than from a live backend.
- List and filter behavior on the same deployment: with one running sandbox, `GET /v2/sandboxes?state=stopped` returned that sandbox on the previous build and returned nothing on this one, while `?state=running` and `?state=unknown` filtered as before and an unrecognised value still returned the unfiltered list. After that sandbox's runtime shim was terminated, both `?state=unknown` and `?state=stopped` returned it, and `?state=running` returned nothing.
- Healthy-path behavior was preserved: a fresh sandbox reported `running` and connected with HTTP 200, a paused sandbox resumed to `running` through connect, and a sandbox that exited afterwards reported `unknown` with connect rejected with HTTP 409.
- The runtime and list/filter checks above ran on the revision before the documentation pass; the final revision changes only comments, the two guide pages, the OpenAPI description of the `state` parameter, a docstring, and the payload of one list-path test, and the test suite and the OpenAPI export were re-run on it.

## Scope

This PR changes how CubeAPI maps backend sandbox states, filters exposed states, and rejects connect on terminal states, together with the client-facing documentation of that vocabulary.

Terminal sandboxes keep HTTP 409 rather than a separate `410 Gone`. The 409 introduced here is already documented in the handler annotation and `openapi.yml`, and it leaves the existing use of 409 for transient lifecycle conflicts untouched; the retry guidance in the lifecycle guide is corrected instead, since a connect 409 is not always retriable. A client that needs to tell the two apart has to read the state named in the message, which the new test pins; a structured error code would be a cleaner discriminator, but CubeAPI's error body carries only the HTTP status as its `code`, so that is left as a follow-up.


liciazhu:
LGTM.

One non-blocking nit: the Python SDK's SandboxState enum (sdk/python/cubesandbox/_models.py) still only declares RUNNING/PAUSED, so pausing/unknown surface as raw strings.

zyl1121:
> LGTM.
> 
> One non-blocking nit: the Python SDK's SandboxState enum (sdk/python/cubesandbox/_models.py) still only declares RUNNING/PAUSED, so pausing/unknown surface as raw strings.

Fixed. The Python SDK's `SandboxState` now declares `PAUSING` and `UNKNOWN`, and a test covers all four states.

Issue #1797, opened 2026-09-21
[Bug Report] Node reboot loses local template inventory and leaves stale CubeS3lvol build-rootfs volume


yongqisun:
## Summary

After rebooting a CubeSandbox node, templates that were previously distributed successfully remain `READY` in CubeMaster, but disappear from the node's `LocalTemplates` inventory.

The durable CubeS3lvol objects still exist on disk, but the local template state is not reconstructed correctly. Running `tpl redo` may then fail because a stale `*-build-rootfs` volume already exists.

This issue was reproduced independently on both a compute node and a control node.

## Environment

- CubeSandbox version / commit: `0.7.1`
- Host OS: Linux x86_64
- Kernel: `6.6.1199-0009-03_2.0.1`
- Deployment mode: two-node cluster
- Template type: `cubebox`
- Template backend: `s3`
- Template version: `v2`
- Relevant components:
  - CubeMaster
  - TemplateCenter
  - Cubelet
  - CubeS3lvol
- Persistent storage remains mounted after reboot.
- All internal IP addresses, registry URLs and actual identifiers are redacted below.

## Initial State

An S3-backed template was created and successfully distributed to both nodes:

```text
template_id: tpl-<redacted>
status: READY

NODE_ID    STATUS    PHASE
node-a     READY     READY
node-b     READY     READY
```

Runtime validation also succeeded on both nodes before reboot:

```text
distribution=2/2 ready, 0 failed
TEMPLATE_RUNTIME_VALIDATION=PASS
```

## Steps to Reproduce

1. Create an S3-backed `cubebox` template.
2. Wait until the template reports `READY` on all nodes.
3. Verify that a sandbox can be created from the template on each node.
4. Destroy all test sandboxes.
5. Reboot one node.
6. Wait for Cubelet, CubeS3lvol and the other CubeSandbox services to become active.
7. Verify that the normal platform health check succeeds.
8. Query the CubeMaster template state and each node's `LocalTemplates` inventory.
9. Run `tpl redo` for the affected template.

The issue reproduces on both compute-role and control-role nodes.

## Expected Behavior

After reboot:

1. CubeS3lvol should recover persistent template objects.
2. Cubelet should reconstruct and report the local template inventory.
3. CubeMaster replica status should reflect the actual node state.
4. A previously `READY` template should remain immediately usable.
5. `tpl redo` should be idempotent.
6. Any stale temporary `*-build-rootfs` volume should be reused or cleaned up safely.
[...]

PR #1820, opened 2026-09-22
fix(cubelet): remount local S3 package metadata after restart

ls-ggg:
## Summary

- After Cubelet restarts, heartbeat recovery scans S3 template homes with no template id. `config.json` lives on a metadata disk that Finalize unmounts, so the scan sees empty homes, drops the local records, and the node reports `LocalTemplates=0`.
- When the id is unknown, list package directories under the S3 snapshot and pause-snapshot roots and mount each metadata disk before the existing scan. A known id still mounts only that package. Mount failures are logged and skipped, same as before.
- This is the inventory half of https://github.com/TencentCloud/CubeSandbox/issues/1797. It does not change CLI exit codes, leftover `*-build-rootfs` objects, or Master replica status versus the scheduler inventory.

## Test plan

- [x] `go test ./pkg/controller/runtemplate/ -count=1` with `CGO_ENABLED=0` in the CubeSandbox builder (`TestS3RecoveryIDsListsPackageHomesWhenIDUnknown` and the existing recovery tests)
- [ ] Restart Cubelet on a node that already has a finalized S3 template and confirm the next heartbeat still lists that template


ls-ggg:
## Test results

Tested on a single node that already had one finalized S3 template and one XFS template. No sandbox was running beforehand. The patched Cubelet is the binary under test.

### Empty Cubelet state, previous release

A process restart keeps the state tmpfs, so `LocalTemplates` stayed populated and did not reproduce the bug. Unmounting that tmpfs, which is what a reboot does to Cubelet state, and starting the previous release left only the XFS template in the next healthy heartbeat. The S3 package directory was still on disk, and its metadata `config.json` was not mounted.

### Empty Cubelet state, this patch

Starting this build against the same empty state restored both templates on the next healthy heartbeat. The S3 metadata disk was mounted and `snapshot/config.json` was readable. A non-`tpl-` directory placed under the S3 snapshots root was not mounted and did not show up in `LocalTemplates`.

`go test ./pkg/controller/runtemplate/ -count=1` (`CGO_ENABLED=0`, builder) passed, including that filter.

### Host reboot

Rebooted the machine. Cubelet became active before the S3 volume service. The first heartbeat after both were up, about 40s after the host accepted SSH, reported the node healthy with both the XFS and the S3 template. The S3 metadata config was mounted.

### Sandbox

After that reboot, created a sandbox from the recovered S3 template. It reached `running` in under a second and executed a command in the guest successfully.

### After reboot: CRUD, pause/resume, snapshot

On the same node after that reboot, with the S3 template still `READY`:

| Step | Result |
|---|---|
| Create from the S3 template | `running` in 0.4s |
| List and get | The new sandbox is listed and its detail shows `running` |
| Update files | Write, append, and read-back matched |
| Update idle timeout | `end_at` changed |
| Pause | `paused` in 5.2s; the file was still there |
| Resume | `running` in 0.6s; the file was still there |
| Create snapshot | Listed in 5.5s |
| Boot from that snapshot | Child reached `running` in 1.1s and still had the file |
| Rollback | A later edit was reverted to the snapshot contents (0.7s) |
| Delete the child sandbox | Gone from the list |
| Delete the snapshot | Gone from the snapshot list |
| Delete the original sandbox | Get returned 404 and it was unlisted |
| Create from the S3 template again, run a command, delete | Succeeded (1.3s) |

No sandbox or snapshot was left on the node.

Issue #1759, opened 2026-09-16
[Bug Report] golang api can't get result


ningmingxiao:
## Environment

- CubeSandbox version / commit:
- Host OS and kernel version:
- KVM info (`modinfo kvm`):
- Deployment mode: 
- Relevant component: 

## Steps to Reproduce

1. cat main.go
```
package main

import (
	"context"
	"fmt"

	cubesandbox "github.com/tencentcloud/CubeSandbox/sdk/go"
)

func main() {
	ctx := context.Background()
	client := cubesandbox.NewClient(cubesandbox.NewConfigFromEnv())
	defer client.Close()

	sb, err := client.Create(ctx, cubesandbox.CreateOptions{})
	if err != nil {
		panic(err)
	}
	defer sb.Kill(ctx)
	info, err := sb.GetInfo(ctx)
	if err == nil {
		fmt.Printf("info %v", info)
	}
	exec, err := sb.RunCode(ctx, "x = 41\nx + 1", cubesandbox.RunCodeOptions{})
	if err != nil {
		panic(err)
	}
	fmt.Printf("exec %+v \n",exec)
	fmt.Println("test command end")

	result, err := sb.Commands().Run(ctx, "sleep 10000", cubesandbox.CommandOptions{})
	if err != nil {
		panic(err)
	}
	fmt.Println(result)
}
```
2. go run main.go
[...]

PR #1823, opened 2026-09-22
fix(sdk/go): diagnose misrouted HTML responses on Connect streams

tokove:
## Summary

When Go SDK streaming requests are misrouted due to configuration issues (e.g. missing `CUBE_PROXY_NODE_IP` or traffic forwarded to a Web UI / gateway landing page), the endpoint returns an HTML document instead of Connect binary envelopes. The SDK previously parsed `<!DOC` as an envelope length of `558124867` bytes and failed with an opaque error (`Connect stream message too large: 558124867 bytes`).

This PR adds content-type validation and envelope detection to clearly diagnose misrouted HTML responses with actionable troubleshooting hints, rather than attempting to decode HTML as binary framing.

Addresses #1759

## Root Cause

Connect streaming framing specifies a 5-byte header: 1 byte flags + 4 bytes big-endian length.

When an endpoint responds with HTML (`<!DOCTYPE html>`), the first 5 bytes are `<` (`0x3c`) + `!DOC` (`0x21 0x44 0x4f 0x43`).

The stream reader interpreted `!DOC` as a length of `558124867` bytes, which exceeded `maxConnectEnvelopeSize` (64 MiB), resulting in the cryptic message.

## Changes

1. **`validateConnectResponse`**

   * For non-200 responses, delegates to `apiErrorFromResponse` to preserve full error messages and classifications. When an HTML error page is received, enriches the message with actionable hints guiding users to check `CUBE_PROXY_NODE_IP`.
   * Clears false `ErrSandboxNotFound` classifications on 404 HTML pages (since a portal 404 does not mean the sandbox was not found), while strictly preserving `ErrAuthentication` on 401/403 responses.
   * For 200 responses, immediately rejects `text/*` media types with an `*APIError` (HTTP 200) without blocking on body reads.
   * Preserves maximum proxy tolerance: non-text media types and omitted `Content-Type` headers fall through to envelope parsing.

2. **`readConnectEnvelope`**

   * Detects envelopes exceeding size limits that begin with `<` (`<!DOCTYPE`, `<html`) or plain JSON (`{`, `[`) .
   * Diagnoses misrouted HTML and un-framed JSON payloads with actionable hints instead of the cryptic `558124867 bytes` size error.

3. **Stream Endpoints**

   * Enforces `validateConnectResponse` on `startProcess` (commands, code execution), `openStream` (PTY), and `watchDir`.

4. **Testing**

   * Unit tests covering 200/502/404 HTML responses, classification preservation, plain JSON envelopes, and true tolerance for omitted Content-Type headers.
   * All tests pass with `-race` (`go test -race ./...`).

Assisted-by: Antigravity:Gemini-3.8-Flash


liciazhu:
@tokove  Thanks for digging into #1759 and proposing a fix.

## Blocking

### 1. The stated fix doesn't hold for `text/plain` (`connect.go:126` vs `:139`)

The commit message says it clears false `ErrSandboxNotFound` on 404s. The two branches use *opposite* strictness — non-200 matches only `text/html` (`:126`), while 200 uses the `text/` prefix (`:139`). Measured:

```
404 text/plain => err="404 page not found (HTTP 404)", Is(ErrSandboxNotFound)=true   # still false-classified
404 text/html  => err="received HTML error page (HTTP 404); ...", Is(...)=false      # cleared
```

`404 page not found` is the default body of Go's `http.NotFound`, served as `text/plain`. Any Go-based gateway or web service sitting at the misrouted endpoint lands in this gap: no cleared classification, and no `CUBE_PROXY_NODE_IP` hint either. Suggest collapsing both branches onto a single shared predicate using `strings.HasPrefix(mediaType, "text/")`.

### 2. `Addresses #1759` overstates the scope

#1759 reports two symptoms, and the one in the title (`can't get result`) is untouched by this PR:

- `RunCode` goes to the Jupyter port `/execute` (`sandbox.go:229`), not through Connect — `sandbox.go:16-23` documents `JupyterPort = 49999` vs `EnvdPort = 49983` as separate ports.
- `startProcess` is reachable only via the `processStarter` interface (`commands.go:12`, `:29`), i.e. only from `Commands.Run`. `RunCode` never touches it, so none of the three new `validateConnectResponse` call sites apply to it.
- `stream.go:27-34` silently `return`s on any line that fails to parse, and `RunCode` only errors on `>= 400` (`sandbox.go:242`). So on a 2xx HTML response it returns an empty `Execution` with `err == nil` — exactly the reported symptom.

Could you change this to `Refs #1759` and note that this PR is a diagnostics improvement rather than a functional fix? Otherwise merging will close an issue whose headline symptom still reproduces. Happy to file the `RunCode` follow-up if that helps.

### 3. User-facing behaviour changes with no docs (CONTRIBUTING.md:90)

CONTRIBUTING.md:90 asks for docs when behaviour is user-visible. This diff contains no `.md` changes, but changes at least: `errors.Is(err, ErrSandboxNotFound)` on 404+HTML (true → false), `ErrTemplateNotFound`/`ErrVolumeNotFound` on 404+HTML, and 200+`text/*` (was "too large", now `APIError`). Two concrete gaps:

- `ErrSandboxNotFound` has zero hits across all `.md` in the repo. `sdk/go/README.md:226-233` documents `errors.Is` only for `ErrVolumeInUse`/`ErrVolumeNotFound` — 2 of the 5 exported sentinels. This PR changes the semantics of one of the undocumented three.
- The new message tells users to "check whether `CUBE_PROXY_NODE_IP` is configured", but there's no troubleshooting page for it — searching the docs for `CUBE_PROXY_NODE_IP` only returns tutorial snippets that set the variable, and `too large` has zero hits. A user following the hint finds nothing. `docs/guide/troubleshooting/` (+ the `docs/zh/` mirror, per CONTRIBUTING.md:139) would be the natural home, and a changelog entry under `docs/changelog/` seems expected given #1836.

## Worth reconsidering: the enumeration strategy can't be exhaustive

This is the part I'd most encourage you to revisit, because the repo already documents a sibling failure. `CubeProxy/conf/includes/envd_streaming_host_route.inc:2-4`:

```nginx
# Prevent nginx gzip-compressing Connect stream responses.
# iter_raw() does not decompress Content-Encoding — gzip
# magic bytes misread as 2.17 GiB Connect frame header.
gzip off;
```

gzip's magic is `1f 8b 08 00` → `header[0] = 0x1f`, `header[1:5] = 8b 08 00 00` → `0x8B080000 = 2,332,557,312` = 2.17 GiB, matching that comment exactly. Since `0x1f` is neither `<` nor `{`/`[`, it falls through to the cryptic "too large" message this PR is trying to remove. Same for a UTF-8 BOM before `<!DOCTYPE`, and for leading whitespace/newline — measured:

```
BOM+DOCTYPE => Connect stream message too large: 3149872161 bytes
LF+html     => Connect stream message too large: 1013478509 bytes
space+html  => Connect stream message too large: 1013478509 bytes
```

The flag space is closed, so the check can be inverted instead of enumerated. Only `connectCompressedFlag` (0x01) and `connectEndStreamFlag` (0x02) exist (`connect.go:21-22`), and consumers always test them bitwise (`envd.go:290`, `:293`, `:564`; `pty.go:605`, `:608`), so legal values are `0x00`–`0x03`:

```go
if header[0] > (connectCompressedFlag | connectEndStreamFlag) {
    return 0, nil, fmt.Errorf("response is not a Connect envelope (first 5 bytes: %x); "+
        "check whether CUBE_PROXY_NODE_IP is configured, the data-plane endpoint points to "+
        "a web service, or the response was compressed", header)
}
```

One line, and it covers HTML, JSON, gzip, BOM, leading whitespace and TLS record headers. Note it must be `> 0x03`, not `> 0x02` — `0x03` (compressed + end-stream) is legal. This would also let you drop the `!DOC` → `!DOCTYPE` special case and the misleading `(%d bytes)`, and it removes the unsanitized `tag` (see below). Worth a test pinning the legal flag range either way.

Happy to re-review quickly once 1–3 are addressed. Thanks again for chasing this one down — the fix derivation is genuinely nice work.

sdk/go/connect.go
```diff
@@ -82,3 +88,51 @@ func parseConnectEndStream(raw []byte) error {
...
+
+// validateConnectResponse ensures the HTTP response represents a valid Connect stream.
+// Non-200 responses are parsed into an APIError via apiErrorFromResponse, enriching
+// HTML and 404 text pages with actionable hints while clearing false
+// ErrSandboxNotFound classifications on those 404s (preserving other kinds).
+// For 200 responses, text/* media types (such as text/html or text/plain from a misrouted
+// Web UI, portal, or gateway landing page) are immediately rejected with an *APIError
+// without blocking on body reads. Other content types (and omitted headers) fall through
+// to envelope parsing for maximum proxy tolerance.
+func validateConnectResponse(resp *http.Response) error {
```
wbzdssm: validateConnectResponse is doing too many things (status handling, content-type parsing, error-kind mutation, and message rewriting

PR #1824, opened 2026-09-22
fix(snapshot): capture sandbox snapshots through shim actions

luzhixing12345:
Fix bad design of #1760

Add SnapshotCapture and SnapshotResume so CommitSandbox and AppSnapshot keep the VM paused while memory and rootfs snapshots are created.

Reject shims that do not support SnapshotCapture, add lease-based recovery if Cubelet exits while the VM is paused, and remove the keep-paused cube-runtime commands.


lkml-likexu:
Wonderful Performance Improvement for "4.1 Snapshot Creation vs Concurrency"

Clipboard_Screenshot_1790134955

fslongjin (changes requested):
The direction is right: memory and rootfs must be captured inside the same pause. What does not hold is who keeps that pause alive.

After the memory dump, the shim freezes the VM for another 30 seconds, and only the **current Cubelet process** renews that lease every 10 seconds. When the process is gone, renewal stops. If nobody renews for about 30 seconds, the shim resumes the VM on its own. The Cubelet that starts later does not adopt this freeze.

That mechanism fits a crash where the process will not be back soon. It does not fit a planned stop.

After a node is isolated, operators stop Cubelet by hand and start it again later. A gap longer than 30 seconds is normal, not an accident. On some machines, coming back to a usable process already takes more than 40 seconds, and the start script itself allows up to 60 seconds. The 30-second lease is shorter than both. If a sandbox is mid-snapshot when Cubelet stops, that snapshot is split apart: the VM is resumed and the guest writes again before the new process is up. Disk and memory are no longer the same instant, so the snapshot cannot be continued, and the half-written result cannot be published.

Graceful shutdown does not close this hole. On SIGTERM, Cubelet cancels its own context and then stops gRPC immediately (`Server.Stop`, not `GracefulStop`, which would wait for in-flight calls). The process then exits. The exit path does none of the following:

- wait for this snapshot to finish
- resume the VM before exit and mark the snapshot failed
- record "this VM is still frozen" for the next start to pick up

The one-click stop script waits at most 20 seconds, and systemd `TimeoutStopSec` is 30 seconds. Those budgets mean "how long the process is allowed to die," not "hold this snapshot's freeze." Once the process has exited, the 30 seconds are counted again from the last renewal, regardless of when an operator starts Cubelet.

So this is not graceful shutdown being so slow that it uses up the 30 seconds. Shutdown never handles the freeze, and leaves the aftermath to a timer that is shorter than a normal stop-to-start gap.

Please do not treat "if the process dies, let the shim resume 30 seconds later" as the plan for a planned stop. Before exit there must be one definite outcome:

1. Finish this snapshot and resume the VM before the process exits.
2. Or resume the VM before exit, fail this snapshot, and do not leave a half-written result for the next start to touch.

The 30-second auto-resume can stay for crashes that leave no time to clean up, such as SIGKILL or OOM. `systemctl stop`, stopping the process by hand, and maintenance after isolation are not that kind of crash.

fslongjin (approved):
After communicating with @luzhixing12345, the issues I mentioned earlier will be resolved in the next version.

luzhixing12345:
this pr still has two problems
1. Hard exiting a cubelet will leave residual files, which need to be persistently recorded before taking a snapshot.
2. When a cubelet gracefully exits, it stops accepting new requests and waits for or explicitly aborts the snapshot being executed.

PR #1828, opened 2026-09-22
fix(deploy): avoid grep -vx empty-input exit aborting install under p…

liciazhu:
fix(deploy): avoid grep -vx empty-input exit aborting install under pipefail


tinklone:
LGTM

PR #1830, opened 2026-09-22
fix(cubeapi): implement cursor pagination for GET /v2/sandboxes

### LINKED ISSUE #1792: [Bug Report] GET /v2/sandboxes accepts nextToken but does not implement cursor pagination
## Bug Description

`GET /v2/sandboxes` declares a `nextToken` query parameter, but the token is currently ignored.

As a result:

- Every request starts from the first item (`start_idx = 0`).
- Supplying different `nextToken` values does not change the returned page.
- The response does not include an `x-next-token` header.
- Clients cannot retrieve sandboxes beyond the first `limit` results.
- The endpoint does not provide the cursor-pagination semantics expected by the E2B SDK.

The `limit` parameter is applied, but this only limits the first page; it does not provide usable pagination.

## Current Implementation

The request model declares `nextToken`:

```rust
pub struct ListSandboxesV2Query {
    pub metadata: Option<String>,
    pub state: Option<String>,
    #[serde(rename = "nextToken")]
    pub next_token: Option<String>,
    pub limit: i32,
}
```

However, the handler only passes `metadata`, `state`, and `limit` to the service:

```rust
state
    .services
    .sandboxes
    .list(
        params.metadata.as_deref(),
        params.state.as_deref(),
        params.limit,
    )
    .await?;
```

The service always sends the following pagination values to CubeMaster:

```rust
start_idx: Some(0),
size: Some(limit.max(1)),
```

The response is returned as a plain JSON array without an `x-next-token` response header.

Relevant source locations:

- [`ListSandboxesV2Query`](https://github.com/TencentCloud/CubeSandbox/blob/master/CubeAPI/src/models/mod.rs#L828-L842)
- [`list_sandboxes_v2` handler](https://github.com/TencentCloud/CubeSandbox/blob/master/CubeAPI/src/handlers/sandboxes.rs#L86-L117)
- [`SandboxService::list`](https://github.com/TencentCloud/CubeSandbox/blob/master/CubeAPI/src/services/sandboxes.rs#L79-L107)

## Steps to Reproduce

Assume more than two sandboxes exist.

Request the first page:

```bash
curl -i \
  "http://localhost:3000/v2/sandboxes?limit=2"
```

Then request another page using an arbitrary token:

```bash
curl -i \
  "http://localhost:3000/v2/sandboxes?limit=2&nextToken=some-token"
```

## Actual Behavior

1. Both requests return the same first-page results.
2. `nextToken` does not affect the CubeMaster request.
3. No `x-next-token` header is returned.
4. There is no way to retrieve the next page.

## Expected Behavior

`GET /v2/sandboxes` should implement cursor pagination compatible with the documented E2B behavior:

1. An omitted `nextToken` returns the first page.
2. A valid `nextToken` continues after the previous page.
3. If more items are available, the response includes the next opaque cursor in the `x-next-token` header.
4. If no more items are available, `x-next-token` is omitted or empty.
5. Invalid or malformed cursors produce a clear `400 Bad Request`.
6. Repeating the same request with the same cursor and unchanged data returns a stable page.

Example:

```http
GET /v2/sandboxes?limit=2
```

```http
HTTP/1.1 200 OK
x-next-token: <opaque-token>
content-type: application/json
```

The next request would be:

```http
GET /v2/sandboxes?limit=2&nextToken=<opaque-token>
```

## Filtering Semantics

Currently, `metadata` and `state` filtering are applied after CubeMaster has already limited the result set.

This can cause a request such as:

```http
GET /v2/sandboxes?state=paused&limit=10
```

to return fewer than 10 items even when additional matching sandboxes exist outside the first unfiltered backend page.

Pagination should therefore be applied over the filtered result set, or the backend query should receive the filters so that page boundaries correspond to matching items.

## Compatibility Impact

E2B exposes sandbox listing as a paginator and expects cursor-based continuation. CubeSandbox currently returns a plain list and supports only partial list semantics.

This difference is also identified in the E2B compatibility documentation work:

- #1784
- #1790
- [PR #1790 pagination-related review discussion](https://github.com/TencentCloud/CubeSandbox/pull/1790#discussion_r4058950038)

Without cursor support, SDK clients may:

- repeatedly receive the first page;
- fail to enumerate all sandboxes;
- fall back to Cube-specific API handling;
- behave incorrectly once the sandbox count exceeds the page limit.

## Suggested Implementation Direction

One possible implementation is:

1. Treat `nextToken` as an opaque CubeAPI cursor.
2. Encode sufficient continuation state, such as the backend offset and any required ordering information.
3. Pass the decoded offset to CubeMaster as `start_idx`.
4. Request enough information to determine whether another page exists, for example `limit + 1` items.
5. Return the next cursor through `x-next-token` when additional results exist.
6. Define a deterministic ordering before relying on offset-based pagination.
7. Add integration tests covering multiple pages and filters.

The public token should remain opaque so that the backend pagination strategy can change later.

## Acceptance Criteria

- [ ] `nextToken` is consumed by `GET /v2/sandboxes`.
- [ ] A second-page request returns different results from the first page.
- [ ] `x-next-token` is returned when additional results exist.
- [ ] The header is omitted or empty on the final page.
- [ ] Invalid tokens return `400 Bad Request`.
- [ ] Pagination uses deterministic ordering.
- [ ] `metadata` and `state` filters work correctly across page boundaries.
- [ ] Tests cover:
  - first, middle, and final pages;
  - empty results;
  - `limit=1`;
  - invalid cursors;
  - state filtering;
  - metadata filtering;
  - data changes between page requests.
- [ ] E2B SDK compatibility tests cover sandbox list pagination.

## Environment

- Branch: `master`
- Observed on: 2026-09-21

ydflow:
Closes #1792

## Problem

`GET /v2/sandboxes` declares a `nextToken` query parameter that nothing reads. `ListSandboxesV2Query` carried the field (it even had `#[allow(dead_code)]` on it), the handler forwarded only `metadata` / `state` / `limit`, and `SandboxService::list` hardcoded `start_idx: Some(0)`. Consequences:

- Every request returns the same first `limit` sandboxes, whatever `nextToken` says.
- No `x-next-token` response header, so a client cannot ask for the next page.
- `metadata` / `state` filters are applied after CubeMaster has already limited the result set, so a filtered page can be short even when more matching sandboxes exist.

## Why the pagination is completed in CubeAPI

The issue's suggested direction — decode the cursor into an offset and pass it to CubeMaster as `start_idx` — does not work, and I want to be explicit about why rather than quietly pick a different design.

CubeMaster's `start_idx` / `size` window is expressed in **nodes**, not sandboxes. In `CubeMaster/pkg/service/sandbox/sandbox_list.go`, `ListSandbox` fills `Total` with `GetHealthyNodesByInstanceType(...).Len()` and resolves the window with `localcache.RangeDBHost(req.StartIdx, req.Size, ...)`; `NodeList.IndexByPage` is 1-based over nodes (`CubeMaster/pkg/base/node/node.go`). Every sandbox on the selected nodes is then fanned out to cubelet and returned. So:

- `limit` was never "sandboxes per page" — on a single-node deployment `Total == 1` and `limit=2` returns every sandbox at once.
- `Total` / `EndIdx` are node-dimension bookkeeping and cannot answer "is there another page of sandboxes".

Forwarding a client offset into `start_idx` would therefore paginate over the wrong unit. Changing CubeMaster's semantics would be a cross-component change touching the scheduler and the inventory path (which depends on `Total` being a node count), so this PR keeps the change inside CubeAPI.

## What this PR does

- Adds `SandboxService::list_v2`, which walks CubeMaster's node windows until one comes back short, then cuts the page over the **filtered** sandbox set. Because `metadata` / `state` filtering happens on this side, page boundaries line up with what the client actually sees.
- Orders the filtered set by `sandbox_id`. A deterministic order removes backend reordering as a source of drift; see the limitations section for what it does *not* buy.
- Returns the continuation cursor in an `x-next-token` response header, mirroring the existing `GET /snapshots` implementation. It is omitted on the last page rather than sent empty.
- Rejects a malformed cursor with `400` instead of silently restarting from page 1 — the exact behaviour being replaced. An empty / whitespace-only `nextToken=` still means "start from the beginning", matching how `GET /snapshots` normalises it.
- Parses `end_idx` / `total` / `size` from CubeMaster's list response. All three were already serialized by `ListCubeSandboxRes`; CubeAPI simply ignored them. Only `size` is used to decide whether another window remains — see the next section.
- Drops the now-stale `#[allow(dead_code)]` on `ListSandboxesV2Query` and on `AppError::BadRequest` (this change makes it a real caller), and syncs `openapi.yml` with the utoipa annotation.

### Why the walk terminates on `size`, not `end_idx`

An earlier revision of this PR stopped when `end_idx < total`. That is not a valid test, and the review caught it:

- `end_idx` is a node **row id** — `Index: int(elem.ID)` in `localcache/db_cache.go:239` — so deleted node rows leave gaps.
- `total` is a healthy-node **count**.

Comparing the two is only sound while ids happen to be dense `1..N`. On a cluster with id gaps, a full window reports `end_idx >= total`, the loop breaks with nodes still un-fetched, and their sandboxes go missing from every page — together with the shimless paused rows CubeMaster only appends on the final node page, so `state=paused` silently under-reports. No error, no cursor, just fewer items.

`size` is the number of nodes the window actually covered (`rsp.Size = len(nodeList)`), so `size < window` is a sound "this was the last window". A missing or non-positive `size` terminates rather than spinning on the same window. A regression test covers the sparse-id shape (`end_idx: 599, total: 120` on a full window) that defeated the old condition.

Worth flagging: the same `EndIdx >= Total` idiom already exists in `cubemastercli/commands/cubebox/list.go:263`, so this may be a CubeMaster contract question rather than only a CubeAPI one. I have not touched that caller.

### Cost model

`list_v2` walks node windows of 100 until one comes back short, so a request costs roughly `ceil(nodes / 100)` rounds, each fanning out one cubelet `List` per node in the window (serialized per node by `CubeletListLock`). This is inherent to paginating over a cubelet-fanned-out set rather than a backend-side cursor, and it is why the sandbox-level slice is taken after collecting instead of forwarding an offset.

`limit` is still unvalidated, so a `limit=1` request pays the same collection cost as `limit=100` — the fan-out amplification a client can trigger cheaply. Bounding or validating `limit` would be a reasonable follow-up.

### Known limitations

- **An offset cursor is not stable across concurrent writes.** The `sandbox_id` order removes backend reordering as a source of drift, but the offset is recomputed against a fresh snapshot on each call, so a create or delete before the boundary makes a later page repeat or skip an item. A cursor immune to that would name the last-seen key rather than a position — a larger change than this fix, and worth its own issue if it matters to you.
- **A short set can be mistaken for the end of the list.** CubeAPI uses `ListSandbox`, not `ListSandboxWithFailOnError`, so an unreachable cubelet shortens the set with a success `ret_code`. A cursor paging over a truncated set can therefore report "end of list" while unseen sandboxes exist. Pre-existing behaviour.
- **The response order changes for existing clients** — now `sandbox_id` order rather than CubeMaster's `types.SortSandboxList`. Harmless for an array response, but visible.

## Testing

`cargo test --bin cube-api` → **162 passed; 0 failed** (152 before this change, 10 new).

New tests, driven through the real router where the assertion is about what a client observes:

- `services::sandboxes::tests::list_v2_*` (8) — first page capped and reports a cursor; second page differs from the first; final page omits the cursor; malformed cursor is a 400; `metadata` filter paginates over matching items only; `state` filter paginates over matching items only; every node window is walked before the page is cut; the walk continues when a full window's node ids already exceed the healthy-node count (sparse-id regression).
- `routes::tests::list_sandboxes_v2_returns_a_usable_next_token_header` — asserts the header value on page 1, that page 2 continues to `sb-3`, and that the last page carries no trailing cursor.
- `routes::tests::list_sandboxes_v2_rejects_an_unknown_next_token` — asserts a 400 body for a foreign cursor.

Verified RED first: reverting `list_v2` to `start = 0` with `next_token = None` fails 5 of the 8 service tests, so the tests do lock the defect rather than passing vacuously.

`cargo fmt --check` is clean. `cargo clippy --bin cube-api --all-targets` reports no warnings in the ranges this PR touches; the warnings it does emit are pre-existing (verified by diff line range).

Still not covered: a client that requests page 2 after an item was removed between the two requests (the duplicate/skip case above), and a backend returning sparse node ids — the latter is what the new regression test pins, the former is the limitation called out above.

## About `openapi.yml`

Regenerating with `cube-api --export-openapi` produces a file that differs from the checked-in one in ways unrelated to this change — it drops the copyright header and the `servers` block. Rather than overwrite those, I ported the `/v2/sandboxes` section by hand so the diff stays scoped to this PR (+12 lines, no deletions). If you would rather have the full regenerated file, say so and I will swap it in. No CI job regenerates and diffs it, so drift here is silent.

This is a single-component change (CubeAPI only) with no behaviour change to any existing endpoint: `GET /sandboxes` (v1) still uses the original `list`.

Per `AGENTS.md` these commits omit `Signed-off-by` and carry `Assisted-by`; a human submitter still needs to review and add the DCO tag before merge.


ydflow:
Thanks for the review — this is the most useful pass this change has had. I went through all of it rather than only the parts I agreed with. Most is now addressed in `54c449b`, with the exceptions called out explicitly below.

## Addressed in `54c449b`

**`start_idx += window` was wrong, and you were right about the direction.** `IndexByPage` resolves `start_idx` as the first node whose `Index >= start_idx`, and `Index` is the `t_host_info` row id (`db_cache.go:239`), so deleted rows leave gaps and `+100` can land back inside the window just returned. I checked both branches in `node.go:371-393`: the index branch returns the last covered node's `Index`, the positional fallback returns the last covered 1-based position. `end_idx + 1` is the continuation past the window in both, so that is now the advance. Termination stays on `size == window`; a missing, zero or negative `end_idx` breaks the loop rather than spinning on the same window — your "stop instead of guess" point.

**Dedup by `sandbox_id`** is in, for the reason `cubemastercli/commands/cubebox/list.go:255-261` does it, plus your note that CubeMaster can append the same shimless pause row on more than one window when `EndIdx >= Total` holds per page.

**You were right that the sparse-id test did not pin anything.** That one stung, because the test *name* claimed a regression it could not catch. The mock is rebuilt to model `IndexByPage` rather than count windows: it keeps a fixed id list, resolves the first position with `Index >= start_idx`, and reports the last returned `Index` as `end_idx`. It fails against the old `start += window` advance (duplicated ids in the result), which is what the regression should have asserted from the start. `list_v2_walks_every_node_window_before_paging` still cannot distinguish the two advances, exactly as you said — `1 + 100 == 100 + 1` — so it is not the test that locks this; the new one is.

**`openapi.yml` is regenerable and my "cannot regenerate" note was wrong.** Dropped. `cargo run --bin cube-api -- --export-openapi` is there, and the spec *does* track the utoipa doc comments (`openapi.yml:552` for `GET /snapshots`). The regenerated file still differs from the checked-in one in unrelated ways (copyright header, `servers` block), so the section is ported by hand — but that is a scoping choice now, not a claim that regeneration is impossible. No CI diffs it, so drift is silent; happy to swap in the full file if you prefer.

**The unused `build_sandbox_routes` import** is gone, which contradicts the clippy claim in the PR description. That claim was wrong when written and is corrected.

## Not done, deliberately

**Clamping `limit`.** I agree with your reasoning — it is the one knob bounding the pages-per-scan multiplier, and `1..=1000` via `#[validate]` the way the write models do is cheap. I have not done it, because it changes the public contract for existing callers and I would rather land it as a visible, separately reviewable change than smuggle a behaviour change into a pagination fix. Same for a round cap that fails predictably instead of hitting the 30 s default. If you want either in this PR, say so and I will add it.

**The four in-repo consumers that drop `x-next-token`.** This is the finding I would rather not fold in here: `sdk/python` `list_v2`, `sdk/node` `listV2`, `sdk/go` `ListV2` and `web`'s `sandboxApi.list` are four languages with four release cadences, and it is the difference between "this endpoint is broken" and "this endpoint now has a real bound". I think it deserves one PR per SDK. But the "List **all** running sandboxes" docstrings are wrong today and I am not leaving those — correcting them and filing the follow-up is part of this PR.

**Keyed cursor instead of offset.** Your `sbx-after-<sandbox_id>` suggestion is right and cheaper than I assumed — the set is already sorted by `sandbox_id`, so `partition_point` is a couple of lines. Still not in this PR: it changes the cursor format, so any client holding a `sbx-offset-*` token breaks at the boundary. Worth its own issue.

## Two things I want to push back on

**`total` in the response struct.** I read the finding as "drop the field", and I would rather not: CubeMaster still serializes it, it is the only signal a caller has that a window was truncated against a node count, and it costs nothing to keep. But I accept the risk you named — a field nobody reads invites someone to compare it against `end_idx`. So it now carries a doc comment stating plainly that it is a node count, not a sandbox count, and must not be compared with `end_idx`. If you would still rather it were gone, say so and it goes.

**The 30 s timeout framing.** `routes.rs:26` is the default route timeout and `GET /snapshots` — the implementation this mirrors — sits on the same default with the same fan-out shape. I am not defending the current cost; I am saying the pagination change did not introduce it, and the bound you want is the `limit` clamp above rather than a per-route timeout change.

## One thing I got wrong in the first pass

`std::collections::HashSet` is still written fully qualified in `list_v2`, even though `sandboxes.rs:5` imports `HashSet`. I had it down as fixed and it is not — it goes in the next commit.

## Process and verification

`AGENTS.md` requires `Signed-off-by`; the commits carry it plus `Assisted-by`. A human submitter still has to review before merge, which the PR description already says.

For `54c449b`: `cargo test --bin cube-api`, `cargo fmt --check` and `cargo clippy --bin cube-api --all-targets` are the same suite as the base PR. The two tests that changed are the ones named above, and I re-ran the sparse-id one against the previous advance to confirm it goes red rather than passing vacuously. Note that `DCO Check`, `Unit Test Check`, `Build Check` and `Format Check` all report `action_required` on this head — I have not re-run them and am not claiming they pass.

